PR #53, opened 2012-12-05
Improved the database query of results-view and the data structures of r...

atsnisov:
...esults view.


skulonen:
Disregard the last 5 references. I accidentally linked to this pull request instead of the issue with the same id in mooc-grader.

PR #158, opened 2015-02-04
Add CourseHook that calls url and SubmissionContentResource to API

LassiHaaranen:
CONTAINS SCHEMAMIGRATIONS!

Adds CourseHook that enables calls to URLs after a submission is
set as ready on a specific course instance.

SubmissionContentResource is an API resource that displays the
contents of the submitted files (if there are any) base64 encoded,
as well as, students ids.

Together these can be used to automate certaing tasks with external
services, e.g. plagiarism detection.


ttsirkia:
:thumbsup:

PR #161, opened 2015-02-13
Make exercise buttons narrower if there is a lot of exercises

ttsirkia:
With this modification, the exercise buttons in the main view are narrower if there are more than 10 exercises in one exercise round. This prevents some layout issues.


LassiHaaranen:
:+1:

PR #163, opened 2015-02-24
Hotfix problem that occurs when user has no email field set in Shibboleth

LassiHaaranen:
Fixes a problem where A+ fails when user has no email set in Shibboleth
attributes. These cases occur at least when students have study rights for
individual courses (erillisopinto-oikeus).


oseppala:
Ok for merge.

ttsirkia:
This PR had an issue which is hopefully fixed in #162

Issue #207, opened 2016-04-14
REST API accepts requests from unauthenticated user


Vilz92:
When making eg. request https://plus.cs.tut.fi/api/v1/userprofile/?format=json&limit=0, a-plus gives list of all users, their emails etc, without any authentication done.

ttsirkia:
Fixed in #208

PR #208, opened 2016-04-14
Fix for unauthenticated REST API request

Vilz92:
Fix for the issue #207. The main issue seemed to be that the implementation was made for Tastypie 0.9.11 or older which is not supported anymore.


teemulehtinen:

teemulehtinen:
add to whitelist

PR #209, opened 2016-04-15
Prevent copying line numbers in code preview

ttsirkia:
When viewing a submission in staff view in the modal panel, copying the code copies also the line numbers. This small tweak prevents this.

The CSS is based on a example in https://developer.mozilla.org/en-US/docs/Web/CSS/user-select


ttsirkia:
:cry: :broken_heart:

teemulehtinen:
retest this please

PR #212, opened 2016-04-26
Small fix in installation instructions

matiaslindgren:


teemulehtinen:

PR #217, opened 2016-06-06
Move authorization to app

raphendyr:
Implement authorization app that uses same structure as django-rest-framework (this is going to be used for API in the future).

Commits 206dbac935e39c65305a8fec11c2e512f686d662 and 3b6ed573ae5c36dd6752440f4ab64657b152e782 have more description of those changes.

Some methods have been deprecated and should be replaced with correct implementation in later date.

This PR has no change in functionality. It is just refactor of authorization and get_url system.


teemulehtinen:

raphendyr:
Apparently I forgot to do smoke testing after splitting code in authorization to multiple files.

teemulehtinen:
add to whitelist

raphendyr:
Squashed the fixes to correct commits.

PR #216, opened 2016-06-02
API uusiksi Django RESTillä

Vilz92:

PR #219, opened 2016-06-06
Implement api v2 with Django REST Framework (replaces #216)

raphendyr:
Rebased and fixed #216 to work better in a-plus and to be more general initial solution.

Also this one doesn't remove the old v1 api.


raphendyr:
TODO: lock users api and allow only authenticated users.

raphendyr:
I think this initial version of v2 api is on ok shape to be merget to develop. It has support for updating the version number so clients can validate that they are not talking to too old version of a-plus and it can be requested with multiple different accept headers that should result correct responses.

This patch and already merged #217 should be good base to start developing more interconnected views and api calls with same permissions and so on.

PR #222, opened 2016-06-17
Fix local settings import

raphendyr:
Local settings were never imported (unless in path). Please check that this doesn't break any production setups.

In the future, warning will be logged if no local settings file is found


teemulehtinen:
I think local_settings.py is more visible in the project root. At least, we should keep compatibility with that arrangement. Try to import a list of paths?

teemulehtinen:
Closing in favor of #224

PR #225, opened 2016-06-21
New REST API resources

Vilz92:
- api/v2/exercises/{exercise_id}/submissions/{submission_id}
- Magical api/v2/me


teemulehtinen:
In exercise/api/views.py one should use BaseExercise.DoesNotExist (or django.core.exceptions.ObjectDoesNotExist). There seems to also be few cases where tests expect different fields, see http://plustest.cs.hut.fi/jobs/A-plus-test-PR/builds/85/log

raphendyr:
There is mix of commits with message in Finnish and some with English. Also there is set of merging from develop and lot of fixup commits. Could you cleanup the history a bit (rebase -i)?

api/urls_v2.py
```diff
@@ -24,8 +33,8 @@
 
 urlpatterns = [
     url(r'^', include(api.urls, namespace='api')),
-    url(r'^learningobject/$', LearningObjectList.as_view()),
-    url(r'^submission/$', SubmissionList.as_view()),
+
```
raphendyr: I don't see need for new line here.

course/api/views.py
```diff
@@ -1,4 +1,5 @@
 from rest_framework import generics, permissions, viewsets
+
```
raphendyr: I don't see point for newline. All importas are from rest_framework / libraries

userprofile/api/serializers.py
```diff
@@ -26,12 +25,27 @@ class UserSerializer(UserBriefSerialiser):
     Add the details of a user.
     """
 
+    courses = serializers.SerializerMethodField('list_courses')
```
raphendyr: Something like `courses = CourseBriefSerializer(source='enrolled', many=True)` should be enough

userprofile/api/serializers.py
```diff
@@ -1,8 +1,7 @@
 from rest_framework import serializers
-
```
raphendyr: new line here would be nice to keep external libraries and project imports separated.

raphendyr:
I added some comments on the code.

PR #226, opened 2016-07-01
Add submission ordinal number and avoid migration for api tokens

teemulehtinen:


raphendyr:
Seems good. We presume here that method `ordinal_number` works as intended in production. If unexpected problems occur, new issue should be created.

PR #229, opened 2016-07-08
Get points of every user in the course

Vilz92:


teemulehtinen:
Merging this BUT the use of ResultTable should be avoided as it is a very heavy operation on large courses. In this case the same functionality can be achieved with course.CourseInstance.get_student_profiles() and exercise.presentation.summary.UserCourseSummary.

PR #231, opened 2016-08-02
Update links for unlisted exercises in user results page

korhona14:


teemulehtinen:

teemulehtinen:
add to whitelist

raphendyr:
Not sure if correct as `order`is number that works only in own level. `number` would resolve full exercise path like `.1.1` but it doesn't include number from module. `LearningObject.__str__`has the closest of correct values.

In the other hand, I have no idea how those id's are generated.

teemulehtinen:
Order should be sufficient here as exercises embedded to parent are on one level. However the html attribute is now a running number and does not necessarily match the exercise.order
https://github.com/Aalto-LeTech/a-plus/blob/master/exercise/protocol/aplus.py#L140
This "i" here can be easily changed to "o.order" and we should be good.

PR #234, opened 2016-08-16
Feature: api permissions, grader authentication and grader endpoints in api

raphendyr:
Permission and access management rework. Basically most `access_control` calls are replaced with correct `get_access_mode` calls and permission classes (`AccessModePermissions` validates the `get_access_mode` result).

**NOTE**: Permission framework is changed, so there can be bugs that are critical. So extra step for testing is recommended.

Points in API is not yet under permission validation. If this version ends in production that endpoint should be disabled.

Permission framework works in simple manner. All views have set of permission classes (returned by `get_permission_classes`). If all permission classes method `has_permission` return True, permission is granted. If any return False, permission is denied.

**Async urls for grader are moved into API using grader authentication tokens**

Grader authentication for exercise urls uses better tokens than before. Token contains time, nonce and signature. More in https://github.com/Aalto-LeTech/a-plus/blob/1df83f5d5e1358e425fdca1f64fc0b03049adfa7/lib/crypto/signed_messages.py#L15

**IMPORTANT**:
 * grader url get parameter `uid` now continas user ids instead of userprofile ids
 * submission_url get request result is different than before
 * old `/rest/` asyncronous endpoints do not work anymore. a-plus adds new correct submission_url variables
 * submission_url can be used only for linked submission or exercise (submission token only for submission and exercise token only for creating new submission)
 * submission_url for exercise works only for 1 day. After that it expires and new url is needed

### DIFF
--- a/api/urls_v2.py
+++ b/api/urls_v2.py
@@ -22,9 +22,10 @@ with api.register(r'courses',
     courses.register(r'students',
                      course.api.views.CourseStudentsViewSet,
                      base_name='course-students')
-    courses.register(r'points',
-                     course.api.views.CoursePointsViewSet,
-                     base_name='course-points')
+    # FIXME: Disabled w:hile there is no correct permission checks
+    #courses.register(r'points',
+    #                 course.api.views.CoursePointsViewSet,
+    #                 base_name='course-points')
 
 with api.register(r'exercises',
                   exercise.api.views.ExerciseViewSet,
--- a/api/urls_v2.py
+++ b/api/urls_v2.py
@@ -46,8 +46,25 @@ urlpatterns = [
     url(r'^me', userprofile.api.views.MeDetail.as_view()),
 ]
 
+
 if settings.DEBUG:
-    _len = max((len(url.name) for url in api.urls))
-    _fmt = "  - %%-%ds %%s" % (_len,)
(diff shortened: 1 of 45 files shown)


teemulehtinen:
I confirm that the permissions are correct. The structure is greatly improved, good work!

Should we disable the Points API before merging? I would like to merge new features on top of this and demonstrate everything in production environment on Monday.

The 1 day grading token will probably break the Rubyric integration. I believe it stores the grading URL along the submitted files until manual grading is finalized and delivered, which can take any time from a day to a month. Any good suggestions for the manual grading workflow?

raphendyr:
The 1 day restriction only aplies to async tokens created without submission object. Before these were used for view "async-new". While tokens generated with submissiion (old name "async-grade" do not have same restriction as the token is based on hash of the submission).

Idea is that the non submission tokens can be generated by the client (presuming hw knows server secret). Adding time makes it a bit more harder to use tokens captured from logs for example. Idea was that these tokens should be temporary or interactive exercises, while longer taking grading would always have submission already in database.

Of course if this premise doesn't apply to how rubyric interacts with a-plus, we can remove the time restriction for now.

raphendyr:
I disabled the points ednpoint for now.

PR #241, opened 2016-09-05
Feature: api token in lti laucnh request

raphendyr:
Adds 3 custom parameters into lti launch parameters. This contain course api url, course id (for db lookup without api url parsing or request) and users api token.

In future this api token should be created per service (allows revoking access to one service).

This patch is required for mooc-jutut.

### DIFF
--- a/external_services/templates/external_services/lti_form.html
+++ b/external_services/templates/external_services/lti_form.html
@@ -17,18 +17,44 @@
 		to the provider of this service.
 		{% endblocktrans %}
 	</p>
+	{% if service.enable_api_access %}
+	<p>
+		{% blocktrans %}
+		For this service also API token is transferred.
+		Thus this service will have full access to A+ API with your privileges and
+		could possibly do harm in you name.
+		System administrator can disable this feature for this service.
+		{% endblocktrans %}
+	</p>
+	</p>
+	{% endif %}
 	<p>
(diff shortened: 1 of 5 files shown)


external_services/templates/external_services/lti_form.html
```diff
@@ -17,38 +17,73 @@
 		to the provider of this service.
 		{% endblocktrans %}
 	</p>
+	{% if service.enable_api_access %}
+	<p>
+		{% blocktrans %}
+		For this service also API token is transferred.
+		Thus this service will have full access to A+ API with your privileges and
+		could possibly do harm in you name.
```
ihantola: This is all true, in theory. However, as in practice many of these services are not truly external but maintained by the same people who run the course, I would like either to be able to edit this text as a user (course owner),  or use softer tone in the phrasing. How about "could possible act as you". Another question is whether students will benefit from "System administrator can disable this feature for this service" - as I would assume there was a reason why the admins enabled it in the first place.

ttsirkia:
I think there should be three options: no API key transfer at all, API key transfer to an external service (really hosted by someone else) or API key transfer to locally administrated service.

raphendyr:
Added solution that should be good enought for above concerns. Basically show different message for course staff and students.

PR #242, opened 2016-09-07
Feature: user tags

raphendyr:
User tags are mainly designed for course managment to tag students with usefull tags like "mooc", "freshmen",. "experienced in programming".

This chane implements models and course management interface for usertags.

Course contains list of available tags and relation that holds link between: course instance, user and tag.
There is no way to  add these "taggings" from web or through api yet.

API interfaces to get list of registered tags and taggings.


teemulehtinen:
retest this please

PR #243, opened 2016-09-07
Better __str__ for UserProfile

ttsirkia:
Provides a better string format for users, shown for example when selecting course assistants.


teemulehtinen:
retest this please

PR #253, opened 2016-09-22
Feature: Add deviations to course module + show exercise category in …

Vilz92:
…infobox

### DIFF
--- a/exercise/templates/exercise/_exercise_info.html
+++ b/exercise/templates/exercise/_exercise_info.html
@@ -23,7 +23,12 @@
     </div>
     <div class="panel-body">
         <dl>
-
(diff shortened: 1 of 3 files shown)


exercise/templates/exercise/_exercise_info.html
```diff
@@ -38,6 +38,10 @@ <h4 class="panel-title">{% trans "Exercise info" %}</h4>
             </dd>
             {% endif %}
 
+            <dt>{% trans "Exercise category" %}</dt>
```
ttsirkia: Categories are not always visible for the students. Therefore, this should be checked before rendering the name of the category.

PR #255, opened 2016-09-28
Make course images clickable

ttsirkia:
Make the course images in the front page clickable. Those are nice and big targets to click.


LassiHaaranen:
👍

PR #261, opened 2016-11-18
Implement Copy to clipboard button

ttsirkia:
Allows easy copying of students' code in Inspect view.

copy

The button text cannot be currently translated and this does not work with model solutions.


teemulehtinen:
retest this please

PR #263, opened 2016-11-24
Add management command to scale student's points on a course instance

LassiHaaranen:
This is useful for example when a single students grades need to be zeroed out on a course.


teemulehtinen:
retest this please

PR #273, opened 2017-01-21
Submission result filtering fixes (by score)

piehei:
This fixes a score filtering bug in aplus.js.

Filtering other fields is fine with partial match but score should be exact.

Filter scores by 0
Matches before: 0, 10, 102, 300, 50 etc.
Matches now: 0

Filter scores by 10 
Matches before: 10, 110, 710, 1000 etc.
Matches now: 10

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -320,6 +320,12 @@ $(function() {
                 pass = false;
                 return false;
               }
+		      if (self.filters[i] && self.filters[3] !== "") {
+                if (self.filters[3] != $(this).text().toLowerCase().trim()) {
+                  pass = false;
+                  return false;
+                }
+              }
             });
             return pass;
           }).show();


Plustest:

assets/js/aplus.js
```diff
@@ -320,6 +320,12 @@ $(function() {
                 pass = false;
                 return false;
               }
+		      if (self.filters[i] && self.filters[3] !== "") {
```
ttsirkia: Bad indentation

ttsirkia:
I would suggest that in HTML a column can be marked as integer. In those columns, the value is converted into integer first and the compared. This solution will explode as soon as the column order changes.

assets/js/aplus.js
```diff
@@ -320,6 +320,12 @@ $(function() {
...
+                if (self.filters[3] != $(this).text().toLowerCase().trim()) {
```
ttsirkia: `if (integerColumn && self.filters[3] === +$(this).text())`

ttsirkia:
BTW, `master` branch has not been updated for ages, it is 375 commits behind `develop` branch which is actually in use.

teemulehtinen:
add to whitelist

teemulehtinen:
retest this please

teemulehtinen:
The logic is flawed. If you enter any filter to Grade-column it will filter EVERY column with that exact filter. Can you implement the alternative logic into if (i == 3) { ... } else { ... }. As ttsirkia suggested it would be good improvement to mark the integer column in markup and solve those in init-function: e.g. find('th') and hasClass("integer") for each. Please, push the correction to your fork to update the pull request.

PR #282, opened 2017-06-01
stripped whitespace from course configure urls

splyysh:
Since usually when you paste configure url from mooc-grader it comes with space in the end and breaks the configure phase. It has been annoying enough for me to create a simple fix, since it is less effort for the server to strip the space than me doing it manually.


Plustest:

teemulehtinen:
add to the whitelist

teemulehtinen:
add to whitelist

PR #283, opened 2017-06-07
Upgrade to Django 1.10 and delete API v1

teemulehtinen:


teemulehtinen:
Behold, Python 3.* compatibility is here.

PR #285, opened 2017-06-19
Feature/course states

zebala:
Course now has additional attributes lifesupport_time and archive_time, which specify the course state. Lifesupport removes visibilibilty of model answers for students and archiving removes possibility to submit solutions. Updated translations accordingly.


exercise/exercise_models.py
```diff
@@ -192,6 +192,9 @@ def is_after_open(self, when=None):
     def is_closed(self, when=None):
         return self.course_module.is_closed(when=when)
 
+    def can_show_solutions(self):
+        return (self.is_closed() and (not self.course_instance.is_on_lifesupport()) and (not self.course_instance.is_archived()))
```
ttsirkia: Should this take account the specific user who is requesting the model answer? The user can be staff member or a student with an extended deadline. Now all of these must be handled separately. The original implementation did not care about roles or anything else but just wondering what is the best approach.

Also the number of braces is quite high. :-)

teemulehtinen:
The model view access sjould use the same can_show_solutions test
https://github.com/Aalto-LeTech/a-plus/blob/master/exercise/views.py#L167

PR #289, opened 2017-07-17
Refactor exercise model answer permissions to be in line with other perms.

zebala:


raphendyr:
Squashaa nää commitit ja droppaa toi gitignore pois.

zebala:
Squashed commits and removed .gitignore change due to it being merged earlier in another PR.

PR #310, opened 2017-12-14
Add active element exercise type

roosapi:
Adds active element which is a new exercise type that consists of separate input and output elements. For inputs, chapter.js handles the constructing of the elements and the mapping of inputs to correct outputs. Output elements behave mostly like normal submit exercises, but submits are triggered from the input elements, the inputs for the submission can come from multiple input elements, and the results are updated straight to the output element without modal/redirect to a results page.


Plustest:

exercise/static/exercise/chapter.js
```diff
@@ -233,42 +356,204 @@
...
-				data: new FormData(form_element),
+				data: formData,
 				contentType: false,
 				processData: false,
 				dataType: "html"
-			}).fail(function() {
+			}).fail(function(xhr, textStatus, errorThrown) {
+					// handle database lock exceptions
+					retry = retry || 0;
+					if (xhr.responseText.indexOf("database is locked") >= 0 && retry < 5) {
```
ttsirkia: Is this a special hack for some specific grader? Could it be more generic?

piehei:
I don't completely understand how this works. Could you provide a simple example?

Specifically:

1. When and how the multiple input fields come?
 --- From the exercise.{yaml,rst} mooc-grader proxies to A+, right?
 --- Do all of them come at the same time (when the page is loaded) or asynchronously some at the beginning, some later on (ajax)?

2. Is this graded somehow asynchronously?
 --- Do you make one POST / submission or multiple POSTs throughout the exercise as you complete more and more input fields? 
 --- Are there multiple results (eg. results to different input fields) or just a single result?

Thanks :-)

roosapi:
1. The input elements are defined as empty divs in the chapter page source (html/rst) pretty much like exercises, except that they have the data-attribute 'data-aplus-active-element="in"' instead of 'data-aplus-exercise' attribute. The actual building of all of the input forms is handled by the javascript when the page loads. Also input elements don't have any .yaml configuration files and are not in any way registered by the grader or in A+ as exercises, but only the output elements are defined as exercises. The inputs are only recognized by the A+ chapter javascript. Each output has a data-attribute that contains the ids of the input it requires and the correct values for an output are collected when an element is submitted.

2. Triggering a submit on an input element attempts to submit all the outputs that depend on that input for grading, and all of those outputs that have all their required inputs filled in are submitted for asynchronous grading (one POST for each output element). One input field can have multiple results as basically any number of output elements can depend on one input, and the results are updated to the output elements as soon as the grading of the submission of that output element has finished.

So basically if there is input A and output B that depends on A, the flow is roughly like:
-> user clicks submit on A
-> collect all the outputs (B) that depend on A
-> check if the outputs have other inputs and collect their values
-> if all the inputs for B have a value, combine the input to a form submission and send ajax post to grader
-> if submission successful init polling for results
-> when results are ready update them to correct output

A simple example of how active elements are defined in rst and configuration yamls can be found [here](https://version.aalto.fi/gitlab/piitulr1/active-element-example) if it helps.

exercise/static/exercise/chapter.js
roosapi: Yes, in our test environment using mooc-grader and sqlite multiple grading submissions caused database lock errors for some reason, and this is really just a hack to make the elements usable in that environment. I'm not sure if it is possible to make it more generic since the error code is just 500 and the response text seems be the only thing that contains relevant information of the type of the error. It also anyway works when the Django debug setting is true; otherwise the database lock exception seems to result in 500 status code and the ajax fail parameters don't really have any other useful information to determine what was the cause of the error.

piehei:
Thanks!

exercise/static/exercise/chapter.js
raphendyr: Just check that if the server doesn't respond with 200, then retry X times. if server doesn't responding in acceptable time frame, then show error in the exercise.

raphendyr:
Could you rebase this on top of current master and possibly squash commits that don't need to be separated. Later part doesn't need as much effort due to pending rewrite of the A+ layer. Rebase is required so we can merge this to current version.

raphendyr:
I'll review this soonish and I'll merge it then.

exercise/static/exercise/chapter.js
```diff
@@ -9,14 +9,15 @@
...
+		active_element_attr: "data-aplus-active-element",
 		loading_selector: "#loading-indicator",
 		quiz_success_selector: "#quiz-success",
 		message_attr: {
 			load: "data-msg-load",
 			submit: "data-msg-submit",
 			error: "data-msg-error"
 		},
-		modal_selector: "#page-modal",
+		modal_selector: "#page-modal"
```
raphendyr: Where has the comma gone?

exercise/static/exercise/chapter.js
```diff
@@ -44,8 +46,12 @@
 			this.messages = this.readMessages();
 			console.log(this.loader, this.messages);
 			this.quizSuccess = $(this.settings.quiz_success_selector);
+
+			// do not include active element inputs to exercise groups
+			this.element.find("[" + this.settings.active_element_attr +	"='in']").aplusExercise(this, {input: true});
+
 			this.exercises = this.element
-				.find("[" + this.settings.exercise_url_attr + "]")
+				.find("[" + this.settings.exercise_url_attr + "]")
```
raphendyr: huh?

exercise/static/exercise/chapter.js
```diff
@@ -125,7 +130,7 @@
  */
 ;(function($, window, document, undefined) {
 	"use strict";
-
+
```
raphendyr: ?

raphendyr:
There is quite many times some white space in the end of the line. This is handy when your ide is not configured correctly: `sed -i 's/\s*$//' <file>`.

roosapi:
Thanks! Trying to get these done on Friday at the latest.

teemulehtinen:
add to whitelist

teemulehtinen:
Looks good to me, @raphendyr will do the honours.

Mankro:
@raphendyr are you waiting for some changes before you could merge this PR? I think that @roosapi is not aware that more changes could be expected here. I fear that this feature will be in trouble if this PR is not merged before other major changes arrive during this spring.

exercise/static/exercise/poll.js
```diff
@@ -63,7 +63,13 @@
 		},
 
 		ready: function() {
-			this.element.hide();
+			//this.element.hide();
+			
+			// For active elements the element to which the poll plugin is attached remains the same, so to
+			// be able to submit the same form several times the plugin data needs to be removed when the 
+			// evaluation and polling is finished.
+			if ($.data(this.element.context, "plugin_" + pluginName)) $.removeData(this.element.context, "plugin_" + pluginName);
```
Mankro: `.context` on jQuery objects is deprecated. Do you mean `this.element.get(0)` or is the context referring to the `document`? Originally, the data is set on the DOM element and since you are removing it, I assume that the document should not be used here.
https://api.jquery.com/context/
PS: jQuery has a higher level `.data` method too, so is it necessary to use the low level `$.data`? I know it was here previously too.

exercise/static/exercise/chapter.js
```diff
@@ -168,56 +179,168 @@
...
+			// Active element can have height settings in the A+ exercise div that need to be
+			// attached to correct DOM-elements before setting the exercise container div height to auto
+			var cur_height = exercise.element.css('height');
+			if (exercise.active_element) {
+				if (exercise.settings.input) {
+					$("#" + exercise.chapterID + " textarea").css("height", cur_height);
+				} else {
+					if (typeof $("#" + exercise.chapterID).data("scale") != "undefined") {
+						var cont_height = $(exercise.settings.ae_result_selector, exercise.element)[0].scrollHeight;
```
Mankro: I think I had a case in which the line 330 with `.scrollHeight` crashed due to an undefined object when the connection to the exercise service failed. Did you check what happens when the response from the exercise service does not contain the normally expected HTML structure due to an error?

raphendyr:
@Mankro I'm just waiting for time window to read these changes. Hopefully next week.

raphendyr:
retest this please

exercise/static/exercise/poll.js
```diff
@@ -73,25 +79,39 @@
...
+				var message = "There was an error while evaluating the element."
+				if (messageType == "timeout") {
+				 message = "Evaluation was timed out.";
+				}
+				var res_elem = this.element.find(".ae_result").text(message);
+				if (res_elem.height() === 0) res_elem.height("auto");
+				if ($.data(this.element.context, "plugin_" + pluginName)) {
+					$.removeData(this.element.context, "plugin_" + pluginName);
+				}
+			} else {
```
raphendyr: `else if (...`

exercise/static/exercise/poll.js
```diff
@@ -73,25 +79,39 @@
...
 
 		message: function(messageType) {
+
 			this.element.removeClass("active").find(this.settings.message_selector)
-				.text(this.element.attr(this.settings.message_attr[messageType]));
-			if (messageType == "error") {
-				this.element.addClass("progress-bar-danger");
-			}
+			.text(this.element.attr(this.settings.message_attr[messageType]));
+			if (this.element.data("aplus-active-element")) {
```
raphendyr: Could this be done with `this.element.attr(this.settings.active_element_attr))`?

exercise/static/exercise/poll.js
```diff
@@ -73,25 +79,39 @@
...
+				var message = "There was an error while evaluating the element."
+				if (messageType == "timeout") {
+				 message = "Evaluation was timed out.";
```
raphendyr: Has space instead of tab..

exercise/static/exercise/chapter.js
```diff
@@ -234,40 +357,207 @@
...
+					feedback.append(exercise.chapter.messages.error);
+					exercise.updateOutput(feedback);
+				}
+			}).done(function (data) {
+				/* Bad fix for database locked problem when saving submitted files:
+					 test if the data contains message that the submission was not saved and resubmit
+				*/
+				retry = retry || 0;
+				if ($(data).find("div:contains('The submission could not be saved for some reason')").length > 0 && retry < 5) {
+					console.log("Submission not saved: trying submitAjax again in 100ms");
```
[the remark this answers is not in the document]
roosapi: When one input triggers and submits several active elements, the multiple submits sometimes caused a "database locked" error in the SQLite database, and if there are any file submit exercises that fail, the submitted files are not stored because of the error. (Apparently the same database locked error returns an error code if the submission does not include files added by  the user.)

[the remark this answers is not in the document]
raphendyr: Does this happen only with SQLite and not with real DB? Are you running multiple django processes with SQLite?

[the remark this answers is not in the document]
roosapi: Yes it does seem to be a problem with SQLite only, though it hasn't been tested that extensively. The setup has been pretty much out-of-the-box A+ & mooc-grader with the addition of active element JS that can send multiple AJAX requests to the exercise urls (still single request per exercise) when the user clicks submit.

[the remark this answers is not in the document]
raphendyr: Apparently django runserver use threads, so when running with sqlite it can try to open the file at the same time. Thus, we need to disable threads with sqlite. That can be done with runserver flag `--nothreading`. That at least should help with the case.

If you could run your test environment with that flag on (or `command: "python3 manage.py runserver 0.0.0.0:8000 --nothreading"` in docker-compose.yml) and inform if this error still appears (or doesn't seem to). If that flag fixes it, then I think it could be turned on by default in our test containers (run-aplus-front).

Any case, you can remove this part of the JS as it's not required in production.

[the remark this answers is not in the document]
roosapi: Everything seems to work fine with the flag, thanks!

PR #312, opened 2018-01-15
add exercise title before block columns

JingzhouOu:
Signed-off-by: Ou Jingzhou <ouj1@cs-245.org.aalto.fi>


Plustest:

exercise/templates/exercise/exercise_base.html
```diff
@@ -30,6 +30,10 @@
 {% endblock %}
 {% endblock %}
 
+{% block exercise_title %}
+<h3 class = "exercise_title">{{ current.name }}</h3>
```
ttsirkia: There shoudn't  be a space between the attribute name, equals sign and the value.

exercise/templates/exercise/exercise_base.html
```diff
@@ -30,6 +30,12 @@
 {% endblock %}
 {% endblock %}
 
+{% block exercise_title %}
+<div>
```
Mankro: Is a `div` around `h3` really needed?

Mankro:
What is the purpose of removing the title from the mooc-grader template and adding it here? In the mooc-grader, the implementation of the exercise may freely define the template and override any content, but now A+ would force a title on even if an exercise in past had the possibility to remove the title.

Furthermore, this affects the Moodle frontend so please do not push this to production before the summer or else there would be no exercise titles in Moodle.

raphendyr:
This is fixing issue around where title is displayed on chapters. It's logically better this way.

@Mankro can you create similar change to https://github.com/Aalto-LeTech/moodle-mod_astra ? Next production update **should** happen on start of the summer (at end of period IV or V).

ps. This PR is related to https://github.com/Aalto-LeTech/mooc-grader/pull/16

Relevant screenshots for the change:
screenshot_20180115_184553
o1_exercise_title

Mankro:
@raphendyr yes OK, no problem as long as we do not change the production version during this spring

PR #313, opened 2018-01-16
prevent double clicking submit of forms

JingzhouOu:
all forms prevented

Signed-off-by: Ou Jingzhou <ouj1@cs-245.org.aalto.fi>

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -624,13 +624,15 @@ $(function() {
     } else {
       // Mark it so that the next submit can be ignored
       $form.data('submitted', true);
+      //disable the button
+      $form.find('[type="submit"]').prop('disabled', 'disabled')
     }
   });
 
   // Keep chainability
   return this;
 };
-
-$('form').preventDoubleSubmission();
+//prevent double submisssion of exercise form
+$('#exercise-page-content').preventDoubleSubmission();
 
 });
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -606,3 +606,31 @@ $.fn.highlightCode = function(options) {
         });
     };
 })(jQuery, window, document);
+
+
+/**
+ * prevent double submit of forms
+ */
+
+$(function() {
+
+    $.fn.preventDoubleSubmission = function() {
+  $(this).on('submit',function(e){
+    var $form = $(this);
+
+    if ($form.data('submitted') === true) {
+      // Previously submitted - don't submit again
+      e.preventDefault();
+    } else {
+      // Mark it so that the next submit can be ignored
+      $form.data('submitted', true);
+    }
+  });
+
+  // Keep chainability
+  return this;
+};
+
+$('form').preventDoubleSubmission();
+
+});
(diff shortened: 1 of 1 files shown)


Plustest:

ttsirkia:
How does this behave if I want to resubmit after the first submission is handled? If I understood correctly, now this would require reload in order to use the same submission box again. I think that the resubmission should be blocked only when the submission is done but before the feedback is received. There can also be forms that might be loaded after DOM has been constructed (ie. loading a form with AJAX).

teemulehtinen:
add to whitelist

JingzhouOu:
Yes, you are right. This is to solve the double submission of exercise forms. Maybe we can give class attributes to the exercise form.

assets/js/aplus.js
```diff
@@ -606,3 +606,31 @@ $.fn.highlightCode = function(options) {
...
+      // Mark it so that the next submit can be ignored
+      $form.data('submitted', true);
+    }
+  });
+
+  // Keep chainability
+  return this;
+};
+
+$('form').preventDoubleSubmission();
```
raphendyr: This could hook to `#exercise-page-content form`. On chapter page, this would be problematic, but there we open modal dialog when submit is clicked. So I think it's good enough to do this kind of limiting only on old style exercise pages. (this the id in jquery selector).

assets/js/aplus.js
```diff
@@ -606,3 +606,31 @@ $.fn.highlightCode = function(options) {
...
+    $.fn.preventDoubleSubmission = function() {
+  $(this).on('submit',function(e){
+    var $form = $(this);
+
+    if ($form.data('submitted') === true) {
+      // Previously submitted - don't submit again
+      e.preventDefault();
+    } else {
+      // Mark it so that the next submit can be ignored
+      $form.data('submitted', true);
```
raphendyr: Disabling the submit button might be a bit better user experience. So something like `$form.find('[type="submit"]').prop('disabled', 'disabled')`.

assets/js/aplus.js
```diff
@@ -606,3 +606,33 @@ $.fn.highlightCode = function(options) {
...
+      //disable the button
+      $form.find('[type="submit"]').prop('disabled', 'disabled')
+    }
+  });
+
+  // Keep chainability
+  return this;
+};
+//prevent double submisssion of exercise form
+$('#exercise-page-content').preventDoubleSubmission();
```
raphendyr: This doesn't match to the form. #exercise-page-content form should do it.

assets/js/aplus.js
```diff
@@ -606,3 +606,33 @@ $.fn.highlightCode = function(options) {
...
 })(jQuery, window, document);
+
+
+/**
+ * prevent double submit of forms
+ */
+
+$(function() {
+
+    $.fn.preventDoubleSubmission = function() {
```
raphendyr: I don't think this needs to be jquery plugin. simple function could be enough.

assets/js/aplus.js
raphendyr: This also needs to be inside `$(function() {...})` so it's not executed before page is fully loaded.

raphendyr:
retest this please

raphendyr:
Squashed, rebased and merged in acd528b5e2a71fbbf3b134313dbe8134162a7b85

Issue #311, opened 2018-01-15
Multilingual support for a single course


raphendyr:
Requires changes to https://github.com/Aalto-LeTech/a-plus-rst-tools and https://github.com/Aalto-LeTech/mooc-grader .

So, add support for course material in multiple languages (e.g. English and Finnish). This requires update on `aplus-json` returned from gitmanager in mooc-grader in addition to support in rst tools. This is master issue to track whole progress of this feature.

- [x] RST-tools can build yaml configurations for gitmanager and mooc-grader
- [x] Mooc-grader outputs urls and titles per language in aplus-json
- [x] Mooc-grader can render exercise in different languages.
- [x] Aplus can store titles and other text fields in multiple languages
- [x] Aplus renders these text fields in currently selected language
- [x] Aplus contains easy buttons to change the language

raphendyr:
Example aplus.json with localized texts for the UI.

```json
{
  "lang": [
    "fi",
    "en"
  ],
  "name": {
    "fi": "Ohjelmointi 1",
    "en": "Programming 1"
  },
  "start": "2017-08-21 14:00",
  "end": "2018-01-01 12:00",
  "categories": {
    "chapter": {
      "name": {
        "fi": "Luku",
        "en": "Chapter"
      },
      "status": "nototal"
    },
    "submit": {
      "name": {
        "fi": "Pisteytetty",
        "en": "Graded"
	  }
    },
    "feedback": {
      "name": {
        "fi": "Palaute",
        "en": "Feedback"
      },
      "status": "nototal"
    }
  },
  "modules": [
    {
      "key": "k01",
      "name": {
        "fi": "Kierros 1",
        "en": "Week 1"
      },
      "status": "ready",
      "open": "2017-08-21 14:00",
      "close": "2017-09-20 12:00",
      "late_close": "2018-04-01 12:00",
      "late_penalty": 1.0,
      "children": [
        {
          "key": "osa01",
          "name": {
            "fi": "Luku 1.1: Kurssin esittely",
            "en": "Chapter 1.1: Introduction to O1"
          },
          "status": "ready",
          "category": "chapter",
          "url": {
            "fi": "http://localhost:8080/static/o1_2018/w01/ch01_fi.html",
            "en": "http://localhost:8080/static/o1_2018/w01/ch01_en.html"
          },
          "use_wide_column": true,
          "children": [
            {
              "key": "k01_osa01_1",
              "title": {
[...]

PR #320, opened 2018-03-12
Multilingual support

madaian:
Based on #311 
- course can have multiple languages
- can store titles and URLs in multiple languages in `|en:Chapter|fi:Luku|` format
- renders text fields in currently selected language
- course menu has buttons to change the language

URLs are not validated, as they are handled as regular text


Plustest:

course/models.py
```diff
@@ -356,7 +356,15 @@ def clean(self):
...
             errors['archive_time'] = _("Archive time must be later than lifesupport time.")
-        if not self.is_valid_language(self.language):
+        if self.language.startswith("|"):
+            langs = list(filter(None, self.language.split("|"))) # remove pipes & empty strings
+            for lang in langs:
+                if not self.is_valid_language(lang):
+                    if "language" in errors:
+                        errors['language'] = "%s%s" % (errors['language'], (", " + lang))
+                    else:
+                        errors['language'] = _("Language code(s) missing from settings: " + lang)
```
raphendyr: This doesn't go well in translation. Text inside function `_()` is translated, but it's now different based on language. `_("plaa plaa: ") + lang` should be good enough here.

course/templatetags/course.py
```diff
@@ -38,6 +39,32 @@ def group_select(context):
...
+def parse_localization(entry):
+    """
+    Picks the currently selected language's value from
+    |lang:value|lang:value| -format text.
+    """
+    text = ""
+    if isinstance(entry, str):
+        text = entry
+    else:
+        text = str(entry) # if not a string, assume entry is a model
```
raphendyr: `text  = entry if isinstance(entry, str) else str(entry)`. Though, `str(a_string_object)` would work too.

Other solution would be setting entry if it's not str:
`if not isinstance(entry, str): entry = str(entry)`

course/viewbase.py
```diff
@@ -66,7 +67,13 @@ def get_resource_objects(self):
 
             # Apply course instance language.
             if self.instance.language:
-                translation.activate(self.instance.language)
+                lang = self.instance.language
+                if lang.startswith("|"):
+                    active = get_language()
+                    if "|" + active + "|" in lang:
+                        translation.activate(active)
```
raphendyr: What happens if active language is not in the list?

edit_course/operations/configure.py
```diff
@@ -239,8 +251,18 @@ def configure_content(instance, url):
...
+    if "lang" in config:
+        langs = config["lang"]
+        if isinstance(langs, list):
+            formatted = ""
+            for lang in langs:
+                if instance.is_valid_language(lang):
+                    formatted += ("|" + lang)
+            if len(formatted) > 1:
+                formatted += "|"
+                instance.language = str(formatted)
```
raphendyr: Would ```langs = [lang for lang in langs if instance.is_valid_language(lang)]
if langs:
	instance.language = "|{}|".format("|".join(langs))``` be a bit cleaner?

edit_course/operations/configure.py
```diff
@@ -68,6 +68,18 @@ def parse_float(value, errors):
...
+def format_localization(element):
+    """
+    Parse localised elements into |lang:val|lang:val| -format strings
+    """
+    if isinstance(element, dict):
+        concat = ""
+        for pair in element.items():
+            concat += ("|" + ":".join(pair))
+        concat += "|"
+        return concat
```
raphendyr: maybe?

```
strings = ("{}:{}".format(k, v) for k, v in element,items())
return "|{}|".format("|".join(strings))
```

edit_course/operations/configure.py
```diff
@@ -276,7 +298,7 @@ def configure_content(instance, url):
         if not "name" in c:
             errors.append(_("Category requires a name."))
             continue
-        category, flag = instance.categories.get_or_create(name=str(c["name"]))
+        category, flag = instance.categories.get_or_create(name=format_localization(c["name"]))
```
raphendyr: This requires more looking into. Technically categories should have `key/slug` and `display string` instead of them both being the same field. We need to look into this a bit more to see if this works as expected.

exercise/templates/exercise/exercise_base.html
```diff
@@ -4,7 +4,7 @@
 {% load exercise %}
 {% load apps %}
 
-{% block title %}{{ current.name }} | {{ block.super }}{% endblock %}
+{% block title %}{{ current.name|parse_localization }} | {{ block.super }}{% endblock %}
```
raphendyr: Need to use translation block with variable and run the filter there

course/viewbase.py
madaian: Fixed to fall back to the first language discovered in the course's settings so that the multilingual fields won't just disappear.

exercise/templates/exercise/exercise_base.html
madaian: I don't understand what the problem here is, it seems to be working fine?

raphendyr: I see. Django works interestingly on that matter, so it really does work. Django strores everything inside `{{}}` as the "key" (think as variable) and uses that with `%` python format to expand the variables in translation string. Only downside from current is that the translation contains `current.name|parse_localization` instead of `name` for example. Thus in real life, it could be obscuring detail for the translator.

Relevant parts in django source:
 * store vars: https://github.com/django/django/blob/master/django/templatetags/i18n.py#L119
 * string format: https://github.com/django/django/blob/master/django/templatetags/i18n.py#L164

So in short, current code is fine and ok.

PR #321, opened 2018-03-13
Replace tag.text_color with tag.font_color

jrp6:
Fixes tag checkboxes being hard to read if the background is dark.
Turns out it was caused by a trivial typo.


Plustest:

PR #322, opened 2018-03-20
Replace tag.text_color with tag.font_color in analytics

jrp6:
Found the same bug in another place. There should be no more, at least
according to grep.


Plustest:

PR #323, opened 2018-03-20
Navbar dropdown

madaian:
Combines the user profile and logout links under a shared dropdown

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -88,17 +88,21 @@
                             {% if user.is_authenticated %}
                             {% group_select %}
                             {% notification_menu %}
-                            <li role="presentation" class="menu-profile">
-                                <a href="{% url 'profile' %}">
-                                    <span class="glyphicon glyphicon-user" aria-hidden="true"></span>
-                                    {{ user.username }}
-                                </a>
-                            </li>
-                            <li role="presentation">
-                                <a href="{% url 'logout' %}">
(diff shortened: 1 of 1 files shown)


Plustest:

templates/base.html
```diff
@@ -88,17 +88,21 @@
...
-                                    <span class="glyphicon glyphicon-user" aria-hidden="true"></span>
-                                    {{ user.username }}
-                                </a>
-                            </li>
-                            <li role="presentation">
-                                <a href="{% url 'logout' %}">
+                            <li class="dropdown">
+                              <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
+                                <span class="glyphicon glyphicon-user" aria-hidden="true"></span>
+                                {{ user.first_name }} {{ user.last_name }} <span class="caret"></span>
```
raphendyr: Maybe one of these would be cleaner: `user.get_full_name` (in django AbstractUser) or `user.userprofile.shortname` (in a-plus/userprofile/models.py).

raphendyr: rofl.. the `shortname` is currently used only in some tests... so maybe go with the former.

raphendyr: I don't know what I'm thinking. It's probably better to have this here. so I'll just merge this.

PR #324, opened 2018-03-28
Menu highlighting

madaian:
Highlights the correct menu entry for an open page.
Removes selection based on data-type-tag, and instead selects those menu elements whose target path matches the current URL.
Tested on the live O1 2017 course with Chrome dev tools, so it should work as intended with chapter/excersise-type menu items.


Plustest:

raphendyr:
add to whitelist

PR #327, opened 2018-04-10
Django colortag

jrp6:
Reimplement UserTag using Aalto-LeTech/django-colortag


Plustest:

jrp6:
A+ could be simplified by rendering the tags using the templatetags from django_colortag. However, it doesn't use font_color so tags with light colour background are difficult to read, so I'm staying with rendering the tags in A+ until that's fixed

raphendyr:
add to whitelist

raphendyr:
Check this line https://github.com/Aalto-LeTech/django-colortag/blob/master/django_colortag/widgets.py#L22

It adds correct class based on what font color should be used.

PR #329, opened 2018-04-16
Render MathJax content in chapters, including modals

Mankro:
Chapter pages load exercises and feedback modal dialogs via AJAX. MathJax does not render math in the page after modifications after the initial page load unless it is explicitly called. This PR adds the necessary calls to MathJax in the JS code so that MathJax works automatically in all chapter content (chapter text, exercises, and feedback) as long as the MathJax library is enabled in the course.


Plustest:

PR #330, opened 2018-04-17
Expose exercise template URLs in API

matiaslindgren:


raphendyr:
**NOTE:** this was PR to production as this fix is needed there by radar.

Issue #336, opened 2018-06-06
Make TOC view clearer


piehei:
TOC version of the course index is a bit hard to read when you've got lots of content.

I suggest either of these two layouts to make it easier for the reader:

![screenshot 2018-06-06 at 20 43 00](https://user-images.githubusercontent.com/8404555/41055575-5f1b5d32-69ca-11e8-83d7-f490a3cb6b2f.png)
![screenshot 2018-06-06 at 20 41 42](https://user-images.githubusercontent.com/8404555/41055577-5f415154-69ca-11e8-8782-8b5041e7accb.png)

Original:

![screenshot 2018-06-06 at 20 44 56](https://user-images.githubusercontent.com/8404555/41055718-aded7076-69ca-11e8-9e50-62278696631d.png)

What do you think? Any other versions?

raphendyr:
Huargh. I see. For the next level.

I would say the second one as the bold subtitles in the first might take the focus from the module title.

PR #337, opened 2018-06-07
Indentation for third level entries in TOC view

piehei:
This implements #336.

Any comments?

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -236,9 +236,6 @@ html, body {
 .toc li {
     list-style: none;
 }
-.third-level-entry {
-  margin-left: 25px;
-}
 
 /* index thumbnails */
 .course-thumbnail {
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -236,6 +236,9 @@ html, body {
 .toc li {
     list-style: none;
 }
+.third-level-entry {
+  margin-left: 25px;
+}
 
 /* index thumbnails */
 .course-thumbnail {
--- a/exercise/templates/exercise/_children.html
+++ b/exercise/templates/exercise/_children.html
@@ -11,8 +11,7 @@
 {% endif %}
 {% elif entry|is_listed %}
 
-<li{% if entry.number|length > 4 %} class="third-level-entry"{% endif %}>
-
+<li>
   {% if accessible %}
   {% if entry.is_empty %}
   {{ entry.name|parse_localization }}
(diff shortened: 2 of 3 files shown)


raphendyr:
That looks a big ugly hack. Any better solutions?

piehei:
Yes it does haha. 

I'm not too sure about the internals so just checking the entry's number was an easy way to implement this without touching the view logic itself.

Would you find it better to add something like indent: True to each entry object and do it in the view?

jrp6:
Isn't the 'correct' way to do it to have a nested list, like this: https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/HTML_text_fundamentals#Nesting_lists

teemulehtinen:
I believe nested lists might/should already be produced. Maybe there is styling issue for ul.toc? Or ul.toc > li > ul.toc etc..

jrp6:
`_children.html` looks to me like it _should_ produce a nested list, but what I'm seeing in the inspector is
```
  <li>
    <h3>
      <a href="/def/current/k01/" class="">1. Kierros 1</a>
    </h3>
    <h4>
      <small>21.8.2017 maanantai 14:00 – 20.9.2017 keskiviikko 12:00</small>
    </h4>
</li><li>
  <a href="/def/current/k01/osa01/" class="">1.1 Luku 1.1: Kurssin esittely</a>
</li>
etc
```

raphendyr:
I'm not sure if we have any courses with third level to test this with.  @jrp example only has first and second levels.

jrp:
Hi all, FYI, that sent an update to @jrp (me) when it should have gone to @jrp6 -- I've never even heard of your project. 

Good luck with the nesting.

raphendyr:
Huargh... Didn't even look. Expected first tab to be person in thread instead of first match... I'm missing irssi times.

raphendyr:
Yep. A+ rendering of that html should be changed. 1.1.1. should be child for 1.1.

And yes, that is 3 levels (1.1.1.). First level is always the title, which contains a list of subtitles (1.1. second level and 1.1.1. for third). Technically, we can have as many levels, so the template should be recursive.

teemulehtinen:
There is a template bug in _children.html
`{% if entry.down == 'level' %}
{% if entry.down %}
<ul class="toc">
{% elif entry.up %}
</ul>
{% endif %}
{% elif entry|is_listed %}`
The first if should be entry.type == 'level'

piehei:
@teemulehtinen's suggestion fixes this. I already implemented it recursively but this is better of course.

screenshot_20180608_170647

raphendyr:
I didn't like those extra newlines :)  If only you would have fixed the indentation instead..

PR #339, opened 2018-06-11
LTI 1.1

Mankro:
LTI Tool Providers may now post grades back to A+ using the Basic Outcomes Service.


Plustest:

raphendyr:
Add to whitelist

external_services/api/parsers.py
```diff
@@ -0,0 +1,106 @@
...
+
+def parse_sourced_id(sourced_id_str):
+    '''Parse a sourcedId value from an LTI 1.1 Outcomes Service request.
+    Return a tuple (exercise, user_profile). Either value may be None if
+    the instance is not found in the database.
+    '''
+    parts = sourced_id_str.split('-', 1)
+    if len(parts) != 2:
+        return (None, None)
+    exercise_id, user_token = parts[0], parts[1]
```
raphendyr:
```
exercise_id, _, user_token = source_id_str.partition('-')
if not exercise_id or not user_token:
  return (None, None)
```
Maybe a bit more clean?

external_services/api/views.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class LTIExerciseBasicOutcomesView(APIView):
+    '''View that responds to LTI 1.1 Basic Outcomes Service requests that
+    are posted by the Tool Provider to A+ in order to return the scores of submissions.
+    '''
+    
+    parser_classes = (LTIOutcomeXMLParser,)
+    authentication_classes = (OAuthBodyHashAuthentication,)
+    permission_classes = (IsAuthenticated,)
```
raphendyr: Hmm.. So, any authenticated user that has authenticated using OAuthBodyHashAuthentication has access to write this resource? Thus, student using SessionAuthentication shouldn't be able to post here.

So, if anyone can get access to LTI secret, they can add submissions to any user to any exercise.

I think we should give some exercise specific secret when launching LTI service, so the service can only update that specific exercise with the secret.

raphendyr: IP check (could be that is not be possible for LTI) - https://github.com/Aalto-LeTech/a-plus/blob/master/lib/api/authentication/grader.py#L28

raphendyr: > I think we should give some exercise specific secret when launching LTI service, so the service can only update that specific exercise with the secret.

Now I remember, this was waiting for Aplus v2, where we will pre-create the submission, so we can have the secret for updating it. Currently, we don't have this secret to share with the grader.

Mankro: The LTI security is based on OAuth1 signature and body hash, which are based on the shared secret key. If the secret key is leaked, it should be changed. Knowing the secret allows one to post new submissions, indeed.

The LTI specification does not guarantee that the Tool Provider (TP) launch URL would correspond to the IP source address of the Basic Outcomes Service request when the TP posts grades back to A+. Therefore, I would not check the IP at all.

raphendyr: Hmm. Yeah. So we go with that the LTI service is restricted to update submissions in exercises where it is linked (I forgot that restriction).

I Still would prefer that this endpoint would have permission that enforces the user is lti service.

external_services/api/authentication.py
```diff
@@ -0,0 +1,124 @@
...
+        headers=headers, exclude_oauth_signature=False)
+    # collect_parameters returns a list of key-value pairs
+    req_oauth_params = [] # the oauth_signature must be separated from the other parameters
+    oauth_signature = None
+    for key, value in all_req_oauth_params:
+        if key == 'oauth_signature':
+            oauth_signature = value
+        else:
+            req_oauth_params.append((key, value))
+    req_oauth_params_dict = dict(req_oauth_params)
```
raphendyr: Would it be cleaner without for loop and with:

```
req_oauth_params_dict = dict(req_oauth_params)
oauth_signature = req_oauth_params_dict.pop('oauth_signature')
```

?

external_services/api/authentication.py
```diff
@@ -0,0 +1,124 @@
...
+        return False, 'oauth_timestamp has expired'
+    
+    # check OAuth nonce: The nonce value MUST be unique across all requests with
+    # the same timestamp, client credentials, and token combinations.
+    # Previously seen nonces are kept in the cache for a few minutes
+    # (the duration must match the accepted timestamp age).
+    nonce = req_oauth_params_dict.get('oauth_nonce')
+    if not nonce:
+        return False, 'oauth_nonce missing'
+    nonce_cache = OAuthNonceCache(nonce, req_oauth_params_dict.get('oauth_timestamp'), client_secret)
```
raphendyr: why `req_oauth_params_dict.get('oauth_timestamp')` and not `timestamp`?

external_services/api/authentication.py
```diff
@@ -0,0 +1,124 @@
...
+    if nonce_cache.nonce_used():
+        return False, 'oauth_nonce has been used'
+    nonce_cache.add_used_nonce()
```
raphendyr: There is possibility for race condition here. nonce_cache should implement only single method to set and verify the nonce. Django cache backend has method [add](https://docs.djangoproject.com/en/2.0/topics/cache/#basic-usage) and it works for this purpose better than get + set.

external_services/api/parsers.py
```diff
@@ -0,0 +1,106 @@
...
+            # do not set None values to the data
+            # the XML API returns None if the query finds nothing
+            if value is not None:
+                data[key] = value
+        
+        # the stream can be read only once, but we need to both compute
+        # the hash of the data and parse the data as XML
+        body_bytes = stream.read()
+        # b64encode returns a byte string so it is decoded to a normal string
+        body_hash = base64.b64encode(hashlib.sha1(body_bytes).digest()).decode('ASCII')
```
raphendyr: Would `sha1().hexdigest()` work here?

external_services/api/authentication.py
Mankro: `req_oauth_params_dict.get('oauth_timestamp')` is a string like `"436372524"` and `timestamp` is a datetime object. I want the string format here.

external_services/api/authentication.py
Mankro: sure

external_services/api/parsers.py
Mankro: yup, base64 is needed for oauth1

external_services/api/authentication.py
raphendyr: If this oauth_timestamp was rendered by us and then signed, then there shouldn't be any problem. I was afraid that non sanitized timestamp could fool nonce cache.

Mankro: The oauth_signature protects the integrity of the oauth headers. The format of oauth_timestamp is checked earlier in the code. Are you concerned that attackers could fill the cache with trash nonces even without having a valid signature?

external_services/lti.py
```diff
@@ -75,27 +57,55 @@ def __init__(self, service, user, instance, host, title, context_id=None, link_i
             "context_label": course.code,
 
             "launch_presentation_locale": get_language(),
+            "launch_presentation_document_target": "window",
+            "launch_presentation_return_url": request.scheme + '://' + request.get_host() + instance.get_absolute_url(),
 
-            "tool_consumer_instance_guid": host + "/aplus",
+            "tool_consumer_instance_guid": request.get_host() + "/aplus",
```
raphendyr: This will need to be refactored to come from django settings in the future or from lti settings. At least in local container environments it would be handy to pass different host for different services (how to call back).

external_services/lti.py
```diff
@@ -75,27 +57,55 @@ def __init__(self, service, user, instance, host, title, context_id=None, link_i
...
             "tool_consumer_instance_name": "A+ LMS",
         })
 
         if service.api_access:
             self.parameters.update({
-                # FIXME: we need request or full host with protocol here!
-                'custom_context_api': '//' + host + api_reverse("course-detail", kwargs={'course_id': instance.id}),
+                'custom_context_api': request.scheme + '://' + request.get_host() + api_reverse("course-detail", kwargs={'course_id': instance.id}),
```
raphendyr: Above comment applies here too :)

external_services/lti.py
```diff
@@ -126,26 +136,32 @@ def sign_get_query(self, url=None):
...
+            student_id, given_name, family_name, full_name, email = self.user_info(self.course_instance, user)
+            d = {
+                'user': student_id,
+                'given_name': given_name,
+                'family_name': family_name,
+                'full_name': full_name,
+                'email': email,
+            }
+            if not self.service.is_anonymous:
+                d['student_id'] = self.true_student_id(profile)
```
raphendyr: This seems like a hack for services that expect real student id instead of internal reference. I think we should check what services use this (I presume rubyric is the only one) and we should change this as it's not consistent with other places.  I think we should deliver real student id in differently named field if possible (and thus we wouldn't over write aplus student id).

external_services/lti.py
```diff
@@ -126,26 +136,32 @@ def sign_get_query(self, url=None):
...
+        super().__init__(service, user, instance, request, title, context_id, link_id, parameters, exercise)
 
     def true_student_id(self, profile):
         return profile.student_id or "A{:d}".format(profile.id)
 
     def group_json(self, profiles):
-        data = [];
+        data = []
         for profile in profiles:
             user = profile.user
```
raphendyr: This could be moved to next line.

userprofile/models.py
```diff
@@ -140,3 +140,13 @@ def _course_instance(self):
...
         return self._course_instance.course
+
+
+class LTIServiceUser(GraderUser):
+    def __init__(self, submission=None, exercise=None, lti_service=None, **kwargs):
+        self._lti_service = lti_service
+        super().__init__(submission=submission, exercise=exercise, **kwargs)
+
+    @property
+    def lti_service(self):
```
raphendyr: No need to have property for this. Let's change it later if needed. We any case can't protect against non accidental writes.

external_services/api/authentication.py
raphendyr: Nope. Just initially though it maybe could have been possible to fool the cache, but due to signature it shouldn't open any vulnerabilities.

external_services/lti.py
Mankro: tool_consumer_instance_guid is not a callback, though. From the specs:

tool_consumer_instance_guid=lmsng.school.edu
This is a unique identifier for the TC.  A common practice is to use the DNS of the organization or the DNS of the TC instance.  If the organization has multiple TC instances, then the best practice is to prefix the domain name with a locally unique identifier for the TC instance.  In the single-tenancy case, the tool consumer data can be often be derived from the oauth_consumer_key.  In a multi-tenancy case this can be used to differentiate between the multiple tenants within a single installation of a Tool Consumer. This parameter is strongly recommended in systems capable of multi-tenancy.

external_services/lti.py
Mankro: This is a non-standard, custom parameter used to provide the URL to the A+ REST API. The host  could be defined in settings (for easier container setup), but are we doing it now or later?

external_services/lti.py
Mankro: The `if not self.service.is_anonymous:` part is not modifying any existing key in the dictionary. It adds a new key, but only if the service is not anonymous since it is sensitive user data.

external_services/lti.py
Mankro: Where? You can not initialize the list inside the loop.

userprofile/models.py
Mankro: Do you really want to remove the whole property? It causes no harm and it is a natural way to read the lti_service in the public API.

external_services/lti.py
raphendyr: Later. Just commented as on general information.

external_services/lti.py
raphendyr: Yep. and thus this parameter shouldn't be generated from client browser (Host header). But I keep it in mind. Should be changed at some point on later date. Actually, I'll create and issue about it.

external_services/lti.py
raphendyr:
```
user = profile.user
student_id, given_name, family_name, full_name, email = self.user_info(self.course_instance, user)
```

to

```
student_id, given_name, family_name, full_name, email = self.user_info(self.course_instance, profile.user)
```

external_services/lti.py
raphendyr: Damn. I got fooled my self as this has the student_id with different name (user).

It actually would be good idea to rename `student_id` to `user_id` in the code as it's more describing. We are not actually sending student ID's to external services (except here and another custom field).

userprofile/models.py
raphendyr: I mean the property part is not needed. `_lti_service` can be renamed to `lti_service` as the property wrap doesn't bring anything extra.

external_services/lti.py
Mankro: I have renamed `student_id` to `user_id`.

Mankro:
@raphendyr I have responded to all of your suggestions thus far. There is still something to consider:

The inspect submission page allows the teacher to upload an old submission to the exercise service for regrading. However, the LTI protocol does not support that and A+ crashes when the regrade button is pressed. The regrade view could cancel the operation if the exercise is an LTI exercise and warn the user.

raphendyr:
Regrade button can be removed for LTI exercises and optionally replaced with information that this is LTI exercise and so on.

Mankro:
@raphendyr I fixed the inspect page regrade button and the points cache for LTI exercise group submissions (#348)

raphendyr:
I would love one pass of rebase and relevant squashes to the history, but I may accept this any case..

Mankro:
I'm not very familiar with rebasing nor squashing yet, so how would I do that?

raphendyr:
Long answer: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

Short:

I recommend you start by creating new branch so it's trivial to restart the process. You use `git rebase -i <commit/branch>` to select what to do with commits in your history. Reorder commits so that fixes to previous commits are after the actual commit and then those fixes can be squashed (rewrite commit message) or fixups (just change files, keep original message).

Another thing to do (what I referred with rebase) is to rebase your history on top of current origin master. `git rebase origin/master` does that, but then you need to fix all conflicts commit by commit (as you are applying your changes on top of master commit by commit. compare to cherry-picking).

Once the history is good in new local branch, you can force push that to this PR with `git push -f myrepo temp_branch:newlti11`.

Mankro:
Rebasing failed on the first try: it changed the checksums of the commits already in master that occurred between my initial commits and the later fixes. I tried to squash my work into a single commit.

jrp6:
@Mankro what, exactly, was the rebase command you used? It shouldn't usually change any commits already in master. Maybe you did the rebase the wrong way (i.e. you accidentally rebased master on top of your branch, not your branch on top of master)?

Mankro:
@jrp6 `git rebase -i 60a46cb03437ea958c4db7b34954455bf33287e6`
I moved all of my commits to the top and squashed them. I have already once done a merge commit by merging newer commits from upstream master, so perhaps that's why I can not rebase my commits without also rewriting the checksums of the existing master commits.

Mankro:
I pushed a new rebased version and it works correctly. You could merge this PR now.

raphendyr:
FYI: I would probably kept commits:
* fix cache invalidation
* add regrade flag
* implement lti 1.1

PR #344, opened 2018-06-20
Excel csv

jrp6:
Add `excel.csv` format (newlines stripped, semicolons instead of commas)
Add dropdown to select format (csv, excel.csv or json) when downloading submission data


ttsirkia:
Actually, it is not anyhow Excel-specific to use semicolons as the separator character. That is because of the Finnish locale which defines that the list separator (luetteloerotin) is semicolon. When using `en-US` it will be comma.

ihantola:
Is there a specific need for multiple CSV output formats? It should be possible to import files with different separators to Excel, regardless of the locale. For me, defining the separator sounds more like an option of csv, rather than a totally new output format.

lib/api/renderers.py
```diff
@@ -0,0 +1,14 @@
+from rest_framework_csv.renderers import CSVRenderer
+
+class CSVExcelRenderer(CSVRenderer):
+    format = 'excel.csv'
+    writer_opts = { 'delimiter': ';' }
+
+    def remove_newlines(self, x):
```
raphendyr: no need to be member, can be a function.

raphendyr: Should also make sure that v is a string, if that is not the case always.

raphendyr:
After extensive research, it was noticed that there isn't any rational way how excel interpretes csv files. Behavior changed based if the file is opened as is (double click for example) or if the import tool is used. In all of these cases new lines are not handled correctly (within data) and character encoding isn't always honored.

Sadly, all the other tools I know handle [RFC 4180](https://tools.ietf.org/html/rfc4180) CSV files correctly and excel is only difference. In our testing. excel can export different kind of valid CSV files, but it's unable to read them back properly.

In light of this, it's just "best" to support excel friendly format, format that conform standard and is easily readably by all other tools (with multi line support) and in addition, to have link to APi for proper implementation (see tuntiassistenttipooli exporter with pdf generator support, when it's available).

Instead of supporting xlsx export, I think this is the simplest and cleanest solution, without requiring people to install LibreOffice.

exercise/templates/exercise/staff/_submissions_table.html
```diff
@@ -6,16 +6,31 @@
...
+        id="download-data"
+      >
+        {% trans "Download" %} <span class="caret"></span>
+      </button>
+      <ul class="dropdown-menu" aria-labeledby="download-data">
+        {% with formats='csv excel.csv json' %}
+        {% for format in formats.split %}
+        <li>
+          <a href="{% url 'api:course-submissiondata-list' version=2 course_id=instance.id %}?exercise_id={{ exercise.id }}&amp;format={{ format }}">
+            <pre>{{ format }}</pre>
```
raphendyr: I see this as a bit too simple solution. Maybe tuple with formats and nicer strings?

raphendyr:
It probably should be tested if we can support utf-8 in some way and also define `sep=;` in the start (to make sure other locales play well. ref: [stackoverflow](https://stackoverflow.com/questions/20395699/sep-statement-breaks-utf8-bom-in-csv-file-which-is-generated-by-xsl)). Presuming Excel still has a but with BOM, then we could maybe (just maybe) fix it by ordering the info `sep=;\n\ufeffField1;Field2;\r\nData1;Data2`.

jrp6:
* Moved `remove_lines` outside class
* Made delimiter configurable via either `EXCEL_CSV_DEFAULT_DELIMITER` in `settings.py` or the `sep` GET parameter
* Added UTF-8 BOM (`EF BB BF` i.e. `'\uFEFF'.encode('UTF-8')`) to start of file so excel recognizes the file as UTF-8

exercise/templates/exercise/staff/_submissions_table.html
jrp6: Any ideas what would be a nice way to implement this?  I don't know how to (easily) define a list of tuples in django templates, but I don't want to do it in python because this template is used in several places and I want to avoid writing everything twice writing everything twice..

lib/api/renderers.py
jrp6: We check if it's a string or not in `remove_newlines`, which acts as an identity function if not `isinstance(x, str)`

exercise/templates/exercise/staff/_submissions_table.html
raphendyr: https://github.com/django/django/blob/master/django/templatetags/i18n.py#L221

Probably template tag that injects nice list of tuples in render context, like above example from i18n tags.

jrp6:
Implemented @raphendyr suggestion of using a templatetag to insert dict containing the name and the user-readable name of a format

PR #349, opened 2018-07-10
Exercise verbal status

evnurm:
Exercise/submission status is displayed in a textual form if difficulty == "-" and maximum points for the exercise <= 2. Otherwise points are displayed normally. The text to be displayed in the badge is chosen with the following principle:

if max-points == 2, the text is either "Accepted/Hyväksytty" (2), "OK" (1) or "Failed/Hylätty" (0 and a submission has been made) based on the number of points received. If max-points == 1, the texts are "Accepted" or "Failed". In the case that no submissions have been made, the status states "Missing/Puuttuu".

### DIFF
--- a/exercise/templates/exercise/_points_badge.html
+++ b/exercise/templates/exercise/_points_badge.html
@@ -5,28 +5,33 @@
     {{ status|submission_status }}
 </span>
 
-{% elif confirm_the_level %}
-{% if submitted %}
-<span class="badge{% if graded and official %} badge-{% if full_score %}success{% elif passed %}warning{% else %}danger{% endif %}{% endif %} {{ classes }}">
-  {% if not official %}{% trans 'Submitted after deadline' %}{% elif passed %}{% trans "Accepted" %}{% elif graded %}{% trans "Rejected" %}{% else %}{% trans "Submitted" %}{% endif %}
-</span>
-{% elif confirmable_points %}
-<span class="badge badge-danger" data-toggle="tooltip" title="{% trans 'You must submit here to confirm your points. Submissions are accepted even after the deadline.' %}">
-  {% trans "Missing" %}
-</span>
-{% elif exercise_page %}
-<span class="badge">
-  {% trans "Not submitted" %}
-</span>
-{% endif %}
-
 {% else %}
-{% if difficulty %}
-<span class="difficulty">{{ difficulty }}</span>
-{% endif %}
-<span class="badge{% if submitted and official %} {% if full_score %}badge-success{% elif passed %}badge-warning{% else %}badge-danger{% endif %}{% if unconfirmed %} unconfirmed-points{% endif %}{% endif %}{% if classes %} {{ classes }}{% endif %}"
-        data-toggle="tooltip" data-placement="bottom"
-        title="{% if not official %}{% trans 'Submitted after deadline' %}{% elif passed %}{% if required > 0 %}{% trans 'Passed' %}{% endif %}{% elif missing_points %}{% blocktrans with points=required %}{{ points }} points required to pass{% endblocktrans %}{% elif submitted %}{% trans 'Required exercises are not passed' %}{% endif %}">
-    {{ points }} / {{ max }}
+    {% if difficulty and difficulty != "-" and not confirm_the_level %}
+    <span class="difficulty">{{ difficulty }}</span>
+    {% endif %}
+    <span class="badge {% get_points_class %} {% if classes %} {{ classes }}{% endif %}"
+      data-toggle="tooltip"
+      {% if confirm_the_level and confirmable_points %}
+        title="{% trans 'You must submit here to confirm your points. Submissions are accepted even after the deadline.' %}"
+      {% else %}
+        data-placement="bottom"
+        title="
+          {% if not official %}
+            {% trans 'Submitted after deadline' %}
+          {% elif passed %}
+            {% if required > 0 %}
+              {% trans 'Passed' %}
+            {% endif %}
+          {% elif missing_points %}
+            {% if difficulty != "-" %}
+              {% blocktrans with points=required %}{{ points }} points required to pass{% endblocktrans %}
+            {% else %}
+              {% trans 'No submissions' %}
+            {% endif %}
+          {% elif submitted %}
+            {% trans 'Required exercises are not passed' %}
+          {% endif %}"
+      {% endif %}>
+		{% get_points_text %}
 </span>
 {% endif %}
--- a/exercise/templatetags/exercise.py
+++ b/exercise/templatetags/exercise.py
@@ -41,6 +41,7 @@ def _prepare_context(context, student=None):
 def _get_toc(context, student=None):
     points = _prepare_context(context, student)
     context = context.flatten()
+
     context.update({
         'modules': points.modules_flatted(),
         'categories': points.categories(),
@@ -48,7 +49,6 @@ def _get_toc(context, student=None):
         'is_course_staff': context.get('is_course_staff', False),
     })
     return context
-
 
 @register.inclusion_tag("exercise/_user_results.html", takes_context=True)
 def user_results(context, student=None):
@@ -117,6 +117,68 @@ def percent(decimal):
 def submission_status(status):
     return Submission.STATUS[status]
 
+def _get_points_type(context):
+    class_name = ""
+
+    if ((context['confirm_the_level']) or
+        context['difficulty'] == "-" and context['max'] <= 2):
+
+        if (context['full_score'] or
+            (context['passed'] and context['confirm_the_level'])):
+            class_name = "accepted"
+        elif context['passed'] and not context['confirm_the_level']:
+            class_name = "ok"
+        elif ((context['points'] == 0 and context['submitted'])
+            or context['graded']):
+            class_name = "rejected"
+        elif context.get('confirmable_points'):
+            class_name = "missing"
+        else:
+            class_name = "not_submitted"
+    else: # normal exercise without textual status
+        if context['full_score']:
+            class_name = "full"
+        elif context['passed']:
+            class_name = "passed"
+        elif context['submitted']:
+            class_name = "not_passed"
+
+    unconfirmed = not context['confirm_the_level'] and context.get('unconfirmed')
+    return (class_name, unconfirmed)
+"""
+    Used for adding relevant classes to the
+    points_badge to enable textual status.
+"""
+@register.simple_tag(takes_context=True)
+def get_points_class(context):
+    points_type, unconfirmed = _get_points_type(context)
+    options = {
+        'accepted': 'badge-success aplus-points accepted',
+        'ok': 'badge-warning aplus-points ok',
+        'rejected': 'badge-danger aplus-points rejected',
+        'full': 'badge-success',
+        'passed': 'badge-warning',
+        'not_passed': 'badge-danger',
+        'missing': 'badge-danger',
+    }
+    class_name = options.get(points_type, "")
+    if unconfirmed:
+        class_name += " unconfirmed-points"
+    return class_name
+
+# Adds the content of the points_badge.
+@register.simple_tag(takes_context=True)
+def get_points_text(context):
+    points_string = str(context['points']) + " / " + str(context['max'])
+    points_type, _uc = _get_points_type(context)
+    options = {
+        'accepted': _("Accepted"),
+        'ok': _("OK"),
+        'rejected': _("Rejected"),
+        'missing': _("Missing"),
+        'not_submitted': _("Not submitted")
+    }
+    return options.get(points_type, points_string)
 
 def _points_data(obj, classes=None):
     if isinstance(obj, UserExerciseSummary):
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -84,6 +84,7 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     'max_points': entry['max_points'],
                     'points_to_pass': entry['points_to_pass'],
                     'confirm_the_level': entry.get('confirm_the_level', False),
+                    'difficulty': entry['difficulty'],
                     'submission_count': 1, # to fool points badge
                     'points': submission.grade,
                     'graded': submission.is_graded,


exercise/templates/exercise/submission_plain.html
```diff
@@ -52,8 +60,8 @@
 					{% endif %}
 				</td>
 				<td class="points-badge">
-					<span class="hidden">{% points_badge summary %}</span>
-					{% points_badge submission %}
+					<span class="hidden">{% points_badge summary %}</span>
```
raphendyr: white space in end of line?...

exercise/templates/exercise/submission_plain.html
```diff
@@ -52,8 +60,8 @@
 					{% endif %}
 				</td>
 				<td class="points-badge">
-					<span class="hidden">{% points_badge summary %}</span>
-					{% points_badge submission %}
+					<span class="hidden">{% points_badge summary %}</span>				
+					{% points_badge submission  %}
```
raphendyr: extra space?

exercise/templatetags/exercise.py
```diff
@@ -53,6 +54,15 @@ def _get_toc(context, student=None):
 @register.inclusion_tag("exercise/_user_results.html", takes_context=True)
 def user_results(context, student=None):
     values = _get_toc(context, student)
+
+    # Add the difficulty information to all submission to enable
+    # textual status text in points badges in submissions list.
+    for module in values['modules']:
+      for child in module['children']:
```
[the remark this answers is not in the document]
evnurm: I printed out the data while figuring out how to implement this and it seemed to work with the course that I tested it on. It might not work in all cases though.

exercise/templates/exercise/submission_plain.html
evnurm: Made a new commit that should address the whitespace issues you pointed out.

exercise/templates/exercise/_points_badge.html
```diff
@@ -6,27 +6,34 @@
...
-  {% trans "Missing" %}
-</span>
-{% elif exercise_page %}
-<span class="badge">
-  {% trans "Not submitted" %}
-</span>
-{% endif %}
+	{% if submitted %}
+	<span class="badge{% if graded and official %} badge-{% if full_score %}success{% elif passed %}warning{% else %}danger{% endif %}{% endif %} {{ classes }}">
+	  {% if not official %}{% trans 'Submitted after deadline' %}{% elif passed %}{% trans "Accepted" %}{% elif graded %}{% trans "Rejected" %}{% else %}{% trans "Submitted" %}{% endif %}
```
raphendyr: mix of tabs and spaces :(
Should use the same indentation as rest of the file.

exercise/templates/exercise/submission_plain.html
```diff
@@ -20,6 +20,7 @@
 
 		<link rel="stylesheet" href="{% static 'css/main.css' %}" />
 		<link rel="stylesheet" href="{% static 'css/submission.css' %}" />
+
```
raphendyr: This change is not needed :/

exercise/templatetags/exercise.py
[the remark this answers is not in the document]
raphendyr: Thus, my comment. E.g. Tampere is using more levels than our courses do.

exercise/templates/exercise/_points_badge.html
```diff
@@ -6,27 +6,34 @@
...
 
 {% else %}
-{% if difficulty %}
-<span class="difficulty">{{ difficulty }}</span>
-{% endif %}
-<span class="badge{% if submitted and official %} {% if full_score %}badge-success{% elif passed %}badge-warning{% else %}badge-danger{% endif %}{% if unconfirmed %} unconfirmed-points{% endif %}{% endif %}{% if classes %} {{ classes }}{% endif %}"
+	{% if difficulty and difficulty != "-" %}
+	<span class="difficulty">{{ difficulty }}</span>
+	{% endif %}
+<span class="badge{% if submitted and official %} {% if full_score %}badge-success{% elif passed %}badge-warning{% else %}badge-danger{% endif %}{% if unconfirmed %} unconfirmed-points{% endif %}{% endif %}{% if classes %} {{ classes }}{% endif %}
```
raphendyr: Should split the old too long line to multiple lines as you are adding your additions to different lines.
Or in the other words, when your change makes difference to an old line, there is no harm to refactor that line at the same time.

exercise/templates/exercise/_points_badge.html
```diff
@@ -6,27 +6,34 @@
...
+            {% if difficulty == "-" and max <= 2 %} aplus-points
+            {% if points == 1 %}{% if max == 2 %}ok{% elif max == 1%}accepted{% endif %}{% elif points == 2 %}accepted{% elif points == 0 and submitted %}rejected{% endif %}
```
raphendyr: This could be a bit more clear as a templatetag. do the logic in python and just get the class out. Thus, we can render the default text with help of the same function below. For example:
`{% get_points_class %}`

exercise/templates/exercise/_points_badge.html
```diff
@@ -6,27 +6,34 @@
...
+            {% endif %}" 
         data-toggle="tooltip" data-placement="bottom"
         title="{% if not official %}{% trans 'Submitted after deadline' %}{% elif passed %}{% if required > 0 %}{% trans 'Passed' %}{% endif %}{% elif missing_points %}{% blocktrans with points=required %}{{ points }} points required to pass{% endblocktrans %}{% elif submitted %}{% trans 'Required exercises are not passed' %}{% endif %}">
-    {{ points }} / {{ max }}
+			    		
+    		{% if not difficulty == "-" or max > 2 %}
+	    		{{ points }} / {{ max }}
+    		{% endif %}
```
raphendyr: And here we could have `{% else %}{% get_points_text %}`

exercise/templates/exercise/_points_badge.html
raphendyr: Btw. this line has different indentation than line before.

exercise/templatetags/exercise.py
```diff
@@ -128,6 +128,33 @@ def submission_status(status):
     return Submission.STATUS[status]
 
 
+"""
+    Used for adding relevant classes to the
+    points_badge to enable textual status.
+"""
+@register.simple_tag(takes_context=True)
+def get_points_class(context):
+    html_class = "" # holds result
```
raphendyr:
```
html_class = set()
if plaah:
  html_class.add('foobar')
return " ".join(html_class)
```
String concatenation isn't so nice looking...

exercise/templatetags/exercise.py
```diff
@@ -128,6 +128,33 @@ def submission_status(status):
...
+    if context['difficulty'] == "-" and context['max'] <= 2:
+        html_class = "aplus-points"
+        if (context['max'] == 2 and context['points'] == 2)\
+        or context['max'] == 1 and context['points']== 1:
+            html_class += " accepted"
+        elif (context['max'] == 2 and context['points'] == 1):
+            html_class += " ok"
+        elif context['points'] == 0 and context['submitted']:
+            html_class += " rejected"
```
raphendyr: and probable classes `points` and `points-ok`. Just `ok` isn't very exact.

exercise/templatetags/exercise.py
```diff
@@ -128,6 +128,33 @@ def submission_status(status):
...
+"""
+    Used for adding relevant classes to the
+    points_badge to enable textual status.
+"""
+@register.simple_tag(takes_context=True)
+def get_points_class(context):
+    html_class = "" # holds result
+    if context['difficulty'] == "-" and context['max'] <= 2:
+        html_class = "aplus-points"
+        if (context['max'] == 2 and context['points'] == 2)\
```
raphendyr: no need for `\` as the brackets do that.

exercise/templatetags/exercise.py
evnurm: Do you have any ideas as to what the middle step should be called in the 0-2 interval? 0-1 is pretty obvious (accepted/rejected).

exercise/templatetags/exercise.py
```diff
@@ -128,6 +128,33 @@ def submission_status(status):
...
+        elif context['points'] == 0 and context['submitted']:
+            html_class += " rejected"
+    return html_class
+
+
+#Adds the content of the points_badge.
+@register.simple_tag(takes_context=True)
+def get_points_text(context):
+    if context['difficulty'] == "-" and context['max'] <= 2:
+        return ""
```
raphendyr: This is the place to have translations. Should use the same code as above for the class. E.g.

```
class = get_class(context)
tests = {
 "ok": _("OK"),
 "accepted": _("Accepted"),
 "rejected": _("Rejected"),
}
return texts[class]
```

exercise/templatetags/exercise.py
raphendyr: Rejected, ok, good. Actually, rejected is probably bad when displayed, so we probably need to find better word for it. For css/js/code it's relatively fine.

exercise/templatetags/exercise.py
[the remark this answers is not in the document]
evnurm: If I was to attempt to generalize this, is there any documentation/reference regarding the structure of the tree? Is it like this: module -> children -> ... -> children?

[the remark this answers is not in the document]
raphendyr: Vague, but yes. Module is only different layer (shouldn't be, but it's for now) and then all childrens can have more of the same type. You can come and chat with @emmahonkamaa who has the best details of the current configuration hierarchy (except nothing about the cache part). Database structure follows that and then cache does something simplifications on top of it.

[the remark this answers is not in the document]
evnurm: Just pushed a commit that implements a recursive way of adding difficulty information to submissions - I tested it by adding a child to a child of a module and the info displayed seemed to be fine.

exercise/templatetags/exercise.py
```diff
@@ -118,6 +134,45 @@ def submission_status(status):
...
+def get_points_text(context):
+    if context['difficulty'] == "-" and context['max'] <= 2:
+        class_name = context['class']
+        texts = {
+          "ok": _("OK"),
+          "accepted": _("Accepted"),
+          "rejected": _("Rejected"),
+          "": _("Missing"),
+        }
+        return texts[class_name]
```
raphendyr: `texts.get(class_name, _("Unknown"))` is a bit more robust method. or `texts.get(class_name) or _("Unknown")` to be bit more optimal when running.

evnurm: Implemented the first option.

exercise/templates/exercise/_points_badge.html
```diff
@@ -6,27 +6,28 @@
...
-</span>
-{% elif exercise_page %}
-<span class="badge">
-  {% trans "Not submitted" %}
-</span>
-{% endif %}
-
+    {% if submitted %}
+        <span class="badge{% if graded and official %} badge-{% if full_score %}success{% elif passed %}warning{% else %}danger{% endif %}{% endif %} {{ classes }}">
+        {% if not official %}{% trans 'Submitted after deadline' %}{% elif passed %}{% trans "Accepted" %}{% elif graded %}{% trans "Rejected" %}{% else %}{% trans "Submitted" %}{% endif %}
```
raphendyr: This is now handled by below case with `get_points_class` and `get_points_text`, so this part can be refactored in some way so there wouldn't bee too much of duplication.

exercise/templatetags/exercise.py
```diff
@@ -118,6 +134,45 @@ def submission_status(status):
...
+        if ((context['max'] == 2 and context['points'] == 2)
+        or context['max'] == 1 and context['points']== 1):
+            class_name = "accepted"
+        elif (context['max'] == 2 and context['points'] == 1):
+            class_name = "ok"
+        elif context['points'] == 0 and context['submitted']:
+            class_name = "rejected"
+
+        html_class.add(class_name)
+        context['class'] = class_name
```
raphendyr: This is acceptable for this case, but it's really bad as general solution. This requires order of operations. Better solution would include e.g. having single helper function for both of this template tags. Helper can use context as cache, but that is not required.

evnurm: So, should I change this part or is it good as it's "acceptable for this case"?

exercise/templates/exercise/_points_badge.html
raphendyr: We can add special case for `confirm_the_level` if it's really needed, but I'm pretty sure this is better approach for O1 too. In other words, this rendering should be related to `confirm_the_level`, except it can be used for the extra messages like `Not submitted` and so on (logical as the exercise is required to get points from the chapter).

exercise/templatetags/exercise.py
raphendyr: It can be refactored when needed, so it's ok for now.

exercise/templates/exercise/_points_badge.html
evnurm: So are you trying to say that I should remove the `confirm_the_level` section from the template? Apparently it has some extra text options which makes it a little harder to just take that section away.

evnurm: Moved most of the logic into the template tags so the points_badge template does not need overly many spans and messy markup.

exercise/templatetags/exercise.py
```diff
@@ -118,6 +134,81 @@ def submission_status(status):
...
+        if context['submitted']:
+            if not context['official']:
+                points_string = _("Submitted after deadline")
+            elif context['passed']:
+                points_string = _("Accepted")
+            elif context['graded']:
+                points_string = _("Rejected")
+            else:
+                points_string = _("Submitted")
+        elif 'confirmable_points' in context and context['confirmable_points']:
```
raphendyr: `if context.get('confirmable_points'):` does the same.

evnurm: Changed the condition to the one you suggested.

exercise/templatetags/exercise.py
```diff
@@ -49,10 +50,25 @@ def _get_toc(context, student=None):
...
+            set_difficulties(child['children'])
 
 @register.inclusion_tag("exercise/_user_results.html", takes_context=True)
 def user_results(context, student=None):
     values = _get_toc(context, student)
+
+    # Add the difficulty information to all submission to enable
+    # textual status in points badges in submissions list.
+    for module in values['modules']:
+        set_difficulties(module['children'])
```
raphendyr: I was reading the code to find why this is actually even needed. What part breaks if this operation is not done?

To me only https://github.com/evnurm/a-plus/blob/ed99c0abe44ec112ca1ac15e51598cc02e827e9e/exercise/templatetags/exercise.py#L256 seems to be possible place where difficulty is now known.

evnurm: issue

I was faced with the issue above: the texts were not showed on the submission list. I assumed that it was because the difficulty information was not configured in the argument given to points_badge. Therefore I decided to add the information this way.

raphendyr: Could you test if you can fix it by editing `_points_data`, because I'm pretty sure the problem lies there.

raphendyr: It seems like https://github.com/evnurm/a-plus/blob/ed99c0abe44ec112ca1ac15e51598cc02e827e9e/exercise/templatetags/exercise.py#L234 should be executed, for that drop down, thus it should include the difficulty. Thus, I presume the code path goes to else branch due to some reason.

evnurm: I can confirm based on print-testing that the execution always seems to go there - the type of `obj` seems to always be `dict` due to which the `isinstance` conditions evaluate to False.

raphendyr: Yeah. Because the data comes from pointscache now-a-days, so the model information is lost. That templatetag should be redesigned to work correctly with pointscache, so it's able to resolve variables correctly. So, `_points_data` should detect the type from dictionary content or we should pass the type as argument from templates.

evnurm: Just realized that those isinstance-blocks do work as intented in the submission_plain and exercise_plain pages (apparently they do not come from cache?). In other words, I think that those should not be changed after all.

I believe all modules, exercises etc. work as intended at the moment in the dict-case so I guess that the easiest fix to this would be to just insert the difficulty information into each submission in the submissions list just the way I had done before. Is there a good reason why it can't be done that way?

raphendyr: Unneeded copying of data, that should be resolvable in `_points_data`. Can you check if the dictionary (`obj`) contains reference to the exercise, that way the code could could get it there if the obj it self doesn't contain the info.

evnurm: Does not seem to be the case  - here are the keys: 

['submission_status', 'confirm_the_level', 'graded', 'date', 'passed', 'id', 'url', 'max_points', 'submission_count', 'points_to_pass', 'points', 'unofficial']

evnurm: I could also perform the addition of difficulty in the phase where the cached data is generated so it does not need to be changed later in the template tag logic.  Is that a better idea? I would call that addition instead of copying.

evnurm: I decided to implement my suggestion as I feel that it's better than the previous solution. It's now pushed into this PR.

raphendyr: Can you move the change after line `confirm_the_level': entry.get('confirm_the_level', False),`. In addition, it a bit looks like entry is reference to the exercise, so check if that contains `difficulty`.

evnurm: Looks like it does contain `difficulty`. I implemented it that way and moved the change onto that line - should be pushed as well.

exercise/templatetags/exercise.py
```diff
@@ -118,6 +118,81 @@ def submission_status(status):
...
+            elif context['graded']:
+                points_string = _("Rejected")
+            else:
+                points_string = _("Submitted")
+        elif context.get('confirmable_points'):
+            points_string = _("Missing")
+        elif context['exercise_page']:
+            points_string = _("Not submitted")
+
+    elif context['difficulty'] == "-" and context['max'] <= 2:
```
raphendyr: If we change this to: `if (context['confirm_the_level'] or context['difficulty'] == "-") and context['max'] <= 2:` and add one block on below list below

```
elif context.get('confirmable_points');:
  points_string = _("Missing")
elif context['confirm_the_level'] and :context['exercise_page']
  points_string = _("Not submitted")
```

Then these codes should yield the same result for both difficulty ' -' and confirm_the_level cases. Submitted after deadline will be dropped, but that doesn't seem any way logical as it does remove the information about the grade (doesn't tell if the submission was accepted or not).

exercise/templatetags/exercise.py
```diff
@@ -118,6 +118,81 @@ def submission_status(status):
     return Submission.STATUS[status]
 
 
+"""
+    Used for adding relevant classes to the
+    points_badge to enable textual status.
+"""
+@register.simple_tag(takes_context=True)
+def get_points_class(context):
```
raphendyr: If you notice, the logic in both functions is the same, so one of them could provide it to the another one. E.g. `_get_points_type` that returns the class string or any other string. Then get_points_text can just use dict to map those strings to texts.

evnurm: Not sure if I managed to address everything you pointed out but I did make an attempt to consolidate the logic - it was not just as straightforward as I had thought. Anyway, you can have a look at the changes and see if there's (still) something to be fixed.

exercise/templatetags/exercise.py
```diff
@@ -117,81 +117,68 @@ def percent(decimal):
...
 
-
+def _get_points_type(context):
+    points_type = ""
+
+    if ((context['confirm_the_level']) or
+        context['difficulty'] == "-" and context['max'] <= 2):
+
+        if (context['full_score'] or
+            (context['passed'] and context['confirm_the_level'])):
```
raphendyr: I need to take a paper and draw this logic paths on it. :D

exercise/templatetags/exercise.py
```diff
@@ -117,81 +117,68 @@ def percent(decimal):
...
+    else: # normal exercise without textual status
+        if context['full_score']:
+            points_type = "full"
+        elif context['passed']:
+            points_type = "passed"
+        elif context['submitted']:
+            points_type = "not_passed"
+    if not context['confirm_the_level'] and context.get('unconfirmed'):
+        points_type += " unconfirmed"
+    return points_type
```
raphendyr: This returns string, but other functions expects a tuple.

evnurm: Changed into a list.

exercise/templatetags/exercise.py
evnurm: I guess it is a little messy but I did not feel like making a huge bunch of `if...elif...elif...else` -type structures where some of the blocks could have the same contents. Also, I tried to copy the logic from the `points_badge` template as accurately as possible but that was a bit messy too so this might be more complicated than it would need to be.

raphendyr: Set would work here bettere, or just two variables. Sets are nice as you can then do:

`s.add("plaah")` and `"plaah" in s` and it's way faster than with lists.

evnurm: I do see your point - on the other hand, the set will contain 1 or 2 elements: some string and, in some cases, "unconfirmed". I can easily check if "unconfirmed" exists in the set with the in-keyword but the other string is one of numerous alternatives and it doesn't seem very appealing to go over all possible alternatives to see if they are contained. Also, if the list has length 1 or 2, does it actually create a big performance problem?

evnurm: I think I got it to work using a set.

exercise/templatetags/exercise.py
```diff
@@ -167,16 +161,16 @@ def get_points_class(context):
...
+    if "unconfirmed" in points_type:
         class_name += " unconfirmed-points"
     return class_name
 
 # Adds the content of the points_badge.
 @register.simple_tag(takes_context=True)
 def get_points_text(context):
     points_string = str(context['points']) + " / " + str(context['max'])
-    points_type = _get_points_type(context)[0]
+    points_type = next(filter(lambda x: x != "unconfirmed", _get_points_type(context)), "")
```
raphendyr: As we can see, set doesn't after all work as well as predicted, even though it would be nice in some similar cases. So, maybe just return tuple of two arguments? That is also trivial to unpack.

so `return (class_, unconfirmed)` and then `points_type, unconfirmed = _get_points_type(context)[0]`

evnurm: Done.

exercise/templatetags/exercise.py
```diff
@@ -118,33 +118,34 @@ def submission_status(status):
...
         elif context['submitted']:
-            points_type.add("not_passed")
-    if not context['confirm_the_level'] and context.get('unconfirmed'):
-        points_type.add("unconfirmed")
-    return points_type
+            class_name = "not_passed"
+
+    # the second element in the returned tuple is a boolean indicating
+    # the confimation status of the points (unconfirmed is indicated by True).
+    return (class_name, not context['confirm_the_level'] and context.get('unconfirmed'))
```
raphendyr: Bit more clear if temporary variable is used. Variable name documents what is going on and then the last line will be trivial to read and understand :)

exercise/templatetags/exercise.py
```diff
@@ -161,24 +162,24 @@ def get_points_class(context):
...
+    if points_type[1]:
         class_name += " unconfirmed-points"
     return class_name
 
 # Adds the content of the points_badge.
 @register.simple_tag(takes_context=True)
 def get_points_text(context):
     points_string = str(context['points']) + " / " + str(context['max'])
-    points_type = next(filter(lambda x: x != "unconfirmed", _get_points_type(context)), "")
+    points_type = _get_points_type(context)
```
raphendyr: You can assign to two variables here. Makes code more readable ;)

evnurm: I made the `_get_points_class` function use that two variable setting. `get_points_text` only needs the first element of the tuple so I did not create a useless variable for the `unconfirmed` status.

exercise/templatetags/exercise.py
evnurm: Done.

exercise/templatetags/exercise.py
```diff
@@ -117,6 +117,68 @@ def percent(decimal):
...
+    class_name = options.get(points_type, "")
+    if unconfirmed:
+        class_name += " unconfirmed-points"
+    return class_name
+
+# Adds the content of the points_badge.
+@register.simple_tag(takes_context=True)
+def get_points_text(context):
+    points_string = str(context['points']) + " / " + str(context['max'])
+    points_type = _get_points_type(context)[0]
```
raphendyr: `points_type, _ = _get_points_type(context)` this is typical and a bit more readable solution :)

evnurm: I got the following error while testing that: "'bool' object is not callable".

raphendyr: Argh... Yep. Based on short review, pylint should be fine with variables starting with `_`, so `_uc` could be fine here.

evnurm: Done.

raphendyr:
This whole logic how we calculate the state of the submission is horrible. This thing (not the PR, the whole file) needs go to the waste bin... So, Well go how it works now and rewrite everything from the ground some beautiful day, when ever that is. Luckily, we will do it with something better than django templates, so we actually can make it readable and logical.

So,

Rebase this to remove intermediate steps and put it on top of current master. I'll test it afterwards to see if something breaks.

evnurm:
I believe that I got that done.

Mankro:
This old PR was developed at the same time with #355. The use case was to use A+ for applications in the student administration where the application is accepted or rejected instead of being graded with 0-N points (compared to normal exercises).

PR #350, opened 2018-07-12
Show warning overlay when exercise cannot be submitted

jrp6:
... or can be submitted with caveats. Also, disable the submit button when the exercise is not submittable.


exercise/exercise_models.py
```diff
@@ -494,8 +512,7 @@ def _check_submission_allowed(self, profile, request=None):
                 if gid > 0:
                     group = profile.groups.filter(
                         course_instance=self.course_instance,
-                        id=gid
-                    ).first()
+                        id=gid).first()
```
raphendyr: Personally, the `id=gid` is just missing `,` from the end ;)
Fine.

exercise/exercise_models.py
```diff
@@ -511,14 +528,13 @@ def _check_submission_allowed(self, profile, request=None):
                     warnings.append(_("You have previously submitted to this exercise alone.") + " " + msg)
                 else:
                     warnings.append(_("You have previously submitted to this exercise with {collaborators}.").format(
-                        collaborators=StudentGroup.format_collaborator_names(s.submitters.all(), profile)
-                    ) + " " + msg)
-                return False, warnings, students
+                        collaborators=StudentGroup.format_collaborator_names(
+                            s.submitters.all(), profile)) + " " + msg)
```
raphendyr: This was originally hard to read, so I would do something like:

```
if plaah:
  warning = _("...")
else:
  warning = _("...").format(collaborators=...)
warnings.append("{} {}".format(warning, msg))
```

exercise/templates/exercise/exercise.html
```diff
@@ -15,8 +15,27 @@
...
 {% endif %}
 
-<div id="exercise-page-content"{% if exercise.is_submittable %} data-aplus-chapter="{{ exercise|url }}" data-aplus-group="{{ exercise.min_group_size }}-{{ exercise.max_group_size }}"{% if summary.get_submission_count > 0 %} data-aplus-group-fixed="{{ summary.get_group_id }}"{% endif %}{% endif %}>
-    {{ page.content|safe }}
+<div
+    id="exercise-page-content"
+    {% if exercise.is_submittable %}
+    data-aplus-chapter="{{ exercise|url }}"
+    data-aplus-group="{{ exercise.min_group_size }}-{{ exercise.max_group_size }}"
+    {% if summary.get_submission_count > 0 %}
```
raphendyr: I would indent these too, so the if-else structure would be more readable.

exercise/templates/exercise/exercise_plain.html
```diff
@@ -38,101 +38,116 @@
...
-					{% elif submission and submission.status == "waiting" %}
-					<div class="alert alert-info">
-						{% trans "Your submission has been accepted for grading." %}
-					</div>
-					{% endif %}
-				</div>
-				{% include "exercise/_exercise_wait.html" %}
-			</div>
-		</div>
+        {% if exercise.is_submittable %}
```
raphendyr: Indentation for this block looks really strange (the whole block)..

PR #352, opened 2018-07-25
Show exercises in correct order on all results page

jrp6:
I.e. in the order they appear in the forest.

Assume that anything interesting is a leaf. It should be relatively simple
to modify this to inspect the whole forest if needed in the future.

It should be possible to implement this also using search_exercises
which returns the non-root nodes of the forest as a list. That might lead
to a simpler solution code-wise but I feel tree traversal is the more intuitive
solution.

### DIFF
--- a/exercise/exercise_summary.py
+++ b/exercise/exercise_summary.py
@@ -1,6 +1,10 @@
+import itertools
+
+from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Max
 
 from course.models import StudentGroup
+from .cache.content import CachedContent
 from .models import BaseExercise, Submission
 
 
@@ -119,9 +123,7 @@ class ResultTable:
         self.course_instance = course_instance
 
         # Exercises on the course.
-        self.exercises = list(BaseExercise.objects \
-            .filter(course_module__course_instance=course_instance) \
-            .order_by("course_module__closing_time", "course_module", "order"))
+        self.exercises = list(self.__get_exercises())
         self.categories = course_instance.categories.all()
 
         # Students on the course.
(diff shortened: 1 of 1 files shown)


exercise/exercise_summary.py
```diff
@@ -143,6 +145,28 @@ def __init__(self, course_instance):
...
+            return (node['id'],)
+
+        root_node = { 'children': content.modules() }
+        ids = get_descendant_ids(root_node)
+
+        # Loop until end of ids raises StopIteration
+        while True:
+            id = next(ids)
+            try:
+                yield LearningObject.objects.get(pk=id).baseexercise
```
raphendyr: Could this resolved multiple Learning objects with a single sql query? Or at least base exercises with get(something_learninh_object_id=id). Pretty sure the sql queries can be optimized a bit someway.

jrp6: You're right, this could be done better with `BaseExercise.objects.get(learningobject_ptr_id=id)`.

In principle, it would be possible to hit the database only once by first doing something like `BaseExercise.objects.filter(learningobject_ptr_id__in=ids)` to get the BaseExercises in essentially random order, and then sort them so that the `learningobject_ptr_id`s are in the order they are in `ids`. However, first of all it is not obvious to me that this is more efficient, since the sort could end up becoming a relatively heavy operation. Second, it would be a _much_ more complex solution, and therefore a premature optimisation: the all results page is infrequently used and it takes me about 2500 ms to render the page for O1, so there doesn't seem to be a performance problem.

raphendyr: Sounds like the first part (directly requesting baseexercises) is good enough solution.

jrp6:
ping @raphendyr

PR #354, opened 2018-07-30
LTI 1.1 fixes

Mankro:
Small fixes are needed.

### DIFF
--- a/external_services/api/serializers.py
+++ b/external_services/api/serializers.py
@@ -11,7 +11,7 @@ from .renderers import LTIOutcomeResponseRenderer
 class LTIOutcomeSerializer(serializers.Serializer):
     # parameters that are parsed from the XML of the request body
     version = serializers.CharField(default='1.0')
-    msgid = serializers.CharField(max_length=32)
+    msgid = serializers.CharField(max_length=127)
     req_type = serializers.ChoiceField(choices=[
         LTIOutcomeXMLParser.TYPE_REPLACE, LTIOutcomeXMLParser.TYPE_READ, LTIOutcomeXMLParser.TYPE_DELETE
     ])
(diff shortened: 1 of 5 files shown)


external_services/api/serializers.py
```diff
@@ -11,7 +11,7 @@
 class LTIOutcomeSerializer(serializers.Serializer):
     # parameters that are parsed from the XML of the request body
     version = serializers.CharField(default='1.0')
-    msgid = serializers.CharField(max_length=32)
+    msgid = serializers.CharField(max_length=50)
```
raphendyr: Is then length requirement exact, or should we just have enough space?

Mankro: Not exact. I thought it would be better to have some upper limit, but it is not required. The LTI specification is not clear about the max length. After reading Moodle source code, they don't seem to use any limit there.

raphendyr: What type of data this field typically seems to contain? Output of random hash string? If so, then it probably could be 255 or 127 (nice classic numbers). In addition, make sure the code raises and logs good error if the msgid is too long and doesn't just ignore it silently. I'm not sure what happened this time.

Mankro: It contains a unique id. It is not specified in more detail. The max length is not clearly specified either. The Tool Providers I have seen thus far use a length of at most 36 characters. If we ever need to use a Tool Provider that uses longer ids, we would need to increase the limit at any rate. I could remove the limit now since it is very unlikely that it would ever cause issues. The Tool Providers are manually enabled in the system anyway.

raphendyr: CharFields require max_length, so put 127 on it.

Mankro:
I have increased the max length of the message identifier and added error logging. Finished fron my perspective.

raphendyr:
DB migration script is missing, otherwise it's good!

Mankro:
There are no migrations. The serializer is not directly connected to any database models, so no model has been modified here.

raphendyr:
My bad. It just looked like a model, so I made a mistake :)

PR #355, opened 2018-08-02
Change submission in force

evnurm:
Enable choice between the exercise points being from the best submission or latest submission.


Mankro:
This old PR intended to support use cases where the latest submission would always be the most important instead of the submission with the best grade. That would make sense at least when the exercise is more like a feedback form or a job application instead of a normal graded exercise. This PR is very old and certainly won't be merged.

Mankro:
In the past year, new features have been developed that replace this pull request.
* manually set the submission that defines the grade for the assignment: #877 
* use the latest submission instead of the best submission for the final grade of the assignment: #880

PR #358, opened 2018-08-10
Sanify language selection dropdown markup

jrp6:
* Use semantically correct UL tag instead of DIV and add aria-labelledby
* Do not wrap the dropdown contents inside a non-semantic H4 tag, and use
  bootstrap's .h4 CSS class instead


jrp6:
On another note, I decided that the new meaning of _sanify_ is _to make sane_.

PR #362, opened 2018-08-23
Implement language-changing links

jrp6:
for user-friendly wrong language alerts

### DIFF
--- a/exercise/static/exercise/language_link.js
+++ b/exercise/static/exercise/language_link.js
@@ -1,0 +1,26 @@
+(function ($) {
+  const set_language_url = (document.currentScript ?
+    $(document.currentScript) :
+    $('script').last()) // Ugly solution for IE11
+    .attr('data-set-lang');
+
+  $(function () {
+    $('a.change-language').each(function (i, elem) {
+      const $elem = $(elem);
+      const target_lang = $elem.attr('data-target-lang');
+      $elem.on('click', function () {
+        $.post(set_language_url, { language: target_lang })
+          .done(function () {
+            const href = $elem.attr('href');
+            // href="#" means "this page, but in another language"
+            if (href === '#') {
+              window.location.reload();
+            } else {
+              window.location.href = href;
+            }
+          });
+        return false;
+      });
+    });
+  });
+})(jQuery);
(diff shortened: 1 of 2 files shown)


exercise/static/exercise/language_link.js
```diff
@@ -0,0 +1,19 @@
+function set_language_links(set_language_url) {
+  $('a.change-language').each(function (i, elem) {
+    const $elem = $(elem);
+    const target_lang = $elem.attr('data-target-lang');
```
raphendyr: Was there some difference to `.data('target-lang')`?

exercise/templates/exercise/exercise.html
```diff
@@ -66,18 +66,14 @@ <h3 class="text-center">{% trans "Submission received." %}</h3>
...
-        $("#submit-progress").removeClass("hide");
-    });
-});*/
-
 $(function() {
     // LTI exercises
     $('[data-external-launch]').aplusExternalLauncher();
+
+    // Links that change language
+    set_language_links('{% url 'set_language' %}');
```
raphendyr: I just removed this block in local change to get rid of inline javascript. We probably should use link or similar element or just hardcode the url in the js file ( :( ).

exercise/static/exercise/language_link.js
```diff
@@ -0,0 +1,19 @@
...
+    $elem.on('click', function () {
+      $.post(set_language_url, { language: target_lang })
+        .done(function () {
+          const href = $elem.attr('href');
+          // href="#" means "this page, but in another language"
+          if (href === '#') {
+            window.location.reload();
+          } else {
+            window.location.href = $elem.attr('href');
```
raphendyr: `... = href`

exercise/static/exercise/language_link.js
jrp6: In this case, there shouldn't be any. I generally prefer `.attr()` because in more complicated cases `.data()` involves some black magic which IMO can make the code hard to read.

exercise/templates/exercise/exercise.html
jrp6: Probably better to hardcode the url in that case. There doesn't seem to be any suitable `rel` value for links or any other appropriate metadata tags.

---

On another note, life would be much easier if there was a sane way to pass django template variables to javascript

PR #363, opened 2018-08-27
Clickable input support and bug fixes for active element

roosapi:
Add type check and onchange event to active element input update to support clickable input extension. Also fixes active element submitting when there is no previous input value stored and the polling timeout message which was broken for active elements due to incorrect data-attribute name.

### DIFF
--- a/exercise/static/exercise/chapter.js
+++ b/exercise/static/exercise/chapter.js
@@ -235,7 +235,7 @@
 			label.html(title);
 
 			var form_field;
-			if (!type) {
+			if (!type || type === "clickable") {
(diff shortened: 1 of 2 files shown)


raphendyr:
Is this needed for autumn courses?

roosapi:
I don't think so, unless someone else is using the active element exercise type and wants the bugs fixed.

raphendyr:
Check, I'll leave this for the winter release, so I'll merge the PR in couple of weeks once the autumn release is final.

raphendyr:
add to whitelist

PR #364, opened 2018-08-29
Send exercise_id to CourseHooks

jrp6:
in addition to submission_id. Makes implementing the automatic tagger easier.


exercise/submission_models.py
```diff
@@ -222,7 +222,10 @@ def set_ready(self):
         # Fire set hooks.
         for hook in self.exercise.course_module.course_instance \
                 .course_hooks.filter(hook_type="post-grading"):
-            hook.trigger({ "submission_id": self.id })
+            hook.trigger({
+                "submission_id": self.id,
+                "exercise_id": self.exercise.id,
```
raphendyr: should also include course_id (course_instance.id). and url to plus.cs.hut.fi (well, to installation), probably in field `site`.

PR #371, opened 2018-09-24
LTI exercise relative URL and missing LTI error in course import

Mankro:
LTI exercises may use relative service URLs to modify the global LTI service URL and the JSON course import shows errors if the LTI service used by an exercise can not be found in A+.

**Relative service URLs in LTI exercises:**
* If an LTI exercise defines its own service URL and it is a relative URL, it is joined to the URL of the LTI service. Relative URLs with an absolute path replace the previous path completely.
* Absolute service URLs are also still supported for backwards compatibility, though their use is discouraged. The URL domain must equal the domain of the LTI service.
* If the exercise defines no service URL, the URL of the LTI service is used.

**Course import error message for missing LTI services**

LTI exercises refer to the LTI service by a label in JSON. JSON course import shows errors if the LTI service configuration is not found in A+ for the LTI exercise. Furthermore, the exercise is not updated/created.

### DIFF
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -1,4 +1,5 @@
 import datetime
+import urllib.parse
 from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ValidationError, PermissionDenied
@@ -27,6 +28,7 @@ from lib.helpers import (
     Enum,
     update_url_params,
     has_same_domain,
+    drop_domain_from_url,
     safe_file_name,
     roman_numeral,
 )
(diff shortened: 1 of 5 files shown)


raphendyr:
I updated the privacy branch, so this needs rebasing. In addition, note that `get_final_url` can now be used for this too.

exercise/exercise_models.py
```diff
@@ -713,7 +724,21 @@ def load(self, request, students, url_name="exercise"):
...
+        if self.service_url:
+            parse_result = urllib.parse.urlparse(self.service_url)
+            if parse_result.netloc:
+                # Deprecated: absolute URL stored in the database. Drop the domain now.
+                # Typically, LTI exercises do not need to specify a service URL since
+                # they usually use the launch URL of the LTI service. If a custom
+                # service URL is needed, it must be in the same domain as the
+                # LTI service. Previously, LTI exercises stored service URLs in the
+                # absolute form, but now only the URL path is stored.
+                replace_url = drop_domain_from_url(parse_result=parse_result)
```
raphendyr: `replace_url = parse_result._replace(scheme='', netloc='').geturl()`

Mankro: `_replace` method is not part of the public API...

raphendyr: Hmm.. True. Result object is documented to be subclass of `tuple` not `namedtuple` and  `_replace` is part of public api for `namedtuple`.

I personally see that `_replace` is the cleanest way, but if we go with just tuples then `type(r)('', '', *r[2:]).geturl()` or `urlunsplit(('', '', *r[2:]))`.

ps. all instances of urlparse can be replaced with urlsplit in the code.

Mankro: I suppose we can assume that `_replace` is part of the public API, since the documentation in `urllib.parse` says that the return value is a tuple that has named attributes. Normal tuples do not have such attributes, but named tuples do, and the implementation of the library uses them.

Mankro:
I made changes (using `urlsplit` and a one-liner for removing the URL domain) and rebased.

Mankro:
Done rebase.

PR #375, opened 2018-11-07
Exercisecollection feature

Donaldi3:
Can specify a category from a separate course.
Displays completion status of category's exercises and provides links to the exercise area.
Possibility to scale points for current course.


Plustest:

Donaldi3:
There might be performance issues from signal based updates.
Can this be verified and is there a better solution?

exercise/templates/exercise/exercise.html
```diff
@@ -35,6 +35,7 @@
   </div>
 </div>
 
+
```
raphendyr: 🤔

exercise/urls.py
```diff
@@ -71,6 +72,7 @@
         staff_views.FetchMetadataView.as_view(),
         name="exercise-metadata"),
 
+
```
raphendyr: 🤔

exercisecollection_readme.txt
```diff
@@ -0,0 +1,16 @@
...
+
+
+Grader configuration:
+
+In yaml-file: 
+
+Add following keys to exercise:
+
+    collection_category: <name of the category>
+    collection_course: <name of the course>;<name of the course instance>
```
raphendyr: Display name or url part?

exercisecollection_readme.txt
```diff
@@ -0,0 +1,16 @@
+Specify exercise that refers to any LearningObjectCategory.
```
raphendyr: This should be under doc/ and _readme is then not needed.

edit_course/operations/configure.py
```diff
@@ -142,6 +148,7 @@ def configure_learning_objects(category_map, module, config, parent,
                 if obj_key in o:
                     setattr(lobject, key, parse_bool(o[obj_key]))
 
+
```
raphendyr: 🤔

exercise/views.py
```diff
@@ -98,6 +107,7 @@ def get(self, request, *args, **kwargs):
         if self.profile:
             LearningObjectDisplay.objects.create(learning_object=self.exercise, profile=self.profile)
 
+
```
raphendyr: 🤔

edit_course/operations/configure.py
```diff
@@ -171,6 +178,34 @@ def configure_learning_objects(category_map, module, config, parent,
...
+                "max_submissions",
+            ]:
+                if key in o:
+                    errors.append("Can't define '{}' for ExerciseCollection".format(key))
+
+            if "max_points" in o and o["max_points"] <= 0:
+                errors.append("ExerciseCollection can't have max_points <= 0")
+                continue
+
+            setattr(lobject, "target_category", target_category)
```
raphendyr: Why setattr and not just dot notation?

edit_course/operations/configure.py
```diff
@@ -429,3 +464,29 @@ def configure_content(instance, url):
...
 
     return errors
+
+def get_target_category(course, category):
+
+    if not category:
+        return None, _("ExerciseCollection object requires collection_category.")
+
+
+    course_name, instance_name = course.split(";")
```
raphendyr: Doesn't support colon in course name or instance name.

Url part is probably better as there is strict set of allowed letters.

exercise/exercisecollection_models.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def get_load_url(self, language, request, students, url_name="exercise"):
+        pass
+
+
+# Updates submissions if new submission is in any ExerciseCollection's target category.
+# ! Probably needs Cache-optimization
+@receiver(post_save, sender=Submission)
+def update_exercise_collection_submission(sender, instance, **kwargs):
+    collections = ExerciseCollection.objects.filter(target_category=instance.exercise.category)
```
raphendyr: This should create only max two db queries when there is no exercisecollection. One to check that out and possibly to resolve exercise.

That should be fine.

raphendyr:
Also rhis needs squashing.

Donaldi3:
Did some cleaning up and squashed to 4 commits.

raphendyr:
add to whitelist

edit_course/operations/configure.py
```diff
@@ -429,3 +468,54 @@ def configure_content(instance, url):
...
+def get_target_category(category, course=None, course_url=None):
+
+    course_name = None
+    instance_name = None
+
+
+    if not category:
+        return None, _("ExerciseCollection object requires collection_category.")
+
+    if (course is None) == (course_url is None):
```
raphendyr: `if course or course_url and not (course and course_url):` couple be a bit readable?

exercise/templates/exercise/exercise_plain.html
```diff
@@ -44,7 +44,7 @@
                     {% include 'exercise/_warnings_overlay.html' %}
                 {% endif %}
 
-                {% if exercise.is_submittable %}
+                {% if exercise.is_submittable and not exercisecollection %}
```
raphendyr: If you reorder this, then we would have `if exercisecollection` and `elif exercise.is_submittable`.

edit_course/operations/configure.py
```diff
@@ -429,3 +468,54 @@ def configure_content(instance, url):
...
+
+        try:
+            course_instance = CourseInstance.objects.get(instance_name=instance_name,
+                                                         course__name=course_name)
+        except ObjectDoesNotExist:
+            return None, _("Course: {}, Instance: {}, not found.".format(course_name, instance_name))
+
+    else:
+        course_slug_begin = 0
+        if not settings.BASE_URL in course_url:
```
raphendyr: Why configuration url includes the site part? Kind of makes sense of course as they are site specifig.

Donaldi3: Copy pasted URLs usually contain also the site.
By checking that the URL is referring to the same system it's being run we can inform users that they can't use external services (like other A+ systems) for this.

I think giving specific error from wrong site in the URL is better than just informing that the course couldn't be found.

raphendyr: Done side is that it doesn't support local testing. Though, local testing uses different url path and course name also, so it's problem any case :disappointed:

raphendyr: We will go with this for now.

Donaldi3: Oh right... I forgot about the local testing of the course content.

Maybe the configuration could just skip adding the exercisecollections when doing local testing?
This could probably be done by checking if BASE_URL is '127.0.0.1' or 'localhost'?

Donaldi3: But then again that would make local testing generally impossible...
The only working solution right now is to do content testing without exercisecollections and test them separately on the production server.

raphendyr: Settings are already different in repo, local containers and in production. See https://github.com/apluslms/run-aplus-front/blob/dev/Dockerfile#L5 for an example.

PR #380, opened 2019-01-18
Import the course head URLs setting from the aplus-json

Mankro:


raphendyr:
Why are shibboleth tests now broken.. :slightly_frowning_face: 
Well, not caused by this so.

PR #381, opened 2019-01-18
Fixes for exercisecollection

Donaldi3:
Configuration wasn't working correctly for exercisecollection.

- Fixed possible issues in course- and instance-slug identification.
- BASE_URL is no longer mandatory, slugs are guessed by default to be first two parts of the path.
- EC-model's target exercise query is now ordered by id.

### DIFF
--- a/edit_course/operations/configure.py
+++ b/edit_course/operations/configure.py
@@ -104,7 +104,7 @@ def configure_learning_objects(category_map, module, config, parent,
         # Select exercise class.
         lobject_cls = (
             LTIExercise if "lti" in o
-            else ExerciseCollection if "collection_category" in o
+            else ExerciseCollection if "target_category" in o
             else BaseExercise if "max_submissions" in o
             else CourseChapter
         )
@@ -181,18 +181,14 @@ def configure_learning_objects(category_map, module, config, parent,
         lobject.parent = parent
 
         if lobject_cls == ExerciseCollection:
-            if 'collection_course' in o and not o['collection_course'] is None:
-                target_category, error_msg = get_target_category(o["collection_category"],
-                                                                 course=o["collection_course"],)
-            else:
-                target_category, error_msg = get_target_category(o["collection_category"],
-                                                                 course_url=o["collection_url"], )
+            target_category, error_msg = get_target_category(o["target_category"],
+                                                             o["target_url"],)
             if error_msg:
                 errors.append("{} | {}".format(o["key"], error_msg))
                 continue
 
             if target_category.id == lobject.category.id:
-                errors.append("ExerciseCollection can't target its own category")
+                errors.append("{} | ExerciseCollection can't target its own category".format(o["key"]))
                 continue
 
             for key in [
@@ -206,6 +202,9 @@ def configure_learning_objects(category_map, module, config, parent,
             if "max_points" in o and o["max_points"] <= 0:
                 errors.append("ExerciseCollection can't have max_points <= 0")
                 continue
+
+            lobject.max_points = o['max_points']
+            lobject.points_to_pass = o['points_to_pass']
 
             lobject.target_category = target_category
             lobject.min_group_size = 1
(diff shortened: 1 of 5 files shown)


edit_course/operations/configure.py
```diff
@@ -509,7 +512,9 @@ def get_target_category(category, course=None, course_url=None):
             course_instance = CourseInstance.objects.get(url=instance_slug,
                                                          course__url=course_slug)
         except ObjectDoesNotExist:
-            return None, _('No course found with URL "{}"'.format(course_url))
+            return None, _('No course found with URL "{}"; \
```
raphendyr: no need for escape and there is wrong character ';' instead of ','.

edit_course/operations/configure.py
```diff
@@ -496,11 +496,14 @@ def get_target_category(category, course=None, course_url=None):
...
 
     else:
-        course_slug_begin = 0
-        if not settings.BASE_URL in course_url:
-            return None, _('ExerciseColletion URL "{}" not in correct domain {}.'.format(course_url,settings.BASE_URL))
 
-        course_slug_begin = len(settings.BASE_URL)
+        course_slug_begin = course_url.find('/', course_url.find('://') + 1) + 1
+
+        if settings.BASE_URL:
```
raphendyr: This is always defined as a-plus doesn't start without it.

edit_course/operations/configure.py
```diff
@@ -496,11 +496,14 @@ def get_target_category(category, course=None, course_url=None):
             return None, _("Course: {}, Instance: {}, not found.".format(course_name, instance_name))
 
     else:
-        course_slug_begin = 0
-        if not settings.BASE_URL in course_url:
-            return None, _('ExerciseColletion URL "{}" not in correct domain {}.'.format(course_url,settings.BASE_URL))
 
-        course_slug_begin = len(settings.BASE_URL)
+        course_slug_begin = course_url.find('/', course_url.find('://') + 1) + 1
```
raphendyr: Would it be more clear using urllib.parse.urlsplit here?

Somethin like:

```
prefix = urlsplit(settings.BASE_URL).path.rstrip('/') + '/'
path = urlsplit(course_url).path.rstrip('/')
if not path.startswith(prefix):
  raise Error("Invalid prefix")
course, instance = path[len(prefix):].split('/', 1)
```
In addition, we should probably validate domain too?

edit_course/operations/configure.py
Donaldi3: Escapes are used to split that one long string to multiple lines. Not using them would cause End Of Line errors.

I wanted to add slugs as debug-aid to the error message. I though that ';' would keep the "main message" separated and provide the slugs as additional information. '.' might be better...

edit_course/operations/configure.py
```diff
@@ -497,22 +498,20 @@ def get_target_category(category, course=None, course_url=None):
...
-        course_slug_begin = course_url.find('/', course_url.find('://') + 1) + 1
+        parsed_url = urlparse(course_url)
+        service_hostname = urlparse(settings.BASE_URL).hostname
 
-        if settings.BASE_URL:
-            # Can correctly identify slug-beginning with BASE_URLs
-            # even if they end in '/'
-            course_slug_begin = course_url.find('/', len(settings.BASE_URL) - 1) + 1
+        if parsed_url.hostname != service_hostname:
+            return None, _("Course URL '{}' doesn't match service's hostname '{}'".format(course_url, service_hostname))
```
raphendyr: You can't do formatting before translation is resolved the format *needs* to be `_("string").format(foo, bar)`. Furthermore, that would resolve the lazy string at this point, but we wan't to delay it until rendering. Thus, you really need to do `format_lazy(_("string"), foo, bar)`.  Sadly, format_lazy is not in a good place under lib. It's under `authorization/permissions.py` and `lib/remote_page.py`, but should be merged to single place e.g. `lib/translation.py`.

edit_course/operations/configure.py
```diff
@@ -497,22 +498,20 @@ def get_target_category(category, course=None, course_url=None):
...
+
+        course_slug = parsed_url.path.split('/')[1]
+        instance_slug = parsed_url.path.split('/')[2]
```
raphendyr: You can do this with single split (python is unable to do optimisation here).
`course_slug, instance_slug = parsed_url.path.split('/')[:2]` (will fail if there is no `/` in the path).

edit_course/operations/configure.py
```diff
@@ -497,22 +498,20 @@ def get_target_category(category, course=None, course_url=None):
...
-        instance_slug_begin = course_url.find('/', course_slug_begin) + 1
-        course_slug = course_url[course_slug_begin : instance_slug_begin - 1]
-        instance_slug = course_url[instance_slug_begin: course_url.find('/', instance_slug_begin)]
-        
         try:
             course_instance = CourseInstance.objects.get(url=instance_slug,
                                                          course__url=course_slug)
         except ObjectDoesNotExist:
-            return None, _('No course found with URL "{}"; \
+            return None, _('No course found with URL "{}".  \
```
raphendyr: Same thins here as above, can't do formatting within translation.

raphendyr:
Tests are now fixed in master, so you should rebase this.

raphendyr:
Manually merged (rebased these first).

PR #384, opened 2019-03-12
Tagging and course hook improvements, API creation of usertags

JoonatanPartanen:
These relate to the development of the automatic tagging script 
at [JoonatanPartanen/o1_auto_tag](https://github.com/JoonatanPartanen/o1_auto_tag)

### DIFF
--- a/course/api/full_serializers.py
+++ b/course/api/full_serializers.py
@@ -1,4 +1,5 @@
-from rest_framework import serializers
+from rest_framework import serializers, exceptions
+from rest_framework_extensions.fields import NestedHyperlinkedIdentityField
 
 from lib.api.fields import NestedHyperlinkedIdentityField
 from lib.api.serializers import AplusModelSerializer, NestedHyperlinkedIdentityFieldWithQuery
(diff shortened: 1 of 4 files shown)


Plustest:

raphendyr:
whitelist

raphendyr:
add to whitelist

course/api/views.py
```diff
@@ -115,7 +124,7 @@ class CourseUsertaggingsViewSet(NestedViewSetMixin,
                        'course_instance__id')
                  .all() )
     parent_lookup_map = {'course_id': 'course_instance_id'}
-
+
```
raphendyr: ?

course/models.py
```diff
@@ -189,13 +190,23 @@ def namegen():
 post_save.connect(pseudonymize, sender=Enrollment)
 
 
+class UserTagManager(models.Manager): 
+    def set(self, **kwargs):
+        return self.get_or_create(**kwargs)
+
+    def unset(self, **kwargs):
+        self.filter(**kwargs).delete()
```
raphendyr: Maybe `get` and `delete` would probably be safer here. Would ensure that only single object is found. Though, not-found needs to be handled.

Also, both should require `course_instance`.

course/models.py
```diff
@@ -189,13 +190,23 @@ def namegen():
...
+
+
 class UserTag(UrlMixin, ColorTag):
     course_instance = models.ForeignKey('CourseInstance', related_name="usertags", on_delete=models.CASCADE)
     visible_to_students = models.BooleanField(default=False)
 
     class Meta:
         ordering = ['course_instance', 'name']
 
+    objects = UserTagManager()
```
raphendyr: If you order this in fields, `objects`, and `Meta`, then it should match rest of the models.

course/models.py
```diff
@@ -564,9 +575,13 @@ def __str__(self):
...
         logger = logging.getLogger("plus.hooks")
+        url, data = url_with_query_in_data(self.hook_url, data)
         try:
-            urllib.request.urlopen(self.hook_url,
-                urllib.parse.urlencode(data).encode('utf-8'), timeout=10)
+            urllib.request.urlopen(
+                url,
+                urllib.parse.urlencode(data).encode('ascii'),
+                timeout=10,
+            )
```
raphendyr: Interesting, I just noticed that we use urllib here and not requests, like we do in every other place. It's fine, just an observation.

course/models.py
JoonatanPartanen: `get_or_create()` and `get()` behave similarly when multiple objects are found. They raise an exception: [`Queryset.get`](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#django.db.models.query.QuerySet.get)

These should have been renamed, I'll do that.
I propose renaming set, leaving it otherwise as it is now, and removing unset.
`unset()` is never used.
`unset()` is not used in taggings either, even though it is defined in the manager:

[definition](https://github.com/JoonatanPartanen/a-plus/blob/master/course/models.py#L235)

[`create()` is overridden, but `delete()` is not](https://github.com/JoonatanPartanen/a-plus/blob/master/course/api/full_serializers.py#L164)

course/models.py
JoonatanPartanen: Done

course/api/views.py
JoonatanPartanen: Will fix

course/api/full_serializers.py
```diff
@@ -109,6 +110,19 @@ class Meta(CourseUsertagBriefSerializer.Meta):
             'taggings',
         )
 
+    def validate(self, data):
+        # Name is enough, slug will be automatically generated 
+        data = super().validate(data)
+        if not data['name']:
+            raise serializers.ValidationError(
+                'At minimum, field \'name\' is required'
```
raphendyr: I think this can be on a single line. Any case, if you use `"` for string, then single quotes don't need to be escaped. It's good to follow [this "rule"](https://stackoverflow.com/a/56190/11090030) when choosing quotes in Python. It makes the code really easy to read :)

raphendyr:
A small styling comment. Oterwise it seems to be good for merging.

raphendyr:
Could you rebase and fix conflicts after #391 has been merged. Thanks.

PR #387, opened 2019-07-08
Improved filtering. Options to filter with comparison operators and regex.

henkka14:
Currently the filtering of results only sees if it can find the typed string inside the data. Jaakko wished that filtering would be improved.

The proposed improvements allow the teacher to filter results with comparison operators (<, >, >=, <=, ==, !=) and create own regex options for filtering.

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -44,7 +44,6 @@ $(function() {
     $('.page-modal').aplusModalLink();
     $('.file-modal').aplusModalLink({file: true});
     $('.search-select').aplusSearchSelect();
-    $('.filtered-table').aplusTableFilter();
 
     // Clear notifications once opened.
     $('#notification-alert li a').on("click", function(event) {
@@ -564,87 +563,6 @@ $(function() {
             }
             if (selectValues) {
                 $.data(this, "plugin_" + pluginName).resetSelection(selectValues);
-            }
-        });
-    };
-})(jQuery);
-
-/**
- * Table row filter.
- */
-(function($) {
-    "use strict";
-
-    const pluginName = "aplusTableFilter";
-    var defaults = {};
-
-    function AplusTableFilter(element, options) {
-        this.element = $(element);
-        this.filters = null;
-        this.timeout = null;
-        if (this.element.prop("tagName") == "TABLE") {
-            this.settings = $.extend({}, defaults, options);
-            this.init();
-        }
-    }
-
-    $.extend(AplusTableFilter.prototype, {
-
-        init: function() {
-            var columnCount = 0;
-            this.element.find('thead').find('tr').each(function() {
-                var count = $(this).find('th').size();
-                columnCount = count > columnCount ? count : columnCount;
-            });
-
-            var self = this;
-            var filterDelay = function(event) {
-                var input = $(this);
-                clearTimeout(self.timeout);
-                self.timeout = setTimeout(function() {
-                    self.filterColumn(input);
-                }, 500);
-            };
-
-            this.filters = [];
-            var filterRow = $('<tr></tr>');
-            for (var i = 0; i < columnCount; i++) {
-                this.filters.push('');
-                var filterInput = $('<input type="text" data-column="' + i + '">')
-                    .on('keyup', filterDelay).on('change', filterDelay);
-                var filterCell = $('<td></td>');
-                filterCell.append(filterInput);
-                filterRow.append(filterCell);
-            }
-            this.element.find('thead').append(filterRow);
-        },
-
-        filterColumn: function(input) {
-            var column = input.attr('data-column');
-            var query = input.val();
-            this.filters[column] = query.trim();
-            this.filterTable();
-        },
-
-        filterTable: function() {
-            var self = this;
-            this.element.find('tbody').find('tr').hide().filter(function() {
-                var pass = true;
-                $(this).find('td').each(function(i) {
-                    if (self.filters[i] && $(this).text().toLowerCase().indexOf(self.filters[i].toLowerCase()) < 0) {
-                        pass = false;
-                        return false;
-                    }
-                });
-                return pass;
-            }).show();
-        }
-    });
-
-    $.fn[pluginName] = function(options) {
-        return this.each(function() {
-            if (!$.data(this, "plugin_" + pluginName)) {
-                $.data(this, "plugin_" + pluginName, new AplusTableFilter(this, options));
             }
         });
     };
--- a/assets/js/filter.js
+++ b/assets/js/filter.js
@@ -1,0 +1,288 @@
+$(function() {
+    "use strict";
+
+    $('.filtered-table').aplusTableFilter();
+});
+
+/**
+ * Table row filter.
+ */
+(function($, document, undefined) {
+    "use strict";
+
+    const pluginName = "aplusTableFilter";
+    var defaults = {};
+    var translationsReady = false;
+
+    function AplusTableFilter(element, options) {
+        this.element = $(element);
+        this.filters = null;
+        this.timeout = null;
+        if (this.element.prop("tagName") == "TABLE") {
+            this.settings = $.extend({}, defaults, options);
+            var self = this;
+
+            // translationsReady allows us to use the plugin multiple times on same page
+            if (translationsReady) {
+                self.init();
+            } else {
+                $(document).on("aplus:translation-ready", function() {
+                    translationsReady = true;
+                    self.init();
+                });
+            }
+        }
+    }
+
+    $.extend(AplusTableFilter.prototype, {
+
+        init: function() {
+            var columnCount = 0;
+            this.element.find('thead').find('tr').each(function() {
+                var count = $(this).find('th').size();
+                columnCount = count > columnCount ? count : columnCount;
+            });
+
+            var self = this;
+            var filterDelay = function(event) {
+                var input = $(this);
+                clearTimeout(self.timeout);
+                self.timeout = setTimeout(function() {
+                    self.filterColumn(input);
+                }, 500);
+            };
+
+            this.filters = [];
+            var filterRow = $('<tr class="tableexport-ignore"></tr>');
+            for (var i = 0; i < columnCount; i++) {
+                this.filters.push('');
+                var filterInput = 
+                    $('<input type="text" data-column="' + i + '">')
+                    .on('keyup', filterDelay).on('change', filterDelay);
+
+
+                var filterCell = $(
+                    '<td data-toggle="tooltip" title="'
+                    + _(
+                        "Comparison operators >, <, >=, <=, !=, == are available — e.g. >=200."
+                        + " Datetime comparison is available in format yyyy-mm-dd hh:mm"
+                        + " — e.g. >=0918-09-20 15:00."
+                        + " Regex is available — e.g. /^\\d+$/g."
+                        + " Use dot as decimal seperator — e.g. 2.1."
+                        )
+                    + '"></td>'
+                );
+                filterCell.append(filterInput);
+                filterRow.append(filterCell);
+            }
+            this.element.find('thead').append(filterRow);
+        },
+
+        filterGetType: function(query) {
+            var isDate = false;
+            if (!query) {
+                // Empty query, no filtering
+                return [undefined, "", query, isDate];
+            }
+
+            if (query.startsWith("<=")) {
+                isDate = checkForDate(query.slice(2));
+                var parsedQuery = (isDate)
+                    ? Date.parse(query.slice(2))
+                    : requireNumber(query.slice(2));
+                var filter = isDate ? lteFilterDate : lteFilter;
+                return [filter, parsedQuery, query, isDate];
+            } else if (query.startsWith("<")) {
+                isDate = checkForDate(query.slice(1));
+                var parsedQuery = (isDate)
+                    ? Date.parse(query.slice(1))
+                    : requireNumber(query.slice(1));
+                var filter = isDate ? ltFilterDate : ltFilter;
+                return [filter, parsedQuery, query, isDate];
+            } else if (query.startsWith(">=")) {
+                isDate = checkForDate(query.slice(2));
+                var parsedQuery = (isDate)
+                    ? Date.parse(query.slice(2))
+                    : requireNumber(query.slice(2));
+                var filter = isDate ? gteFilterDate : gteFilter;
+                return [filter, parsedQuery, query, isDate];
+            } else if (query.startsWith(">")) {
+                isDate = checkForDate(query.slice(1));
+                var parsedQuery = (isDate)
+                    ? Date.parse(query.slice(1))
+                    : requireNumber(query.slice(1));
+                var filter = isDate ? gtFilterDate : gtFilter;
+                return [filter, parsedQuery, query, isDate];
+            } else if (query.startsWith("==")) {
+                isDate = checkForDate(query.slice(2));
+                var parsedQuery = (isDate)
+                    ? Date.parse(query.slice(2))
+                    : query.slice(2);
+                var filter = isDate ? eFilterDate : eFilter;
+                return [filter, parsedQuery, query, isDate];
+            } else if (query.startsWith("!=")) {
+                isDate = checkForDate(query.slice(2));
+                var parsedQuery = (isDate)
+                    ? Date.parse(query.slice(2))
+                    : query.slice(2);
+                var filter = isDate ? neFilterDate : neFilter;
+                return [filter, parsedQuery, query, isDate];
+            } else if (query.match(/\//g) && query.match(/\//g).length >= 2) {
+                var parts = query.split('/');
+                var regex = query;
+                var options = "";
+
+                if (parts.length > 1) {
+                    regex = parts[1];
+                    options = parts[2];
+                }
+                try {
+                    // If no error is thrown here, the query is valid regex
+                    return [regexFilter, new RegExp(regex, options), query, isDate];
+                } catch (e) {
+                    // If the query is not valid regex, use normal filter
+                    isDate = checkForDate(query);
+                    return [normalFilter, undefined, query, isDate];
+                }
+            } else {
+                isDate = checkForDate(query);
+                return [normalFilter, query, query, isDate];
+            }
+        },
+
+        filterColumn: function(input) {
+            const column = input.data('column');
+            const query = input.val();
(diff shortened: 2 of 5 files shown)


Plustest:

raphendyr:
add to whitelist

templates/base.html
```diff
@@ -45,7 +45,9 @@
         <link rel="stylesheet" href="{% static 'css/submission.css' %}" />
         <link rel="stylesheet" href="{% static 'django_colortag.css' %}" />
         <script src="{% static 'js/polyfill-find.js' %}"></script>
-        <script src="{% static 'js/aplus.js' %}"></script>
+        <script src="{% static 'js/aplus.js' %}"
+                filterHelperTrans="{% trans 'Comparison operators >, <, >=, <=, !=, == are available — i.e. >=200. Use dot as decimal seperator — i.e. 2.1. Regex is also available — i.e. /^\\d+$/g' %}"
```
raphendyr: Name like `data-filter-tooltip-text` would discribe the purpose a bit better and also would be in more correct format. It is read using `currentScript.data('filter-tooltip-text')` or `currentScript.data('filterTooltipText')` for example (or `currentScript.attr('data-filter-tooltip-text')`.

p.s.
> the name must not contain capital A to Z letters.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*

assets/js/aplus.js
```diff
@@ -13,6 +13,10 @@
  window.CustomEvent = CustomEvent;
 })();
 
+let currentScript =  (document.currentScript ?
```
raphendyr: This shouldn't be in the global namespace. Move it inside the local namespace.

assets/js/aplus.js
```diff
@@ -584,65 +588,182 @@ $.fn.highlightCode = function(options) {
...
+
+        this.filters = [];
+        var filterRow = $('<tr class="tableexport-ignore"></tr>');
+        for (var i = 0; i < columnCount; i++) {
+          this.filters.push('');
+          var filterInput = $('<input type="text" data-column="'+i+'">')
+            .on('keyup', filterDelay).on('change', filterDelay);
+
+
+          var filterCell = $('<td data-toggle="tooltip" title="' + currentScript.attr("filterHelperTrans") + '"></td>');
```
raphendyr: This could use the JS translation api and thus have only `_("english test")` instead of using currentScript and django translations. A+ uses the translation api for usertags, so you can look how those are done. I think we could have two json files, "common" and "course staff".

assets/js/aplus.js
raphendyr: Actually, maybe not. We can move the translations to use JS api, so we don't need to pass arguments to single component.

assets/js/aplus.js
```diff
@@ -584,65 +588,182 @@ $.fn.highlightCode = function(options) {
...
+            }
+          }
+
+          else {
+            return [this.normalFilter, query];
+          }
+      },
+
+      lteFilter: function(text, query) {
+        return isNaN(text) || isNaN(query) ? text <= query :
```
raphendyr: `isNaN(query)` and `parseFloat(query)` should be done in `filterGetType`

raphendyr: I see the problem that the query is `<10` and the sell value is `foobar`. Good question is what should be done then. Should we show warning in the filter box "you can't use numeric filters with non-numeric data" or so.

assets/js/aplus.js
```diff
@@ -584,65 +588,182 @@ $.fn.highlightCode = function(options) {
...
-              if (self.filters[i] && $(this).text().toLowerCase().indexOf(self.filters[i].toLowerCase()) < 0) {
-                pass = false;
-                return false;
-              }
-            });
-            return pass;
-          }).show();
-        }
-    });
+     // Commented out inner if. Could not use the plugin second time in same page otherwise.
```
raphendyr: Don't comment out code with git, just remove it. The change is visible here :)

assets/js/aplus.js
```diff
@@ -584,65 +588,182 @@ $.fn.highlightCode = function(options) {
...
+      regexFilter: function(text, query) {
+        var parts = query.split('/');
+        var regex = query;
+        var options = "";
+        if (parts.length > 1) {
+            regex = parts[1];
+            options = parts[2];
+        }
+
+        let regExp = new RegExp(regex, options);
```
raphendyr: `regExp` should be build only once, and then used again and again. The same thing, build regExp in `filterGetType`

raphendyr:
Please fix indentation to match the rest of the file. Keeps changes more visible.

assets/js/aplus.js
henkka14: Why do we want two json files? Currently there is `teacher.fi.json` for translations.

assets/js/aplus.js
henkka14: Check for isNaN is done in filterGetType. Still unsure if there are need for the extra warning. Obviously not hard to implement.

Currently the query 10 would be turned to number 10, since it has only digits. If cell value is "foobar", then comparison "foobar" < 10 tries to convert "foobar" to number, giving NaN and the evaluation is false always. I.E. both "foobar" > 10 and "foorbar" < 10 results in false currently.

assets/js/aplus.js
henkka14: Done.

assets/js-translations/common.fi.json
```diff
@@ -0,0 +1,2 @@
+{
```
raphendyr: No need to add this here yet.. Only when we actually start to use it :)

assets/js-translations/course_staff.fi.json
```diff
@@ -0,0 +1,7 @@
+{
```
raphendyr: No need to rename the file, even though the original file isn't perfectly named.

assets/js/aplus.js
```diff
@@ -1,89 +1,92 @@
...
-  const bubbles = params.bubbles !== undefined ? params.bubbles : false;
-  const cancelable = params.cancelable !== undefined ? params.cancelable : false;
-  const detail = params.detail !== undefined ? params.detail : undefined;
-  const evt = document.createEvent( 'CustomEvent' );
-  evt.initCustomEvent(event, bubbles, cancelable, detail);
-  return evt;
- }
- CustomEvent.prototype = window.Event.prototype;
- window.CustomEvent = CustomEvent;
+(function() {
```
raphendyr: Please don't edit indentation, when it's not required to for the PR.

assets/js/filter.js
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+
+  $.fn[pluginName] = function(options) {
+    return this.each(function() {
+      $.data(this, "plugin_" + pluginName, new AplusTableFilter(this, options));
+    });
+  };
+
+})(jQuery, window, document);
```
raphendyr: Can you add `/* vim: set et ts=2 sw=2: */` to the end of the file and also add a new line (check the symbol at the end of the above line).

Mankro:
If there are more than 100 submissions in the all submissions page of an exercise, does the filtering apply only to the submissions currently visible in the page? The JS code does not seem to query the server, so it only uses the data available in the HTML at that time. However, users would assume that filters in the submission list page query all submissions to the exercise and not only the ones that are initially listed. (If I remember correctly, the page at first only shows the 50 most recent submissions in the table.)

If the user enters an invalid regex (e.g., a small typo in the syntax), then there is no warning that the filter could not be parsed as regex and I guess it then behaves like normal substring matching. It could be nice to warn the user if the query starting with `/` can not be parsed as a regex.

It would be nice to sort tables (ascending or descending) by a column selected by the user and it is somewhat connected to filtering them, though it is a bit different topic. Is it feasible to add sorting here?

If the table column contains dates, such as the submission times in the submission list page, the current filters do not really handle dates. You can currently filter dates like normal strings, but it would be useful to filter dates with other comparisons (e.g., dates greater/less than the 2nd of January 2019). Dates in the Unix timestamp format (epoch) are comparable as integers, but users just don't handle dates in that format, so this would require some extra work to implement well.

Mankro:
I think I found a bug. Filters like `>0` and `==0` do not behave correctly when used in a column that contains zeros and ones. Zero in the filter seems to break something.

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+
+        filterGetType: function(query) {
+            if (!query) {
+                // Empty query, no filtering
+                return [undefined, ""];
+            }
+
+            var querySliceOne = query.slice(1);
+            var querySliceTwo = query.slice(2);
+            querySliceOne = isNaN(querySliceOne) ? querySliceOne : parseFloat(querySliceOne);
```
raphendyr: Variables like "one" and "two" are not explaining what they are. You sohuld name them a bit better.

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+
+/**
+ * Table row filter.
+ */
+(function($, window, document, undefined) {
+    "use strict";
+
+    var pluginName = "aplusTableFilter";
+    var defaults = {};
+    var translationsReady = false;
```
raphendyr: Hmm... Maybe there should be a global status register. Then events like `aplus:translation-ready` would indicate that global `aplus.translationReady` variable has changed.

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+        this.timeout = null;
+        if (this.element.prop("tagName") == "TABLE") {
+            this.settings = $.extend({}, defaults, options);
+            var self = this;
+
+            // translationsReady allows us to use the plugin multiple times on same page
+            if (translationsReady) {
+                self.init();
+            }
+            $(document).on("aplus:translation-ready", function() {
```
raphendyr: Logically this should be inside an else block.

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+                return [undefined, ""];
+            }
+
+            var querySliceOne = query.slice(1);
+            var querySliceTwo = query.slice(2);
+            querySliceOne = isNaN(querySliceOne) ? querySliceOne : parseFloat(querySliceOne);
+            querySliceTwo = isNaN(querySliceTwo) ? querySliceTwo : parseFloat(querySliceTwo);
+
+            if (query.startsWith("<")) {
+                if (query.startsWith("<=")) {
```
raphendyr: I would put these on the same level. For example,

```
if (query.startsWith("<=")) {
  ...
} else if (query.startsWith("<")) {
  ...
} else if (query.startsWith(">="){
  ...
```

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+                    return [this.lteFilter, querySliceTwo];
```
raphendyr: I would create a helper `function require_number(string) { return isNaN(string) ? undefined : parseFloat(string); }` and then here `return [this.lteFilter, require_number(query.slice(2)];`

raphendyr: That helper can then be extended to parse dates and to return error, date (epoch seconds) or a float.

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+                if (parts.length > 1) {
+                    regex = parts[1];
+                    options = parts[2];
+                }
+                try {
+                    // If no error is thrown here, the query is valid regex
+                    return [this.regexFilter, new RegExp(regex, options)];
+                } catch (e) {
+                    // If the query is not valid regex, use normal filter
+                    return [this.normalFilter, query];
```
raphendyr: We need a way to get this error to the user. Same for the number cells. We need to indicate in some way if the user made a mistake.

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+                }
+
+            } else {
+                return [this.normalFilter, query];
+            }
+
+        },
+
+        lteFilter: function(text, query) {
```
raphendyr: You can move this out of the class, but keep them inside the anonymous function scope. E.g. put these below the "class".

assets/js/filter.js
```diff
@@ -0,0 +1,196 @@
...
+            this.filters[column] = this.filterGetType(query.trim());
+            this.filterTable();
+        },
+
+        filterTable: function() {
+            var self = this;
+            this.element.find('tbody').find('tr').not('.no-filtering').hide().filter(function() {
+                var pass = true;
+                $(this).find('td').each(function(i) {
+                    if (self.filters[i][1] && !self.filters[i][0].apply(this, [$(this).text().trim(), self.filters[i][1]])) {
```
raphendyr: Can you break these to vars `const function = self.filters[i][0], argument = self.filters[i][1];`

raphendyr:
I think we can support dates relativel easily. If the date is in format `yyyy-dd-mm hh:mm:ss` (or any part starting from left), then we can parse that to a epoch seconds. 

Filter functions could take three parameters: original string, parsed value (epoch seconds from date or float for number fields) and the input cell argument. Comparsion functions would use the parsed value, while string and regexp would use the original one. In theory, you could parse the cell values once and store the parsed value in the elements (`if (this._parsed === undefined) this._parsed = parse($(this).text().trim()); return this._parsed`). Last part is not needed, just pointing out.

We should implement some way to show errors on input fields. I think bootstrap tooltip could be used for this (with open when called). When you write `<foo` we could add tooltip "foo is not a number and thus can't be used for comparison".

henkka14:
> I think I found a bug. Filters like `>0` and `==0` do not behave correctly when used in a column that contains zeros and ones. Zero in the filter seems to break something.

Ah good catch. Apparently filtering with zero as query was not working, since I was checking if the query is defined in the filter with simple `if (self.filter[i][1] && ....) ` in case of empty string. However, this also resulted for no filtering with if self.filter[i][1] is simply 0, which is not what we wanted. This was easy fix.

assets/js/filter.js
henkka14: I added popover that is shown if user tries to use strings on filter or the regex if invalid.

henkka14:
I added initial support for datetime. Please note that the current format that the table data uses for time stamp does not support Date.parse, thus it will need to reformatted to support the comparison.

There is now a popover that is always on with helper text, when the code detects possible issues on filter.

To answer the question on pagination. The submission page paginates 50 results per page, however if the pagination is in use, there is currently no filtering available as seen in `_submissions_table.html`. There exists is an option to Show all results, and after selecting this option then filtering is enabled. This was also the reason why Jaakko was wondering why he did not see filtering that one time when we were discussing this topic.

I fixed the bug with zero and also changed other parts according to the reviews. Feel free to review the current version :)

Mankro:
OK, so the dates in the submission table can not be parsed yet, but they could be compared otherwise. The date shown in the table should be easy for humans to read, so maybe we shouldn't change that date format. The machine-readable timestamps could be added to the data as data attributes in the server so that JS could easily read them. Alternatively, we could use a JS library that is able to reliably parse a wider range of date formats. Thoughts?

henkka14:
> OK, so the dates in the submission table can not be parsed yet, but they could be compared otherwise. The date shown in the table should be easy for humans to read, so maybe we shouldn't change that date format. The machine-readable timestamps could be added to the data as data attributes in the server so that JS could easily read them. Alternatively, we could use a JS library that is able to reliably parse a wider range of date formats. Thoughts?

I used the idea of adding the date as an attribute. I add the date in a format that is readable by Date.parse JS function to the td elements in _submissions_table.html. The format is parsed from the original submission_time with the Django's built-in date template filter.

exercise/templates/exercise/staff/_submissions_table.html
```diff
@@ -61,7 +61,7 @@
             <td>
                 {% profiles submission.submitters.all instance is_teacher %}
             </td>
-            <td>
+            <td datetime="{{ submission.submission_time|date:'Y-m-d H:i:s' }}">
```
Mankro: Is `datetime` supposed to be a data attribute `data-datetime`? Is it not invalid HTML to use custom non-standard attributes? Even if this works, a data attribute would be the correct approach according to the web standards, right?

raphendyr: Yep. This needs to be  prefixed with `data-`.

henkka14: Aah so that is why the `data-` prefix is used, did not know the reason for using data attributes.

I assume we do not want to use the <time> tag that has its own datetime attribute (which I did now know about when I used my custom datetime attribute).

Mankro: The data attribute is fine. The `<time>` tag is new in HTML5 and it would fit semantically to this usecase, but it is not supported by IE and offers no real benefit over the data attribute. Note that jQuery can read data attributes with the `attr()` method and also the special `data()` method that does not take the data prefix in the attribute name. You can also change the stored data without modifying the actual attribute at runtime, in which case jQuery stores the data key-value pairs elsewhere in memory.

https://api.jquery.com/data/

raphendyr:
I merged #388, so can you rebase this on top of master.

raphendyr:
I think this is pretty goiod now. `isDate` could select one of two implementations for the filters, so the final if/else wouldn't be inside the filter. That is pretty minor, so doesn't matter for now.

assets/js/filter.js
henkka14: Sounds good idea for the future. I also ran into this issue with #390 where some translations require me to wait for the aplus:translation-ready event and others do not require it. It does not seem correct that some translations inside a file wait for the event and others do not, because translations become ready while the script is running. I think this could also bugs with translations in the future.

henkka14:
Please review the current changes. Not making modifications currently unless until next comments.

assets/js/filter.js
raphendyr: It would be amazing if we had implementation for `_`, which returns lazy object. When ever that object is converted to a string, it would request the translation. Then only render actions would need to wait for translations to become ready,

raphendyr: p.s. jQuery ready implements basically the same as here: https://github.com/jquery/jquery/blob/master/src/core/ready.js#L69

raphendyr: I'm pretty sure we need to implement these in A+ JS API, like `aPlus.after('translations-ready', self.init);`. TODO: aplus js api :)

raphendyr:
I think this is good for merging.

Mankro:
I think it is good, assuming that the final changes didn't break anything that worked previously :)

PR #388, opened 2019-07-12
Wrap all functions with anonymous functions inside aplus.js.

henkka14:
Wrap the remaining plugins inside aplus.js with anonymous functions. The closures help to keep the scope less populated inside the aplus.js.

Requested by Jaakko.

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -1,31 +1,38 @@
-/* add CustomEvent for IE 11 */
-(function () {
- if (typeof window.CustomEvent === "function") return false;
- function CustomEvent(event, params) {
-  const bubbles = params.bubbles !== undefined ? params.bubbles : false;
-  const cancelable = params.cancelable !== undefined ? params.cancelable : false;
-  const detail = params.detail !== undefined ? params.detail : undefined;
-  const evt = document.createEvent( 'CustomEvent' );
-  evt.initCustomEvent(event, bubbles, cancelable, detail);
-  return evt;
- }
- CustomEvent.prototype = window.Event.prototype;
- window.CustomEvent = CustomEvent;
-})();
+/**
+ * Add CustomEvent for IE 11
+ */
+(function(window, document, undefined) {
+    "use strict";
+
+    if (typeof window.CustomEvent === "function") return false;
+
+    function CustomEvent(event, params) {
+        const bubbles = params.bubbles !== undefined ? params.bubbles : false;
+        const cancelable = params.cancelable !== undefined ? params.cancelable : false;
+        const detail = params.detail !== undefined ? params.detail : undefined;
+        const evt = document.createEvent('CustomEvent');
+        evt.initCustomEvent(event, bubbles, cancelable, detail);
+        return evt;
+    }
+    CustomEvent.prototype = window.Event.prototype;
+    window.CustomEvent = CustomEvent;
+})(window, document);
+
 
 $(function() {
-
+    "use strict";
+ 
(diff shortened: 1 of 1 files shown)


raphendyr:
Not all of the changed lines follow the indentation rules. Feel free to fix indentation for all of the effected blocks.

p.s. Ths file uses a lot of different indentation (sadly). For interested, rundown is 383 lines indented with 4 spaces, 105 with 2, 68 with 3 and 31 didn't match any of those. Thus, we should continue with 4 (numbers are from commit before this PR).

assets/js/aplus.js
```diff
@@ -86,217 +68,240 @@ $(function() {
...
+        const bubbles = params.bubbles !== undefined ? params.bubbles : false;
+        const cancelable = params.cancelable !== undefined ? params.cancelable : false;
+        const detail = params.detail !== undefined ? params.detail : undefined;
+        const evt = document.createEvent('CustomEvent');
+        evt.initCustomEvent(event, bubbles, cancelable, detail);
+        return evt;
+    }
+    CustomEvent.prototype = window.Event.prototype;
+    window.CustomEvent = CustomEvent;
+})(jQuery, window, document);
```
raphendyr: This should be the first thing on the page. The function wrap doesn't require any parameters as it's just polyfill. Not sure why it's different, but it should be from here: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill

raphendyr: Ah. Yeah. I wanted defaults for all parameters and not as a whole.

assets/js/aplus.js
```diff
@@ -86,217 +68,240 @@ $(function() {
...
+                });
 
-      var clipboard = new Clipboard('#copy-button-' + copyTargetCounter);
-      clipboard.on("error", function(e) {
-          hiddenTextarea.hide();
-      });
-      clipboard.on("success", function(e) {
-          hiddenTextarea.hide();
-      });
+                var clipboard = new Clipboard('#copy-button-' + copyTargetCounter);
```
raphendyr: Can you also replace `var`s that are really `const`s with consts?

raphendyr: At least for short and simple parts of the code (places where it's super clear when just scrolling the document.

assets/js/aplus.js
henkka14: I added the same format for all function wraps as some functions already had "useless" parameters that they didn't use. Should we keep the CustomEvent wrap parameterless and stick to the

raphendyr: I have no idea why it's there. To my knowledge, it doesn't serve any purpose. Personally, all scopes should include only the variables they need. This documents the external interface the function depends on.

henkka14: Apparently it is used to prevent overriding of undefined. However, since we are using strict mode and it is supported in IE11, I will remove undefined from parameter list alongside with parameters that are not used in the scope.

https://stackoverflow.com/questions/33383647/javascript-undefined-as-a-function-parameter

henkka14: Actually, I think having the functions in strict mode would still not prevent defining undefined as variable outside of the function, so maybe I will just leave the functions to take undefined as parameter.

raphendyr: Cool that you found the reason for it. Makes sence.. well, considering javascript is a pretty stupid language :D

raphendyr:
Yeah. I think it's good. Thanks :)

PR #389, opened 2019-07-19
Validation for status

alalav1:


Plustest:

Mankro:
Now the status values are validated when the course JSON is imported, but not elsewhere. Let's try adding the validation to the model classes next in their `clean` methods. In addition, the Enum class could have a new values method for reading the list of valid values so that you do not need to repeatedly create that list elsewhere when checking the status.

Mankro:
Actually, Django 2.2 automatically validates the choices of the model fields when the choices parameter is used in the model field definition. This has changed since the old Django 1.9 according to our tests and the wording in the Django documentation. Therefore, this PR does not need to check at all whether the given status is in the list of accepted values.

https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices

alalav1:
Since Django does validation automatically, purpose of this pull request is to remove hardcoded values and remove unnecessary length requirements from [configure.py](https://github.com/Aalto-LeTech/a-plus/blob/master/edit_course/operations/configure.py). Also 2 methods added to Enum class in helper.py by @raphendyr .

Mankro:
Upon further inspection, the old Django 1.9 actually seems to also validate model field choices automatically, but the old A+ code did not call `model.full_clean()` in configure.py (course JSON import), so that is why the the current production version does not validate the status choices and the master version does (full_clean was added to master in March 2019).

Mankro:
@alalav1 Please squash the commits. Don't forget to fix the final commit message since there is a co-author there too.

PR #390, opened 2019-07-26
Improved the results page for end of course for teachers.

henkka14:
Currently the Results page for teachers is very minimal and quite frankly, useless. Tasked by Juho and Jaakko, I have been making some improvements to it. Added functionality to export the table data. Option to select from the exercise pool. Option to filter based on tags. Option to see different data indicators.

NOTE: Filtering relies on my other pull requests. I would also hope someone could test this with more student data. I have not found a way to test this reliably on large student result data.

### DIFF
--- a/assets/js-translations/teacher.fi.json
+++ b/assets/js-translations/teacher.fi.json
@@ -1,9 +1,104 @@
 {
-  "Toggle filtering by tag": "Suodata merkinnällä",
-  "Remove tagging": "Poista merkintä",
-  "Add new tagging": "Lisää merkintä",
-  "Comparison operators >, <, >=, <=, !=, == are available — e.g. >=200. Datetime comparison is available in format yyyy-mm-dd hh:mm:ss — e.g. >=0918-09-20 15:00. Regex is available — e.g. /^\\d+$/g. Use dot as decimal seperator — e.g. 2.1.":
-  "Vertailuoperaattorit >, <, >=, <=, !=, == ovat käytettävissä, esimerkiksi >=200. Ajan vertailu on mahdollista muodossa yyyy-mm-dd hh:mm:ss, esimerkiksi >=0918-09-20 15:00. Regex on myös käytettävissä, esimerkiksi /^\\d+$/g. Käytä desimaalierottimena pistettä, esimerkiksi 2.1.",
+  "Toggle filtering by tag":
+  "Suodata merkinnällä",
+
+  "Remove tagging":
+  "Poista merkintä",
+
+  "Add new tagging":
+  "Lisää merkintä",
+
+  "Comparison operators >, <, >=, <=, !=, == are available — e.g. >=200. Datetime comparison is available in format yyyy-mm-dd hh:mm — e.g. >=0918-09-20 15:00. Regex is available — e.g. /^\\d+$/g. Use dot as decimal seperator — e.g. 2.1.":
+  "Vertailuoperaattorit >, <, >=, <=, !=, == ovat käytettävissä, esimerkiksi >=200. Ajan vertailu on mahdollista muodossa yyyy-mm-dd hh:mm, esimerkiksi >=0918-09-20 15:00. Regex on myös käytettävissä, esimerkiksi /^\\d+$/g. Käytä desimaalierottimena pistettä, esimerkiksi 2.1.",
+
   "There might be something wrong with the filter. Some comparison operators only accept numbers. If you are using datetime, make sure the format is correct. If you are using regex, make sure it is correct.":
-  "Suodattimissa voi olla virhe. Jotkut vertailuoperaattorit hyväksyvät vain numeroita. Mikäli vertailet aikaa, varmista että formaatti on oikea. Mikäli käytät regexiä, niin varmista että se on oikein."
+  "Suodattimissa voi olla virhe. Jotkut vertailuoperaattorit hyväksyvät vain numeroita. Mikäli vertailet aikaa, varmista että formaatti on oikea. Mikäli käytät regexiä, niin varmista että se on oikein.",
+
+  "Total":
+  "Yhteensä",
+
+  "Tags":
+  "Merkinnät",
+
+  "Student name":
+  "Opiskelijan nimi",
+
+  "Student ID":
+  "Opiskelijanumero",
+
+  " students selected":
+  " opiskelijaa valittuna",
+
+  "No difficulty":
+  "Ei vaikeusastetta",
+
+  "Total submissions":
+  "Palautuksia yhteensä",
+
+  "Average submissions per student with submissions":
+  "Palautusten keskiarvo per opiskelija, jolla on vähintään yksi palautus",
+
+  "Maximum submissions":
+  "Palautusmahdollisuuksia",
+
+  "Students with submissions":
+  "Tehtävän palauttaneet opiskelijat",
+
+  "Students with max points":
+  "Opiskelijoiden lkm joilla on täydet pisteet",
+
+  "Average points per student with submissions":
+  "Pisteiden keskiarvo per opiskelija jolla on vähintään yksi palautus",
+
+  "Maximum points":
+  "Enimmäispisteet",
+
+  "Total number of submissions. Calculates all student submission counts together.":
+  "Palautusten lukumäärä. Laskee kaikkien opiskelijoiden kaikkien palautusten lukumäärän yhteen.",
+
+  "How many submissions a single student has used on the exercise on average. Only accounts for students with one or more submissions.":
+  "Kuinka monta palautusta opiskelija on käyttänyt tehtävään keskimäärin. Lasketaan vain opiskelijat, joilla on vähintään yksi palautus.",
+
+  "Maximum number of available submissions for the exercise.":
+  "Kuinka monta palautusta tehtävässä on mahdollista käyttää.",
+
+  "Number of students that have one or more exercise submissions.":
+  "Opiskelijoiden lukumäärä, joilla on vähintään yksi palautus.",
+
+  "Number of students that have received maximum points from the exercise.":
+  "Opiskeljoiden lukumäärä, jotka ovat saaneet täydet pisteet tehtävästä.",
+
+  "Average points received for the exercise. Only accounts for students with one or more submissions.":
+  "Kuinka monta pistettä opiskelijat ovat saaneet tehtävästä keskimäärin. Lasketaan vain opiskelijat, joilla on vähintään yksi palautus.",
+
+  "Maximum points for the exercise.":
+  "Tehtävän enimmäispisteet.",
+
+  "Export to xlsx (Excel)":
+  "Vie xlsx tiedostoon (Excel)",
+
+  "Export to csv (LibreOffice)":
+  "Vie csv tiedostoon (LibreOffice)",
+
+  "Export to txt":
+  "Vie txt tiedostoon",
+
+  "None selected":
+  "Nolla valintaa",
+
+  "All selected":
+  "Kaikki valittu",
+
+  "selected":
+  "valittu",
+
+  "Select all":
+  "Valitse kaikki",
+
+  "student_results":
+  "opiskelijoiden_tulokset",
+
+  "Some students have unofficial points that are greater than zero. However, they are calculated as zero in the table until they are confirmed. This may be caused for example by missing feedback form.":
+  "Osalla opiskelijoista on epävirallisia pisteitä, jotka ovat enemmän kuin nolla. Nämä pisteet on silti laskettu nollana taulukossa, kunnes ne on vahvistettu. Tämä voi johtua esimerkiksi puuttuvasta palautelomakkeesta."
+
 }
--- a/exercise/staff_views.py
+++ b/exercise/staff_views.py
@@ -12,12 +12,15 @@ from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 
 from course.viewbase import CourseInstanceBaseView, CourseInstanceMixin
+from course.models import (
+    USERTAG_EXTERNAL,
+    USERTAG_INTERNAL,
+)
 from deviations.models import MaxSubmissionsRuleDeviation
 from lib.helpers import settings_text
 from lib.viewbase import BaseRedirectView, BaseFormView, BaseView
 from notification.models import Notification
 from authorization.permissions import ACCESS
-from .exercise_summary import ResultTable
 from .models import LearningObject
 from .forms import (
     SubmissionReviewForm,
@@ -178,8 +181,11 @@ class AllResultsView(CourseInstanceBaseView):
 
     def get_common_objects(self):
         super().get_common_objects()
-        self.table = ResultTable(self.instance)
-        self.note("table")
+        self.tags = [USERTAG_INTERNAL, USERTAG_EXTERNAL]
+        self.tags.extend(self.instance.usertags.all())
+        self.note(
+            'tags',
+        )
 
 
 class AnalyticsView(CourseInstanceBaseView):
--- a/exercise/static/exercise/results_staff.js
+++ b/exercise/static/exercise/results_staff.js
@@ -1,0 +1,833 @@
+(function($, document, window, undefined) {
+    let currentScript = (document.currentScript ?
+        $(document.currentScript) :
+        $('script').last()); // Ugly solution for IE11
+
+    const exercisesUrl = currentScript.data("exercisesUrl");
+    const studentsUrl = currentScript.data("studentsUrl");
+    const usertagsUrl = currentScript.data("usertagsUrl");
+    const pointsUrl = currentScript.data("pointsUrl");
+
+    let _exerciseSelection;
+    let _allExercises = [];
+    let _exercises;
+    let _students;
+    let _users = {};
+    let _usertags;
+    let _points = {};
+    let _ajaxCompleted = false;
+
+    let tableExportVar;
+
+    /* TODO: Use better logic for translations.
+     * Currently only some translations wait for aplus:translation-ready and most of them do not,
+     * since the translations become available during the script.
+     */
+    $(document).on("aplus:translation-ready", function() {
+        // TableExport plugin
+        // https://tableexport.v5.travismclarke.com/#tableexport
+        TableExport.prototype.formatConfig.xlsx.buttonContent = _("Export to xlsx (Excel)");
+        TableExport.prototype.formatConfig.csv.buttonContent = _("Export to csv (LibreOffice)");
+        TableExport.prototype.formatConfig.txt.buttonContent = _("Export to txt");
+        tableExportVar = $("#table-points").tableExport({
+            "position": "top",
+            "bootstrap": "true",
+            "filename": _("student_results"),
+        });
+
+        // Move the default TableExport download buttons inside a single dropdown menu
+        $("caption.tableexport-caption").children("button").each(function() {
+            $("#export-button-menu").append($(this).detach());
+        });
+    });
+
+    // Make the table headings and student IDs stick when scrolling table
+    $('#table-points-div').scroll(function(ev) {
+        $('thead#table-heading th').css('transform', 'translateY(' + this.scrollTop + 'px)');
+        $('tbody td.stick-on-scroll').css('transform', 'translateX(' + this.scrollLeft + 'px)');
+    });
+
+
+    // Booleans to allow teacher to choose with checkboxes what data indicators they want to see
+    let _totalSubmTrue = false;
+    let _avgSubmTrue = false;
+    let _maxSubmTrue = false;
+    let _totalStuSubmTrue = false;
+    let _totalStuMaxTrue = false;
+    let _avgPTrue = false;
+    let _maxPTrue = false;
+    let _showOfficial = true;
+
+    $("input.total-subm-checkbox").change(function() {
+        _totalSubmTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.avg-subm-checkbox").change(function() {
+        _avgSubmTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.max-subm-checkbox").change(function() {
+        _maxSubmTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.total-stu-subm-checkbox").change(function() {
+        _totalStuSubmTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.total-stu-max-checkbox").change(function() {
+        _totalStuMaxTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.avg-p-checkbox").change(function() {
+        _avgPTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.max-p-checkbox").change(function() {
+        _maxPTrue = this.checked;
+        exerciseSelectionChange();
+    });
+
+    $("input.official-checkbox").change(function() {
+        _showOfficial = this.checked;
+        exerciseSelectionChange();
+    });
+
+
+    // Set up the tooltip for checkboxes, look docs below for tooltip
+    // https://getbootstrap.com/docs/3.3/javascript/
+    $('[data-toggle="tooltip"]').tooltip({
+        "trigger": "hover",
+    });
+
+    $('[data-toggle="tooltip"]').on('click', function() {
+        $(this).tooltip('hide');
+    })
+
+
+    // Event listener for tag filters
+    $('.filter-users button').on('click', function(event) {
+        event.preventDefault();
+        let icon = $(this).find('.glyphicon');
+        if (icon.hasClass('glyphicon-unchecked')) {
+            icon.removeClass('glyphicon-unchecked').addClass('glyphicon-check');
+        } else {
+            icon.removeClass('glyphicon-check').addClass('glyphicon-unchecked');
+        }
+        exerciseSelectionChange();
+    });
+
+    /*
+     * Creates the html for indicator data row. This is called for each row that has
+     * checkbox checked for its indicator data.
+     * @param  {string} tooltipTitle The text that is shown when hovering over heading.
+     * @param  {string} headingTitle The heading text.
+     * @return {string} Returns the html that creates the data indicator rows in table.
+     */
+    function createIndicatorRow(tooltipTitle, headingTitle, dataValues) {
+        // Data indicator headings have title and tooltip info that is shown when hovering mouse on the heading title
+        let indicatorHeadingHtml = "";
+        if (dataValues && dataValues.length > 0) {
+            indicatorHeadingHtml +=
+                '<tr class="no-filtering"><td style="border-right: none !important; border-bottom: none !important;"' +
+                'class="indicator-heading stick-on-scroll" data-toggle="tooltip" ' +
+                'title="' + tooltipTitle + '">' + headingTitle +
+                '</td><td style="border-left: none !important; border-bottom: none !important;"' +
+                'class="indicator-heading stick-on-scroll" data-toggle="tooltip" ' +
+                'title="' + tooltipTitle + '"></td>';
+        } else {
+            indicatorHeadingHtml +=
+                '<tr class="no-filtering"><td style="border-right: none !important;" colspan="2"' +
+                'class="indicator-heading stick-on-scroll" data-toggle="tooltip" ' +
+                'title="' + tooltipTitle + '">' + headingTitle +
+                '</td><td style="border-left: none !important; border-top: none !important;"' +
+                'class="indicator-heading stick-on-scroll" data-toggle="tooltip" ' +
+                'title="' + tooltipTitle + '"></td>';
+        }
+
+        let normalValuesHtml = '';
+        let pctValuesHtml =
+            '<tr class="no-filtering tableexport-ignore">'
+        let sumValue = 0;
+        dataValues.forEach(function(value) {
+            sumValue += value[0];
+        });
+
+        // Empty data cells for total and tags in data indicators
+        if (dataValues && dataValues.length > 0) {
+            normalValuesHtml +=
+                '<td class="indi-normal-val"></td>'
+                + '<td class="indi-normal-val">' + sumValue + '</td>';
+            pctValuesHtml +=
+                '<td style="border-right: none !important; border-top: none !important;"'
+                + 'class="indicator-heading stick-on-scroll" data-toggle="tooltip" '
+                + 'title="' + tooltipTitle + '"></td>'
+                + '<td style="border-left: none !important; border-top: none !important;"'
+                + 'class="indicator-heading stick-on-scroll" data-toggle="tooltip" '
+                + 'title="' + tooltipTitle + '"></td>'
+                + '<td class="indi-pct-val"></td>'
+                + '<td class="indi-pct-val"></td>';
+        }
+
+        // Normal and percentage values on seperate rows in data indicators
+        dataValues.forEach(function(value) {
+            normalValuesHtml += '<td class="indi-normal-val">' + value[0] + '</td>';
+            pctValuesHtml += '<td class="indi-pct-val">' + value[1].toFixed(2) + '%</td>';
+        });
+
+        normalValuesHtml += '</tr>';
+        pctValuesHtml += '</tr>';
+
+        return indicatorHeadingHtml + normalValuesHtml + pctValuesHtml;
+    }
+
+
+    /*
+     * Creates the table with student points based on
+     * selected exercises, data indicators, tags and grouping method.
+     * @param {string} showMethod The grouping method: show all, show by difficulties, show by modules
+     */
+    window.createPointTable = function(showMethod) {
+        if (!_ajaxCompleted) {
+            return;
+        }
+
+        // Pick only students that have the selected tags
+        // Use same logic for tag filtering as in participants.js
+        let = filteredStudentPool = [];
+        _students.results.forEach(function(student) {
+            const tagSlugFilters = $.makeArray($('.filter-users button:has(.glyphicon-check)'))
+            .map(function(elem) {
+                return $(elem).data('tagSlug');
+            });
+            let studentTagSlugs = student.tag_slugs;
+
+            // Set intercetion tags ∩ filters
+            const intersect = studentTagSlugs.filter(function (tag) {
+                return tagSlugFilters.indexOf(tag) >= 0;
+            });
+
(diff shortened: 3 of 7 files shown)


exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+let _totalSubmTrue = false;
+let _avgSubmTrue = false;
+let _maxSubmTrue = false;
+let _totalStuSubmTrue = false;
+let _totalStuMaxTrue = false;
+let _avgPTrue = false;
+let _maxPTrue = false;
+
+$("#total-subm-checkbox").change(function() {
+    _totalSubmTrue = this.checked ? true : false;
```
Mankro: `_totalSubmTrue = this.checked ? true : false;`
Is this any different from simply assigning like this `_totalSubmTrue = this.checked;`

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,185 @@
...
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/FileSaver.js/1.3.8/FileSaver.min.js"></script>
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/TableExport/5.2.0/js/tableexport.min.js"></script>
+
+<!--Custom for loading animation (loading_animation.css) and other extra customizations (results_staff.css)
+    results_staff.js is the main script for this page. -->
+<link rel="stylesheet" href="{% static 'exercise/css/loading_animation.css' %}" />
+<link rel="stylesheet" href="{% static 'exercise/css/results_staff.css' %}" />
+<script src="{% static 'exercise/results_staff.js' %}"
+        allExercisesUrl="/api/v2/exercises/"
+        exercisesUrl="/api/v2/courses/{{course.id}}/exercises"
```
raphendyr: These require url reverse resolving. No hardcoding urls in the files :)

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3485,3 +3485,74 @@ msgstr "Unohda automaattinen uudelleenohjaus."
 #: userprofile/views.py:80
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
+
+#: results.html
```
raphendyr: Use `./manage.py makemessages -i 'venv/*'` to update this file.

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,185 @@
...
+                <tbody id="table-body">
+
+                </tbody>
+            </table>
+        </div>
+
+            <!--Loading animation-->
+    <div class="row">
+        <div class="col-md-12">
+            <div id="loading-animation" class="loader">
```
raphendyr: The class is probably a bit generic (`loader`)? Also, can we use the same css for this and exercise loading? Should they have the same? Should we actually use this one for the exercises?

exercise/staff_views.py
```diff
@@ -179,7 +179,12 @@ class AllResultsView(CourseInstanceBaseView):
     def get_common_objects(self):
         super().get_common_objects()
         self.table = ResultTable(self.instance)
-        self.note("table")
+        self.tags = list(self.instance.usertags.all())
```
raphendyr: Follow #391 for how to do tags better now. Internal and external are handled via the same interface now.

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
+let currentScript = (document.currentScript ?
```
raphendyr: Missing the anonymous function scope wrapping!

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+let _totalStuMaxTrue = false;
+let _avgPTrue = false;
+let _maxPTrue = false;
+
+$("#total-subm-checkbox").change(function() {
+    _totalSubmTrue = this.checked ? true : false;
+    exerciseSelectionChange();
+});
+
+$("#avg-subm-checkbox").change(function() {
```
raphendyr: I would prefer mostly classes here. E.g. `$(this).find('input.results.avg-submissions').change(...`

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+
+    let normalValuesHtml = '';
+    let pctValuesHtml = 
+        '<tr class="no-filtering tableexport-ignore">'
+    let sumValue = 0;
+    dataValues.forEach(function(value) {
+        sumValue += value[0];
+    });
+
+// Empty data cells for total and tags in data indicators
```
raphendyr: Why these are not indented?

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+    _students.results.forEach(function(student) {
+        let sid = student.id;
+        const tagFilters = $.makeArray($('.filter-users button:has(.glyphicon-check)'))
+        .map(function(elem) {
+            return $(elem).attr('data-id');
+        });
+
+        let tag_ids = [];
+        _taggings.results.forEach(function(tagging) {
+        if (tagging.user.id === sid) {
```
raphendyr: indentation!

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+            tag_ids.push(tagging.tag.id);
+            }
+        });
+
+        const tags = tag_ids.map(function (id) {
+        return id.toString();
```
raphendyr: there are apparenly multiple indentation problems here :(

exercise/templates/exercise/staff/results.html
henkka14: How would I find the url reversers for them all? I can only find url for course api

`courseUrl="{% url 'api:course-detail' version=2 course_id=course.id %}"`

And still would have to do `courseUrl + "students"`

exercise/static/exercise/results_staff.js
henkka14: I assume I want to wrap the whole script?

exercise/static/exercise/results_staff.js
henkka14: Some style guide recommended putting all comments starting from left margin. Not sure what I should follow.

exercise/templates/exercise/staff/results.html
Mankro: I think you need to find the view class in the code and look for some URL key parameter that becomes the URL name for the URL resolver (maybe like this https://github.com/Aalto-LeTech/a-plus/blob/master/exercise/api/views.py#L66). The name is used in the static template tag.

In Python code, there is the function `api_reverse` that likewise uses the URL names to reverse them. 
https://github.com/Aalto-LeTech/a-plus/blob/master/aplus/api/__init__.py#L4

exercise/static/exercise/results_staff.js
henkka14: Actually I dont want to do that, but i am unsure how I would handle these variables I use in functions.

exercise/templates/exercise/staff/results.html
raphendyr: Basenames are here: https://github.com/Aalto-LeTech/a-plus/blob/master/api/urls_v2.py . Url name is `api:basename-[list|detail]` depending if it's the list or detail version of the view handler.

```
exercisesUrl="{% url 'api:course-exercises-list' version=2 course_id=course.id %}"
studentsUrl="{% url 'api:course-students-list' version=2 course_id=course.id %}"
```

and so on

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,185 @@
...
+<link rel="stylesheet" href="{% static 'exercise/css/results_staff.css' %}" />
+<script src="{% static 'exercise/results_staff.js' %}"
+        allExercisesUrl="/api/v2/exercises/"
+        exercisesUrl="/api/v2/courses/{{course.id}}/exercises"
+        studentsUrl="/api/v2/courses/{{course.id}}/students"
+        pointsUrl="/api/v2/courses/{{course.id}}/points/"
+        usersUrl="/api/v2/users/"
+        usertagsUrl="/api/v2/courses/{{course.id}}/usertags"
+        taggingsUrl="/api/v2/courses/{{course.id}}/taggings"
+        externalLabel="{{ external_user_label|safe }}"
```
raphendyr: Are these needed still?

Mankro: If they are needed, they should be data attributes

exercise/static/exercise/results_staff.js
raphendyr: sounds pretty interesting style guide. this seemed pretty nice https://github.com/airbnb/javascript  with exceptions on those features that are not supported by ie 11.

exercise/static/exercise/results_staff.js
raphendyr: Wrap the whole file and export functions that really need it. You can set variables to global scope with `window.variable = foo`. `window` is the global scope in browsers.

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+            exerciseSelectionChange();
+            $("#loading-animation").hide();
+            $("#table-points-div").show();
+        }
+    }
+
+    _students.results.forEach(function(student) {
+        let sid = student.id;
+        $.ajax({
+            url: usersUrl + sid
```
raphendyr: `url: student.url`, should work here, I presume.

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,840 @@
...
+            }
+            checkIfAllAjaxCompleted();
+        });
+    });
+
+    _exercises.results.forEach(function(module) {
+        module.exercises.forEach(function(exercise) {
+            let eid = exercise.id;
+            $.ajax({
+                url: allExercisesUrl + eid
```
raphendyr: `exercise.url` should have this value?

exercise/static/exercise/results_staff.js
henkka14: https://www.crockford.com/code.html

This was the style guide I looked at for comments. I assume the rest are same as in airbnb guide aside from comments, which I changed.

exercise/templates/exercise/staff/results.html
henkka14: These were not needed

exercise/static/exercise/results_staff.js
henkka14: Currently using `$("input.avg-subm-checkbox").change(function() {`

exercise/static/exercise/results_staff.js
henkka14: I used this to do it. My question is how should I indent the file. Should I have one whole indent for the rest of the file just for the wrap function as I have in the current revision?

exercise/staff_views.py
henkka14: Changed the tags to include  USERTAG_EXTERNAL, USERTAG_INTERNAL.

exercise/templates/exercise/staff/results.html
henkka14: changed the name to loading-animation. Did not do anything else yet. Not sure what would be good option.

locale/fi/LC_MESSAGES/django.po
henkka14: Done. However, there are some translations that are basically the same in `django.po` and `teacher.fi.json`. Would there be a clean way to use `_()` notation inside the html files?

exercise/templates/exercise/staff/results.html
henkka14: Changed these.

exercise/static/exercise/results_staff.js
henkka14: It is same. Currently using `_totalSubmTrue = this.checked;`

henkka14:
Please review the current changes. Not making modifications currently unless until next comments.

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,783 @@
...
+            // Create the table row for a single student from the points above
+            htmlTablePoints += '<tr><td class="student-id stick-on-scroll">' + student.student_id + '</td>';
+            htmlTablePoints += '<td class="student-name stick-on-scroll">' + _users[sid].full_name + '</td>';
+
+            if (pointKeys.length > 0) {
+                let tagHtml = "";
+                let studentTags = student.tag_slugs;
+    
+                // Sort tags in order. Put internal and external tags in the beginning.
+                studentTags.sort(function(left, right){
```
raphendyr: This should be handled, if you keep the order from the API.

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,175 @@
...
+    results_staff.js is the main script for this page. -->
+<link rel="stylesheet" href="{% static 'exercise/css/loading_animation.css' %}" />
+<link rel="stylesheet" href="{% static 'exercise/css/results_staff.css' %}" />
+<script src="{% static 'exercise/results_staff.js' %}"
+        data-exercises-url="{% url 'api:course-exercises-list' version=2 course_id=course.id %}"
+        data-students-url="{% url 'api:course-students-list' version=2 course_id=course.id %}"
+        data-usertags-url="{% url 'api:course-usertags-list' version=2 course_id=course.id %}"
+        data-taggings-url="{% url 'api:course-taggings-list' version=2 course_id=course.id %}"
+        data-points-url="{% url 'api:course-points-list' version=2 course_id=course.id %}"
+        data-external-label="{{ external_user_label|safe }}"
```
raphendyr: I don't think these are used anymore. I don't think they are provided by the view either anymore.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3511,3 +3592,17 @@ msgstr "Unohda automaattinen uudelleenohjaus."
 #: userprofile/views.py:80
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
```
raphendyr: Add `--no-obsolete` to `makemessages`. Old messages are tracked by the git, so we don't need them in the file. They are used for translation programs, so they can find similar translations. Or if you remove and then add a string back.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2623,23 +2624,97 @@ msgstr ""
...
+#: exercise/templates/exercise/staff/results.html:112
+msgid "Average points per student"
+msgstr "Pisteiden keskiarvo per opiskelija"
+
+#: exercise/templates/exercise/staff/results.html:116
+msgid "Maximum points for the exercise."
+msgstr "Tehtävän maksimipisteet."
+
+#: exercise/templates/exercise/staff/results.html:117
+msgid "Max points"
```
raphendyr: Should this be `Maximum points` as it was before?

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,175 @@
...
+            <button class="btn btn-primary dropdown-toggle" type="button" data-toggle="dropdown">{% trans "Export Data" %}
+                  <span class="caret"></span></button>
+            <ul id="export-button-menu" class="dropdown-menu">
+
+            </ul>
+        </div>
+
+    </div>
+
+        <!--Data table-->
```
raphendyr: Indentation starting from this line, doesn't seem to be consistent. Are there mix of tabs and spaces? I prefer tabs fro html files btw.

exercise/static/exercise/results_staff.js
```diff
@@ -0,0 +1,783 @@
+(function($, document, window, undefined) {
+    let currentScript = (document.currentScript ?
+        $(document.currentScript) :
+        $('script').last()); // Ugly solution for IE11
+
+    let exercisesUrl = currentScript.data("exercisesUrl");
```
raphendyr: These probably should be `const`.

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,175 @@
...
+<link rel="stylesheet" href="{% static 'exercise/css/results_staff.css' %}" />
+<script src="{% static 'exercise/results_staff.js' %}"
+        data-exercises-url="{% url 'api:course-exercises-list' version=2 course_id=course.id %}"
+        data-students-url="{% url 'api:course-students-list' version=2 course_id=course.id %}"
+        data-usertags-url="{% url 'api:course-usertags-list' version=2 course_id=course.id %}"
+        data-taggings-url="{% url 'api:course-taggings-list' version=2 course_id=course.id %}"
+        data-points-url="{% url 'api:course-points-list' version=2 course_id=course.id %}"
+        data-external-label="{{ external_user_label|safe }}"
+        data-internal-label="{{ internal_user_label|safe }}"
+        defer>
```
raphendyr: If I understand right, this means that the script is executed after html is parsed, but before document is ready (before `DOMContentLoaded`). All other scripts are hooked to jQuery document ready, which is run after `DOMContentLoaded` event (https://github.com/jquery/jquery/blob/cf84696fd1d7fe314a11492606529b5a658ee9e3/src/core/ready.js#L77). This includes translations. Thus, the script in question is always run before translation, which is kind of fine. That said, translations actually has a code that moves itself as a last document ready function, so that all other functions have time to register translations-ready listeners.

So, the defer vs. document ready here mostly is just different as anything else in A+. but technically it should be the same (at least when considering translations). In addition, as javascript runs all scripts in a single thread, there is no benefit from multi-threading.

So, I guess this is fine, just wrote the analysis down.

locale/fi/LC_MESSAGES/django.po
raphendyr: One day, teacher.fi.json will be generated from django.po or so. Untill then, we need to copy translations two two files.

exercise/static/exercise/results_staff.js
raphendyr: It's good now

exercise/static/exercise/results_staff.js
raphendyr: Works as long as there is only one result table (or table with those class names) on a single page. So, probably untill it's rewritten.

raphendyr:
Looks pretty good. Some small comments. In addittio, a lot of variables, which could be marked as `const`, are now `let`. It would be a good practice to always use `const` by default, and replace it with `let` if really needed. Typically, there is no need for `let` and most of algorithms using `let` can be written to be clearer with `const`s.

Mankro:
Has anyone tested whether this works correctly with unofficial or unconfirmed points? (unofficial = submissions after the deadline or max submissions limit if the category has the "allow unofficial" setting; unconfirmed = exercise points before the mandatory feedback exercise has been completed if the feedback exercise has the "confirm the level" setting)

Since the points are retrieved from the API, it might work correctly anyway, but it is worth checking.

raphendyr:
> Since the points are retrieved from the API, it might work correctly anyway, but it is worth checking.

Yeah it should work. API _should_ pull data from points cache, which should implement this logic for all different views.

exercise/static/exercise/results_staff.js
henkka14: Yep, seems to work without the sorting.

exercise/templates/exercise/staff/results.html
henkka14: Currently retrieving exercises, students, usertags and points URLs from API.

locale/fi/LC_MESSAGES/django.po
henkka14: Done

locale/fi/LC_MESSAGES/django.po
henkka14: Changed to the suggestions

exercise/templates/exercise/staff/results.html
henkka14: Should be all tabs now the whole file, please note I use vs code for indentation that has its own customizations available. Feel free to check that all works in some other ide.

exercise/static/exercise/results_staff.js
henkka14: Changed some variables to const.

henkka14:
Also feel free to check if all looks good. Changed some things based on last comments. Also corrected a few minor issues, how `exercise` is found from `_points` and also fixed issue with modules that had no exercises, which gave NaN for their points.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2108,6 +2113,18 @@ msgstr "Palautusta arvostellaan..."
...
+#, fuzzy
+#| msgid "Total points"
+msgid "Total score"
+msgstr "Yhteispisteet"
+
+#: exercise/templates/exercise/_exercisecollection.html:29
+#, fuzzy
+#| msgid "Teacher of the course"
+msgid "Score on this course"
+msgstr "Kurssin opettaja"
```
raphendyr: I don't think  the automatic translation went so well :)

The `#, fuzzy` means that this was automatically translated from a string that was pretty close. Using a translation program helps to track which translations are final. Any case, check translations which include `fuzzy`, fix them and remove the two lines (fuzzy and source msgid).

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1131,48 +1132,52 @@ msgstr ""
...
+#, fuzzy
+#| msgid "ExerciseCollection object requires collection_category."
+msgid "ExerciseCollection requires URL to target course instance."
+msgstr "Tehtäväkokoelmaolio tarvitsee collection_category:n."
 
-#: edit_course/operations/configure.py:586
-msgid "Course: {}, Instance: {}, not found."
-msgstr "Kurssi: {}, Instanssi: {}, ei löytynyt."
+#: edit_course/operations/configure.py:576
+msgid "Course URL '{}' doesn't match             service's hostname '{}'"
```
raphendyr: Dman. These are actually from previous changes, which did not update translations. I'll try to update them and then you can rebase again, so we get rid of them here.

raphendyr:
Master needs to update translations first, then we need to check this again. I think it's pretty much good. The unofficial points should be tested though.

On a side note, I don't personally like the method for building html, but I guess it's the simplest solution without better tools. So I guess we go with that.

raphendyr:
Master updated with translations for a missing set. Can you rebase and fix translations file.

henkka14:
Rebased and fixed translations.

henkka14:
Unofficial submissions are not accounted for, so if there are 7 submissions total where 5 are official and 2 are unofficial, the data indicators would only account for the 5 official submissions.

With the uncorfirmed points I am not sure, since I do not know how they are supposed to work. If you look at the picture below, the feedback form has the confirm_the_level option selected, yet it still gives points for the first exercise?

confirm_the_level

henkka14:
Also note that assistants will be listed as students in the Results table if the assistant has enrolled to the course. I was under the assumption that this is not a problem, since assistants should not enroll to the course. However, I assume that the system allows assistants to enroll to the course if they want for whatever reason? Not sure exactly though

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,176 @@
...
+				<thead id="table-heading">
+					<tr id="table-heading-row"></tr>
+				</thead>
+				<tbody id="table-body">
+
+				</tbody>
+			</table>
+		</div>
+
+    </div>
```
raphendyr: Indent

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,176 @@
...
+				<tr>
+					<td style="padding: 5px 10px 5px 10px;">
+						<label for="exercise-collection">{% trans "Select exercises: " %}</label>
+					</td>
+					<td style="padding: 5px 10px 5px 10px;">
+						<select id="exercise-selection" multiple="multiple">
+						</select>
+					</td>
+				</tr>
+				</tbody>
```
raphendyr: Indent

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,176 @@
...
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/xlsx/0.14.3/xlsx.core.min.js"></script>
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/FileSaver.js/1.3.8/FileSaver.min.js"></script>
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/TableExport/5.2.0/js/tableexport.min.js"></script>
+
+<!--Custom for loading animation (loading_animation.css) and other extra customizations (results_staff.css)
+    results_staff.js is the main script for this page. -->
+<link rel="stylesheet" href="{% static 'exercise/css/loading_animation.css' %}" />
+<link rel="stylesheet" href="{% static 'exercise/css/results_staff.css' %}" />
+<script src="{% static 'exercise/results_staff.js' %}"
+		data-exercises-url="{% url 'api:course-exercises-list' version=2 course_id=course.id %}"
```
raphendyr: One tab would be sufficient

raphendyr:
> However, I assume that the system allows assistants to enroll to the course if they want for whatever reason? Not sure exactly though

The teacher can also enroll to the course :)

raphendyr:
> Thus, it also shows 200 points from 3.1.1. in the Results table.

How does it show module and A, B, C points? If those are correct, then I think it's good.

Mankro:
About the unconfirmed points (confirm the level): it is an awkward feature. It may require that the feedback questionnaire is in the same chapter as the exercises. Then it should require that the points of the chapter become active only after the feedback has been submitted and the "exercise results" page should be clear about it. The point badges are gray and the mouse hover text says something about mandatory feedback.

henkka14:
> > Thus, it also shows 200 points from 3.1.1. in the Results table.
> 
> How does it show module and A, B, C points? If those are correct, then I think it's good. Of course, it would be good if the table would highlight that "these points are not actually counted, but as you reaquested to show exercise points, I did".

Uncorfirmed points do not work for any of the point categories, since the API always accepts the points in my testing, even if the feedback form is not submitted as in the picture above. Isn't it in the same chapter in the picture?

Anyway, fixed the indentation

Mankro:
I have a feeling that the A+ API does not know about unconfirmed points and it uses them immediately as normal points. If that is true, then only the user interface in the exercise results page really knows about the unconfirmed points feature. (It is only a hunch since I haven't tested it now.) Fixing/changing this is not part of this PR and I doubt it is urgent anyway since it has worked this way for a couple of years.

raphendyr:
I added `official` boolean to exercise entry in the api (in points). The boolean is true, if these exercise points are used for the grade (e.g. they are calculated to the module sum of points).

```json
   "modules": [
        {
            "id": 1,
            "max_points": 67,
            "points": 0,
            "points_by_difficulty": {
                "": 0
            },
            "exercises": [
                {
                    "id": 10,
                    "max_points": 2,
                    "points": 2,
                    "passed": true,
                    "official": false
```

Grade page could a) include button to select if unofficial points are shown and effecting summary calculations or b) calculate those as 0, but show the unofficial points in brackets (`0 (2)`).

For the teacher, it might be relevant to be able to see that some students are not getting as good grade as they have unconfirmed points (aka. unofficial due to not being confirmed).

Changes pushed to master.

raphendyr:
> raphendyr closed this now

miss click :(

Mankro:
The way you mixed the words "unofficial" and "unconfirmed" was confusing since they are separate features and I think the API supports unofficial points, but unconfirmed points maybe not.

raphendyr:
> The way you mixed the words "unofficial" and "unconfirmed"

They are used to refer to those specific features. Teacher would like to see unconfirmed points, but the API now exposes only combination of unofficial and unconfirmed. Though, only case when the `official = False` is true for a exe4rcise, which is confirmed, is when you submitted it after the deadline, but the points are confirmed.

Thus, we can show unofficial points, when teacher is looking for unconfirmed points as they should most of the time be the same thing. In addition, it's just good that the teacher will check the case when a student has submitted exercise after the DL.

And ofr the last part, the first submission to a exercise must be after the dl, otherwise the ready submission is used (submission is "official", when it's ready and confirmed).

raphendyr:
> I think the API supports unofficial points, but unconfirmed points maybe not.

API did expose ponts as long as they were not unofficial, but it didn't indicate in any way if they were unconfirmed. This was per exercise. For modules, API exposed the points values calculated internally, which did care if the points were unconfirmed or not.

Since this new page recalculates the summaries, it included also the unconfirmed points. Hence, I added the `official` flag to the API, so e.g. this page can filter per exercise points out, if they are not official points (included in the module points).

henkka14:
> I added `official` boolean to exercise entry in the api (in points). The boolean is true, if these exercise points are used for the grade (e.g. they are calculated to the module sum of points).
> 
> ```json
>    "modules": [
>         {
>             "id": 1,
>             "max_points": 67,
>             "points": 0,
>             "points_by_difficulty": {
>                 "": 0
>             },
>             "exercises": [
>                 {
>                     "id": 10,
>                     "max_points": 2,
>                     "points": 2,
>                     "passed": true,
>                     "official": false
> ```
> 
> Grade page could a) include button to select if unofficial points are shown and effecting summary calculations or b) calculate those as 0, but show the unofficial points in brackets (`0 (2)`).
> 
> For the teacher, it might be relevant to be able to see that some students are not getting as good grade as they have unconfirmed points (aka. unofficial due to not being confirmed).
> 
> Changes pushed to master.

I used this logic to show them as (`0 (2)`).

henkka14:
Rebased with master. I figured out my issue with confirm the level feedback form. Apparently it only works if the feedback gives points, so I override the feedback form to have 2 `max_points` in `conf.py`. It is a bit hacky for sure, but works I guess

I did use the `official` flag to see unconfirmed points inside parentheses. However, using the `official` flag did not let me get the unofficial points. For example, exercise with 5 submissions, if student first gets 5 submissions with 0 poins and with unofficial submission he scores 100 points, currently the option to show unofficial points would still show 0 points.

I could of course go through each exercise `submission` directly and check their `official` status and keep track of the highest point, but this would make the page much slower as the amount of ajax calls will be large.

Perhaps there could be `unofficial_points` in the `points` in API?

raphendyr:
> I used this logic to show them as `0 (2)`.

How that works with excel export? Can it drop the `(2)` part out from the excel, thus the cell would be plain number?

> Apparently it only works if the feedback gives points, so I override the feedback form to have 2 max_points in conf.py. It is a bit hacky for sure, but works I guess

Yep. This is how it works. Zero points can't effect anything as it means excercise is optional.

> If student first gets 5 submissions with 0 points and with unofficial submission he scores 100 points, currently the option to show unofficial points would still show 0 points.

> Perhaps there could be unofficial_points alongside the points in API?

I think current behavior is fine for now. Best probably would be to have best official and best unofficial submissions separately. We need to think that later if it seems to be something teachers would want. Though, I presume if the student has better unofficial score, they will contact staff via email and request them to be nice. On the other hand, there has been problems with unconfirmed points as people have missed the feedback often.

henkka14:
> > I used this logic to show them as `0 (2)`.
> 
> How that works with excel export? Can it drop the `(2)` part out from the excel, thus the cell would be plain number?
> 
> > Apparently it only works if the feedback gives points, so I override the feedback form to have 2 max_points in conf.py. It is a bit hacky for sure, but works I guess
> 
> Yep. This is how it works. Zero points can't effect anything as it means excercise is optional.
> 
> > If student first gets 5 submissions with 0 points and with unofficial submission he scores 100 points, currently the option to show unofficial points would still show 0 points.
> 
> > Perhaps there could be unofficial_points alongside the points in API?
> 
> I think current behavior is fine for now. Best probably would be to have best official and best unofficial submissions separately. We need to think that later if it seems to be something teachers would want. Though, I presume if the student has better unofficial score, they will contact staff via email and request them to be nice. On the other hand, there has been problems with unconfirmed points as people have missed the feedback often.

It is very easy to ignore whole `<tr>` and `<td>` tags in the table. However, since the ` (2)` is a `<span>` tag inside `<td>` tag, it would be harder. Would mean that I would have to write own row for the exported table as hidden and one row for the display of the table, ie duplicate all rows. Seems a bit unnecessary at this point, since teacher can still just unclick the checkbox?

assets/js-translations/teacher.fi.json
```diff
@@ -1,9 +1,104 @@
...
+
+  "No difficulty":
+  "Ei vaikeusastetta",
+
+  "Total submissions":
+  "Palautuksia yhteensä",
+
+  "Average submissions per student with submissions":
+  "Palautusten keskiarvo per opiskelija, jolla on vähintään yksi palautus",
+
```
raphendyr: There is whitespace here (not visible in github :( )

raphendyr:
Can you run `sed -i 's/\s*$//' <file>` for all the files in the PR, thanks.

exercise/templates/exercise/staff/results.html
```diff
@@ -10,45 +11,187 @@
...
+				{% trans "Maximum points" %}
+			</label>
+
+			<div class="has-error">
+				<div class="checkbox">
+					<label style="margin-left: 10px;" class="checkbox-inline" data-toggle="tooltip"
+						title="{% trans 'Show unofficial points next to the official points in parentheses.' %}">
+						<input type="checkbox" class="unofficial-checkbox" value="unofficial">
+						{% trans "Show unofficial points" %}
+				  	</label>
```
raphendyr: This line contains space plus tabs. The spaces are not visible, as the tab masks them out.

raphendyr:
> Seems a bit unnecessary at this point, since teacher can still just unclick the checkbox?

The checkbox is pretty good. Though, it should only add brackets for those cells that actually have unofficial points. In addition, I would prefer if the text would be `Show only official points` and would be selected by default. Negation of official is not unofficial in this case as `official = not unofficial and not unconfirmed` or `official = ready and confirmed`.

exercise/staff_views.py
```diff
@@ -179,7 +183,11 @@ class AllResultsView(CourseInstanceBaseView):
     def get_common_objects(self):
         super().get_common_objects()
         self.table = ResultTable(self.instance)
```
raphendyr: This can be dropped as it's not used. Am I right?

Mankro: `ResultTable` builds the whole result table in the server side, i.e., fetches all submission data in the course from the database. Since the new results page fetches the data from the REST API, ResultTable is absolutely unnecessary in the view and must be removed.

raphendyr:
I think it's good enough for now. Ain't testing in production a thing?

PR #391, opened 2019-07-29
Replace multi student cache with per user cache and include internal/external tags in most places

raphendyr:
This cache change removes limit on how many tags per student a course can have. There isn't a cache of a students, thus number if students should be unbound. Of course, per student point cache and this tags cache are both limited by the memcached 1M size limit.

I also merged external/internal student tags with usertags for the most part. Visualization page is not updated, but it's under rewrite and will be replaced in a near future (won't make for 1.5 release). Any case, these chenges pave a way for other user value based tags, including tag per organization or things like https://fitech.io/ .


raphendyr:
retest this please

raphendyr:
Damn. I need to fix some selenium tests :(

Mankro:
I read this through. I did not execute any tests myself. This looks fine to me so no suggestions for changes.

PR #392, opened 2019-07-29
New Shibboleth authentication

raphendyr:
* Dictionary style settings for shibboleth configuration
* Implement a parser for shib2 encoded multivalue fields.
* Parse only a single student number, based on a config value
* Authentication backend is rewritten, but should be more robust
  and handle problems better than before.
  e.g. before, oo long input values were just cut, and thus multiple
  different inputs might have resulted same output.
  Now we raise errors instead and require someone to fix the actual
  problem.
* Better debug output.. So we can actually find out these problems :)

Note that the `SHIBBOLETH_ENVIRONMENT_VARS.STUDENT_DOMAIN` configuration variable is now required with shibboleth! Check the documentation.

### DIFF
--- a/shibboleth_login/auth_backend.py
+++ b/shibboleth_login/auth_backend.py
@@ -1,27 +1,15 @@
 import logging
 import urllib.parse
 
-from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
+from django.shortcuts import redirect
+
+from .apps import app_settings, env_settings
+from .parser import Parser
 
 
 logger = logging.getLogger('aplus.shibboleth')
-
-USER_ID_KEY = 'SHIB_USER_ID_KEY'
-FIRST_NAME_KEY = 'SHIB_FIRST_NAME_KEY'
-LAST_NAME_KEY = 'SHIB_LAST_NAME_KEY'
-MAIL_KEY = 'SHIB_MAIL_KEY'
-STUDENT_ID_KEY = 'SHIB_STUDENT_ID_KEY'
-
-# These keys may be overridden in Django settings.
-DEFAULT_KEYS = {
-    USER_ID_KEY: 'SHIB_eppn',
-    MAIL_KEY: 'SHIB_mail',
-    FIRST_NAME_KEY: 'SHIB_givenName',
-    LAST_NAME_KEY: 'SHIB_sn',
-    STUDENT_ID_KEY: 'SHIB_schacPersonalUniqueCode',
-}
 
 
 class ShibbolethAuthBackend(ModelBackend):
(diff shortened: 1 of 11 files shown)


raphendyr:
retest this please

raphendyr:
retest this please

raphendyr:
retest this please

shibboleth_login/auth_backend.py
```diff
@@ -35,62 +23,138 @@ def authenticate(self, request, shibd_meta=None):
...
+            first_, __, last_ = full_name.partition(' ')
+            if not user.first_name:
+                user.first_name = first_[:first_name_len]
+            if not user.last_name:
+                user.last_name = last_[:last_name_len]
+            user_save_flag = True
+
         if user_save_flag:
+            # TODO: write better error reporting, when there is a validator to raise something
+            user.clean()
```
Mankro: Shouldn't the `full_clean()` method be used instead of `clean()`?

shibboleth_login/parser.py
```diff
@@ -0,0 +1,82 @@
...
+                    "Content is %s" % (name, fields))
+            ns, _, value = field[3:].partition(urn)
+            values = tuple(reversed(value.split(':'))) + (ns.lstrip(':'), field)
+            if filters and not all(values[i] == v for i, v in filters.items()):
+                continue
+            data.append(values)
+        return data
+
+
+class HakaParser(Parser):
```
Mankro: Is `HakaParser` actually used anywhere?

shibboleth_login/auth_backend.py
```diff
@@ -35,62 +23,138 @@ def authenticate(self, request, shibd_meta=None):
...
 
+        # TODO: support multiple domains
         profile = user.userprofile
-        student_id = self._parse(shibd_meta, STUDENT_ID_KEY)
-        if student_id:
-            student_id = student_id.split(':')[-1][:25]
-            if student_id and student_id != profile.student_id:
-                profile.student_id = student_id
-                profile.save()
+        sid_filters = env_settings.STUDENT_FILTERS.copy()
```
Mankro: Just to check, does the parser for student ids work with non-integer values? Some student ids mix digits and letters. The long Shibboleth URN seems to include the word `int` even when the value really isn't an integer.

jykesaviaGNH:
I can confirm that there's data around (e.g. "A57575757") with still the claim 'int' in the field.


On 31.7.2019 12.44, Markku Riekkinen wrote:

@Mankro commented on this pull request.

________________________________

In shibboleth_login/auth_backend.py<https://github.com/Aalto-LeTech/a-plus/pull/392#discussion_r309130075>:

>          profile = user.userprofile
-        student_id = self._parse(shibd_meta, STUDENT_ID_KEY)
-        if student_id:
-            student_id = student_id.split(':')[-1][:25]
-            if student_id and student_id != profile.student_id:
-                profile.student_id = student_id
-                profile.save()
+        sid_filters = env_settings.STUDENT_FILTERS.copy()


Just to check, does the parser for student ids work with non-integer values? Some student ids mix digits and letters. The long Shibboleth URN seems to include the word int even when the value really isn't an integer.

shibboleth_login/auth_backend.py
raphendyr: I think the `int` refers to international in this case. I thought about integers first too.

shibboleth_login/auth_backend.py
raphendyr: Yes. Fix coming up.

shibboleth_login/parser.py
raphendyr: Nope. Was used in tests under `if __name__ == '__main__'`, but that was removed later on. Thanks for noticing this.

raphendyr:
Updated. If no comments, I'll merge it.

Mankro:
I think it's ready.

raphendyr:
Just reminding that the `SHIBBOLETH_ENVIRONMENT_VARS.STUDENT_DOMAIN` configuration variable is now required with shibboleth! Check the documentation.

PR #393, opened 2019-08-09
Store the latest enrollment exercise submission per user

Mankro:
Store a link to the latest enrollment exercise submission for
each user. It is then possible to prefill enrollment questionnaire
forms on other courses if they use a similar questionnaire as
a course that the student has already enrolled in.
The enrollment questionnaire form may be pre-filled by
frontend JavaScript. The backend embeds the necessary submission
data in the enrollment exercise page as a JSON dump.

The migration for HardcodedUserTag is here because somebody seems to have forgotten to make that migration previously.


raphendyr:
> The migration for HardcodedUserTag is here because somebody seems to have forgotten to make that migration previously.

The hardcoded tag feature was supposed not to create any migrations, but apparently the only working set of flags did require one. In first iteration, it was marked as `abstract`, but that created some problems and I forgot migration after that.

exercise/views.py
```diff
@@ -143,6 +149,10 @@ def post(self, request, *args, **kwargs):
                     LearningObject.STATUS.ENROLLMENT_EXTERNAL,
                 ) and new_submission.status == Submission.STATUS.READY:
                     self.instance.enroll_student(self.request.user)
+                    # Save a link to the latest submission of an enrollment
+                    # exercise per user.
+                    self.profile.latest_enrollment_submission = new_submission
```
raphendyr: Can you make method for this. E.g. `self.profile.set_latest_enrollment_submission(new_submission) . That way, it's easier to merge user models and to create a proxy for `.profile`.

exercise/templates/exercise/exercise.html
```diff
@@ -25,6 +25,9 @@
         {% endif %}
     {% endif %}
 >
+  {% if latest_enrollment_submission %}
+    {{ latest_enrollment_submission.submission_data|json_script:"enroll-submission-data" }}
```
raphendyr: This exposes all the fields from the previous form. Do we have anyway to limit the fields? For example, could you read the form_spec and only include the fields which have the same key?

Mankro: What is the danger in exposing all the fields? It is the same user that made the submission.

raphendyr: Working toward a place where teacher of a course can't access students data from another course. Currently, this is of course possible via the API, which is a huge security issue :(

raphendyr: As a minimum, add a comment with `TODO: fix leaking of personal details by limiting amount of informations exposed here` or so here, so this problematic part of a code can be found later on.

Mankro:
I implemented the two suggested changes. I rebased against master but did not squash yet.

exercise/exercise_models.py
```diff
@@ -275,6 +275,32 @@ def get_templates(self):
...
+                key = item.get('key')
+                typ = item.get('type')
+                if not include_static_fields and typ == 'static':
+                    continue
+                if key and typ: # avoid empty or missing values
+                    pairs.add((key, typ))
+            return pairs
+
+        keys1 = make_key_type_pairs(self.exercise_info['form_spec'])
+        keys2 = make_key_type_pairs(another_lobject.exercise_info['form_spec'])
```
raphendyr: Notice that `form_spec` might be missing, in which case, I would presume an empty dictionary, In addition, `exercise_info`, might not be a dictionary (it's what ever json data, e.g. list or just a string).

Maybe `spec = self.exercise_info.get('form_spec', {}) if isinstance(self.exercise_info, dict) else {}`

exercise/exercise_models.py
```diff
@@ -275,6 +275,32 @@ def get_templates(self):
...
+        intersected = keys1.intersection(keys2)
+        return set(map(lambda p: p[0], intersected)) # return only the keys
```
raphendyr: Maybe a more pythonic way: `return set(key for key, typ in intersected)`

exercise/exercise_models.py
```diff
@@ -275,6 +275,32 @@ def get_templates(self):
...
 
+    def intersect_form_spec_keys(self, another_lobject, include_static_fields=False):
+        """Return the intersection of form field keys from this learning object and another.
+        In other words, keys that exist in both forms. The form keys are read
+        from the form_spec list of the exercise_info JSON field.
+        """
+        if not (self.exercise_info and another_lobject.exercise_info \
+                and 'form_spec' in self.exercise_info \
+                and 'form_spec' in another_lobject.exercise_info):
+            return set()
```
raphendyr: Sorry, missed this line. Yeah. This handles 90% of the problems. Only thing to check more, is to ensure exercise_infos are dicts.

exercise/exercise_models.py
raphendyr: This is mostly irrelevant, relevant commit: https://github.com/Aalto-LeTech/a-plus/pull/393/files#r313386334

Mankro:
I changed the implementation. Now there is no new database field or other database schema changes. When the enrollment exercise is opened, all enrollment submissions are read from the database and their submission data are combined. The keys are then filtered against the current exercise form spec before the submission data is embedded into the HTML page.

exercise/views.py
```diff
@@ -93,6 +95,16 @@ def get(self, request, *args, **kwargs):
...
+        elif self.exercise.status in \
+                (LearningObject.STATUS.ENROLLMENT, LearningObject.STATUS.ENROLLMENT_EXTERNAL):
+            # Retrieve the data of the user's latest enrollment exercise submissions
+            # in case the data could be reused in this enrollment exercise.
+            all_enroll_data = Submission.objects.get_combined_enrollment_submission_data(self.request.user)
+            # Find the intersection of form field keys that are used by the current
+            # exercise and all_enroll_data.
+            # Filter all_enroll_data to only contain these keys.
+            all_enroll_data = filter_dict_with_keys(all_enroll_data,
+                self.exercise.get_form_spec_keys().intersection(all_enroll_data.keys()))
```
raphendyr: The intersection could be calculate inside filter_dict_with_keys. Alternatively, I don't see need for the helper, as you could do

```
include_fields = self.exercise.get_form_spec_keys().intersection(all_enroll_data.keys())
all_enroll_data = {key, value for key, value in all_enroll_data.items() if key in include_fields}
```

Mankro: I removed the helper and used this one-liner dict compehension.

Mankro:
I made that one fix that was requested. I rebased and squashed the commits.

PR #395, opened 2019-09-18
Fix JS bug in the removal of user taggings

Mankro:
Previously, deleting user taggings in the course participants page
could fail because the JS code did not handle paginated responses
from the API. The code retrieved all user taggings for the tag
even though it was only supposed to remove one tagging.

The commit adds a new endpoint to the API so that it is easy to
delete taggings without retrieving all taggings (which was
previously done to find out the ID of the tagging to delete).

The new JS code for the deletion may also delete multiple taggings
in a single HTTP DELETE request.

### DIFF
--- a/api/urls_v2.py
+++ b/api/urls_v2.py
@@ -1,3 +1,5 @@
+import copy
+
 from django.conf import settings
 from django.conf.urls import url, include
 from rest_framework_extensions.routers import ExtendedDefaultRouter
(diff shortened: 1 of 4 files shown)


Mankro:
The URL for the new API endpoint is like `/api/v2/courses/1/taggings/delete_taggings/?tag_slug=tomato&user_id=3`.
Django REST framework does not enable the HTTP DELETE method by default for the list views, so I had to create an "action detail" method that adds another word to the URL. If it were easy to modify the router and add DELETE support to the list view, then the URL would not need the word "delete_taggings".

Mankro:
The new commit changes the URL conf for the delete taggings API endpoint: the new URL is like `/api/v2/courses/1/taggings/?tag_slug=tomato&user_id=3`. It is the same URL as the default URL for the list view of the view set.

The manual definition in the URL conf looks a little awkward, but I don't see any other way to change the behaviour of the router. Another option would be to define a new router class that adds the DELETE action to the list view. The default router specifically does not enable DELETE for the list view: https://github.com/encode/django-rest-framework/blob/3.9.4/rest_framework/routers.py#L136

Mankro:
Eventually, we decided to define a new router class with just 3 lines of code. The new router adds HTTP DELETE to the list view. It does look a little hacky though, since it assumes that the `routes` list in the router base class defines the list view as the first item.

PR #397, opened 2019-10-15
Material can be opened for reading before the opening time

alalav1:
Modules have the `reading_opening_time` setting which allows the teacher to set time to open material for reading. The exercises in the module are not shown before the actual opening time.


Plustest:

course/migrations/0046_coursemodule_reading_opening_time.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.3 on 2019-10-09 13:52
+
+from django.db import migrations, models
+import django.utils.timezone
+
+
+class Migration(migrations.Migration):
```
Mankro: You don't need two migration files for adding one field.

alalav1:
Javascript string translation is still missing (one string in chapter.js).

Mankro:
When this is rebased against upstream master, we need to check the dependencies on the migrations. Migration files refer to to the previous migrations since they must be executed in the right order. The upstream master has new migrations and the migrations of this PR should come after those.

course/models.py
```diff
@@ -685,6 +685,7 @@ class CourseModule(UrlMixin, models.Model):
     introduction = models.TextField(blank=True)
     course_instance = models.ForeignKey(CourseInstance, on_delete=models.CASCADE,
         related_name="course_modules")
+    reading_opening_time = models.DateTimeField(verbose_name=_("Opening time for reading"), null=True, blank=True)
```
Mankro: Is the verbose name of the field rendered in the form page? Or is there another keyword for the help string? The form page should display a help text that says that the reading opening time may be left empty if no material should be available before the module opening time.

course/models.py
```diff
@@ -722,17 +723,28 @@ def clean(self):
...
             errors['late_submission_deadline'] = _("Late submission deadline must be later than the closing time.")
+        if self.reading_opening_time and self.reading_opening_time > self.opening_time:
+            errors['reading_opening_time'] = _("Opening time of reading material "
+                "must be earlier than the opening time of the exercises.")
         if errors:
             raise ValidationError(errors)
 
     def is_open(self, when=None):
         when = when or timezone.now()
+        if self.reading_opening_time:
```
Mankro: Are there any consequences that the module is considered to be open already during the reading time? Now, exercises may not be submitted even though this method says that the module is open.

course/permissions.py
```diff
@@ -118,8 +118,6 @@ def has_permission(self, request, view):
         return True
 
     def has_object_permission(self, request, view, module):
-        if not isinstance(module, CourseModule):
```
Mankro: Why did you remove this `isinstance` check? Was there a problem with it?

course/permissions.py
```diff
@@ -138,6 +136,11 @@ def has_object_permission(self, request, view, module):
             return module.are_requirements_passed(points)
         return True
 
+        if module.requirements.count() > 0:
```
Mankro: `if module.requirements.count() > 0:` is done two times in a row now. The second one should be removed, right?

course/templates/course/module.html
```diff
@@ -43,7 +43,7 @@
 	</p>
 	{% endif %}
 
-  {% include "exercise/_children.html" with children=children accessible=True %}
+  {% include "exercise/_children.html" with children=children accessible=True exercise_accessible=current|is_open:now %}
```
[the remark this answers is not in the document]
Mankro: Is `exercise_accessible` used for chapters too? The word exercise is often used for both exercises and chapters.

course/models.py
```diff
@@ -722,17 +723,28 @@ def clean(self):
...
     def is_after_open(self, when=None):
         """
         Checks if current time is past the round opening time.
         """
+        when = when or timezone.now()
+        if self.reading_opening_time:
+            return self.reading_opening_time <= when
+        return self.opening_time <= when
+
+    def exercises_are_after_open(self, when=None):
```
Mankro: `exercises_are_after_open` is a bad name for the method. How about `have_exercises_been_opened`?

course/models.py
Mankro: I see you added a separate method for checking the exercise opening time. I just wonder if old code uses the right method for each use case then.

course/templatetags/course.py
```diff
@@ -81,6 +81,11 @@ def is_open(entry, now):
     return entry['opening_time'] <= now <= entry['closing_time']
 
 
+@register.filter
+def exercises_open(entry, now):
```
Mankro: In this file, function `exercises_open` looks identical to `is_open`. Are both needed?

Mankro: In the course module class, `is_open` uses the reading time, so now the `is_open` template tag is different when it uses the opening time. Inconsistency causes confusion.

exercise/templates/exercise/_children.html
```diff
@@ -11,7 +11,17 @@
 {% endif %}
 {% elif entry|is_listed %}
 <li>
-  {% if accessible %}
+  {% if accessible and not entry.submittable%}
```
Mankro: Missing space after `entry.submittable`.

Mankro: Can you explain what these if statements do? I'd like to know what is used for chapters and exercises, and what happens before and after the opening time and reading opening time.

exercise/templatetags/exercise.py
```diff
@@ -213,6 +213,18 @@ def min_group_size(context):
 
 @register.simple_tag(takes_context=True)
 def module_accessible(context, entry):
+    t = entry.get('reading_opening_time', entry.get('opening_time'))
```
Mankro: Your new function `exercise_accessible` seems to be almost identical to the old `module_accessible`? What is the difference and is it really necessary to duplicate code? What if you add new parameters to the old function?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -268,32 +268,42 @@ msgstr "Huolto"
 msgid "Input an URL identifier for this module."
 msgstr "Syötä URL-osoitteessa käytettävä tunnus moduulille."
 
-#: course/models.py:699
+#: course/models.py:688
+msgid "Opening time for reading"
+msgstr "Lukumateriaalin avautuminen"
```
Mankro: "avautuminen" vai "avautumisaika" vai jotain muuta?

Mankro:
This should be rebased against upstream master. The commit history must be cleaned. Database migrations and Finnish translations must be cleaned. Currently, changes in the translation django.po file cause git merge conflicts, but they should be easy to fix by runnig `python manage.py makemessages` again.

Note previous comments about migrations and the missing JavaScript string translation.

course/templatetags/course.py
alalav1: The method `is_open` has been removed.

exercise/templatetags/exercise.py
alalav1: The difference is whether the function uses the `reading_opening_time` or the `opening_time`. Those functions have now their common parts in the `is_accessible`.

course/models.py
alalav1: Done.

course/permissions.py
alalav1: Done.

course/templates/course/module.html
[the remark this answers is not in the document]
alalav1: Exercise accessible uses the `opening_time` and is used to modify links and other elements.

exercise/templates/exercise/_children.html
alalav1: `entry.submittable` checks if the children is exercise or chapter.

exercise/templatetags/exercise.py
```diff
@@ -213,14 +213,14 @@ def min_group_size(context):
...
+def exercise_accessible(context, entry):
     t = entry.get('opening_time')
-    if t and t > _prepare_now(context):
-        return False
-    if entry.get('requirements'):
-        points = _prepare_context(context)
-        module = CourseModule.objects.get(id=entry['id'])
-        return module.are_requirements_passed(points)
-    return True
+    return is_accessible(context, entry, t)
```
Mankro: It is confusing that `context` and `entry` are given to the `is_accessible` function call in a different order than in the definition of the function.

Mankro: Looks like that it is OK to switch the order of the two variables in the definition of the `is_accessible` function. The `context` variable has been the first one in other functions too.

course/templates/course/_siblings.html
```diff
@@ -4,7 +4,11 @@
 <nav class="row siblings">
 
     {% if next %}
-    {% module_accessible next as accessible %}
+     {% if next.submittable %}
```
Mankro: The indentation is wrong here. Don't use a single space. Our indentation in HTML templates has been inconsistent, but you could follow the style already used in the same file. Here, you could keep `{% if next.submittable %}` on the same level with `{% if next %}` and indent `{% exercise_accessible next as accessible %}` with four additional spaces so that it shows it is inside the if.

exercise/templates/exercise/_children.html
```diff
@@ -11,18 +11,17 @@
 {% endif %}
 {% elif entry|is_listed %}
 <li>
-  {% if accessible %}
-  {% if entry.is_empty %}
-  {{ entry.name|parse_localization }}
+  {% if accessible and not entry.submittable or exercise_accessible %}
+   {% if entry.is_empty %}
```
Mankro: It looks like you have only changed one if condition here, but because you changed the indentations, many lines show up here in the git diff. Using one space is not good for indentation, and it is good to avoid making such whitespace changes that pollute the diff unless there is a good reason to do it. Either fix the indentations properly or do not change them on existing lines. I think you should remove your indentation changes so that the diff shows only the one changed if condition in the end after squashing.

exercise/static/exercise/chapter.js
```diff
@@ -311,8 +311,18 @@
...
 			 } else {
 				$.ajax(this.url, {dataType: "html"})
-					.fail(function() {
-						exercise.showLoader("error");
+					.fail(function(jqXHR, textStatus, errorThrown) {
+						if (jqXHR.status == 403) {
+							exercise.hideLoader();
+							exercise.update($(
+								'<div id="exercise-all"><div class="info admonition" style="margin: 0px">' +
+								_('There will be an exercise when the module opens for submissions.') +
```
Mankro: `'There will be an exercise when the module opens for submissions.'` sounds a little off.
`'There will be an exercise HERE when the module opens for submissions.'` is slightly better. This could still be improved.

assets/js-translations/exercise.fi.json
```diff
@@ -0,0 +1,4 @@
+{
+  "There will be an exercise when the module opens for submissions.":
+  "Tähän tulee tehtävä kierroksen avautuessa."
```
Mankro: The Finnish string is OK, but the English string requires improvement.

course/templates/course/_siblings.html
```diff
@@ -4,7 +4,11 @@
...
+      {% exercise_accessible next as accessible %}
```
Mankro: (I can't target this comment to the bottom of the file.) In the end of this file, there is the link to the "previous" item. This part is for the "next" item. Did you test the previous link? It could require the same separation of submittable and non-submittable items. Now, it always uses the `module_accessible` value.

exercise/templatetags/exercise.py
```diff
@@ -278,6 +279,17 @@ def get_format_info(format):
     except KeyError as e:
         raise RuntimeError('Invalid format: \'{}\''.format(format)) from e
 
+
 @register.simple_tag
 def get_format_info_list(formats):
     return [get_format_info(format) for format in formats.split()]
+
+
+def is_accessible(entry, context, t):
```
Mankro: Let's name the function `_is_accessible` and move it to the top of the file since the other helper functions are defined there. The underscore in the beginning implies that the function is not intended to be used outside this module (private function).

exercise/permissions.py
```diff
@@ -31,6 +31,10 @@ def is_object_visible(self, request, view, exercise):
             self.error_msg(_("The exercise is not currently visible."))
             return False
 
+        if exercise.is_submittable and not exercise.course_module.have_exercises_been_opened():
+            self.error_msg(_("The exercise opens for submissions in {}.".format(exercise.course_module.opening_time)))
```
Mankro: I suspect that the date is formatted in the wrong way in the translated string. Shouldn't the `format` method be called on the string only after the `_` function has been called? Now you are trying to translate a string that contains a specific date value. Does it really work at all when you test it? The trivial fix is to move the parenthesis:
`_("The exercise opens for submissions in {}.").format(exercise.course_module.opening_time)`

Mankro:
@alalav1 when you make these small changes, do not squash the commits yet into the original commit so that I can see the latest changes. Preferably, keep these new changes in one commit.

exercise/templatetags/exercise.py
alalav1: Switched parameters in the definition of `is_accessible` function.

course/templates/course/_siblings.html
alalav1: Identation is changed so that the number of spaces in the beginning of the row is multiple of 4.

exercise/templates/exercise/_children.html
alalav1: In August there was a problem when my atom changed spaces in tabs or other way around when I saved this file. I tried to fix them manually then. I'll use the proper indentation this time but git diff might still be little dirty.

course/templates/course/_siblings.html
alalav1: Added the same separation between exercises and modules.

exercise/templatetags/exercise.py
alalav1: Renamed.

exercise/permissions.py
alalav1: Moved the format out of the function.

exercise/static/exercise/chapter.js
alalav1: "An exercise will be opened here once the round opens for submissions."?
Something else?
Or do you prefer your suggestion? @Mankro

assets/js-translations/exercise.fi.json
alalav1: New ideas:

Tässä julkaistaan tehtävä kierroksen avautuessa.

An exercise will be published here once the round opens for submissions.

Mankro: I approve these.

exercise/static/exercise/chapter.js
Mankro: See new ideas under assets/js-translations/exercise.fi.json

PR #398, opened 2019-10-21
Show the deadline deviations to the students

alalav1:
Student should see the deadline deviations.

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -780,7 +780,8 @@ class CourseModule(UrlMixin, models.Model):
         Returns the percentage (0-100) that late submission points are worth.
         """
         point_worth = 0
-        if self.late_submissions_allowed:
+        if self.late_submission_penalty is not None:
+            # can be set to zero
             point_worth = int((1.0 - self.late_submission_penalty) * 100.0)
         return point_worth
 
--- a/course/models.py
+++ b/course/models.py
@@ -780,8 +780,7 @@ class CourseModule(UrlMixin, models.Model):
         Returns the percentage (0-100) that late submission points are worth.
         """
         point_worth = 0
-        if self.late_submission_penalty is not None:
-            # can be set to zero
+        if self.late_submissions_allowed:
             point_worth = int((1.0 - self.late_submission_penalty) * 100.0)
         return point_worth
 
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -480,6 +480,10 @@ class BaseExercise(LearningObject):
                     deviation = d
         return deviation
 
+    def student_has_deadline_deviation(self, student):
+        dl = self.deadlineruledeviation_set.filter(submitter=student)
+        return dl[0].get_new_deadline() if dl else None
+
     def number_of_submitters(self):
         return self.course_instance.students\
             .filter(submissions__exercise=self).distinct().count()
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -482,7 +482,7 @@ class BaseExercise(LearningObject):
 
     def student_has_deadline_deviation(self, student):
         dl = self.deadlineruledeviation_set.filter(submitter=student)
-        return dl[0].get_new_deadline() if dl else None
+        return dl[0] if dl else None
 
     def number_of_submitters(self):
         return self.course_instance.students\
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -480,10 +480,6 @@ class BaseExercise(LearningObject):
                     deviation = d
         return deviation
 
-    def student_has_deadline_deviation(self, student):
-        dl = self.deadlineruledeviation_set.filter(submitter=student)
-        return dl[0] if dl else None
-
     def number_of_submitters(self):
         return self.course_instance.students\
             .filter(submissions__exercise=self).distinct().count()
--- a/exercise/templates/exercise/_exercise_info.html
+++ b/exercise/templates/exercise/_exercise_info.html
@@ -50,8 +50,13 @@
             <dd class="exercise-info-required-points">{{ exercise.points_to_pass }}</dd>
             {% endif %}
 
+            {% if exercise|has_deviation:profile %}
+            <dt>{% trans "Extended deadline" %}</dt>
+            <dd class="exercise-info-deadline">{{ exercise|has_deviation:profile }}</dd>
+            {% else %}
             <dt>{% trans "Deadline" %}</dt>
             <dd class="exercise-info-deadline">{{ exercise.course_module.closing_time }}</dd>
+            {% endif %}
 
             {% if exercise.course_module.late_submissions_allowed %}
             <dt>{% trans "Late submission deadline" %}</dt>
--- a/exercise/templates/exercise/_exercise_info.html
+++ b/exercise/templates/exercise/_exercise_info.html
@@ -54,11 +54,13 @@
 
             <dt>{% trans "Deadline" %}</dt>
             <dd class="exercise-info-deadline">{{ exercise.course_module.closing_time }}</dd>
-            {% if exercise|deviation_deadline:profile %}
+            {% deviation_deadline exercise as dl %}
+            {% has_deviation_penalty exercise as has_penalty %}
+            {% if dl %}
             <dt>{% trans "Extended deadline" %}</dt>
             <dd class="exercise-info-deadline">
-              {{ exercise|deviation_deadline:profile }}
-              {% if exercise|has_deviation_penalty:profile|percent > 0 %}
+              {{ dl }}
+              {% if has_penalty %}
                 (-{{ exercise.course_module.late_submission_penalty|percent }}%)
               {% endif %}
             </dd>
--- a/exercise/templates/exercise/_exercise_info.html
+++ b/exercise/templates/exercise/_exercise_info.html
@@ -50,12 +50,18 @@
             <dd class="exercise-info-required-points">{{ exercise.points_to_pass }}</dd>
             {% endif %}
 
-            {% if exercise|has_deviation:profile %}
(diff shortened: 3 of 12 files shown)


Plustest:

course/models.py
```diff
@@ -758,7 +758,8 @@ def get_late_submission_point_worth(self):
         Returns the percentage (0-100) that late submission points are worth.
         """
         point_worth = 0
-        if self.late_submissions_allowed:
+        if self.late_submission_penalty is not None:
```
Mankro: Why did you make this change? Late submissions are not accepted at all if `self.late_submissions_allowed` is `False`.

exercise/exercise_models.py
```diff
@@ -470,6 +470,10 @@ def one_has_deadline_deviation(self, students):
                     deviation = d
         return deviation
 
+    def student_has_deadline_deviation(self, student):
```
Mankro: Is this method really needed? The method `one_has_deadline_deviation` can be called with a list of one student.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2071,21 +2071,26 @@ msgstr "Palautuksesi"
...
+#: exercise/templates/exercise/exercise_plain.html:114
 msgid "Deadline"
 msgstr "Määräaika"
 
-#: exercise/templates/exercise/_exercise_info.html:57
-#: exercise/templates/exercise/exercise_plain.html:98
+#: exercise/templates/exercise/_exercise_info.html:58
+#: exercise/templates/exercise/exercise_plain.html:97
+msgid "Extended deadline"
+msgstr "Pidennetty palautusaika"
```
Mankro: "Deadline" has been translated to "määräaika" previously, so we should use the same word here ("pidennetty palautusaika" is changed to "pidennetty määräaika").

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2350,30 +2355,45 @@ msgid "Inspect submission"
...
 #: exercise/templates/exercise/exercise_base.html:97
-#: exercise/templates/exercise/exercise_plain.html:134
+#: exercise/templates/exercise/exercise_plain.html:144
 msgid "View all submissions"
 msgstr "Näytä kaikki palautukset"
 
-#: exercise/templates/exercise/exercise_plain.html:93
-#, python-format
+#: exercise/templates/exercise/exercise_plain.html:92
+#, fuzzy, python-format
```
Mankro: You should not commit "fuzzy" translation changes. The gettext tool got confused about some changes and you need to check what has changed and fix the problems. The "fuzzy" word and the lines starting with `#|` must not be committed. You can fix and remove them by modifying the file in a text editor. The point is that you do not leave the fuzzy parts lying around but instead you fix them.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2350,30 +2355,45 @@ msgid "Inspect submission"
...
 "Suoritusmerkintään vaaditaan %(points)s pistettä"
 
-#: exercise/templates/exercise/exercise_plain.html:119
+#: exercise/templates/exercise/exercise_plain.html:101
+msgid "Penalty after"
+msgstr "Pistevähennykset alkaen"
+
+#: exercise/templates/exercise/exercise_plain.html:112
+#, fuzzy
+#| msgid "Exercise info"
```
Mankro: Fuzzy again. See the previous comment for more explanations.

exercise/templates/exercise/_exercise_info.html
```diff
@@ -50,8 +50,19 @@ <h4 class="panel-title">{% trans "Exercise info" %}</h4>
             <dd class="exercise-info-required-points">{{ exercise.points_to_pass }}</dd>
             {% endif %}
 
+
+
```
Mankro: No reason to leave these extra empty lines here.

course/models.py
Mankro: I don't think `self.late_submission_penalty` could ever be `None` since the model field does not have null=True nor blank=True.

exercise/templates/exercise/_exercise_info.html
```diff
@@ -50,8 +50,19 @@ <h4 class="panel-title">{% trans "Exercise info" %}</h4>
...
             <dt>{% trans "Deadline" %}</dt>
             <dd class="exercise-info-deadline">{{ exercise.course_module.closing_time }}</dd>
+            {% if exercise|deviation_deadline:profile %}
+            <dt>{% trans "Extended deadline" %}</dt>
+            <dd class="exercise-info-deadline">
+              {{ exercise|deviation_deadline:profile }}
+              {% if exercise|has_deviation_penalty:profile|percent > 0 %}
```
Mankro: It looks like that `exercise|has_deviation_penalty:profile` returns a boolean, so how can you call the `percent` filter after it?

exercise/templates/exercise/_exercise_info.html
```diff
@@ -50,8 +50,19 @@ <h4 class="panel-title">{% trans "Exercise info" %}</h4>
             <dd class="exercise-info-required-points">{{ exercise.points_to_pass }}</dd>
             {% endif %}
 
+
+
             <dt>{% trans "Deadline" %}</dt>
             <dd class="exercise-info-deadline">{{ exercise.course_module.closing_time }}</dd>
+            {% if exercise|deviation_deadline:profile %}
```
Mankro: There are three template tags here that each query the database for the same deviation. Only one database query should be used for the same thing.

Mankro: The `with` template tag should be useful for saving the deviation into a variable so that you can reuse it: 
https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#with

exercise/templates/exercise/exercise_plain.html
```diff
@@ -87,24 +87,34 @@
                             </li>
                             {% if not exercise.category.confirm_the_level %}
                                 <li class="navbar-text">
-                                    {% if not exercise.is_closed %}
```
Mankro: Did you remove `{% if not exercise.is_closed %}` completely? Why?

exercise/templates/exercise/exercise_plain.html
```diff
@@ -87,24 +87,34 @@
...
-                                                {% if module.late_submission_penalty > 0 %}
-                                                (-{{ module.late_submission_penalty|percent }}%)
+                                      <small>
+                                          {% if exercise.points_to_pass > 0 %}
+                                          {% blocktrans with points=exercise.points_to_pass %}
+                                              {{ points }} points required to pass.
+                                          {% endblocktrans %}
+                                          {% endif %}
+                                          {% if exercise|deviation_deadline:profile %}
+                                              {% trans "Extended deadline" %} {{ exercise|deviation_deadline:profile }}
```
Mankro: Like I commented in the other HTML template, it is not good to query the database multiple times for the same deviation. The `with` template tag can be used to create a template variable that holds the return value of `exercise|deviation_deadline:profile`.

Mankro:
Currently, you have implemented these updates with template tags. That means that every time the templates are rendered, there is at least one additional database query compared to the past state. We should think if we want to proceed with that approach or if we want to add deadline deviations to the student cache in order to avoid additional database queries on each exercise page load. Chapter pages even contain multiple exercises and the deadline deviation would be checked for each exercise every time.

Furthermore, if we render submission limit deviations in the exercise pages too (they are now missing, right?), then it would add even more database queries if no caching is added.

exercise/templates/exercise/exercise_plain.html
Mankro: Actually, looks like it was moved to the end.

Mankro:
@alalav1 If we decide to use the student cache, then there will larger changes and some of the previous comments become irrelevant.

Mankro:
Currently, A+ has a course-level cache for student taggings. I suppose we could similarly store the student's deviations in the cache too. Most students have no deviations and some students have deviations for some exercises in the course.

Student cache for tags:
https://github.com/Aalto-LeTech/a-plus/blob/3c3579b52ff05ba62283a07bd9975f0d0246b289/course/cache/students.py#L15

@raphendyr quick comment about storing deadline and submission limit deviations in the cache? Is it feasible?

oseppala:
On courses where there are a large number of exercises, the amount of deviations can grow large. On O1 there are currently ~1000 deviations.

Mankro:
After a quick chat with raphendyr, we could say that the student cache should have enough space for storing the deviations.

alalav1:
This PR is extended to deal with bonus submissions, too. They are now shown correctly to the student in the results page.

raphendyr:
If needed, the PR can be split to two to address the cache and the UI separately.

Mankro:
@skulonen  in https://jira.aalto.fi/browse/EDIT-148:

>    The PR adds the student's deviations to CachedStudent. However, since the delayed feedback update, the personal deadline and personal max submissions of  a student have been stored in CachedPoints. This makes the CachedStudent changes unnecessary.
    In addition, the PR doesn't seem to consider the deviations received by other members in the student's group.
    The PR mostly uses template tags to add the deviation data to the UI. In each template tag's code, a CachedStudent instance is constructed. In my opinion, it would be better to construct a CachedPoints instance once (perhaps in ExerciseMixin.get_common_objects, which already contains cache-related code) and store the relevant values as attributes of the view.
    The exercise frame has changed somewhat, so the template code may not be directly usable.

>  In conclusion, I think it would be better to re-implement these changes using CachedPoints.

Great conclusion and I agree. We can close the old PR and the UI changes can be implemented from scratch.

Mankro:
This pull request was replaced by https://github.com/apluslms/a-plus/pull/908.

PR #401, opened 2019-10-29
Add possibility to override deviations

alalav1:
Previously teacher needed to remove deviations manually before they could be substituted.
Add view to select deviations to be overridden.

Also add a form to remove deadline deviations.


Plustest:

Mankro:
Comments copied from chat:

onko vanhan muokkaaminen vaikeaa? (jonka takia poistetaan ja luodaan uusi)

self.note(a, b, c) on tehokkaampaa kuin self.note(a); self.note(b); self.note(c) ja kommentit voisi sisentää koodin mukaan

locale/fi/LC_MESSAGES/django.po
```diff
@@ -882,49 +887,93 @@ msgstr "Lisää uusia määräajan muutoksia"
 msgid "Submit"
 msgstr "Lähetä"
 
-#: deviations/templates/deviations/list_dl.html:23
+#: deviations/templates/deviations/list_dl.html:22
+#: deviations/templates/deviations/remove_dl.html:6
+#: deviations/templates/deviations/remove_dl.html:12
+#: deviations/templates/deviations/remove_dl.html:19
+msgid "Remove deadline deviations"
+msgstr "Ei yhtään määräajan muutosta."
```
Mankro: This translation is incorrect.

Mankro: Note that this file contains these lines elsewhere.

```
msgid "No deadline deviations."
msgstr "Ei yhtään määräajan muutosta."
```

locale/fi/LC_MESSAGES/django.po
```diff
@@ -882,49 +887,93 @@ msgstr "Lisää uusia määräajan muutoksia"
...
+msgstr "Osa määräajan muutoksista oli jo olemassa!"
+"Valitse ylikirjoitettavat muutokset!"
+
+#: deviations/templates/deviations/override.html:16
+msgid "Select deviations to be overridden"
+msgstr "Valitse ylikirjoitettavat muutokset"
+
+#: deviations/templates/deviations/override.html:22
+msgid "Extra time before (minutes)"
+msgstr "Lisä aika ennen (minuuteissa)"
```
Mankro: "extra time before": do users understand that "before" means the previous old value that would be overwritten? Is there a better word for this? The wording may mislead people into thinking that this is some extra time given before something else.

"Lisä aika" should be "lisäaika".

Mankro: Words to replace "before": "Current extra time"

locale/fi/LC_MESSAGES/django.po
```diff
@@ -882,49 +887,93 @@ msgstr "Lisää uusia määräajan muutoksia"
...
+
+#: deviations/templates/deviations/override.html:23
+msgid "New extra time (minutes)"
+msgstr "Uusi lisäaika (minuuteissa)"
+
+#: deviations/templates/deviations/override.html:24
+msgid "Without late penalty before"
```
Mankro: The word "before" is similarly confusing like in the other comments.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -882,49 +887,93 @@ msgstr "Lisää uusia määräajan muutoksia"
...
-"trying to add a new one."
-msgstr ""
-"Määräajan muutos on jo olemassa opiskelijalle {user} tehtävässä {exercise}! "
-"Poista se ennen uuden muutokse lisäämistä."
+#: deviations/views.py:177
+msgid "Nothing removed!"
+msgstr "Mitään ei poistettu!"
+
+#: deviations/views.py:179
+msgid "Removed one deviation!"
```
Mankro: "Removed one deviation!", "Removed {number} deviations!"
You have two strings for the same thing, but singular and plural forms are separated. Plural forms should be implemented in the same string since gettext supports it. In the HTML template, you would use `{% blocktrans %}` with the `plural` section.

https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#blocktrans-template-tag

Mankro: In Python code, there are functions for doing the same: https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#pluralization

deviations/views.py
```diff
@@ -95,3 +148,48 @@ def get_resource_objects(self):
...
+                exercises = BaseExercise.objects.filter(course_module=module)
+                deviation_exercises = chain(deviation_exercises, exercises)
+            for exercise in deviation_exercises:
+                if (self.remove_deviation(exercise, profile)):
+                    number_of_removed += 1
+        if number_of_removed == 0:
+            messages.warning(self.request, _("Nothing removed!"))
+        else:
+            message = (_("Removed one deviation!") if number_of_removed == 1 else _(
+                "Removed {number} deviations!"
```
Mankro: Pluralization of translated strings must be done with the API, not with a custom if statement:
https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#pluralization

deviations/views.py
```diff
@@ -95,3 +148,48 @@ def get_resource_objects(self):
...
+
+    def remove_deviation(self, exercise, profile):
+        deviation = DeadlineRuleDeviation.objects.filter(
+                exercise=exercise,
+                submitter=profile,
+                exercise__course_module__course_instance=self.instance)
+        if deviation:
+            deviation.delete()
+            return True
+        else:
```
Mankro: Let's remove the `else` and move the `return False` one indentation level to the left.

deviations/views.py
```diff
@@ -95,3 +148,48 @@ def get_resource_objects(self):
...
+
+    def get_success_url(self):
+        return self.instance.get_url("deviations-list-dl")
+
+    def form_valid(self, form):
+        number_of_removed = 0
+        for profile in form.cleaned_data["submitter"]:
+            deviation_exercises = form.cleaned_data["exercise"]
+            for module in form.cleaned_data["module"]:
+                exercises = BaseExercise.objects.filter(course_module=module)
```
Mankro: I think that the exercises should be queried from the database outside the `for profile` loop. The exercises are the same for all users so you do not need to query the exercises repeatedly for each user.

deviations/views.py
```diff
@@ -44,39 +94,42 @@ def form_valid(self, form):
...
-                )
-                for exercise in exercises:
-                    if new_date:
-                        minutes = exercise.delta_in_minutes_from_closing_to_date(
-                                new_date)
-                    self.add_deviation(
-                        exercise, profile, minutes, without_late_penalty)
-
-            for exercise in form.cleaned_data["exercise"]:
+                exercises = BaseExercise.objects.filter(course_module=module)
```
Mankro: Does this query the exercises from the database for each user? It is unnecessary to query the same exercises multiple times.

deviations/views.py
```diff
@@ -44,39 +94,42 @@ def form_valid(self, form):
...
+            for exercise in deviation_exercises:
                 if new_date:
                     minutes = exercise.delta_in_minutes_from_closing_to_date(
                                 new_date)
-                self.add_deviation(
-                    exercise, profile, minutes, without_late_penalty)
+                if (not self.add_deviation(exercise, profile, minutes, without_late_penalty)):
+                    already_have_deviation += [(exercise.id, profile.id)]
+
+        self.request.session['minutes'] = minutes
```
Mankro: Session data keys could have a prefix that makes it less likely that the key collides with other session keys used elsewhere in the system. E.g., `deviations_minutes`

deviations/views.py
```diff
@@ -26,14 +72,18 @@ class AddDeadlinesView(CourseInstanceMixin, BaseFormView):
     access_mode = ACCESS.TEACHER
     template_name = "deviations/add_dl.html"
     form_class = DeadlineRuleDeviationForm
+    already_have_deviation = []
```
Mankro: What is the purpose of this class variable `already_have_deviation`? Why is it a list and later you do `            self.already_have_deviation = True`?

deviations/views.py
```diff
@@ -44,39 +94,42 @@ def form_valid(self, form):
...
                                 new_date)
-                self.add_deviation(
-                    exercise, profile, minutes, without_late_penalty)
+                if (not self.add_deviation(exercise, profile, minutes, without_late_penalty)):
+                    already_have_deviation += [(exercise.id, profile.id)]
+
+        self.request.session['minutes'] = minutes
+        self.request.session['without_late_penalty'] = without_late_penalty
+
+        if already_have_deviation:
```
Mankro: The difference between `already_have_deviation` and `self.already_have_deviation` feels confusing. What do they do?

deviations/views.py
```diff
@@ -2,13 +2,59 @@
...
+from .forms import DeadlineRuleDeviationForm, RemoveDeadlineRuleDeviationForm
 from .models import DeadlineRuleDeviation
 from exercise.models import BaseExercise
+from itertools import chain
+
+
+class OverrideDeadlinesView(CourseInstanceMixin, BaseFormView):
+    access_mode = ACCESS.TEACHER
+    template_name = "deviations/override.html"
+    form_class = DeadlineRuleDeviationForm
```
Mankro: Does this view really use `DeadlineRuleDeviationForm`?

deviations/views.py
```diff
@@ -2,13 +2,59 @@
...
+        return kwargs
+
+    def get_success_url(self):
+        return self.instance.get_url("deviations-list-dl")
+
+    def get_common_objects(self):
+        super().get_common_objects()
+        deviation_list = []
+        for e, s in self.request.session['already_have_deviation']:
+            deviation_list += [DeadlineRuleDeviation.objects.get(
```
Mankro: I think `DeadlineRuleDeviation.objects.get` raises an exception if the object does not exist in the database anymore. You must handle the exception. It is possible that users delete rows at the same time somebody tries to add more rows.

Is it also unnecessary to use the `+=` operator of lists here? You can `append` single elements to the list. That would be a better implementation here, right?

deviations/forms.py
```diff
@@ -51,3 +51,35 @@ def __init__(self, *args, **kwargs):
...
+        queryset=CourseModule.objects.none(),
+        required=False
+    )
+    exercise = forms.ModelMultipleChoiceField(
+        queryset=BaseExercise.objects.none(),
+        required=False
+    )
+    submitter = forms.ModelMultipleChoiceField(
+        queryset=UserProfile.objects.none(),
+        required=False,
```
Mankro: If the `submitter` field is not required, what happens when it is left empty? Are all deviations removed for the given exercises or is nothing removed?

deviations/templates/deviations/override.html
```diff
@@ -0,0 +1,54 @@
...
+      <td>
+        <input type="checkbox" name="override" value="{{ deviation.exercise.id }}-{{ deviation.submitter.id }}" checked>
+      </td>
+    </tr>
+    {% empty %}
+    <tr>
+      <td colspan="4">{% trans "No deadline deviations." %}</td>
+    </tr>
+    {% endfor %}
+	</table>
```
Mankro: Some indentations in this file look wrong.

deviations/views.py
```diff
@@ -2,13 +2,59 @@
...
+        without_late_penalty = self.request.session['without_late_penalty']
+        for string in deviation_list:
+            e, s = string.split('-')
+            DeadlineRuleDeviation.objects.filter(
+                exercise=BaseExercise.objects.get(id=e),
+                submitter=UserProfile.objects.get(id=s)).update(
+                extra_minutes=minutes,
+                without_late_penalty=without_late_penalty)
+
+        del self.request.session['already_have_deviation']
```
Mankro: Shouldn't all session keys (the ones that are used for this deviation page) be deleted at this stage?

deviations/forms.py
alalav1: Those fields are not required in the add view either. Leaving submitter empty removes nothing. This could be changed in adding of deviations, too.

deviations/views.py
```diff
@@ -1,14 +1,59 @@
...
+        super().get_common_objects()
+        deviation_list = []
+        for e, s in self.request.session['already_have_deviation']:
+            try:
+                deviation_list.append(DeadlineRuleDeviation.objects.get(
+                        exercise=BaseExercise.objects.get(id=e),
+                        submitter=UserProfile.objects.get(id=s),
+                        exercise__course_module__course_instance=self.instance))
+            except:
+                pass
```
alalav1: @Mankro how should we handle this exception?

deviations/forms.py
Mankro: Submitting empty values does not make sense so it should not validate. The user should press the cancel button if they want to cancel everything. Does the form have a cancel button?

The form validation can be changed so that it requires submitters and either modules or exercises. The form class has a method that is very suitable for validating data that depends on multiple fields ("either module or exercise must be set"). That is the `clean` method.

https://docs.djangoproject.com/en/2.2/ref/forms/validation/

deviations/views.py
Mankro: You must name the exception, something like `DeadlineRuleDeviation.DoesNotExist`.

Mankro: The user tries to add deviations here, so let's create the deviations since they don't exist for these students (even though they existed before this view was loaded).

deviations/static/deviations/remove_deviation.js
```diff
@@ -0,0 +1,20 @@
+$('table form').on('submit', function(event) {
```
Mankro: The script should be wrapped inside `$(function() { .... });` so that it waits for the DOM to finish loading.

The selector `table form` could be made more specific. E.g., `table[data-deviations-list]` if we add a data attribute to the table in the HTML template.

alalav1:
Deadline override template sometimes shows wrong minutes, fix is in the making.

alalav1:
The new deviation forms should use the AJAX search.

Mankro:
About the URL reverse issue in broken unit tests: you forgot to add the `url` field to the test exercise you created in the unit test, so naturally URL reversal does not work for it. All URLs of the course content are built in the cache code when you test the course views that usually always trigger the content cache.

PR #404, opened 2019-11-14
Fix chapter links with nested directories

alalav1:
Rst-tools write links with full directory path
`/module_name/some_folder/chapter_name`. However, in A-plus, the URL
directory is `/module_name/chapter_name`, so the folder names between must be
removed. Links can be written with or without the ending `/` so it must be checked, too.

This solution expects the URLs of the chapters to be form
`https://plus.cs.aalto.fi/course/version/module/chapter/`, and at least the courses currently in production seem to use that form.

### DIFF
--- a/lib/remote_page.py
+++ b/lib/remote_page.py
@@ -194,12 +194,14 @@ class RemotePage:
                 # If chapter is in a nested directory, folders names between
                 # the module and the chapter must be removed
                 # from the link since a-plus urls do not have them.
-                if len(element[attr_name].split('/')) > 3:
-                    path_to_module = ('/').join(element[attr_name].split('/')[0:3])+'/'
+                split_path = element[attr_name].split('/')
+                if len(split_path) > 3:
+                    path_to_module = ('/').join(split_path[0:3])+'/'
                     # Links can be written with our without the ending slash ('/')
-                    index = 1 if element[attr_name].split('/')[-1] else 2
-                    chapter_name = element[attr_name].split('/')[-index]
+                    chapter_name = '_'.join(split_path[3:])
                     element[attr_name] = path_to_module + chapter_name
+                    if element[attr_name][-1] == '_':
+                        element[attr_name] = element[attr_name][:-1]
 
             elif value and not test.match(value):
 
--- a/lib/remote_page.py
+++ b/lib/remote_page.py
@@ -191,6 +191,16 @@ class RemotePage:
                 elif not value.startswith('/'):
                     element[attr_name] = '../' + value
 
+                # If chapter is in a nested directory, folders names between
+                # the module and the chapter must be removed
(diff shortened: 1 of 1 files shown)


Plustest:

lib/remote_page.py
```diff
@@ -191,6 +191,16 @@ def _fix_relative_urls(self, url, tag_name, attr_name):
...
                     element[attr_name] = '../' + value
 
+                # If chapter is in a nested directory, folders names between
+                # the module and the chapter must be removed
+                # from the link since a-plus urls do not have them.
+                if len(element[attr_name].split('/')) > 3:
+                    path_to_module = ('/').join(element[attr_name].split('/')[0:3])+'/'
+                    # Links can be written with our without the ending slash ('/')
+                    index = 1 if element[attr_name].split('/')[-1] else 2
+                    chapter_name = element[attr_name].split('/')[-index]
```
raphendyr: Maube `element[attr_name].rstrip('/').rpartition('/')[-1]`? Though, if we already have `path_parts`, we can do `chatper_name = path_parts[-1] or part_parts[-2]`

Mankro:
This might not be the approach we want. RST-tools could compile chapter keys differently when the chapter RST files are nested inside the module directories. Currently, chapter keys collide if you have files such as `module01/mychapter.rst` and `module01/subdir/mychapter.rst`.

raphendyr:
add to whitelist

raphendyr:
As a coding style comment, I would use following code for better (presumed) performance and readability:

```
path_parts = element[attr_name].split('/') # cache the string operation (python is unable to optimize this)
if len(path_parts) > 3:
  path_to_module = '/'.join(path_parts[:3]) # no / here
  chapter_name = path_parts[-1] or path_parts[-2] # or is ok for case `x = a if a else b`
  element[attr_name] = path_to_module + '/' + chapter_name # include / here, so it's clear it's there always
```

alalav1:
Chapters having a same name in different subfolders don't work with current RST-tools, since the chapter keys would be the same. Thus the keys of such chapters must be modified, so this PR would need modifications, too. It might be better to let RST-tools to fix it's own links, so this PR is most likely useless.

lib/remote_page.py
```diff
@@ -191,6 +191,18 @@ def _fix_relative_urls(self, url, tag_name, attr_name):
                 elif not value.startswith('/'):
                     element[attr_name] = '../' + value
 
+                # If chapter is in a nested directory, folders names between
```
alalav1: This comment is outdated.

alalav1:
Links and chapter keys should be modified in the rst-tools, so this pull request is unnecessary. Work related to this issue can be found in Aalto-Letech/a-plus-rst-tools#58

PR #405, opened 2019-11-18
Load all student data in results, show progress

JoonatanPartanen:


exercise/static/exercise/results_staff.js
```diff
@@ -733,17 +749,23 @@
         let successFirstStudent = false;
 
         let checkIfAllAjaxCompleted = function() {
+            const users_progress = completedUserAjaxCalls + " / " + requiredUserAjaxCalls;
+            const exercises_progress = completedExerciseAjaxCalls + " / " + requiredExerciseAjaxCalls;
+            const points_progress = completedPointAjaxCalls + " / " + requiredPointAjaxCalls;
+            const progress_report = users_progress + "<br>" + exercises_progress + "<br>" + points_progress;
+            $("#results-loading-progress").html(progress_report)
```
Mankro: Missing `;` at the end of the line.

exercise/static/exercise/results_staff.js
```diff
@@ -705,23 +705,39 @@
...
-        _students = studentsResolve[0];
-        _usertags = usertagsResolve[0];
-
-        let requiredPointAjaxCalls = _students.results.length;
-        let requiredUserAjaxCalls = _students.results.length;
+    $.when($.ajax(exercisesUrl), gatherFromAPIPaging(studentsUrl), $.ajax(usertagsUrl))
+        .done(function(exercisesResolve, studentsPagedResults, usertagsResolve) {
+        _exercises = exercisesResolve[0].results;
+        _students = studentsPagedResults;
+        _usertags = usertagsResolve[0].results;
```
Mankro: `_usertags = usertagsResolve[0].results;`
Does this mean that only the first page is used for the user tags and if there are over 100 tags, it would break?

Mankro:
Is there anything to say about the user ID vs. profile ID confusion? Is there any issue there?

exercise/static/exercise/results_staff.js
raphendyr: Thus, should both `exercisesUrl` and `usertagsUrl` be wrapped in `gatherFromAPIPaging`?

Mankro: Even though no course uses over 100 tags, it would be safer to just deal with the potential pagination.

JoonatanPartanen:
Wrapped all paginated endpoints, added semicolon.
I created a test script that creates students and submissions and has an option for ensuring the desync of ids. It didn't seem to affect this page.

JoonatanPartanen:
If some (or all) of the requests fail, they are still counted as being done. This causes the table to load only partially or to not load at all. The partial load might be missing the points columns, from what I've seen.

raphendyr:
> If some (or all) of the requests fail, they are still counted as being done. This causes the table to load only partially or to not load at all. The partial load might be missing the points columns, from what I've seen.

I was thinking a bit about this when I was reading the code. We probably should address this somehow. Question is, is it ok enough before next iteration or should we fix this before even merging this version.

Mankro:
It would be nice to show some error to the user if the data is invalid/incomplete. Isn't it quite easy to add some visible error message somewhere? A big red banner in the top?

raphendyr:
Plan: change the loader to try 10 times with 1 second delay after an error. if the process can't complete with this, then instead of showing the table, show an error.

JoonatanPartanen:
Missing translation on error message (don't know how to do that), and there are quite a few errors in the console (I think new requests are blocked after too many failures, probably already sent requests).

exercise/static/exercise/results_staff.js
```diff
@@ -705,9 +705,30 @@
...
+          $("#ajax-failed-alert").show();
+          $("#results-loading-animation").hide(); 
+        }
+	  }
+	};		  
+
     function gatherFromAPIPaging(url) {
         function gather(cur_result, url) {
-            return $.ajax(url)
+            return $.ajax({...ajaxSettings, url: url})
```
raphendyr: Is this syntax supported by internet explorer 11 or Edge (new in ECMAScript 2018). Should be replaced with `$.extend({}, ajaxSettings, {url: url})`. ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Browser_compatibility

exercise/static/exercise/results_staff.js
```diff
@@ -742,7 +763,7 @@
                 requiredExerciseAjaxCalls++;
             });
         });
-
+
```
raphendyr: snif..

exercise/static/exercise/results_staff.js
```diff
@@ -767,20 +788,18 @@
 
         _students.forEach(function(student) {
             $.ajax({
-                url: student.url
+                ...ajaxSettings, url: student.url
```
raphendyr: https://github.com/Aalto-LeTech/a-plus/pull/405/commits/4a52003d425eeb6d76442fee16cfcb2970a074fd#r355409189

exercise/static/exercise/results_staff.js
```diff
@@ -767,20 +788,18 @@
...
                 checkIfAllAjaxCompleted();
-            }, function() {
-                completedUserAjaxCalls++;
-                checkIfAllAjaxCompleted();
             });
 
             const sid = student.id;
             $.ajax({
-                url: pointsUrl + sid
+                ...ajaxSettings, url: pointsUrl + sid
```
raphendyr: https://github.com/Aalto-LeTech/a-plus/pull/405/commits/4a52003d425eeb6d76442fee16cfcb2970a074fd#r355409189

exercise/static/exercise/results_staff.js
```diff
@@ -833,23 +852,17 @@
...
-                completedPointAjaxCalls++;
-                checkIfAllAjaxCompleted();
             });
         });
 
         _exercises.forEach(function(module) {
             module.exercises.forEach(function(exercise) {
                 $.ajax({
-                    url: exercise.url
+                    ...ajaxSettings,  url: exercise.url
```
raphendyr: https://github.com/Aalto-LeTech/a-plus/pull/405/commits/4a52003d425eeb6d76442fee16cfcb2970a074fd#r355409189

exercise/templates/exercise/staff/results.html
```diff
@@ -168,6 +168,13 @@ <h4>{% trans "Summaries" %}</h4>
 			</ul>
 		</div>
 
+		<!--Alerts-->
+		<div id="alerts" style="margin: 30px">
+		   <div id="ajax-failed-alert" class="alert alert-danger collapse" role="alert">
+  				Data couldn't be fetched. Please check your internet connection.
```
raphendyr: For django translations, wrap that to `{% trans "Data couldn't be fetched. Please check your internet connection." %}`. I can add translations after the merge.

exercise/static/exercise/results_staff.js
```diff
@@ -705,9 +705,30 @@
...
+		if (ajaxRetryCount <= ajaxRetryLimit) {
+		  console.log("Retrying ajax:", statusText);
+		  $.ajax(this);
+		  return;
+		} else {
+          $("#ajax-failed-alert").show();
+          $("#results-loading-animation").hide(); 
+        }
+	  }
+	};
```
raphendyr: Indentation is broken in this block.

exercise/templates/exercise/staff/results.html
JoonatanPartanen: Marvelous :+1:

exercise/static/exercise/results_staff.js
Mankro: We do have lots of JS code that does not work in IE 11, but if the latest Edge does not work either, then that's bad.

JoonatanPartanen:
I hope this is close to the final revision. There is an interesting problem where the AJAX timeout timer starts when the request $.ajax() call is made, but the requests are stalled for quite a long time. Any timeout I set will likely be exceeded as the number of students increases.

JoonatanPartanen:
I'll remove the timeout altogether. There will be other errors if there is no internet connection. Worth noting though.

exercise/static/exercise/results_staff.js
```diff
@@ -705,23 +705,64 @@
...
+        timeout: 0,
+        beforeSend: function(xhr, settings) {
+            return ajaxEnabled;
+        },
+        error: function(xhr, statusText, errorThrown) {
+            this._retryCount++;
+            if (this.retryCount <= this._retryLimit) {
+                console.log("Retrying ajax:", statusText);
+                //const req = this;
+                //setTimeout(function() {$.ajax(req);}, 1000);
```
raphendyr: This comment was not supposed to come here, but maybe it can stay.

exercise/templates/exercise/staff/results.html
```diff
@@ -160,14 +160,21 @@ <h4>{% trans "Summaries" %}</h4>
...
 				<span class="caret"></span>
 			</button>
 			<ul id="export-button-menu" class="dropdown-menu">
 
 			</ul>
 		</div>
 
+		<!--Alerts-->
+		<div id="alerts" style="margin: 30px">
+			<div id="ajax-failed-alert" class="alert alert-danger collapse" role="alert">
```
raphendyr: `hidden` would probable be a better class or did `collapse` look better?

raphendyr:
Good for merging.

exercise/templates/exercise/staff/results.html
JoonatanPartanen: Hidden doesn't work, apparently.

PR #406, opened 2019-11-29
Teacher can submit in any group

alalav1:
Teacher's submissions don't get rejected, when teacher is already in some group.
If student can't do submissions, but teacher for some reason wants to give points for a certain exercise, they can submit on their own and add the student in the submitters. However, this previously could be done only once per exercise since teachers submissions in different groups got rejected. In this change, teacher's submissions will never get rejected for invalid group.

### DIFF
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -566,7 +566,7 @@ class BaseExercise(LearningObject):
         # Support group id from post or currently selected group.
         group = None
         group_id = request.POST.get("_aplus_group") if request else None
-        if not group_id is None:
+        if group_id is not None:
             try:
                 gid = int(group_id)
                 if gid > 0:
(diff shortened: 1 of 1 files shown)


exercise/exercise_models.py
```diff
@@ -566,7 +566,7 @@ def _check_submission_allowed(self, profile, request=None):
         # Support group id from post or currently selected group.
         group = None
         group_id = request.POST.get("_aplus_group") if request else None
-        if not group_id is None:
```
Mankro: `request.POST.get("_aplus_group")`
Did you check if the code finds any group in this case? Normally, students create a group in a separate page first and then they can choose the group when they are making the submission to the exercise. However, when the teacher modifies the submitters of a submission in the "inspect submission" page, the submission has a group even though nobody has created the group in the database. Is the `_aplus_group` hidden field empty then?

exercise/exercise_models.py
```diff
@@ -649,6 +653,14 @@ def _detect_submissions(self, profile, group):
             ))
         return False
 
+    def group_contains_staff(self, group, profile):
+        if group:
+            return any((
+                self.course_instance.is_course_staff(p.user)
```
Mankro: Does this allow the student to make new submissions with any group after the student has been added to a teacher's submission?

alalav1: No, the student is not able to make submissions after the teacher has added the student to the submission. The group_id is read from post data so I assume it is the group that made the submission which is currently in the validation process. But I have to admit that I don't know how the group made by teacher works, e.g. how is it saved in the database etc.

Mankro: I had a feeling that there is no group in the database since nobody has created the group when the teacher modifies the submitters afterwards in the "inspect submission" page.

Mankro:
We probably do not want to merge this PR because it is dirty to allow teachers to break normal group submission constraints. Let's leave this PR open for now, though.

Mankro:
Closing as outdated.

PR #408, opened 2019-12-13
Add css class to the neutral option

alalav1:
This class is needed to separate correct answers from neutral. Mooc-grader PR: Aalto-LeTech/mooc-grader#51


Mankro:
I clarified the commit message a little.

PR #464, opened 2020-01-22
Feature exam app

alalav1:


Mankro:
The old version of this PR was in the forked repository (https://github.com/motonki/a-plus/pull/1) and it was based on an outdated version of the master branch. This new PR contains new commits and fixes.

Before this is merged, the database migrations must be removed and recreated so that they are based on the up-to-date upstream master branch. Migrations are executed in order and each migration refers to its predecessor.

Before merge, Finnish translations need to be checked and the missing translations must be added.

Mankro:
The A+ exam app development has been inactive since January 2020. It is still unfinished and it seems unlikely that it would be continued and finished. The code is falling behind the newer changes in the upstream master branch. I merged this PR so that the unfinished code can be archived in the feature-exam-app branch of the upstream repo. Unfortunately, finding the discussion chain of this PR becomes a little harder when the PR has been closed, but it can be found with the name "feature exam app".

Issue #432, opened 2020-01-17
Bug: student's unofficial submissions are counted in the points under the Visualizations page and in the API. The total results table shows the correct points. (Unofficial submissions must not affect grading even though the submission may be better than previous ones.)


Mankro:
**Status in January 2020**: someone should check if there are still any bugs left concerning unofficial submissions (in the new all results page, points cache, API, visualizations page).

Status in 17th April 2019:

- A+ visualizations page includes unofficial submissions in the data, which is misleading and should be fixed. The graphs show unofficial points amongst the normal points without differentiation. The visualizations page seems to download data from the A+ API, e.g., /api/v2/courses/1/aggregatedata/

- user results API (/api/v2/courses/COURSEID/points/USERID/) gives correct points in the points_by_difficulty field. (This API takes data from the points cache.) The submission_count does not include unofficial submissions, either. However, the exercises list shows the best submissions for each exercise and if the student has only unofficial submissions in an exercise, then the best exercise is unofficial. If there are also official submissions, then unofficial submission is not the best even if it got more points (which is what we want).
SIMPLE IDEA FOR A FIX: When there are only unofficial submissions, the exercise object could contain a field "unofficial true/false" that quickly shows that the points are unofficial. The complete submission status can be queried from the submission detail API endpoint, but it requires another HTTP query. (exercise/api/custom_serializers.py ExercisePointsSerializer)
ANOTHER IDEA: is it necessary for the points cache to ever show unofficial submissions as the best? What if the cache shows always zero points if there are only unofficial submissions? Would this break something (the points cache is used in many places)?

- the hacky final course grade calculator used in the O1 course does not query the API. It uses data that is embedded into the HTML page. The data originates from the points cache, which gives correct (total) points by difficulty.

- The all results page for the course staff shows a table of the results. They are queried from the database without the points cache. The query uses the filter submission status ready, which means it works correctly with unofficial submissions.

Mankro:
Old comment:

The aggregatedata API endpoint used by the visualizations page does not use the points cache. It queries the database itself. A new simple fix was made that removes unofficial submissions from the results. However, the implementation should maybe use the points cache instead of custom queries. Link to the small fix:
Aalto-LeTech/a-plus: 3324bfa

Mankro:
@alalav1 can you work on this?

alalav1:
I start investigating.

Mankro:
[...]

PR #466, opened 2020-01-27
Confirm the level exercises need to be submitted.

alalav1:
# Description
Previously, all exercises that had `points_to_pass` as 0 had the `passed`
value `True` as a default, even without any submissions. If the chapter
had an exercise with `confirm_the_level` as `True`, it was often considered
passed and it didn't have to be submitted. By changing the default `passed`
value for the level confirming exercises, those exercises really have to be
submitted to get any points for the chapter.

This PR is related to the issue #432 and even though some visual updates were suggested,
this change already fixes the issue where points were given when they shouldn't.
# Testing

- [ ] All unit tests pass
- [x] I have tested manually
- [x] I have tested with my own test data generated by a python script

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -45,6 +45,8 @@ class CachedPoints(ContentMixin, CachedAbstract):
                         'graded': False,
                         'unofficial': False, # TODO: this should be True, but we need to ensure nothing breaks when it's changed
                     })
+                if entry.get('confirm_the_level'):
+                    entry['passed'] = False
                 r_augment(entry.get('children'))
         for module in modules:
             module.update({


alalav1:
The visualizations page still includes unconfirmed exercises. It might be better to have it that way, since most people will anyway return the mandatory exercises and thus including the unconfirmed points gives more accurate information of the collective progress.

Mankro:
We suspect that somebody has broken unconfirmed points within the past two years, but this PR fixes again how unconfirmed points are rendered in the exercise results page (i.e., the page with the module-level points progress bars). With this fix, exercise point badges are rendered with the unconfirmed-points CSS class which adds some opacity to make the points look "unconfirmed". The badge for the feedback exercise (that confirms the points) shows a red alert until it has been submitted. These visual effects have existed in the code for years, but apparently they have been broken for the past 1-2 years due to a change in the cache code.

alalav1:
Update: This PR has now some visual updates. Unconfirmed points are shown with the grey colour in the same progress bar as the confirmed points. The tool tip has been updated to show unconfirmed points:
`Points: 20 (+40) / 100`
If there are no unconfirmed points (the usual case) the points are shown as before.

Issue #467, opened 2020-01-30
Teacher should be able to manually enrol students in the course


Mankro:
Sometimes teachers need to manually add students to the course after the normal enrollment period has ended. The student can not enroll in the A+ course himself/herself at that point since A+ does not allow it.

There could be a new subpage in the A+ course participants section. It would have a form in which the teacher could select students to enroll in the course. The teacher would search for students with their names or student IDs. This manual enrollment could always be done by the teacher even outside the course enrollment open and close times.

jsorva:
N.B. When addressing this, it's important to consider the issue of background questionnaires (which some courses require students to fill in upon enrolling).

Mankro:
@jsorva I would say that the enrollment questionnaire is skipped and ignored if the teacher manually enrolls a student afterwards. Otherwise, the implementation would become awkward and complicated. This feature is only intended to be used in rare cases when the student must be added to the course after the normal enrollment period. I would keep this implementation very simple.

Mankro:
@alalav1 started to work on this.

Mankro:
Duplicate enrollments may become a bigger issue when teachers can enroll students manually: #512.

PR #469, opened 2020-02-05
Enroll students

alalav1:
# Description
Add teachers possibility to enrol students even after the enrolment has ended.

Participants page has a new button which directs to the form in which the teacher can select students
to be added to the course. Only existing students that are not yet enrolled are shown in the choices.

Related issue #467 

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant


alalav1:
This PR still requires translations.

course/forms.py
```diff
@@ -93,3 +94,35 @@ def __init__(self, *args, **kwargs):
...
+        queryset=UserProfile.objects.none(),
+        required=True,
+        error_messages=custom_errors
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.instance = kwargs.pop('course_instance')
+        super().__init__(*args, **kwargs)
+        self.fields['user_profiles'].widget.attrs['class'] = 'search-select'
+        self.fields['user_profiles'].queryset = UserProfile.objects.exclude(
```
raphendyr: This is too heavy. We have around 14k people in the database and loading this widget will take couple of minutes.

course/forms.py
```diff
@@ -93,3 +94,35 @@ def __init__(self, *args, **kwargs):
...
+            id__in=[profile.id for profile in self.instance.get_student_profiles()]
+        )
+
+    def clean(self):
+        value = self.data.getlist('user_profiles')
+        existing_students = [i.id for i in self.instance.get_student_profiles()]
```
raphendyr: Check existence with a database query. E.g. `Enrollment.objects.filter(course=, student=).exists()`

raphendyr: Or you can retrieve all existing (`Enrollment.objects.filter(.., profile__in=[])`) and then return the error page.

course/forms.py
alalav1: Using `ModelMultipleChoiceField` is not possible without setting possible choices in the queryset.
We still propably want that multiple students can be added at a time. Is it better to have the students added by student ids? Or can we do something else to have the field suggest the possible values?

Mankro: To my understanding, the "add assistants" field in the edit course page uses the same logic. It includes all users in the HTML.

course/forms.py
alalav1: It would indeed be better that all the students didn't get rejected, if one was already in the course. The validation in the for loop seems sufficient, the students not in course could be added to the `cleaned_data` in some new value.

course/forms.py
alalav1: @raphendyr Do you have any suggestions, how we should handle this efficiency problem.

My proposition is that we used some other field which could handle multiple strings. The user would add new students by student ids. Validation would not require that much modification. It would work fast, but the teacher wouldn't be able to see, if the id they had written was valid.

Mankro:
@raphendyr We should decide on how we approach user experience and performance on these kinds of forms. Ideally, the page should not contain all users of the database in the HTML and the user should be able to select users easily without writing their full names.

course/long_urls.py
```diff
@@ -23,6 +23,9 @@
     url(EDIT_URL_PREFIX + r'participants/$',
         staff_views.ParticipantsView.as_view(),
         name="participants"),
+    url(EDIT_URL_PREFIX + r'enroll_students/$',
```
Mankro, line 26: Let's use a hyphen in URLs instead of an underscore: `r'enroll-students/$'`

course/views.py
```diff
@@ -198,3 +199,17 @@ def form_valid(self, form):
...
+
+
+class EnrollStudentsView(CourseInstanceMixin, BaseFormView):
+    access_mode = ACCESS.TEACHER
+    form_class = EnrollStudentsForm
+    template_name = "course/staff/enroll_students.html"
+
+    def form_valid(self, form):
+        for profile in form.cleaned_data["user_profiles"]:
+            self.instance.enroll_student(profile.user)
```
Mankro, line 211: Is there any warning if the selected user was already enrolled in the course?

course/templates/course/staff/enroll_students.html
```diff
@@ -0,0 +1,30 @@
...
+{% block view_tag %}participants,enroll-students{% endblock %}
+
+{% block breadcrumblist %}
+{{ block.super }}
+<li><a href="{{ instance|url:'participants' }}">{% trans "Participants" %}</a></li>
+<li class="active">{% trans "Add students" %}</li>
+{% endblock %}
+
+{% block columns %}
+<h3>{% trans "Add students to this course" %}</h3>
```
Mankro, line 17: "Enroll students in the course" is a better title.

Mankro:
This PR was rebased onto the upstream master so that it can use the new AJAX search for selecting users. Therefore, some previous comments about the performance of the user selection form are outdated.

Mankro:
#512 duplicate enrollments are annoying, but not really part of this PR.

Mankro:
I rebased this onto the upstream master and added some fixes in new commits.

PR #471, opened 2020-02-06
Show initialized and waiting submissions in the submission count.

### LINKED ISSUE #407: Different submission count between course "exercise results" page and assignment submission view
I spotted this in Aalto's production A+ (I'm a student enrolled on O1).

When I view the "exercise results" page, for one assignment I see the following (note the "0/10" submissions):

Screenshot 2019-11-30 at 19 29 42

When I follow the link to the chapter, and the view the assignment, it states I've used one submission (which is currently ungraded).
 
Screenshot 2019-11-30 at 19 29 55

It seems to only count graded submissions (this particular assignment is human marked).

This may be the intended behaviour, but it felt wrong that they were not consistent and it caused me a brief moment of intense "did I forget to submit the big assignment before the deadline" panic.

alalav1:
Even though the submission with statuses `initialized` or `waiting`
can be rejected, it is better to have them counted in the submission
count. It is more clear for the students so they can see how many
submissions they've made. If the submission gets rejected, it will be
removed from the total submission count.

Adding these statuses to the total count benefits especially exercises
that are graded manually, such as Rubyric exercises. Also not counting
those statuses could give false information of the submissions left, if the
waiting submission should pass.

Fixes #407 
# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -87,7 +87,7 @@ class CachedPoints(ContentMixin, CachedAbstract):
                 entry = tree[-1]
                 ready = submission.status == Submission.STATUS.READY
                 unofficial = submission.status == Submission.STATUS.UNOFFICIAL
-                if ready:
+                if ready or submission.status in (Submission.STATUS.WAITING, Submission.STATUS.INITIALIZED):
                     entry['submission_count'] += 1
                 entry['submissions'].append({
                     'id': submission.id,


alalav1:
Previously, @raphendyr made pull request which removed submissions with statuses `rejected`, `waiting` or `initialized` from total submission count https://github.com/apluslms/a-plus/commit/0a9fd39a92bad5cb4fd92fba4b6506b5e6f67d5f stating that waiting and initialized submissions shouldn't count, since they might be rejected later.

raphendyr:
This should be good. Most important change before was to drop error statuses. Can you list all locations which react to submission_count from cache? If i remember right, `ordinal` in grader url was calculated differently?

alalav1:
The `ordinal_number` is calculated in the submission_models using submissions with status `ready`, `initialized`, `waiting` or `unofficial`, so this change do not affect it. Previously the actual ordinal number could be different than ordinal number shown to students, since `waiting` and `initialized` submissions were not counted. 

In general this change does not affect back end or grading. Following templates use the changed `submission_count`: `_children.html`, `_points_badge.html` and `_user_results.html` .

PR #472, opened 2020-02-06
Add a custom filter to list future deadlines in the admin page

alalav1:
Add a filter object to course modules to show only the modules with deadline
in the future. This object also orders the modules by `closing_time`.

# Description

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/course/admin.py
+++ b/course/admin.py
@@ -1,5 +1,6 @@
 from django.contrib import admin
 from django.utils.translation import ugettext_lazy as _
+from django.utils import timezone
 
 from course.models import (
     Course,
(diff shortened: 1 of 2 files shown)


raphendyr:
I guess this solves the immediate problem, even though the goal is to reduce use of django admin. :+1:

course/admin.py
```diff
@@ -62,11 +63,26 @@ class EnrollmentAdmin(admin.ModelAdmin):
     list_filter = ("course_instance",)
 
 
+# Filtering obejct to help scheduling maintenance breaks
```
Mankro: Misspelled object in the comment line.

course/admin.py
```diff
@@ -62,11 +63,26 @@ class EnrollmentAdmin(admin.ModelAdmin):
...
+class FutureDeadlines(admin.SimpleListFilter):
+    title = _('Future deadlines')
+    parameter_name = 'future_deadlines'
+
+    def lookups(self, request, model_admin):
+        return (
+            ('now', _("Show only modules with 'closing_time in the' future.")),
```
Mankro: Why are the quotes like this:  'closing_time in the'

course/admin.py
```diff
@@ -62,11 +63,26 @@ class EnrollmentAdmin(admin.ModelAdmin):
     list_filter = ("course_instance",)
 
 
+# Filtering obejct to help scheduling maintenance breaks
+class FutureDeadlines(admin.SimpleListFilter):
+    title = _('Future deadlines')
```
Mankro: Is "Coming deadlines" a better title?

course/admin.py
```diff
@@ -62,11 +63,26 @@ class EnrollmentAdmin(admin.ModelAdmin):
...
+    parameter_name = 'future_deadlines'
+
+    def lookups(self, request, model_admin):
+        return (
+            ('now', _("Show only modules with 'closing_time in the' future.")),
+        )
+
+    def queryset(self, request, queryset):
+        if self.value() == 'now':
+            return queryset.filter(closing_time__gte=timezone.now()).order_by('closing_time')
```
Mankro: The `late_submission_deadline` should also be included here in the query set filter since students may still submit at that time. It must also check that the late submissions are enabled since otherwise, the late deadline would be meaningless (it could have a value but the date is never used if the late submissions are disabled).

course/admin.py
```diff
@@ -62,11 +63,26 @@ class EnrollmentAdmin(admin.ModelAdmin):
...
+
+
 class CourseModuleAdmin(admin.ModelAdmin):
     list_display_links = ("__str__",)
     list_display = ("course_instance", "__str__",
         "opening_time", "closing_time", instance_url)
```
Mankro: Let's add `late_submission_deadline` to the `list_display`.

alalav1:
I accidentally moved changes from the other branch here.

alalav1:
We fixed the broken git history with @Mankro . This branch contains the correct history now and it has been rebased onto upstream master.

Issue #473, opened 2020-02-06
The exercise results page does not show that a module is open for reading while the exercises are still closed.


Mankro:
The UI leaves the impression that the module is completely closed (between the reading opening time and the opening time). The "opens" badge in the module box also only shows the normal open time. The UI should show clearly that a module is open for reading. The collapsing box should be opened by default for read-open modules.

Mankro:
@alalav1 you can work on this.

PR #474, opened 2020-02-07
Show modules open in the results page, if they are open for reading

alalav1:
# Description
If module is open for reading, it is now opened by default in the
exercise results page. Add a badge to inform that this module is for
reading only. Add correct prepositions with dates (Opens on saturday).
Template tag `module_accessible` fixed to work properly, even when
opening time is not set.
Related issue #473 

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/exercise/templates/exercise/_user_results.html
+++ b/exercise/templates/exercise/_user_results.html
@@ -24,14 +24,21 @@
 {% module_accessible module as accessible %}
 {% exercise_accessible module as exercise_accessible %}
 <div class="panel panel-primary module-panel{% if not open and after_open %} module-expired{% endif %}">
-  <a class="panel-heading{% if not open or not accessible %} collapsed{% endif %}"
+  <a class="panel-heading{% if not accessible or not open and after_open %} collapsed{% endif %}"
     role="button" href="#module{{ module.id }}" data-toggle="collapse"
-    aria-expanded="{% if open and accessible %}true{% else %}false{% endif %}" aria-controls="#module{{ module.id }}">
+    aria-expanded="{% if accessible and not after_open or open %}true{% else %}false{% endif %}" aria-controls="#module{{ module.id }}">
     <h3 class="panel-title">
       {% points_badge module "pull-right" %}
-      {% if not open and not after_open %}
+      {% if not accessible %}
       <span class="badge pull-right">
-        {% trans "Opens" %} {{ module.opening_time }}
+        {% trans "Opens on" %} {{ module.opening_time }}
+      </span>
+      {% elif not after_open %}
+      <span class="badge pull-right">
+        {% trans "Exercises open on" %} {{ module.opening_time }}
+      </span>
+      <span class="badge badge-info pull-right">
+        {% trans "Open for reading" %}
       </span>
       {% endif %}
       {% if module.requirements|length > 0 %}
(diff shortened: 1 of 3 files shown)


alalav1:
I found out that links to the exercises are not functional, if the module closing time is passed, but the late submissions are still allowed. Sadly, the currently used `exercises_open` function cannot be modified, since it is used to separate late and open exercises in the other templates. The `have exercises_been_opened` function could be used.

exercise/templatetags/exercise.py
```diff
@@ -223,8 +223,11 @@ def min_group_size(context):
 
 @register.simple_tag(takes_context=True)
 def module_accessible(context, entry):
-    t = entry.get('reading_opening_time', entry.get('opening_time'))
-    return _is_accessible(context, entry, t)
+    t = entry.get('reading_opening_time')
```
Mankro: Does this really change anything? Looks like it does the same stuff as the old code.

Mankro:
Have the exercise links been fixed when late submissions are allowed?

exercise/templatetags/exercise.py
alalav1: The old code didn't work, if `reading_opening_time` was not set. I think it was maybe the nested get method that caused malfunction. (The outer get method would always find the object even if it was empty.)

alalav1: But the logic should be same, and the changed version works as it should.

alalav1:
> Have the exercise links been fixed when late submissions are allowed?

That fix is made in another branch with other late submission related changes.

Mankro:
Are you referring to https://github.com/apluslms/a-plus/pull/475?

alalav1:
Yes.

exercise/templates/exercise/_user_results.html
```diff
@@ -24,14 +24,21 @@
...
+    aria-expanded="{% if accessible %}true{% else %}false{% endif %}" aria-controls="#module{{ module.id }}">
     <h3 class="panel-title">
       {% points_badge module "pull-right" %}
-      {% if not open and not after_open %}
+      {% if not accessible %}
       <span class="badge pull-right">
-        {% trans "Opens" %} {{ module.opening_time }}
+        {% trans "Opens on" %} {{ module.opening_time }}
+      </span>
+      {% elif not open and not after_open %}
```
Mankro: `not after_open` would be shorter and semantically identical, right?

exercise/templates/exercise/_user_results.html
```diff
@@ -44,7 +51,7 @@ <h3 class="panel-title">
       {{ module.name|parse_localization }}
     </h3>
   </a>
-  <div class="collapse{% if open and accessible %} in{% endif %}" id="module{{ module.id }}">
```
Mankro: Previously, when a module got closed (after the close time), the div was also closed/collapsed. After your change, the div is always open forever after the open time. This should be fixed. The div should be closed after the module close time. This probably affects the `a.panel-heading` element in the top too.

exercise/templates/exercise/_user_results.html
```diff
@@ -24,14 +24,21 @@
...
+      {% if not accessible %}
       <span class="badge pull-right">
-        {% trans "Opens" %} {{ module.opening_time }}
+        {% trans "Opens on" %} {{ module.opening_time }}
+      </span>
+      {% elif not after_open %}
+      <span class="badge pull-right">
+        {% trans "Exercises open on" %} {{ module.opening_time }}
+      </span>
+      <span class="badge badge-success pull-right">
```
Mankro: Change the class on the "open for reading" badge from `badge-success` to `badge-info`, please.

Mankro:
I squashed, slightly modified the commit message and force-pushed before merging.

PR #475, opened 2020-02-10
Show the time of the late deadlines

### LINKED ISSUE #382: Missing time on late submission deadline
Time is not shown on late submissions, only the date can be seen. Jaakko tasked me to create an issue for it.

alalav1:
# Description
Previous version showed only the date of the deadline.
Additionally, activate the links to the exercises between the closing time and the late deadline.
Fix #382 
# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -443,29 +443,31 @@ class BaseExercise(LearningObject):
         in consideration.
         """
         timing,d = self.get_timing(students, when or timezone.now())
+
+        formatted_time = date_format(timezone.localtime(d), "DATETIME_FORMAT")
         if timing == self.TIMING.OPEN:
             return True,[]
         if timing == self.TIMING.LATE:
             # xgettext:no-python-format
             return True,[_("Deadline for the exercise has passed. Late submissions are allowed until {date} but points are only worth {percent:d}% of normal.").format(
-                date=date_format(d),
+                date=formatted_time,
                 percent=self.course_module.get_late_submission_point_worth(),
             )]
         if timing == self.TIMING.UNOFFICIAL:
             return True,[_("Deadline for the exercise has passed ({date}). You may still submit to receive feedback, but your current grade will not change.").format(
-                date=date_format(d)
+                date=formatted_time,
             )]
         if timing == self.TIMING.CLOSED_BEFORE:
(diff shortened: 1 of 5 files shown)


exercise/exercise_models.py
```diff
@@ -443,29 +443,37 @@ def one_has_access(self, students, when=None):
...
         """
         timing,d = self.get_timing(students, when or timezone.now())
+
+        # Use format that shows the hours and minutes
+        # TODO: This should use formats set in specific language settings
+        if get_language() == 'fi':
+            time_format = 'j.n.Y l H:i'
+        else:
+            time_format = 'D, M j Y, P'
+        formatted_time = date_format(d, time_format)
```
alalav1: Setting the `time_format` manually is not the cleanest way to do this. The `DATETIME_FORMAT` set in the language files of the `aplus` folder should be used instead.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2279,10 +2279,9 @@ msgid "Requires"
 msgstr "Vaaditaan"
 
 #: exercise/templates/exercise/_user_results.html:55
-#, python-format
```
Mankro: Did you generate this file with `python manage.py makemessages`? I am a little surprised that the `python-format` flag has been removed since the string contains `%(deadline)s`. Just checking.

Mankro:
@alalav1 This looks nice. Please write a better commit message and then you can squash the commits.

exercise/exercise_models.py
```diff
@@ -443,29 +443,31 @@ def one_has_access(self, students, when=None):
         in consideration.
         """
         timing,d = self.get_timing(students, when or timezone.now())
+
+        formatted_time = date_format(d, "DATETIME_FORMAT")
```
Mankro: It looks like this messes up time zones. The message is printed in the normalized UTC+0 time zone even if the server is set for the Finnish UTC+2 time zone.

Mankro: You can see it when you submit to an exercise during the late time. I wrote the hours 17:00 but the message shows 15:00.

Mankro:
I squashed and force-pushed before merging.

Issue #477, opened 2020-02-10
Update text on the Edit exercise button


Mankro:
In the A+ edit course page, the settings of exercises and chapters can be modified manually. The text on the button is always "Edit exercise". This text should be changed so that it is still "edit exercise" for real exercises, but for content chapters it should be different since they are not really exercises that students submit solutions into. I suggest the text "Edit content chapter".

Mankro:
Fixed in #478

PR #478, opened 2020-02-11
Separate chapters and exercises in the Edit course page

alalav1:
# Description
The exercises and chapters are now separated in the Edit course page.
Fixes the previous implementation of  the `is_submittable` function in the template.
Related issue #477 

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/edit_course/templates/edit_course/edit_content.html
+++ b/edit_course/templates/edit_course/edit_content.html
@@ -106,13 +106,14 @@
                 </td>
                 <td>
                     <a class="btn btn-default btn-xs" href="{{ lobject|editurl:'exercise' }}">
-                        <span class="glyphicon glyphicon-edit" aria-hidden="true"></span> {% trans "Edit exercise" %}
+                      <span class="glyphicon glyphicon-edit" aria-hidden="true"></span>
+                      {% if lobject.is_submittable %}{% trans "Edit exercise" %}{% else %}{% trans "Edit content chapter" %}{% endif %}
                     </a>
                     <a class="btn btn-default btn-xs" href="{{ lobject|removeurl:'exercise' }}">
                         <span class="glyphicon glyphicon-remove" aria-hidden="true"></span> {% trans "Remove" %}
                     </a>
                     <a class="btn btn-link btn-xs" href="{{ lobject|url }}">
-                    	{% trans "Open exercise" %}
+                      {% if lobject.is_submittable %}{% trans "Open exercise" %}{% else %}{% trans "Open content chapter" %}{% endif %}
                     </a>
                     {% if lobject.is_submittable %}
                     <a class="btn btn-link btn-xs" href="{{ lobject|url:'submission-list' }}">
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-02-14 10:27+0200\n"
+"POT-Creation-Date: 2020-02-18 18:47+0200\n"
 "PO-Revision-Date: 2019-08-14 12:16+0200\n"
 "Last-Translator: Jaakko Kantojärvi <jaakko.kantojarvi@aalto.fi>\n"
 "Language-Team: Finnish <>\n"
@@ -733,7 +733,7 @@ msgstr "Poista kaikki"
 #: deviations/templates/deviations/list_dl.html:44
 #: edit_course/templates/edit_course/edit_content.html:53
 #: edit_course/templates/edit_course/edit_content.html:91
-#: edit_course/templates/edit_course/edit_content.html:112
+#: edit_course/templates/edit_course/edit_content.html:113
 #: edit_course/templates/edit_course/remove_model.html:51
 #: edit_course/templates/edit_course/usertag_delete.html:29
 #: edit_course/templates/edit_course/usertag_list.html:48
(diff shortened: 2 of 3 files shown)


edit_course/templatetags/editcourse.py
```diff
@@ -34,6 +35,13 @@ def removeurl(model_object, model_name):
     ))
 
 
+@register.simple_tag
+def submittable(entry):
+    if entry.content_type_id == 21:
```
alalav1: There must be a better way to do this. Somehow using `entry.is_submittable` did not work, even though the `is_submittable` is defined in the `LearningObject` class. Entries are not handled as `BaseExercises`, even though the objects would be members of that class.

Mankro: `entry.is_submittable` may also refer to the values in the content cache or some template tag. So it might not directly call the method in the model class (LearningObject). Can you find out which method it really calls and where the function is defined?

edit_course/templatetags/editcourse.py
```diff
@@ -34,6 +35,13 @@ def removeurl(model_object, model_name):
     ))
 
 
+@register.simple_tag
+def submittable(exercises, lobject):
```
alalav1: Opinions on this @Mankro ?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1323,81 +1323,89 @@ msgstr "Muokkaa moduulia"
...
+msgstr "Muokkaa kappaletta"
+
+#: edit_course/templates/edit_course/edit_content.html:117
 msgid "Open exercise"
 msgstr "Avaa tehtävä"
 
-#: edit_course/templates/edit_course/edit_content.html:119
+#: edit_course/templates/edit_course/edit_content.html:117
+msgid "Open chapter"
+msgstr "avaa kappale"
```
Mankro: Change avaa to Avaa.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1323,81 +1323,89 @@ msgstr "Muokkaa moduulia"
...
-#: edit_course/templates/edit_course/edit_content.html:109
+#: edit_course/templates/edit_course/edit_content.html:111
 #: exercise/templates/exercise/exercise_base.html:91
 msgid "Edit exercise"
 msgstr "Muokkaa tehtävää"
 
-#: edit_course/templates/edit_course/edit_content.html:115
+#: edit_course/templates/edit_course/edit_content.html:111
+msgid "Edit content chapter"
+msgstr "Muokkaa kappaletta"
```
Mankro: We have been uncertain about how to translate "content chapter" to Finnish. Sometimes it is called "sisältökappale". "Kappale" alone does not feel sufficient. You should use "sisältökappale" here unless we get better ideas.

edit_course/templatetags/editcourse.py
Mankro: It is bad and we need to find a better approach. Templates should be able to use a variable like `lobject.is_submittable`. It can't be necessary to build a list of exercise IDs in order to do this.

edit_course/templates/edit_course/edit_content.html
```diff
@@ -106,15 +107,16 @@ <h3 class="panel-title">{% trans "Course modules" %}</h3>
...
                     </a>
                     <a class="btn btn-default btn-xs" href="{{ lobject|removeurl:'exercise' }}">
                         <span class="glyphicon glyphicon-remove" aria-hidden="true"></span> {% trans "Remove" %}
                     </a>
                     <a class="btn btn-link btn-xs" href="{{ lobject|url }}">
-                    	{% trans "Open exercise" %}
+                    	{% if is_exercise %}{% trans "Open exercise" %}{% else %}{% trans "Open chapter" %}{% endif %}
                     </a>
-                    {% if lobject.is_submittable %}
+                    {% if is_exercise %}
```
Mankro: It looks like that `{% if lobject.is_submittable %}` is currently broken in production and it always evaluates to False. (The "view submissions" link is never visible.) This must also be fixed now. This used to work in the past, perhaps over a year ago.

We must find the code that really defines the value of `lobject.is_submittable` and hopefully it will be easy to fix it. Has the name of the attribute been changed to something else than `is_submittable`? Is the code that defines the value broken?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1323,81 +1323,89 @@ msgstr "Muokkaa moduulia"
...
+msgstr "Muokkaa sisältökappaletta"
+
+#: edit_course/templates/edit_course/edit_content.html:116
 msgid "Open exercise"
 msgstr "Avaa tehtävä"
 
-#: edit_course/templates/edit_course/edit_content.html:119
+#: edit_course/templates/edit_course/edit_content.html:116
+msgid "Open content chapter"
+msgstr "Muokkaa sisältökappaletta"
```
Mankro: Open is "avaa", not "muokkaa".

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3635,3 +3643,6 @@ msgstr "Unohda automaattinen uudelleenohjaus."
 #: userprofile/views.py:80
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
+#~ msgid "Open chapter"
```
Mankro: The old string "open chapter" should be removed if it is not used anywhere. I guess the makemessages command commented it out automatically.

PR #479, opened 2020-02-20
Use ajax in search fields.

alalav1:
# Description
Search Users from API instead of loading all 10000+ UserProfiles to the HTML template.
The pages containing search fields should download much faster.

# Testing

- [ ] All unit tests pass
- [x] I have tested manually

- [ ] Yes
- [x] Not relevant

### DIFF
--- a/assets/js/ajax_search_select.js
+++ b/assets/js/ajax_search_select.js
@@ -1,0 +1,182 @@
+$(function() {
+    "use strict";
+
+    $('.search-select-ajax').aplusSearchSelectAjax();
+});
+
+(function($) {
+    "use strict";
+
+    const pluginName = "aplusSearchSelectAjax";
+    var defaults = {
+        widget_selector: "#search-select-widget",
+        field_selector: 'input[type="text"]',
+        search_selector: '.dropdown-toggle',
+        result_selector: '.search-options',
+        selection_selector: '.search-selected',
+    };
+
+    function AplusSearchSelectAjax(element, options) {
+        this.element = $(element);
+        this.timeout = null;
+        if (this.element.prop("tagName") == "SELECT" && this.element.prop("multiple")) {
+            this.settings = $.extend({}, defaults, options);
+            this.init();
+        }
+    }
+
+    $.extend(AplusSearchSelectAjax.prototype, {
+
+        init: function() {
+            this.widget = $(this.settings.widget_selector).clone()
+                .removeAttr("id").removeClass("hide").insertBefore(this.element);
+            this.element.hide();
+            var self = this;
+            this.selection = this.widget.find(this.settings.selection_selector);
+            this.api_url = self.element.attr("data-search-api-url");
+            this.parameter_list = self.element.attr("data-key-parameter-list").split(",");
+            this.selection_li = this.selection.find("li").remove();
+            this.element.find("option:selected").each(function(index) {
+                $.ajax({
+                    url: self.api_url + $(this).attr("value"),
+                }).done(function(data) {
+                    self.addSelection(
+                        data['id'],
+                        self.addResultInfo(self.parameter_list, data)
+                    )
+                });
+            });
+            this.result = this.widget.find(this.settings.result_selector);
+            this.field = this.widget.find(this.settings.field_selector)
+                .on("keypress", function(event) {
+                    if (event.keyCode == 13) {
+                        event.preventDefault();
+                        self.searchOptions(true);
+                    }
+                }).on("keyup", function(event) {
+                    if (event.keyCode != 13) {
+                        clearTimeout(self.timeout);
+                        self.timeout = setTimeout(function() {
+                            self.searchOptions(true);
+                            self.field.focus();
+                        }, 500);
+                    }
+                });
+            this.search = this.widget.find(this.settings.search_selector)
+                .on("show.bs.dropdown", function(event) {
+                    self.searchOptions();
+                });
+            this.element.parents("form").on("submit", function(event) {
+                self.finish();
+            });
+            this.element.removeAttr('required');
+        },
+
+        searchOptions: function(show_dropdown) {
+            if (show_dropdown && this.result.is(":visible") === false) {
+                this.search.find("button").dropdown("toggle");
+            }
+            var selector = "option";
+            const query = this.field.val().trim();
+            const self = this;
+            if (query.length > 0) {
+                $.ajax({
+                    url: this.api_url,
+                        data: {
+                            "search": query,
+                        },
+                }).done(function(data) {
+                    const numOfEntries = data.count;
+                    self.result.empty();
+                    if (numOfEntries > 0) {
+                        var shownFields = (numOfEntries > 10) ? 5 : numOfEntries;
+                        for (let i = 0; i < shownFields; i++) {
+                            const result_info = data.results[i];
+                            self.result.append(
+                                $('<li>').append(
+                                    $('<a>').text(
+                                        self.addResultInfo(self.parameter_list, result_info)
+                                    )
+                                ).click(function() {
+                                    self.addSelection(
+                                        result_info['id'],
+                                        self.addResultInfo(self.parameter_list, result_info)
+                                    )
+                                })
+                            );
+                        }
+                        if (numOfEntries > shownFields){
+                            self.result.append(
+                                $('<li>').append(
+                                    $('<a>').text(numOfEntries - 5 + _(" results more"))
+                                )
+                            );
+                        }
+                    } else {
+                        self.result.append(
+                            $('<li>').append(
+                                $('<a>').text(_("No matches!"))
+                            )
+                        );
+                    }
+                });
+            }
+        },
+
+        addResultInfo: function(parameter_list, result_info) {
+            return parameter_list.map(
+                item => result_info[item]
+            ).join(', ');
+        },
+
+        addSelection: function(value, name) {
+            if (this.selection.find('[data-value="' + value + '"]').size() === 0) {
+                var li = this.selection_li.clone();
+                li.find(".name").text(name);
+                li.find("button").attr("data-value", value).on('click', function(event) {
+                    $(this).parent("li").remove();
+                });
+                this.selection.append(li);
+            }
+        },
+
+        resetSelection: function(values) {
+            this.selection.empty();
+            var self = this;
+            $.each(values, function(index, value) {
+                $.ajax({
+                    url: self.api_url + value,
+                }).done(function(data) {
+                    self.addSelection(
+                        value,
+                        self.addResultInfo(self.parameter_list, data)
+                    )
+                });
+            });
+        },
+
+        finish: function() {
+            this.widget.remove();
+            var select = this.element.show();
+            select.empty();
+            select.find("option:selected").prop("selected", false);
+            this.selection.find("button").each(function(index) {
+                jQuery('<option/>', {
+                    value: $(this).attr("data-value"),
+                    "selected": 'true',
+                }).appendTo(select);
+            });
+        }
+    });
+
+    $.fn[pluginName] = function(options, selectValues) {
+        return this.each(function() {
+            if (!$.data(this, "plugin_" + pluginName)) {
+                $.data(this, "plugin_" + pluginName, new AplusSearchSelectAjax(this, options));
+            }
+            if (selectValues) {
+                $.data(this, "plugin_" + pluginName).resetSelection(selectValues);
+            }
+        });
+    };
+})(jQuery);
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -21,7 +21,7 @@
 
 $(function() {
     "use strict";
- 
+
     // Mark active menu item
     $("[class^=menu-] a").each(function() {
         if ($(this)[0].pathname === location.pathname) {
--- a/edit_course/course_forms.py
+++ b/edit_course/course_forms.py
@@ -3,8 +3,10 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 from django_colortag.forms import ColorTagForm
 
+from aplus.api import api_reverse
 from course.models import LearningObjectCategory, CourseModule, CourseInstance, UserTag
 from lib.validators import generate_url_key_validator
+from lib.fields import UsersSearchSelectField
 from userprofile.models import UserProfile
 
 
@@ -68,6 +70,8 @@ class CourseModuleForm(FieldsetModelForm):
 
 class CourseInstanceForm(forms.ModelForm):
 
+    assistants = UsersSearchSelectField(queryset=UserProfile.objects.none())
+
     class Meta:
         model = CourseInstance
         fields = [
@@ -91,8 +95,8 @@ class CourseInstanceForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields["assistants"].widget.attrs["class"] = "search-select"
-        self.fields["assistants"].help_text = ""
+        self.fields['assistants'].queryset = self.instance.assistants.values_list('user', flat=True)
+        self.fields['assistants'].widget.attrs["data-search-api-url"] = api_reverse("user-list")
         if self.instance and self.instance.visible_to_students:
             self.fields["url"].widget.attrs["readonly"] = "true"
             self.fields["url"].help_text = _("The URL identifier is locked "
--- a/edit_course/course_forms.py
+++ b/edit_course/course_forms.py
@@ -177,13 +177,10 @@ class UserTagForm(ColorTagForm):
         return obj
 
 class SelectUsersForm(forms.Form):
-    user = forms.ModelMultipleChoiceField(queryset=UserProfile.objects.none())
+    user = UsersSearchSelectField(queryset=UserProfile.objects.none())
 
     def __init__(self, *args, **kwargs):
-        # This is copied from deviations/forms.py, which itself is not DRY.
-        # TODO: refactor this and the aforementioned form to avoid repetition
         course_instance = kwargs.pop('instance')
         super(SelectUsersForm, self).__init__(*args, **kwargs)
-        self.fields['user'].widget.attrs['class'] = 'search-select'
-        self.fields['user'].help_text = ''
-        self.fields['user'].queryset = course_instance.get_student_profiles()
+        self.fields['user'].widget.attrs["data-search-api-url"] = api_reverse(
+            "course-students-list", kwargs={'course_id': course_instance.id})
--- a/lib/fields.py
+++ b/lib/fields.py
@@ -1,10 +1,13 @@
 import json
 
 from django import forms
+from django.contrib.auth.models import User
 from django.core import exceptions
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
+from aplus.api import api_reverse
+from userprofile.models import UserProfile
 
 class PercentField(models.FloatField):
     """
@@ -17,6 +20,38 @@ class PercentField(models.FloatField):
                 _("The number must be between 0.0 and 1.0")
             )
         return value
+
+
+class SearchSelectField(forms.ModelMultipleChoiceField):
+    """
+    A generic field to 'search-select' widget.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.widget.attrs["class"] = "search-select-ajax"
+
+
+class UsersSearchSelectField(SearchSelectField):
+    """
+    A field to search and select multiple UserProfiles from API.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        #self.widget.attrs["data-users-api-url"] = api_reverse("user-list")
+        self.widget.attrs["data-key-parameter-list"] = "full_name,student_id"
+
+    def clean(self, value):
+        if not isinstance(value, list):
+            raise exceptions.ValidationError(
+                _("Invalid input type.")
+            )
+        for key in value:
+            if not User.objects.get(id=key):
+                raise exceptions.ValidationError(
+                    _("{} is not a valid key.").format(key)
+                )
+        users = User.objects.filter(id__in=value)
+        return UserProfile.objects.filter(user__in=users)
 
 
 class JSONField(models.TextField):
--- a/userprofile/api/views.py
+++ b/userprofile/api/views.py
@@ -1,4 +1,4 @@
-from rest_framework import permissions, viewsets
+from rest_framework import filters, permissions, viewsets
 from rest_framework.views import APIView
 from rest_framework.response import Response
 from rest_framework.settings import api_settings
@@ -7,7 +7,7 @@ from lib.api.mixins import ListSerializerMixin, MeUserMixin
 from lib.api.constants import REGEX_INT_ME
 
 from ..models import UserProfile
-from ..permissions import IsAdminOrUserObjIsSelf
+from ..permissions import IsAdminOrUserObjIsSelf, IsTeacherOrAdminOrSelf
 from .serializers import *
 from .full_serializers import *
 
@@ -16,11 +16,13 @@ class UserViewSet(ListSerializerMixin,
                   MeUserMixin,
                   viewsets.ReadOnlyModelViewSet):
     permission_classes = api_settings.DEFAULT_PERMISSION_CLASSES + [
-        IsAdminOrUserObjIsSelf,
+        IsTeacherOrAdminOrSelf,
     ]
     filter_backends = (
-        IsAdminOrUserObjIsSelf,
+        IsTeacherOrAdminOrSelf,
+        filters.SearchFilter,
     )
+    search_fields = ['user__first_name', 'user__last_name', 'student_id', 'user__email']
     lookup_field = 'user_id' # UserProfile.user.id
     lookup_url_kwarg = 'user_id'
     lookup_value_regex = REGEX_INT_ME
--- a/userprofile/permissions.py
+++ b/userprofile/permissions.py
@@ -1,6 +1,7 @@
 from authorization.permissions import SAFE_METHODS, Permission, FilterBackend
 
 from .models import UserProfile, GraderUser, LTIServiceUser
+from course.models import CourseInstance
 
 class IsAdminOrUserObjIsSelf(Permission, FilterBackend):
     def is_super(self, user):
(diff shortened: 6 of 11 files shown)


alalav1:
This PR is not complete, even thought it functionally works. There are two known issues yet:
	- Current assistants may have wrong id (UserProfile.id instead of User.id)
	- TThe hardcoded url in Javascript works only in the testing environment

alalav1:
> This PR is not complete, even thought it functionally works. There are two known issues yet to be solved:
> - Current assistants may have wrong id (UserProfile.id instead of User.id)
> - The hardcoded url in Javascript works only in the testing environment

Only the `user_id`s of the initial assistants is loaded to the HTML template. The other information of the assistants is read from the API, to avoid confusion between different IDs between `users` and `Userprofiles`. 
URL is now added to element as attribute and read in JS.

assets/js/aplus.js
```diff
@@ -463,9 +464,21 @@ $(function() {
...
             var self = this;
             this.selection = this.widget.find(this.settings.selection_selector);
+            this.api_url = self.element.attr("data-users-api-url")
             this.selection_li = this.selection.find("li").remove();
             this.element.find("option:selected").each(function(index) {
-                self.addSelection($(this).attr("value"), $(this).text());
+              $.ajax({
+                url: self.api_url,
+                data: {
+                  "search": $(this).attr("value"),
```
alalav1: This makes a GET request for each existing assistant in the course. It would be more efficient if they could be combined to single GET request that retrieves the user data for the given user IDs. 
If we keep this approach that makes GET requests for each user, then it must use the user detail API URL instead of making a useless seach in the API. (User details is something like `/api/v2/users/ID`).

assets/js/aplus.js
```diff
@@ -500,31 +513,43 @@ $(function() {
...
+                      var text = $('<a></a>')
+                      text.append(document.createTextNode(student_info['full_name'] + ", " + student_info['student_id']))
+                      li.append(text)
+                      self.result.append(li)
+                    }
+                }
+                else {
+                  var li = $('<li></li>')
+                  var text = $('<a></a>')
+                  text.append(document.createTextNode(_("No matches!")))
```
alalav1: Check if the "no matches" string translation works and if the string needs to be added to the teacher.fi.json.

assets/js/aplus.js
```diff
@@ -500,31 +513,43 @@ $(function() {
...
+                var numOfEntries = data.count;
+                self.result.empty();
+                if(numOfEntries > 0) {
+                  for (let i = 0; i < numOfEntries; i++) {
+                      const student_info = data.results[i];
+                      var li = $('<li></li>').click(function() {
+                        self.addSelection(student_info['id'], student_info['full_name'] + ", " + student_info['student_id']);
+                      });
+                      var text = $('<a></a>')
+                      text.append(document.createTextNode(student_info['full_name'] + ", " + student_info['student_id']))
```
alalav1: document.createTextNode can be replaced with the jquery text method.

assets/js/aplus.js
```diff
@@ -500,31 +513,43 @@ $(function() {
...
+                      li.append(text)
+                      self.result.append(li)
+                    }
+                }
+                else {
+                  var li = $('<li></li>')
```
alalav1: End lines with the semicolon `;`

assets/js/aplus.js
```diff
@@ -500,31 +513,43 @@ $(function() {
...
+                  text.append(document.createTextNode(_("No matches!")))
+                  li.append(text)
+                  self.result.append(li)
             }
+            });
+          }
         },
 
-        addSelection: function(value, name) {
+        addSelection: function(value, name, initial) {
```
alalav1: Remove the initial argument that is not used and was forgotten here.

edit_course/course_forms.py
```diff
@@ -1,13 +1,36 @@
...
 from django.utils.translation import ugettext_lazy as _
 from django_colortag.forms import ColorTagForm
 
+from aplus.api import api_reverse
 from course.models import LearningObjectCategory, CourseModule, CourseInstance, UserTag
 from lib.validators import generate_url_key_validator
 from userprofile.models import UserProfile
 
 
+class SelectManyField(forms.ModelMultipleChoiceField):
```
alalav1: Move the `SelectManyField` class to the file `lib/fields.py`. The class name must be changed since it is not descriptive enough now. Note that the clean method hardcodes references to the User model here. Can the model class be abstracted somehow or is it really necessary to hardcode it?

userprofile/api/views.py
```diff
@@ -12,6 +12,13 @@
 from .full_serializers import *
 
 
+class CustomSearchFilter(filters.SearchFilter):
```
alalav1: This filter is related to the initial_selection AJAX in aplus.js. This is useless in this format since single user details can be fetched from the API without any search. If we could fetch multiple given users, then this might be changed.

assets/js/aplus.js
alalav1: For now there is separate GET request for each assistant. It will probably take 50-200ms to load per assistant, so even in the larger courses the delay shouldn't be unbearable, since the 'Edit course details' page is not used very often.

Mankro:
@alalav1 Did you directly modify the old JavaScript code of the search-select widget (jQuery extension)? If yes, we should consider if the old widget is still useful somewhere and where it is used in the old code base. Basically, old components get broken if the JS is changed but nobody updates the backend Python code of the component. You should investigate where this JS widget is used in the old code base and list those features here. Those other components could be updated if it seems feasible. If the search-select widget has been used for other purposes besides selecting users, then take that into account. 

It would be sensible to separate git commits that add this new interface for doing AJAX search forms from the commits that change old components so that they use the new interface. This is particularly necessary if there are more old components than expected depending on the search-select JS.

The A+ Django admin interface has many dropdown menus in the edit model views that become too massive due to the amount data we have in the database. I don't think all those cases are tied to selecting users since Django admin creates those dropdown menus automatically for any foreign key or many-to-many relationship. This is why we would want to have a generic solution that can be connected to any form instead of hardcoding the new search-select backend with users.

One annoyance in the Django admin is that if I remember correctly, it does not include jQuery while our JS widget depends on jQuery. Perhaps we can include jQuery in the admin pages when needed.

Once all old components are fixed, I would guess that we don't need the old version of the search-select JS widget. It is better to query the server with AJAX than to include hundreds of entries in the HTML, and if there are very few options to choose from, then a normal select dropdown menu would be more appropriate than a textual search box.

assets/js/aplus.js
```diff
@@ -463,9 +463,14 @@ $(function() {
             this.element.hide();
             var self = this;
             this.selection = this.widget.find(this.settings.selection_selector);
+            this.api_url = self.element.attr("data-users-api-url");
             this.selection_li = this.selection.find("li").remove();
             this.element.find("option:selected").each(function(index) {
-                self.addSelection($(this).attr("value"), $(this).text());
+              $.ajax({
+                url: self.api_url+$(this).attr("value"),
+              }).done( function( data ) {
```
Mankro: We haven't used extra whitespace around this kind of anonymous functions, have we? Have you seen this elsewhere in A+ code? Without whitespace, the single line would be `}).done(function(data) {` instead of `}).done( function( data ) {`

assets/js/aplus.js
```diff
@@ -500,28 +505,40 @@ $(function() {
...
-                });
+              $.ajax({
+                url: this.api_url,
+                data: {
+                  "search": query,
+                },
+              }).done( function( data ) {
+                const numOfEntries = data.count;
+                self.result.empty();
+                if(numOfEntries > 0) {
```
Mankro: Add space between `if` and the condition: `if (condition)`

assets/js/aplus.js
```diff
@@ -500,28 +505,40 @@ $(function() {
...
+                        self.addSelection(student_info['id'], student_info['full_name'] + ", " + student_info['student_id']);
+                      });
+                      const text = $('<a></a>');
+                      text.text(student_info['full_name'] + ", " + student_info['student_id']);
+                      li.append(text);
+                      self.result.append(li);
+                    }
+                }
+                else {
+                  var li = $('<li></li>');
```
Mankro: You could check if jQuery allows you to create the li and a elements at once without separating them like this. In addition, the `var` could be `const` here. Const is better if there is no need to use var.

assets/js/aplus.js
```diff
@@ -500,28 +505,40 @@ $(function() {
...
+                data: {
+                  "search": query,
+                },
+              }).done( function( data ) {
+                const numOfEntries = data.count;
+                self.result.empty();
+                if(numOfEntries > 0) {
+                  for (let i = 0; i < numOfEntries; i++) {
+                      const student_info = data.results[i];
+                      var li = $('<li></li>').click(function() {
```
Mankro: `var li` should be `const li` if there is no need to use var.

assets/js/aplus.js
```diff
@@ -463,9 +463,14 @@ $(function() {
             this.element.hide();
             var self = this;
             this.selection = this.widget.find(this.settings.selection_selector);
+            this.api_url = self.element.attr("data-users-api-url");
             this.selection_li = this.selection.find("li").remove();
             this.element.find("option:selected").each(function(index) {
-                self.addSelection($(this).attr("value"), $(this).text());
+              $.ajax({
+                url: self.api_url+$(this).attr("value"),
```
Mankro: Please add whitespace around the plus operator.

edit_course/course_forms.py
```diff
@@ -1,10 +1,14 @@
 from django import forms
+from django.conf import settings
+from django.contrib.auth.models import User
```
Mankro: Are some of the new imports unused now?

edit_course/course_forms.py
```diff
@@ -91,8 +97,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields["assistants"].widget.attrs["class"] = "search-select"
-        self.fields["assistants"].help_text = ""
+        self.fields['assistants'].queryset = self.instance.assistants.values_list('user', flat=True)
+        self.fields['assistants'].widget.attrs["data-users-api-url"] = api_reverse("user-list")
```
Mankro: If the new field class hard codes the User model in it, then the `self.fields['assistants'].widget.attrs["data-users-api-url"] = api_reverse("user-list")` part could also be defined there inside the field class (`self.widget` instead of `self.fields['assistants']`).

lib/fields.py
```diff
@@ -19,6 +19,27 @@ def clean(self, value, model_instance):
         return value
 
 
+class SelectManyField(forms.ModelMultipleChoiceField):
```
Mankro: This field class name seems to be from the old version, so you somehow did not update it here.

assets/js/aplus.js
```diff
@@ -500,28 +505,40 @@ $(function() {
...
+                const numOfEntries = data.count;
+                self.result.empty();
+                if(numOfEntries > 0) {
+                  for (let i = 0; i < numOfEntries; i++) {
+                      const student_info = data.results[i];
+                      var li = $('<li></li>').click(function() {
+                        self.addSelection(student_info['id'], student_info['full_name'] + ", " + student_info['student_id']);
+                      });
+                      const text = $('<a></a>');
+                      text.text(student_info['full_name'] + ", " + student_info['student_id']);
```
alalav1: The new JS code works only with the User search, since the search results should have attributes like 'student_id' or 'full_name'. Otherwise, there should not be any problem to add the new AJAX search to other models as well. Currently the 'search-select' -widget is only used with models other than users in the `DeadlineRuleDeviationForm` with exercises and modules. The original efficiency issue was with downloading thousands of users in to the HTML template, so those fields don't necessarily need the AJAX search.

alalav1: The deviation forms and views are going to have somewhat major refactoring in the open PR #401.

alalav1:
TODO:
Hakutulosten määrää tulee rajoittaa, opettajille pitää antaa oikeudet tehdä haku Users sivulla.

Mankro:
If I understood correctly, the old `DeadlineRuleDeviationForm` uses the search-select JS for something else than users. Thus, we want to keep the old search-select JS code and the new AJAX version can named something like search-select-ajax. They can both exist in the source code. Actually, we don't want to put all JS widgets in the same aplus.js file, so the new AJAX widget can be moved to a new file under `a-plus/assets/js/`.

assets/js/aplus.js
Mankro: Let's change the search select AJAX JS so that it does not hardcode references to the User model. Basically, the attributes 'student_id' and 'full_name' can not be hardcoded in the JS since they refer to the User API output. Let's read those values from HTML data attributes so that only the server side needs to define which forms are for users and which forms are something else.

alalav1:
The new AJAX widget can now be used with other Models than Users, too. The old search-select JS code has been left unmodified so that search fields that do not require ajax (ie. deviation module and exercise search) work correctly.

assets/js/ajax_search_select.js
```diff
@@ -0,0 +1,168 @@
...
+                          self.addSelection(
+                            result_info['id'], self.parameter_list.map(
+                                item => result_info[item]
+                            ).join(', ')
+                          )
+                        })
+                      );
+                    }
+                }
+                else {
```
Mankro: `else` should be on the same line as the ending `}` of the `if` block. Please check the indentation here since it looks suspicious.

assets/js/ajax_search_select.js
```diff
@@ -0,0 +1,168 @@
...
+                });
+                this.selection.append(li);
+            }
+        },
+
+        resetSelection: function(values) {
+            this.selection.empty();
+            var self = this;
+            $.each(values, function(index, value) {
+                var opt = self.element.find('option[value="' + value + '"]');
```
Mankro: Does this `resetSelection` method work in the new AJAX version? It appears this method is not often called.

assets/js/ajax_search_select.js
```diff
@@ -0,0 +1,168 @@
...
+            var self = this;
+            this.selection = this.widget.find(this.settings.selection_selector);
+            this.api_url = self.element.attr("data-search-api-url");
+            this.parameter_list = self.element.attr("data-key-parameter-list").split(",");
+            this.selection_li = this.selection.find("li").remove();
+            this.element.find("option:selected").each(function(index) {
+              $.ajax({
+                url: self.api_url + $(this).attr("value"),
+              }).done(function(data) {
+                self.addSelection(data['id'], self.parameter_list.map(
```
Mankro: It looks like the same `parameter_list.map(...).join(', ')` invocation is done in three places, so could you move it to a new small helper function in order to avoid repeating the same code?

lib/fields.py
```diff
@@ -19,6 +22,38 @@ def clean(self, value, model_instance):
...
+    A generic field to 'search-select' widget.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.widget.attrs["class"] = "search-select-ajax"
+
+
+class UsersSearchSelectField(SearchSelectField):
+    """
+    A field to search and select multiple UsersProfiles from API.
```
Mankro: Isn't the model name `UserProfile`, not `UsersProfile`?

edit_course/course_forms.py
```diff
@@ -91,8 +95,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.fields["assistants"].widget.attrs["class"] = "search-select"
-        self.fields["assistants"].help_text = ""
+        self.fields['assistants'].queryset = self.instance.assistants.values_list('user', flat=True)
+        self.fields['assistants'].widget.attrs["data-search-api-url"] = api_reverse("user-list")
```
Mankro: I think the attr data-search-api-url can be defined in the UsersSearchSelectField constructor since it is the same for all user search forms. The queryset is kept here in the form class because it requires the course instance object.

assets/js/ajax_search_select.js
alalav1: It seems that the Atom editor automatically uses 2 spaces indentation in the JavaScript files, while the A-plus code uses 4 spaces.

assets/js/ajax_search_select.js
alalav1: I don't think that it works right now, but I have to admit that I don't know in which scenarios it is called and why.
It is probably useless in this new code.

assets/js/ajax_search_select.js
Mankro: Let's use the same indentation that the old code has used. That keeps it consistent. You may modify your Atom editor settings.

assets/js/ajax_search_select.js
Mankro: You can see in the bottom that the jQuery extension may call resetSelection if you supply it an extra parameter. I mean the `$.fn[...]` part in the bottom. It is called when you execute `$('.search-select-ajax').aplusSearchSelectAjax({}, true)`.
I think it is best to understand the purpose of resetSelection and then fix it for the new ajax version even if we are not really using it now. I assume this would be quite easy to do anyway.

alalav1:
The teachers need permission to the API for this feature to work. It will allow all teachers to access all students, student IDs and emails.

assets/js/ajax_search_select.js
alalav1: This method is called only when editing submitters of an exercise. The edit submitters page has quick select for groups, and selecting a group will call the `resetSelection` method to empty the currently selected students.

The search field uses the old version of `search-select` widget. I noticed that many search fields wants to use only the students enrolled to the course, such as this, and it wouldn't make any sense that the submitters field would search students from all registered accounts. Perhaps we could add an option to search from courses students only using URL `/api/v2/courses/1/students/`. Or we can let those fields to use HTML templates with all the couse students, as they currently do. Opinions on this @Mankro ?

alalav1: So in short: should we use ajax-search in fields that need to use only the enrolled students?

Mankro: >  Perhaps we could add an option to search from courses students only using URL /api/v2/courses/1/students/

Yes indeed, we want to support different API endpoints. If a form should only search enrolled students in the course, then it uses the API endpoint for that. The JS code does not need to be changed since it reads the API URL from the HTML data attribute. The server can set the right API URL to the data attribute for each form.

alalav1:
There are several search fields that use old `search-select` widget:
- Deadline deviation module, exercise and student fields
- Group editing, student field
- Adding tag to, student field
I think that the student fields should use this new version, and the deadline deviations could use the old for module and exercise.  (There are probably 10 modules and 100 exercises.)
I could apply this new Ajax search to those student fields in this PR.

alalav1:
There are several search fields that use old `search-select` widget:
- Deadline deviation module, exercise and student fields
- Group editing, student field
- Adding tag to, student field
I think that the student fields should use this new version, and the deadline deviations could use the old for module and exercise.  (There are probably 10 modules and 100 exercises.)
I could apply this new Ajax search to those student fields in this PR.

alalav1:
The 'IsTeacherOrAdminOrSelf' permission and filter class uses an inefficient and somewhat questionable way to retrieve the information if the user is a teacher, i.e. it checks if the user is teacher in any course, since the permission class doesn't know the current course instance.

userprofile/permissions.py
```diff
@@ -27,6 +28,13 @@ def filter_queryset(self, request, queryset, view):
         return queryset
 
 
+class IsTeacherOrAdminOrSelf(IsAdminOrUserObjIsSelf):
+    def is_super(self, user):
+        every_course = CourseInstance.objects.all()
+        is_teacher = any([course.is_teacher(user) for course in every_course])
```
Mankro: Some course permissions avoid looping over all courses in order to check the teachers by using custom attributes of the `view` variable in the `has_object_permission` method. For example, `view.is_teacher`:

* https://github.com/apluslms/a-plus/blob/6da3b36fbc583e40057c8119b25fc1a8997721f8/course/permissions.py#L150
* https://github.com/apluslms/a-plus/blob/6da3b36fbc583e40057c8119b25fc1a8997721f8/course/viewbase.py#L62 (this sets the `is_teacher` attribute to the view instance if the view class inherits `CourseInstanceBaseMixin`)
* https://github.com/apluslms/a-plus/blob/6da3b36fbc583e40057c8119b25fc1a8997721f8/course/api/views.py#L123 (example viewset in the API that inherits `CourseInstanceBaseMixin` via `CourseResourceMixin`)
* https://github.com/apluslms/a-plus/blob/6da3b36fbc583e40057c8119b25fc1a8997721f8/course/api/mixins.py#L13

There is still the problem that `UserViewSet` is not tied to any single course and of course you can't make it inherit `CourseInstanceBaseMixin`.

edit_course/course_forms.py
Mankro: Actually, the form should define the API URL like it is now because some forms use the all users list from the API and some only use the enrolled students of one course.

Mankro:
I reordered some commits so that the base code for the AJAX search is added first and then most forms are modified in other commits. The add assistants field of the edit course form was too tightly coupled with the base code commits and could not be separated. I squashed the commits that modified the base code.

I did not change any code. There are plenty of issues with this PR that need to be fixed soon. I merged this PR only so that other related PRs can be rebased onto the master branch now.

Mankro:
List of small code changes that should still be done (in a new PR):

* templates/base.html: script element for js/ajax_search_select.js should be moved to the templates that really need it. The JS script is not needed on every A+ page.
* lib/fields.py: SearchSelectField and UsersSearchSelectField: clarify and improve comments (used with the custom JS class; user ID versus profile ID conversion because API uses user IDs; html data attributes have to be set in the Python form class because JS depends on them)
* lib/fields.py UsersSearchSelectField: looping over users twice is inefficient. Can this be improved?
* assets/js/ajax_search_select.js line 40: explain in comments why old values are fetched from the API one-by-one (hack for user full names that are displayed in the form for previous existing values; user and profile ID mix)
* assets/js/ajax_search_select.js Add docstring (comment); explain the purpose of the JS class and its connection to Python forms and special Python form field classes; mandatory html data attributes; API endpoint requires the search filter
* assets/js/ajax_search_select.js row 92: should the user always see at least 10 results? Why would you sometimes show 9 results and then only 5 if there are very many results?
* assets/js/ajax_search_select.js method addResultInfo has a bad name since it does not add anything; it combines strings
* assets/js/ajax_search_select.js row 79: variable selector unused?
* course/forms.py GroupEditForm: clean up and improve comments, fix typos
* course/forms.py GroupEditForm: does members queryset mix user and profile IDs? Will this break when the IDs are not identical?
* exercise/forms.py EditSubmittersForm: same as above, members queryset may mix users and profiles
* course/api/views.py CourseStudentsViewSet: this does not allow teachers to use the API, right? Should it allow teachers when the JS uses this API in some forms?
* string translations are missing (lib/fields.py has some strings)

About testing:

* did you test that students can't access the changed APIs? Students must not have access to all users' info while teachers now need to have access. (`/api/v2/users`)
* testing API pagination when there are over 100 results. Is there a limit on the search filter or does it use the normal pagination setting?
* check that `/api/v2/users` includes all users while `/api/v2/courses/ID/students` includes only students enrolled in one course instance
* testing all the modified forms carefully (Adding new groups, editing groups, adding tags, edit submitters, add assistants)

More complex tasks:

* it is important to test that the user forms (e.g., add assistants under edit course) work correctly since the code needs to juggle between user IDs and profile IDs (because the API uses user IDs while many models have foreign keys to userprofiles). The test must use **a large number of users** in the database and **the profile IDs must not be equal to the user IDs**. That should reveal errors. If the IDs are identical in both user and profile tables, then many bugs could be hidden.
* It is inefficient to retrieve details for old values individually from the API after the page load. E.g., details for each existing assistant are fetched one-by-one with AJAX requests when the edit course form is loaded. We should try to include the existing users in the HTML of the form when it is loaded from the server. This is only problematic due to the user and profile ID mix. The queryset used in the Python code of the form can be initialized with the existing values. We should **try to implement a custom queryset class** that can be used in the form in place of the UserProfile queryset, but the new queryset would be able to use user IDs (not profile IDs) while it would still print the correct user full names and student IDs to the form HTML at page load.
* the permission `IsTeacherOrAdminOrSelf` is inefficient since it loops over all courses to check if the user is a teacher anywhere. We could require that the request URL includes a GET query parameter `course_instance` in order to limit the check to one course. If the user is not a teacher on that course or if the parameter is missing, permission is denied. Admins are allowed through without this parameter like before. The JavaScript using the API needs to supply the GET query parameters, which means it needs to find out the ID of the current course instance. HTML data attributes are one option for sharing the course instance ID to JavaScript.

Mankro:
> the permission IsTeacherOrAdminOrSelf is inefficient since it loops over all courses to check if the user is a teacher anywhere. We could require that the request URL includes a GET query parameter course_instance in order to limit the check to one course. If the user is not a teacher on that course or if the parameter is missing, permission is denied. Admins are allowed through without this parameter like before. The JavaScript using the API needs to supply the GET query parameters, which means it needs to find out the ID of the current course instance. HTML data attributes are one option for sharing the course instance ID to JavaScript.

Another solution would be to cache the user's teacher state in the UserProfile class. Basically, a `cached_property` `is_teacher` that can loop over all courses, but it would not be computed repeatedly because Django caches the result. This might be much cleaner than adding awkward extra GET query parameters to the users API and `IsTeacherOrAdminOrSelf`.

raphendyr:
To secure the API from cross course actions, I was planning that we should add JWT token to the html or headers. This token would include course id, so we can require that course action (e.g. adding user tags or adding submissions) are done only to the course user has open. In other words, ajax shouldn't be able to create a submission to any other course than the one open. If that would be implemented, then the permission could get the course from the token easily.

Similar information should be embeded to API tokens exposed to LTI services, IDEs or command line. Thus, when using token authentication, the course would come from the auth token. In addition, the API token should be a tuple of (user, course, service/title), so one could remove token per service.

alalav1:
For the list above, I have made all of the changes excluding: 
- moving AJAX code to templates that need it
- removed double loop in the lib/fields
- course/api/views.py CourseStudentsViewSet: teacher API allowing
I think that @Mankro will still have some better alternatives to the comments or docstrings.
In assets/js/ajax_search_select.js row 92 the different amount of showed users is on purpose. I think that showing 10 results all the time is useless and takes too much space in the page, so 5 is sufficient (if there are not all the results, there is no sense browsing them). But if there are less than 10 results, it makes sense to show them all, since then browsing through them is sensible. That is just my opinion however.
I have tested manually the listed cases, excluding pagination.
I have not attempted to do the complex tasks.

Mankro:
The new fixes from alalav1 are in a new PR: https://github.com/apluslms/a-plus/pull/515

Mankro:
The remaining issues that have been discussed here have been resolved except the performance of the `IsTeacherOrAdminOrSelf` permission. The issue #514 was opened for that.

Issue #546, opened 2020-06-04
Show the currently focused element distinctively


mtthwhggns:
When navigating A+ using the keyboard only, elements on the pages receive focus one by one. For example, the site name gets a narrow white border:

![Screenshot 2020-06-04 at 9 47 35](https://user-images.githubusercontent.com/6674046/83724137-6e809c80-a648-11ea-97f3-c971d34f64c4.png)

The course items on the index page get their elements highlighted like this:

![Screenshot 2020-06-04 at 9 48 20](https://user-images.githubusercontent.com/6674046/83724210-88ba7a80-a648-11ea-8fd5-16f73c47273e.png)

But some elements, like the select course menu, don't change their visible state at all:

![Screenshot 2020-06-04 at 9 54 59](https://user-images.githubusercontent.com/6674046/83724737-768d0c00-a649-11ea-8f3f-4225709dfdc9.png)

Criterion 2.4.7 of WCAG 2.1 (included in A level, the most basic) (https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html) requires that:

> Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible.

It's also in the draft of WCAG 2.2 at level of AA, along with requirements for it's colour contrast and size: https://www.w3.org/WAI/WCAG21/Understanding/focus-visible-enhanced

A consistent focus state can be implemented with CSS, as per this technique: https://www.w3.org/WAI/WCAG21/Techniques/css/C15

I'd aim for something that works similar to the excellent gov.uk example: https://design-system.service.gov.uk/get-started/focus-states/

This has little impact on users not using the keyboard for navigation, except the colour may be visible for a tiny fraction of a second while an element is clicked. 

The meta issue: #501

PR #547, opened 2020-06-05
Implement a consistent focus state

mtthwhggns:
# Description

This PR adds a focus state to links, buttons, form controls and focusable `div`s. The state is a yellow background or outline, which should be highly visible on both light and dark backgrounds.

Due to the quirks of CSS inheritance, and the fact that we are using CSS from multiple sources (CDN Bootstrap/Course-specific CSS/etc), I've gone with `!important` to override these. I can't foresee this being an issue elsewhere. 

In the future, maybe we could consider something more explicit, like preprocessing CSS with SCSS or similar, to allow for more fine-grained control of things like this and avoiding having the same "magic number" in a handful of different places.  

Here are a few examples:

Screenshot 2020-06-05 at 8 29 29

Screenshot 2020-06-05 at 8 29 39

Screenshot 2020-06-05 at 8 31 32

Remember that this is for keyboard navigation only, and these are not replacing the `:hover` state. That doesn't mean they're hidden to mouse users, but shouldn't get in the way. 

Fixes https://github.com/apluslms/a-plus/issues/546

# Testing

As far as I'm aware, we don't have a good way of using automated tests for this type of visual-only change.

I've tested it manually with the course template, and in non-course-specific interfaces.

There is a comment above the changes specifying their purpose. I wasn't able to find documentation relevant to this at the moment. Please advise if there is somewhere I should document this. It's especially relevant to those creating course content to know not to override it. 


raphendyr:
>  I've gone with !important to override these

HAve you considered if you could use more exact CSS maches? Is there reasonable CSS match that would drop the need for `!important`, while being some what acceptable? For example, you mention in the code that `a:focus` os mostly for the text. In that case, would something like `.content p a:focus` work without important?

I think yellow is really good color for this purpose. However, I have no idea how that woks from color theory point of view (i.e. compared to the blue). Would there be other possible colors we would be able to consider?

assets/css/main.css
```diff
@@ -58,6 +58,27 @@ table.src tr td.src {
...
+   components) */
+
+a:focus { /* Mostly for links in text */
+    background-color: #ffdd00 !important;
+    color: #333 !important;
+    outline: 3px solid rgba(0, 0, 0, 0);
+    text-decoration: underline;
+}
+
+a.btn:focus, button:focus, input.btn:focus {
```
raphendyr, line 71: How does this work when one clicks a button and the button opens a modal dialog for example. Does that show the yellow highligh too? Is that ok and is there anything we could do about that?

mtthwhggns: The highlight applies to the thing that has keybord focus. So, you focus on the button, then activate it. The modal appears, but initially nothing in the modal has keyboard focus. When you press tab, you jump to the next thing that should be reachable.

Ideally, this would be inside the model. We'd remove the page underneath from the tab order while it's unreachable. So maybe there are fields or links in the modal, or a close button. Tab should cycle the user around the visible, usable elements until they close the modal.

At present, I've not touched this and it doesn't work that way. Focus is kept by the button after the button is clicked. Pressing tab takes you further along the page. I recorded a 90-second example for you: https://aalto.cloud.panopto.eu/Panopto/Pages/Viewer.aspx?id=887cee86-b69b-4754-a897-abd100ca4658

The whole modal shouldn't get focus. 

In short, this PR only concerns how we show what currently has focus, not how we manage where the focus should be at a given moment.

raphendyr: Thanks. That video was a good example (basically, how focus breaks). You could create an issue about that, or you can add tha to the list of issues for now. For sure, thing to look into later!

I was more considering a case where a person clicks "submit" with a mouse and, if I remember right, the "focus" flag is left in the button until something else takes it. However, I might be wrong here. Nevertheless, based on the video, I think this seems to work as expected (i.e. the submit loses focus and doesn't stay yellow).

Furthermore, if this change happens to introduce unintended consequences, then we can address those later. Benefits are good.

Mankro:
This looks nice! `!important` is maybe a little concerning, though we could use that too. It is worth to try more exact CSS rules like Jaakko suggested.

mtthwhggns:
Hi @Mankro and raphendyr,

I'll try and answer your comments in turn.

### The use of `!important`

In answer to @raphendyr 's question, yes, I spent almost two hours attempting to implement this without `!important`. I share @Mankro 's concern, but after sinking some time into it, decided it was the least worst way to go. I'll try and explain why. 

My comment in the code was unclear, as my definition of "text" includes menus and links, everything that's not images or interactive elements.

The way it's written, the CSS rule covers:

- The links on the front page (which are `h3`s)
- The front page archive link (not wrapped in `content`
- The course archive items (which is an `h4` inside an `li`)
- The left menu on the course pages
- All links in the course points page (each `a` is inside a `td`.

There are probably a lot more, these are the ones I found in about 5 minutes.

We want the same, _consistent_ state to apply to every element that can be interacted with, that's the point of the guideline. If we specify each of these cases now, everyone adding or removing an interface element will have to update this part of the CSS.

Is it sensible to create so much future work?

If someone adds a link and doesn't add the corresponding CSS selector, the keyboard focus will look like it disappears, then reappears later having skipped an element. Leaving this more general, but depending on `!important` means that we don't have to worry about that happening. 

I raised the possibility of using SASS on Friday because I've used SASS mixins to create focus states before. Here's an example of how it can work: https://gist.github.com/nickcolley/791d084ff263786acb433ed0bd13c381  In the long term, I believe it would make most sense to move to something like this. 

In the meantime, my opinion is that `!important` is a good compromise. There's less risk, and less future work.

In summary, we have three options:

- **Switch to the hyper-specific rules** - slightly cleaner, but more risk and future work.
- **Switch to SASS, use a mixin** - the cleanest option, in my opinion, but a huge task. @Mankro has said, it's "very low priority"
- ** Stick with ** `!important`  - feels slightly hacky, lower risk of breaking the focus states, no extra future work. 

Now that I've explained my thinking, do you still think we should switch to the more specific rules?

### The colours

There's currently no universally accepted "correct" colour for focus state (the way users expect links tp be blue, for example). I went with this colour as it's high visibility, against light backgrounds, darker backgrounds and the blue we use. 

The most important thing is that we pick a colour we can avoid using elsewhere, so it is very clear which item is in focus. So, we currently use blue, red, green, orange in different places, so out of common groups of colours, yellow was one of a relatively narrow range of options. 

If you want to try another colour, that's OK with me. Also, if you want to try a slightly different visual style (many websites use a thicker box without a background), that would be fine too. My only concern is that it is consistent and recognisable.

raphendyr:
> Now that I've explained my thinking, do you still think we should switch to the more specific rules?

I think `!important` is the cleanest solution here. In addition, I think it suties well to solve cases like these. Thanks for the reaseach of different options.

> The most important thing is that we pick a colour we can avoid using elsewhere, so it is very clear which item is in focus. So, we currently use blue, red, green, orange in different places, so out of common groups of colours, yellow was one of a relatively narrow range of options.

Sounds reasonable. We could go with this, unless someone else has some unheard opinions.

For a different visual style, I think we can iterate that later if someone raises opinions about that. For now, I think what is in the PR is good. :+1:

etanttila:
Is it on purpose that some of the buttons and input fields don't turn yellow nor get a yellow border?
Screenshot from 2020-06-11 11-37-42
Screenshot from 2020-06-11 11-39-26
Screenshot from 2020-06-11 11-40-41

mtthwhggns:
Thanks for spotting that @etanttila.

I'd not used `!important` for non-links, and hadn't noticed that it was missing those. It should be fixed now.

assets/css/main.css
```diff
@@ -58,6 +58,27 @@ table.src tr td.src {
...
+    background-color: #ffdd00 !important;
+    color: #000 !important;
+    border: 2px solid #33 !important;
```
etanttila, line 74: Was this border color changed to `#33` (from `#333`) on purpose?

mtthwhggns: It was not! Thanks!

mtthwhggns:
Good points @etanttila ,

I did explain this more clearly in the issue, where I wrote: 

> This has little impact on users not using the keyboard for navigation, except the colour may be visible for a tiny fraction of a second while an element is clicked.

I didn't consider select boxes specifically, they're a good example, but in my opinion, this doesn't hinder mouse users from using anything. 

I'd suggest it's not so useful to consider a brand colour here. If we get a University-specific template in the future, we should probably use a focus state that doesn't look like part of the identity, so that it's easy to spot compared to the other elements on the page.

I consider focus to be a "functional" or "purposeful" colour. In the same way that we have red = error, orange = warning, green = success, yellow just means the focused element. This should be distinct from any branding. As I explained above, I chose yellow as we aren't using it anywhere else at the moment. That said, I don't mind which yellow we use. If you have a "nicer" yellow, I'm happy to experiment with it.

raphendyr:
> --, although the yellow color does show up when pressing things (especially when the focus remains on say, a dropdown). It might be advantagous to make sure that this too is communicated through in the description.

This was the concern I tried to raise earlier. The highligh might take the users focus and might be confusing or feel like it breaks the design (i.e. might feel like a bug), for users who don't know that there is feature to help navigate with the keyboard.

I'm not sure should we find a way to address this, or is that concern irrelevant. Any case, discussion if the impact to just plain mouse users is acceptable or not should be done. I would love someone who has better views on usability to comment more about that.

mtthwhggns:
> I'm not sure should we find a way to address this, or is that concern irrelevant. Any case, discussion if the impact to just plain mouse users is acceptable or not should be done. I would love someone who has better views on usability to comment more about that.

I don't profess to be an expert on this, but my opinion is:

- The "impact" is that a thin yellow outline is visible when using form fields where the browser would usually show a blue one (depending on the browser, some black).
- All elements have a focus state in plain HTML - this task is necessary because they're inconsistent and not always very clear. As I type this in GitHub, the text area has an extra blue outline that disappears if I click off it. 
- Many websites do very similar things, including some (like gov.uk) which place huge emphasis on usability testing with real users, so I'm fairly confident that it doesn't introduce any especially problematic issues for other users.
- Even if a few users are initially surprised by it, it's quite easy to get used to, apart from a little visible yellow, everything works as it did before. Our users typically take courses lasting many weeks, and most of them take multiple courses over a 3/5 year degree. I'm confident anyone who isn't used to seeing a visible focus state won't even notice it after a couple of visits.

etanttila:
>     * Many websites do very similar things, including some (like gov.uk) which place huge emphasis on usability testing with real users, so I'm fairly confident that it doesn't introduce any especially problematic issues for other users.

@mtthwhggns I noticed that it seems like gov.uk doesn't have any dropdowns on their website, which I assume could be due to several things:
(1) It works without JavaScript (or when JS is disabled)
(2) There are no dropdowns to maintain their focus and remain yellow when navigating with a mouse. Hopefully having the dropdowns remain yellow even for mouse users wouldn't be a problem, but if it is, there probably would be ways around that.
image

etanttila:
> I did explain this more clearly in the issue, where I wrote:
> 
> > This has little impact on users not using the keyboard for navigation, except the colour may be visible for a tiny fraction of a second while an element is clicked.
I forgot to read the issue, oops! Good that it's been acknowledged there.

> I'd suggest it's not so useful to consider a brand colour here. If we get a University-specific template in the future, we should probably use a focus state that doesn't look like part of the identity, so that it's easy to spot compared to the other elements on the page.
> 
> I consider focus to be a "functional" or "purposeful" colour. In the same way that we have red = error, orange = warning, green = success, yellow just means the focused element. This should be distinct from any branding.

That's a great point!

mtthwhggns:
Given the number of concerns, I'll investigate a couple of alternatives and see if anything could work better here. 

I'll update this later with what I find.

raphendyr:
> Given the number of concerns, I'll investigate a couple of alternatives and see if anything could work better here.

That is very nice, thank you. This is important, even though we wouldn't change the PR at all. Having options is good for discussion and for deciding a good path forward.

jaguarfi:
I am still not sure about the use of `!important`. I think there are ways to apply the `focus` class to the most relevant elements inside the page.  For example,

```css
.nav > li > a:focus {
background-color: #ffdd00;
color: #000;
border: 2px solid #333;
outline: 3px solid rgba(0, 0, 0, 0);
}
```
and
```css
.filter-categories > button:focus {
    background: yellow;
}
```
I know this may be time-consuming, but I think it is better to fix the issues in the right way now. 
In addition, many elements here use bootstrap, therefore there are some classes you could use in order to be more specific. I was also thinking, that in some cases you might use `tabindex` to address the problem.

**In conclusion**, I think there are some things we could try before going with the `!important` solution (Maybe this is the fastest, but not sure if it is the safest). 

**As a side note**: I was changing some styles in the Aplus theme before, and I learnt that using general selectors is a bad idea here. There are some [rst directives](https://github.com/apluslms/a-plus-rst-tools/) and other elements that expect certain behaviour from the CSS rules, and those should be taken into consideration

>We want the same, consistent state to apply to every element that can be interacted with, that's the point of the guideline. If we specify each of these cases now, everyone adding or removing an interface element will have to update this part of the CSS.

How adding new elements to the interface would affect the style? if we have specific rules for certain elements and classes all the new elements should inherit the CSS style.
It is also important to notice that many of the elements come from a Sphinx compilation, therefore most of the new elements will have the same structure. Of course, there are some cases were teachers add custom elements, but those cases are uncommon.
I also think that if in the future we decide to switch to sass or less, it might be good to define many of the individual rules now, so that the person implementing Sass of Less will have some specific styles to implement.  You could also use one of those tools to generate current changes.

jaguarfi:
I am still not sure about the use of the  `!important` rule here. I think there are ways to apply the `focus` class to the most relevant elements inside the page by using other CSS selectors.  For example,

```css
.nav > li > a:focus {
background-color: #ffdd00;
color: #000;
border: 2px solid #333;
outline: 3px solid rgba(0, 0, 0, 0);
}
```
and
```css
.filter-categories > button:focus {
    background: yellow;
}
```
I know this may be time-consuming, but I think it is better to fix the issues in the right way now, instead of fixing this in a rushed manner. In addition, many elements here use bootstrap, therefore there are some bootstrap classes you could use in order to be more specific. I was also thinking, that in some cases you might use `tabindex` to address the problem.

**In conclusion**, I think there are some things we could try before going with the `!important` solution (Maybe this is the fastest, but not sure if it is the safest). 

**As a side note**: I was changing some styles in the Aplus theme before, and I learnt that using general selectors is a bad idea here. There are some [rst directives](https://github.com/apluslms/a-plus-rst-tools/) and other elements that expect certain behaviour from the CSS rules, and those should be taken into consideration

>We want the same, consistent state to apply to every element that can be interacted with, that's the point of the guideline. If we specify each of these cases now, everyone adding or removing an interface element will have to update this part of the CSS.

How adding new elements to the interface would affect the state? if we have specific rules for certain elements and classes all the new elements should have the CSS style (Maybe, I did not understood the point here).
It is also important to notice that many of the elements come from a Sphinx compilation, therefore most of the UI elements will have the same structure including the same classes. Of course, there are some cases were teachers add custom elements, but those cases are uncommon.

>I raised the possibility of using SASS on Friday because I've used SASS mixins to create focus states before. Here's an example of how it can work: https://gist.github.com/nickcolley/791d084ff263786acb433ed0bd13c381 In the long term, I believe it would make most sense to move to something like this.

I also think that in case we decide to start using SASS or LESS. The person who start working on that will start by trying to replicate the current CSS, and if we have a bad CSS it might not help that person. it might be good to define many of the individual rules now so that the person implementing SASS or LESS will have some specific guidelines to follow.  

It is just an opinion, and you should follow @raphendyr and @Mankro suggestions because they know more about the whole system.

mtthwhggns:
I will return to @jaguarfi 's questions later, once I've had the chance to clarify them a little with him. 

In the meantime, I've been playing with [the focus-visible polyfill](https://github.com/WICG/focus-visible), and I think this is a way we can achieve what we want to. It would allow us to use `:focus-visible`instead of `:focus`, which is only true if the browser thinks the user is navigating using the keyboard (this is a bit of an oversimplification, but hopefully you get the idea). Unfortunately, [no browsers properly support it yet](https://caniuse.com/#feat=css-focus-visible), so a polyfill is the only option. 

This makes the task bigger, but worth it if it eliminates the type of surprised reaction @etanttila had on reviewing.

raphendyr:
@mtthwhggns Would it be too much to ask to get a video showing how the UI works differently with the current PR compared to the focus-visible solution?

mtthwhggns:
@raphendyr Here you go, this only swaps it on one group of elements: https://aalto.cloud.panopto.eu/Panopto/Pages/Viewer.aspx?id=6bfa772b-1ead-4dec-b2bf-abd800d3455d

raphendyr:
Thanks @mtthwhggns! That was very enlightening. I didn't read the polyfill code, but assuming it ok from performance point of view, I would prefer it. I think the result with `focus-visible` is exactly what we want, or what do you think?

mtthwhggns:
I just chatted with @jaguarfi. We talked about:

- There are some inherent risks to `!important`
- There are no perfect solutions without it either - using specific selectors risks people accidentally removing the focus state from focusable elements.
- There's a risk of confusion as to why this has been used, which is especially concerning in the future if we do migrate to SASS.

We agreed that the following sounds like an acceptable plan:

- I'll check the documentation, especially for Sphinx, to see which elements we can expect to be consistently marked up - these should work with a specific selector
- For others, we can move to `:focus-visible` with the polyfill, with `!important`. This isn't ideal, but it's targeted at such a small proportion of our users that I hope it's not a serious issue.
- To make sure there's no confusion from this, a clear comment can be added to the file stating why it's been used, so that anyone working on it in the future understands.
- I can test it by swapping the CSS/JS in production, to check it doesn't have an adverse effect on courses like O1 which make heavy use of custom CSS/JS.

I'll do these quite soon and it can be reviewed after that.

raphendyr:
> To make sure there's no confusion from this, a clear comment can be added to the file stating why it's been used, so that anyone working on it in the future understands.

You can add text `/* !important is needed to apply to all elements */` or such and add more explanation to the git commit. Thus, when someone asks "why", they can check the git history.

mtthwhggns:
I've made a new version based on the polyfill.

It's worth noting that they are visible on `input` elements for all types on focus. I don't think this is an issue, as the browser default focus states do that too - this just changes the colour. 

@etanttila would you like to review it again? You should probably start over, as everything has been replaced, although it should look and work mostly the same. After that, maybe the others who are interested can take a look?

etanttila:
The user experience seems absolutely wonderful! However @mtthwhggns do you have an unminimized version of the focus-visible-polyfill JavaScript file? It would probably be easier to read and review than the minimized version.

raphendyr:
Can you add all three: the original version, minified version and the mapping file, so the browser development tools can read the file, thanks.

mtthwhggns:
I've added these now - @etanttila feel free to take a look.

I'm not sure of the merits of thoroughly reviewing library code - if we had the tools, this would come directly from the NPM package as it's designed to, this is just a workaround where I tried to copy the technique that was used when `polyfill-find.min.js` was added.

etanttila:
> I've added these now - @etanttila feel free to take a look.
> 
> I'm not sure of the merits of thoroughly reviewing library code - if we had the tools, this would come directly from the NPM package as it's designed to, this is just a workaround where I tried to copy the technique that was used when `polyfill-find.min.js` was added.

Thanks! Originally I thought you might have written the code yourself (in which case reviewing the code would have been necessary), now however it's more to see how the library works in order to check its use.

etanttila:
Your changes seem fine to me! I don't know how the library, mapping and minimized file should be included, one of the instructors probably has a better idea and a vision.

raphendyr:
> I don't know how the library, mapping and minimized file should be included

Should be fine as they are now (3 files in the same path next to each other with the correct extensions).

Mankro:
I merged the pull request using Github's "squash and rebase" option just to see how it works. I also thought that it is better to squash those two commits into one.

Thanks for your great work @mtthwhggns and for the reviews @raphendyr, @etanttila, and @jaguarfi .

PR #548, opened 2020-06-05
LTI: enable branding for instance name Fixes #423

### LINKED ISSUE #423: LTI: enable branding for instance name
Enable branding name for  `tool_consumer_instance_name`.

HenrikSundell:
# Description

This PR changes how  `tool_consumer_instance_name` is set and adds two new parameters: 
 `tool_consumer_instance_description` and  `tool_consumer_instance_url`. Setting the the instance name using BRAND_NAME enables easier branding.

# Testing

Testing has been done manually with the help of https://github.com/Aalto-LeTech/django-lti-login
Not relevant.

### DIFF
--- a/aplus/local_settings.example.py
+++ b/aplus/local_settings.example.py
@@ -5,6 +5,9 @@
 #ADMINS = (
 #    ('Your Name', 'your_email@domain.com'),
 #)
+## Branding
+#BRAND_NAME = 'Your brand name (default is A+)'
+#BRAND_DESCRIPTION = 'Description of your service (default is Virtual Learning Environment)'
 # Show red alert on top of all pages
 #SITEWIDE_ALERT_TEXT = "Maintenance on Monday"
 #SITEWIDE_ALERT_TEXT = {'en': "Maintenance on Monday", 'fi': "Maanantaina on palvelukatko"}
--- a/aplus/settings.py
+++ b/aplus/settings.py
@@ -29,6 +29,7 @@ ALLOWED_HOSTS = ["*"]
 SITEWIDE_ALERT_TEXT = None
 SITEWIDE_ADVERT = None
 BRAND_NAME = 'A+'
+BRAND_DESCRIPTION = 'Virtual Learning Environment'
 
 WELCOME_TEXT = 'Welcome to A+ <small>modern learning environment</small>'
 SHIBBOLETH_TITLE_TEXT = 'Aalto University users'
--- a/external_services/lti.py
+++ b/external_services/lti.py
@@ -66,7 +66,9 @@ class LTIRequest(object):
             "launch_presentation_return_url": urljoin(settings.BASE_URL, instance.get_absolute_url()),
 
             "tool_consumer_instance_guid": base_url_parts.netloc + "/aplus",
-            "tool_consumer_instance_name": "A+ LMS",
+            "tool_consumer_instance_name": settings.BRAND_NAME,
+            "tool_consumer_instance_description": settings.BRAND_DESCRIPTION,
+            "tool_consumer_instance_url": settings.BASE_URL,
         })
 
         if service.api_access:


external_services/lti.py
```diff
@@ -63,7 +63,10 @@ def __init__(self, service, user, instance, request, title, context_id=None, lin
             "launch_presentation_return_url": request.scheme + '://' + request.get_host() + instance.get_absolute_url(),
 
             "tool_consumer_instance_guid": request.get_host() + "/aplus",
-            "tool_consumer_instance_name": "A+ LMS",
+            "tool_consumer_instance_name": settings.BRAND_NAME + " LMS",
```
raphendyr, line 66: I don't like adding `LMS` here. If `A+` would be too simple, then we should add another settings variable. Consider case where BRAND_NAME is `Plussa`, wihich would render `Plussa LMS` here, and that is not a thing.

external_services/lti.py
```diff
@@ -63,7 +63,10 @@ def __init__(self, service, user, instance, request, title, context_id=None, lin
...
+            "tool_consumer_instance_description": settings.WELCOME_TEXT,
```
raphendyr, line 67: I think we should add a new field to the settings for this. For consideration, we have added following data to HAKA.

```
DisplayName/en: plus.cs.aalto.fi, virtual learning environment
DisplayName/sv: plus.cs.aalto.fi, virtual provmiljö
DisplayName/fi: plus.cs.aalto.fi, virtuaalinen oppimisympäristö
Description/sv: Virtual lärmiljö vid Aalto-universitet institutionen för datateknik
Description/en: Virtual Learning Environment at Aalto University Department of Computer Science
Description/fi: Aalto-yliopiston Tietotekniikan laitoksen virtuaalinen oppimisympäristö
```

I think the later strings would be good here, but we don't have them in the settings.

raphendyr:
You can add `Fixes <issue url>` in the end of a commit message and/or you could add `Fixes #<issu number>` to end of the PR description.

external_services/lti.py
HenrikSundell: I didn't really like it either, I just wasn't sure if "A+" is enough. Is it ok if it's just "A+"?

external_services/lti.py
HenrikSundell: I assume you mean we should use these instead of the WELCOME_TEXT?

HenrikSundell:
> You can add `Fixes <issue url>` in the end of a commit message and/or you could add `Fixes #<issu number>` to end of the PR description.

Should Fixes <issue url> be added to every commit message?

external_services/lti.py
raphendyr: Yes, except those exact strings are specific to plus.cs.aalto.fi. `settings.py` should include some generic defaults and then the admin would write installation specific strings to local_settings. However, I'm not sure yet what we should do about translations. It's possible that IMS Global has not considered that.

HenrikSundell: Ok, but isn't "A+" specific to plus.cs.aalto.fi too? Should I set it to just something like "Virtual Learning Environment"?

external_services/lti.py
raphendyr: Let's go with just the BRAND_NAME. If that is not suitable, we can change it later.

raphendyr:
> Should Fixes <issue url> be added to every commit message?

Those that it applies to. Basically, you start with commits that implement the PR. During review you would create new commits in reaction to the discussion. After reviewers are happy, you would rebase and merge commits, so that we have again concrete logical commits and no "fixup" commits. Hence, before and after review, you shouldn't have commits which edit code created in previous commits in the same PR.

Thus, The specific commit which fixes a specific issue would include hte link. PR can sometimes include miltiple commits which would fix multiple issues. And there isn't strick rules how to split commits. E.g. this PR could be a single commit with a longer body explaining few changes. Or parts that fix specific issues have their own commits. In any case, the software should work after any commit, i.e., first commit can't depend on the second.

If you have not read the git/pr etiquette references in the wiki, I recommend that. They include lot of reasoning why that is a good practice (in short: we do this so that the git history helps use later).

external_services/lti.py
Mankro: Yes, "A+" is specific to plus.cs, but we can use it as a default value. The advantage of using Django settings is that they can be easily overridden on another site/installation.

Mankro:
Jaakko didn't mean that you would add "Fixes #XYZ" to the pull request **title**. The **description** is the longer text below the title.

aplus/settings.py
```diff
@@ -29,6 +29,7 @@
 SITEWIDE_ALERT_TEXT = None
 SITEWIDE_ADVERT = None
 BRAND_NAME = 'A+'
+BRAND_DESCRIPTION = 'Virtual Learning Environment at Aalto University Department of Computer Science'
```
Mankro, line 32: I think the default settings should avoid very Aalto-specific contents like this description that mentions "Aalto University". Should the default be just "Virtual learning environment"? We should also document the settings that should be overridden in the `local_settings.py` of a real production installation. Is it enough to add them to the `local_settings.example.py` file?

Mankro:
The body of the git **commit message** should have at most 80 characters per line.

aplus/settings.py
raphendyr: >  Is it enough to add them to the local_settings.example.py file?

I think it is, but I'm not sure. I personally read that file first, so if it's documented somewhere else, then we need to duplicate the information. Most important values are mentioned in the deployment document.

raphendyr:
> The body of the git **commit message** should have at most 80 characters per line.

To be exact, at most 72. A really good stackoverflow discussion about the topic: https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

aplus/local_settings.example.py
```diff
@@ -79,3 +79,7 @@
 #        'level': 'DEBUG',
 #    },
 #})
+
+## Branding
+#BRAND_NAME = 'Your brand name'
```
raphendyr, line 84: Could you move these immediately after `BASE_URL`. I think that could be a better location. If it feels too crowded, then some spacing could be added.

The default values (those written in settings.py) could be mentioned here. How you feel about `=  'Your brand name (default is A+)'` or such?

raphendyr: Hmm.. Maybe just after `ADMINS` would be a better location :thinking: . Any case, I think in the first block of variables. would be better than in the end of the file.

Plustest:
test comment

raphendyr:
test this please

PR #560, opened 2020-06-10
Increase usability of front page with instance names, longer course visibility and sectioning

### LINKED ISSUE #425: A+ front page
The A+ front page should display the name of the course instance in addition to the course name and code; Multiple instances of the same course are hard to separate in the A+ front page because the only different parts are the open/close dates and the image. Course instances have names, so that should be displayed in the front page too. Jaakko has some ideas; contact him!

saarasat:
# Description

Fixes #425 

1. Adds the name of the instance to the module-box of a course on Front page. To accommodate longer course and instance names bootstrap-thumbnails have been changed to cards. 

2. Divides the front page into two sections for students and staff:
   - In the upper section there are all the courses the student is currently enrolled in OR a staff member (assistant/teacher) is currently teaching. These will show 1 month after the set instance ending time.
   - In the bottom section there are all the currently ongoing courses regardless whether student 
has enrolled in them or not. These have the same cut-off point timewise as the ones above. The bottom section also includes the possible adverts.
- One course can only show on either of the sections once to avoid confusion.
- As these now use the same html-template-part a new template has been created for that
- A default message is shown for both sections, in case there are no course instances. 
3. Styling changes have been made to enhance both the appearance and accessibility of the front page:

- Removed alt-texts from decorative images
- Decreased the links of a course-card from three to one, by making most of the card clickable. This means removing the button "view course" as well. 
- Added hover-effect when the user places their mouse on a card
- Added the h2-tag between h1- and h3. 
- Increased the size of the archive-link for it to be more easily noticed
- Increased the size of a card slightly to make sure that overly long course or course instance names don't drop over the card. 
- Added language-attribute to course-titles

**Before**

image

**After**

image

# Testing

- The changes cannot be really tested with unit tests
- All the existing selenium- and unit-tests pass, a few locators had to be altered for the tests to pass
- Changes can be tested by adding course instances with different timeframes, and by logging in as a teacher, an assistant and a student and check that the front page displays the correct courses. Tests like this have been manually done. Eg. setting the ending date of a course instance 29 days in the past, or setting the date into the future and checking that those both show on the front page.
- Additionally longer course or course-instance names can be added

A+ README does not need to be updated here. In case these changes pass and the PR is merged, the teachers' course templates might need some additional documentation.


Mankro:
Hmm, your PR description does not use the phrase "fix #425" and thus GitHub did not automatically link the issue to this PR.

raphendyr:
> Hmm, your PR description does not use the phrase "fix #425" and thus GitHub did not automatically link the issue to this PR.

To continue on that, you could move the "Fixes" from the title to the end of the description block. I added that to the template, so future pull requests will point that out with a placeholder.

course/templates/course/index.html
```diff
@@ -12,85 +12,144 @@ <h1>{{ welcome_text|safe }}</h1>
...
-                        <img src="{{ advert.image }}" alt="">
-                    {% else %}
-                        <h1 class="glyphicon glyphicon-comment"></h1>
-                    {% endif %}
-                </div>
+            <div class="thumbnail course-thumbnail">
+                <a href="{{ instance_url }}">
+                    <div class="image" aria-hidden="true">
+                        {% if instance.image %}
+                        <img src="{{ instance.image.url }}" alt="{{ instance.course.name|parse_localization }}" />
```
raphendyr: I persoanlly prefer that block inside `if` is indented too. I need to look if django has any style guides for that...

course/views.py
```diff
@@ -33,18 +35,16 @@ def get_common_objects(self):
...
-        prio2 = []
-        start_treshold = timezone.now() - datetime.timedelta(days=10)
+        self.all_instances = []
+        self.my_instances = []
         end_threshold = timezone.now() - datetime.timedelta(days=30)
         for instance in CourseInstance.objects.get_visible(self.request.user)\
                 .filter(ending_time__gte=end_threshold):
-            if instance.starting_time > start_treshold:
-                self.instances += [instance]
+            if (instance.is_student(self.request.user) or instance.is_course_staff(self.request.user)):
```
raphendyr, line 41: This looks like this will generate an SQL query per instances, which would be _bad_.  You can show SQL queries in the log with the following settings snippet:

```
LOGGING['loggers'].update({
    'django.db.backends': {
        'level': 'DEBUG',
    },
})
```

I think you need to craft a query which would find the information with a single sql query.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -670,23 +670,15 @@ msgstr "Ei vielä ryhmiä"
...
-msgstr "All courses"
-
-#: course/templates/course/index.html:90
-msgid "Lue lisää"
-msgstr "Read more"
-
-#: course/templates/course/index.html:136
-msgid "View course"
-msgstr "Näytä kurssi"
+msgid "All courses"
```
raphendyr, line 674: Should this translation be part of previous commits?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3696,3 +3696,6 @@ msgstr "Unohda automaattinen uudelleenohjaus."
 #: userprofile/views.py:80
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
+#~ msgid "View course"
```
raphendyr, line 3700: Could you run compilemssages with argument `--no-obsolete`, because these don't need to be kept in the file (i.e. git stores old valuesi in the history).

assets/css/main.css
```diff
@@ -246,46 +246,71 @@ html, body {
...
 
-/* index thumbnails */
-.course-thumbnails .advert-thumbnail,
-.course-thumbnails .course-thumbnail {
-    height: 330px;
+
+/* index frontpage */
+
+.frontpage-panel {
+    margin: 3%;
```
mtthwhggns, line 253: This looks a bit strange to me. Is the extra space what you were aiming for?

Screenshot 2020-06-11 at 12 17 24

course/views.py
```diff
@@ -35,10 +35,11 @@ def get_common_objects(self):
...
         prio2 = []
-        treshold = timezone.now() - datetime.timedelta(days=10)
+        start_treshold = timezone.now() - datetime.timedelta(days=10)
+        end_threshold = timezone.now() - datetime.timedelta(days=30)
         for instance in CourseInstance.objects.get_visible(self.request.user)\
-                .filter(ending_time__gte=timezone.now()):
-            if instance.starting_time > treshold:
+                .filter(ending_time__gte=end_threshold):
+            if instance.starting_time > start_treshold:
                 self.instances += [instance]
```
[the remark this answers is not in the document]
raphendyr: This is a good place to show how slow python method resolution is too:

```
>>> timeit('for i in range(50): a.append(i)', 'a = []', number=1000000)
1.7041928516700864
>>> timeit('for i in range(50): _(i)', 'a = []; _ = a.append', number=1000000)
1.206302935257554
``

In this test, doing the attribute resolution adds 40% overhead, hence when same attribute is resolved _multi_ times, e.g., in a loop, caching that attribute resolution might be beneficial. Of course, this makes code often less readable, thus one should consider if the performance benefit is more or less imporant than the readability. I just think this is important fact to know about Python.

File: course/templates/course/index.html [hunk 7]
```diff
@@ -12,85 +12,148 @@ <h1>{{ welcome_text|safe }}</h1>
...
+                        {{ instance.course.code }}
+                        {% if not instance.visible_to_students %}
+                        <span class="label label-danger">{% trans "Hidden from students" %}</span>
+                        {% endif %}
+                        <br />
+                        {{ instance.starting_time|date:"SHORT_DATE_FORMAT" }} &ndash;
+                        {{ instance.ending_time|date:"SHORT_DATE_FORMAT" }}
+                        <br />
+                        {% if instance.enrollment_audience == 1 or instance.enrollment_audience == 3 %}
+                        <span class="label label-success">{{ internal_user_label|safe }}</span>
```
mtthwhggns, line 53: Do we benefit from showing a label on courses for "home institution" students? That's the default, I'd assume, and we could clean up the block a lot by skipping it. 

It's likely that this is also out of scope for this issue, but definitely worth considering in the future.

course/templates/course/index.html
```diff
@@ -12,85 +12,148 @@ <h1>{{ welcome_text|safe }}</h1>
...
+                        {{ instance.instance_name|parse_localization }}
+                    </h4>
+                    <p class="card-text">
+                        {{ instance.course.code }}
+                        {% if not instance.visible_to_students %}
+                        <span class="label label-danger">{% trans "Hidden from students" %}</span>
+                        {% endif %}
+                        <br />
+                        {{ instance.starting_time|date:"SHORT_DATE_FORMAT" }} &ndash;
+                        {{ instance.ending_time|date:"SHORT_DATE_FORMAT" }}
```
mtthwhggns, line 50: This PR may be too specific for this issue, but in my opinion, we should not be changing the date format by language, nor should we be using the American date format at all. 

Would you like to look at that, or create a new issue somewhere else?

course/templates/course/index.html
```diff
@@ -12,85 +12,148 @@ <h1>{{ welcome_text|safe }}</h1>
...
+        {% with instance_url=instance|url %}
         <div class="col-sm-6 col-md-4 col-lg-3">
-            <div class="thumbnail advert-thumbnail">
-                <div class="image" aria-hidden="true">
-                    {% if advert.image %}
-                        <img src="{{ advert.image }}" alt="">
-                    {% else %}
-                        <h1 class="glyphicon glyphicon-comment"></h1>
-                    {% endif %}
+            <div class="frontpage-course card">
```
mtthwhggns, line 22: As previously discussed, it would possibly make sense if the whole card was clickable - similar to, for example, the news items on the aalto.fi homepage. It's much nicer for touch navigation and possibly the expected behaviour given the panel size. 

How about making this an issue to revisit later, as it's not necessary to change now?

assets/css/main.css
```diff
@@ -246,46 +246,71 @@ html, body {
...
-.course-thumbnails .advert-thumbnail .image,
-.course-thumbnails .course-thumbnail .image {
-    height: 130px;
-    overflow: hidden;
-    text-align: center;
-}
-.course-thumbnails .course-thumbnail .image {
-    background: #eee;
+
+.frontpage .course-image {
```
mtthwhggns, line 276: Should the image have a hover state? Maybe something like 5% lighter, as it's clickable, but currently now shown at all.

course/templates/course/index.html
```diff
@@ -12,85 +12,148 @@ <h1>{{ welcome_text|safe }}</h1>
...
+                <a href="{{ instance_url }}">
```
mtthwhggns, line 23: Here we've got a link with no content. 

`aria-hidden` means assistive technology will completely ignore the element, so this `a` tag is actually empty. 

It might be worth just moving the next line's `aria-hidden="true"` up to here so screen readers skip the link completely.

course/templates/course/index.html
raphendyr: Good discussion. Ideally there could be courses that are only available to the external users or such. In addition, some courses might not be open for anyone from aalto, e.g., if they have manua enrollment done by the teacher. Thus, I would wish there would be filters that could be used to limit what is shown. These labels are currently kind of placeholders for that kind of "who can do the course" information. Actually desigining and implementing that would be cool.

course/templates/course/index.html
raphendyr: Please create issue about that. The django date formatting is used in multiple places, so I would prefer we fix all of them to be consistent.

course/templates/course/index.html
raphendyr: Hmm. Aalto.fi actually has nice hover effect there to indicate that is a link. That said, issue would help us to have the discussion in a better place.

course/templates/course/index.html
mtthwhggns: Created #572

course/views.py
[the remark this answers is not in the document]
saarasat: That's a good point and good to know. Actually I had not myself seen the `+= [instance]` used before, it was readily done in the previous code. I could change those to append, it is no problem.

course/templates/course/index.html
```diff
@@ -12,85 +12,144 @@ <h1>{{ welcome_text|safe }}</h1>
...
 {% endif %}
 
-<div class="row course-thumbnails">
-    {% site_advert as advert %}
-    {% if advert %}
+{% if my_instances %}
+<div class="frontpage-panel">
+    <h2 class="panel-heading">{% trans "My courses" %}</h2>
+    <div class="row course-thumbnails">
+        {% for instance in my_instances %}
```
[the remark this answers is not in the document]
saarasat: They are using the same template-part. I could change that.

course/templates/course/index.html
```diff
@@ -12,85 +12,144 @@ <h1>{{ welcome_text|safe }}</h1>
...
-            <div class="thumbnail advert-thumbnail">
-                <div class="image" aria-hidden="true">
-                    {% if advert.image %}
-                        <img src="{{ advert.image }}" alt="">
-                    {% else %}
-                        <h1 class="glyphicon glyphicon-comment"></h1>
-                    {% endif %}
-                </div>
+            <div class="thumbnail course-thumbnail">
+                <a href="{{ instance_url }}">
```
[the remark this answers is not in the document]
saarasat: Yes, will do! When the commits here are sorted out and code ready for mergning, I will add descriptive body-texts for the commits.

course/templates/course/index.html
```diff
@@ -12,85 +12,144 @@ <h1>{{ welcome_text|safe }}</h1>
...
-                    {% if instance.image %}
-                    <img src="{{ instance.image.url }}" alt="{{ instance.course.name|parse_localization }}" />
-                    {% else %}
-                    <h1 class="glyphicon glyphicon-book"></h1>
-                    {% endif %}
+<div class="frontpage-panel">
+    {% if all_instances or site_advert %}
+    <h2 class="frontpage panel-heading">{% trans "All courses" %}</h2>
+    <div class="row course-thumbnails">
+        {% site_advert as advert %}
```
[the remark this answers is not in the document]
saarasat: I thought about this. My idea was that since not all are logged in straightaway, this way they could get to see the advert as well. 

If the advert would be in the my courses -section, when a not logged-in person comes to the front page, they would see only an otherwise empty "my courses"-section with the advert, and then below that the section for all courses. 

I also thought about separating the advert as an own separate instance on the right-hand side, but that seemed like out of the scope of this issue. And something that could maybe be done when the filtering and search are implemented.

course/templates/course/index.html
saarasat: There were some mixed examples in the current source code. Also within the current index.html-file: 
With:
https://github.com/apluslms/a-plus/blob/master/course/templates/course/index.html#L21
Without:
https://github.com/apluslms/a-plus/blob/master/course/templates/course/index.html#L47

So my code was a bit mixed there as well. I couldn't find a source for the correct way for this in the django-documentation, but definitely we should have a common way of doing the indentation.

course/views.py
saarasat: Hmm.. This was done in the previous source-code like this way, but I can come up with another way for sure :) ! https://github.com/apluslms/a-plus/blob/master/course/views.py#L41

locale/fi/LC_MESSAGES/django.po
saarasat: Probably yes, I will correct that

locale/fi/LC_MESSAGES/django.po
saarasat: All right, will do

assets/css/main.css
saarasat: Yes it was, to create a sub-menu-kind-of-feeling :D But maybe we could ask someone to be a tiebreaker and go with that :)

course/templates/course/index.html
saarasat: Definitely, I think that was a good point Matthew made. I could continue with the filters and search once I get this issue wrapped up. I will create an issue on that, after this one closes, and I would like both of your input on what those filters, searches and badges should do so that I can implement them properly. 

From the point of this issue and PR, I think we can skip this now, and continue with this work in the next issue.

course/templates/course/index.html
saarasat: Great, that'll get fixed in a separate place then

course/templates/course/index.html
saarasat: I almost implemented the card clickability here to see how you both liked it. It was an easy fix, and I felt that it worked nice and smoothly. But in the end, the selenium-tests prohibited my attempts to make the whole card clickable, they kept breaking. And I thought that I can continue working on those, if we decide to make the card clickable. I agree with Matthew here, that the clickable card with a hover would be nice. 

I'll create an issue for this discussion and/or decision. 

With the changes in this PR, the button has been removed, the image and course name stay clickable. A hover-effect has been added to the course name. Are those good for now?

assets/css/main.css
saarasat: Yes, a good idea, I will add this for now :)

course/templates/course/index.html
saarasat: Oh, yes, you are correct, I will make this change!

course/templates/course/index.html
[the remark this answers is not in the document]
raphendyr: Yeah. I didn't consider it could be bit complicated to show the advert in both cases, which would require to add it to different div in each case. I think your solution is fine.

course/views.py
raphendyr: The `start_time` works as that is variable from the list of object returned by the query, but is_student creates [a second query.here](https://github.com/apluslms/a-plus/blob/master/course/models.py#L504). This should be visible with the logging :)

course/templates/course/index.html
raphendyr: As talked #elsewhere, current codebase has quite different styles in different files, thus we should slowly fix that. To be written in the style guid, html should be indented with a tab character and tags and elements both add to the indentation.

course/templates/course/index.html
raphendyr: > From the point of this issue and PR, I think we can skip this now, and continue with this work in the next issue.

Yes. I think that is good. Let's continue the discussion in a dedicated issue.

course/templates/course/index.html
raphendyr: > the button has been removed

I missed that. In that case I think the whole box should be the "button". Thus, I would prefer to keep the butten as is for now in this PR and address the course link side in another issue where we can decide how it should be done. For example, I think we should have similar hover effects as aalto.fi for the activatable box, thus it's not just moving the `a` element.

To conclude, if the button is removed, then the whole box should be the activator, but then ithe behavior must be clear for the user too.

mtthwhggns: I would suggest that this is not the correct call. I highlighted the buttons last week as they have identical link text while pointing to different pages. 

The course name is still formatted as a link, so I don't believe there is a risk of it not being identified as clickable. 

This pull request is a proposal that would fix all currently identified WCAG violations, and your instructions are to re-add one. 

I would strongly advocate either making the whole box clickable _at this point_, or accepting that the button has been removed.

course/views.py
```diff
@@ -35,10 +35,11 @@ def get_common_objects(self):
         self.external_user_label = settings_text('EXTERNAL_USER_LABEL')
         self.instances = []
         prio2 = []
-        treshold = timezone.now() - datetime.timedelta(days=10)
+        start_treshold = timezone.now() - datetime.timedelta(days=10)
```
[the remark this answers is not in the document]
saarasat: Good point. In the other commit which implements the sections that was actually reduced to using the timezone.now() to just one time, so this has been fixed. Probably should have done it already in that commit too.

course/views.py
saarasat: Oh you are correct, I wasn't sure which line you were refering to. But yes, will fix that :)

course/templates/course/index.html
```diff
@@ -12,85 +12,148 @@ <h1>{{ welcome_text|safe }}</h1>
...
+                        {% endif %}
+                    </div>
+                </a>
+                <div class="card-body">
+                    <h3 class="card-title" lang="{{ instance.language }}">
+                        <a href="{{ instance_url }}">
+                            {{ instance.course.name|parse_localization }}<br />
+                        </a>
+                    </h3>
+                    <h4 class="card-subtitle" lang="{{ instance.language }}">
```
mtthwhggns, line 40: Only just spotted this one. Have you considered that this may work better if not marked up as a heading. It looks good visually, but I don't think it's behaving like a subheading to the `h3` above it.

assets/css/main.css
```diff
@@ -246,46 +246,71 @@ html, body {
...
+.frontpage-panel .panel-heading {
+    border-bottom: solid 1px #eee;
+    font-weight: 500;
+}
+
+.frontpage .card {
+    margin-bottom: 20px;
+    border: solid 1px #ccc;
     position: relative;
+    height: 355px;
```
raphendyr, line 266: This probably should be rechecked, assuming the button is removed.

saarasat: I tried to implement a smaller height, but then it has the same problem as the current front page, thus if the instance or course name is very long it does not fit well the text flows over. Or when trying to level up the footers of cards, the grid gets a bit messy on some screen widths.  

I also tried substituting the end of a longer instance name with dots or just hiding it, but Finnish words are very long and cutting (all of) them in the correct spots for that to look good on mobile and desktop was not really possible. 

Therefore I ended up increasing the height to accommodate longer course- and instance names. Alternatively we could add a restraint for the length of the names when teachers are setting them up or risk some of them flowing a bit over.

course/templates/course/index.html
saarasat: In my mind it actually kind of makes sense, there are several instances of the same course. So a sort of "sub-heading". And it looked good style-wise :) But hmm.. I guess there is a point, I could just change it to <p>-tag and style it to look like it does now

course/templates/course/index.html
```diff
@@ -12,85 +12,148 @@ <h1>{{ welcome_text|safe }}</h1>
...
+                        <img class="card-img-top" src="{{ instance.image.url }}" alt="" />
+                        {% else %}
+                        <div class="card-img-top">
+                            <i class="glyphicon glyphicon-book"></i>
+                        </div>
+                        {% endif %}
+                    </div>
+                </a>
+                <div class="card-body">
+                    <h3 class="card-title" lang="{{ instance.language }}">
```
raphendyr, line 35: Btw. this will yield `|en|`, because the language is a list and is encoded as a string like `|en|fi|` and so on.

assets/css/main.css
raphendyr: One possibility is to allow some space and possibly use scroll for extra info. I don't know. Pretty hard to find a good solution for this... Would it be ok if different boxes would have different heights (I guess not)? Nevertheless, good arguments for the sice :+1:

course/templates/course/index.html
raphendyr: This topic was discussed elsewhere, and we decided to go with card sized buttons. Saara will work on it and update the PR.

assets/css/main.css
mtthwhggns: Has anyone tried this with `flex`? We could have a minimum height and all boxes grow if one needs it to be taller. 

It might mean a little re-writing styles if we lose the ready-made Bootstrap component.

course/views.py
raphendyr: Hei, checkout variable `request.user.userprofile.enrolled`. Similar thing is done for the drop down: https://github.com/apluslms/a-plus/blob/master/course/cache/menu.py#L37

course/views.py
```diff
@@ -35,10 +35,11 @@ def get_common_objects(self):
         self.external_user_label = settings_text('EXTERNAL_USER_LABEL')
```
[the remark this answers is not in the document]
saarasat: These have been added as links: https://github.com/apluslms/a-plus/pull/560/commits/d3a365277ea0073954bc4dcc388bc65d2d227a1e

assets/css/main.css
saarasat: With the footer added as we talked during the daily on Friday, these look good to me at least now. The footer balances it out. Most of the courses or course-instance names on A+ seem to be longish, so in any case there is not much extra room. And the grid works nicely now too. In any case, in my opinion it is better than how they currently run over such as:

image

course/views.py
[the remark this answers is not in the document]
saarasat: This has been fixed now, the end_threshold checks the timezone only once.

course/views.py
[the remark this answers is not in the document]
saarasat: I tried with list comprehension but it looked quite messy, like you said it might, so I went with append. In my opinion it looks now clean, the code is readable and it works well in this case.

course/templates/course/index.html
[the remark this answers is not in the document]
saarasat: This has been now fixed. I left the advert to be on the front page (thus index.html) as in my mind it doesn't belong with the course-instances. And it is more an item of the front page, so it can be easily changed from there if needed.

course/templates/course/index.html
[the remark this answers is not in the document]
saarasat: Now all (except the one which just fixes a few indendations) have commit bodies specifying all changes

course/templates/course/index.html
saarasat: These should be fixed. There might be still parts that use spaces instead of tabs (like talked #elsewhere) but at least they appear correct.

course/views.py
saarasat: Did not actually notice this comment, but found the same example myself :D So this has been fixed. For the teaching-part I had to be a little bit more innovative since the relation goes through the course-model. And now it does a few more hits to the database. But only a few more, the code is readable and this affects only teachers who have courses they have teached.

locale/fi/LC_MESSAGES/django.po
saarasat: Translations are in the part where they belong now

locale/fi/LC_MESSAGES/django.po
saarasat: There shouldn't be any more those extra values left in django.po

course/templates/course/index.html
saarasat: The cards are now clickable apart from the footer as discussed. Hover-effect has been added to text and image to indicate the clickability.

assets/css/main.css
saarasat: Fixed with clickable cards

course/templates/course/index.html
saarasat: This has been changed.

course/templates/course/index.html
saarasat: I checked this by logging it in the views.py-file and to me it yielded a string. Also when I look at the code from browser, it seems correct:

image


If this does not work for some reason, let me know :)

raphendyr: Let me look into this. I think it is a simple string for single language courses and encoded for multi lingual courses. Though, I'm not sure how did I get `|en|` there.. Nevertheless, you can apparently force the A+ to think it's a multilingual course. Just write `|en|fi|` in the language box in the coruse edit page (edit course > course | muokkaa kurssia > kurssi).

assets/css/main.css
raphendyr: > Has anyone tried this with flex? We could have a minimum height and all boxes grow if one needs it to be taller.

I think it would be a good solution. No one has been experienced enough to do that update with limited amount of time allocation. In the end, static sizes work most of the time. Assuming, it would be relatively simple to implement, then please poke me or markku.

course/templates/course/index.html
raphendyr: > but at least they appear correct.

People usually have tab width set to 2 or 4 spaces, thus it becomes pretty clear when it's rendered with 8 spaces by git and github. Hence, if it looks nice in the editor and here, then it should be fine.

course/views.py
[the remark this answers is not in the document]
raphendyr: >  the code is readable

This is the most important aspect. A code which is fast, but unmaintainable, would still cost more, because the maintenance is one of the biggest costs for a program.

assets/css/main.css
```diff
@@ -246,46 +246,93 @@ html, body {
...
-    text-align: center;
+
+.frontpage a:hover {
+    text-decoration: none;
 }
-.course-thumbnails .course-thumbnail .image {
-    background: #eee;
+
+.frontpage .card-clickable:hover {
+    filter: brightness(85%);
```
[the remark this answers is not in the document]
raphendyr: I'm thinking something like tje following (mouse hover on top of the clickable). Please note, this is _just_ a visualization of my mind, and it is _not_ the solution.

image

However, I'm not so happy with the sharp edge between clickable area and the footer.

assets/css/main.css
```diff
@@ -246,46 +246,93 @@ html, body {
...
 }
-.course-thumbnails .advert-thumbnail .image .glyphicon,
-.course-thumbnails .course-thumbnail .image .glyphicon {
+
+.frontpage .course-image i,
+.frontpage .advert-image i {
     font-size: 90px;
+    position: absolute;
+    left: 34%;
+    top: 7%;
```
raphendyr, line 304: I'm not sure why, but when I played around with the hover, I noticed that the closest anchor for `position: absolute` changes. The following patch made it behave a bit better:

```diff
diff --git a/assets/css/main.css b/assets/css/main.css
index ddc587b..bda1ee4 100644
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -293,6 +293,7 @@ html, body {
 }
 
 .frontpage .course-image {
+    position: relative;
     color: #fff;
 }
 
@@ -301,7 +302,7 @@ html, body {
     font-size: 90px;
     position: absolute;
     left: 34%;
-    top: 7%;
+    top: 15%;
 }
 
 .frontpage .card-body {
```

course/templates/course/course_cards.html
```diff
@@ -0,0 +1,50 @@
+{% load i18n %}
+{% load course %}
+
+{% for instance in instances %}
+{% with instance_url=instance|url %}
+	<div class="col-sm-6 col-md-4 col-lg-3">
+		<div class="frontpage-course card">
```
raphendyr, line 7: Apparently there is a little bit of mixed tabs and spaces. The file starts and ends with tabs. I assume you copypasted the middle part? You can check the tabs in the file with `cat -t` (or all interesting bytes with `cat -A` on Linux). Here is what it looks now (tabs are printed with string `^I`):

image

You can change indentation to one of them (I would prefer tab, but haven't asked what @Mankro prefers: tabs or spaces for html files?).

For example, to replace all  sequences of 4 spaces to tabs, you could execute command `sed -i 's/    /\t/g' course/templates/course/course_cards.html`.

course/templates/course/index.html
```diff
@@ -12,81 +12,48 @@ <h1>{{ welcome_text|safe }}</h1>
...
-                            {% trans "Read more" %}
-                        </a>
-                    </div>
-                </div>
-            </div>
+{% if my_instances %}
+    <div class="frontpage-panel">
+        <h2 class="panel-heading">{% trans "My courses"%}</h2>
+        <div class="row frontpage">
+            {% include "course/course_cards.html" with instances=my_instances %}
```
raphendyr, line 19: Some parts of the A+ use template naming similar to how Ruby on Rails handles partials. I.e. this included file would be named `course/_course_cards.html` to indicate that it's supposed to be included from another template.

:thinking:  Something for me to include in the stuleguide..

course/templates/course/course_cards.html
```diff
@@ -0,0 +1,50 @@
...
+                        </h3>
+                        <p class="card-subtitle" lang="{{ instance.language }}">
+                            {{ instance.instance_name|parse_localization }}
+                        </p>
+                        <p class="card-text">
+                            {{ instance.course.code }}
+                            {% if not instance.visible_to_students %}
+                                <span class="label label-danger">{% trans "Hidden from students" %}</span>
+                            {% endif %}
+                            <br />
```
raphendyr, line 31: As you are moving these to a new file, you can change these to `<br>` (i.e. drop `/` from it). Does not matter though.

course/templates/course/course_cards.html
```diff
@@ -0,0 +1,50 @@
...
+                            {{ instance.starting_time|date:"SHORT_DATE_FORMAT" }} &ndash;
+                            {{ instance.ending_time|date:"SHORT_DATE_FORMAT" }}
+                            <br />
```
raphendyr, line 34: The `<br>` at the end should be implemented with paddings and margins. [Some background in stackoverflow](https://stackoverflow.com/a/12788551). However, this element doesn't appear to effect the rendering,

assets/css/main.css
mtthwhggns: Here's an example I wrote in ~5 minutes; https://codepen.io/mtthwhggns/pen/ExPgoxq

raphendyr: Seems a really good solution. The flex idea _can_ be implemented within the PR or can be left to a second PR.

saarasat: A flex-box would probably be a good solution :) I prefer leaving it to another pull request, as:
- The code now works and looks good and is in my opinion an improvement to the earlier. 
- The appearance changes from the point of a user would be minimal, even though that would be implemented later on, the grid-changes happen in the background, so it wouldn't confuse the users.
- The frontpage inherits the content block from a base-template, so it might need a little rework there and might not be a five-minute-fix. Just a hunch based on working with the grid of the front page :D

course/templates/course/index.html
saarasat: I made an issue for this:
https://github.com/apluslms/a-plus/issues/580

And few other related issues: 
https://github.com/apluslms/a-plus/issues/579
https://github.com/apluslms/a-plus/issues/578

course/templates/course/course_cards.html
saarasat: Yes, I looked that those as well, but since they were in the code already, I didn't see the point of removing them at this point.

course/templates/course/course_cards.html
saarasat: Alright, good to know for future :)

course/templates/course/index.html
saarasat: Oh, I didn't know that. But can definitely change it!

assets/css/main.css
saarasat: Oh, that's strange. I did not notice that, I will make the patch

assets/css/main.css
raphendyr: Can you create an issue about that too? You can assign it to you, link it here and you can then resolve this conversation, thanks.

course/templates/course/index.html
raphendyr: Yeah. I don't think it's so common with Django, but it does help when listing templates in the filesystem and to not make mistakes when using the template filenames in the python code (as underscores shouldn't appear there).

course/views.py
```diff
@@ -33,17 +33,33 @@ def get_common_objects(self):
...
+                    self.my_instances.append(instance)
+            
+            for instance in user.userprofile.enrolled.all().filter(ending_time__gte=end_threshold):
+                if instance not in self.my_instances:
+                    self.my_instances.append(instance)
+        
+        all_open_instances = CourseInstance.objects.get_visible(user)\
+            .filter(ending_time__gte=end_threshold)
+        
+        for instance in all_open_instances:
```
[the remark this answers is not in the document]
saarasat: Alright, I was looking at the previous code on this and how it appended the new instances and fed the list to the template, since that clearly seemed to work at the very least. But setting the instance variables is something I can change now.

course/views.py
```diff
@@ -33,17 +33,33 @@ def get_common_objects(self):
...
-                prio2 += [instance]
-        self.instances += prio2
-        self.note("welcome_text", "internal_user_label", "external_user_label", "instances")
+        self.my_instances = []
+        self.all_instances = []
+        end_threshold = timezone.now() - datetime.timedelta(days=30)
+        user = self.request.user
+
+        if user and user.is_authenticated:
+            for course in user.userprofile.teaching_courses.all():
```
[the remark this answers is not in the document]
saarasat: Yeah, the teacher-part was trickier because it went through the Course so I couldn't find any other way to do it as simply as with the other two categories through the reverse relation. So in the end ended up doing it the way I found it here with the CachedTopMenu: https://github.com/apluslms/a-plus/blob/master/course/cache/menu.py. I am not sure is it a problem there as well? Maybe the caching is a discussion for another day :)  

And in any case, now that I see there is a readymade related field on the CourseInstance-level as well, that seems like a really good fix. Will do

assets/css/main.css
```diff
@@ -246,46 +246,93 @@ html, body {
...
 }
-.course-thumbnails .course-thumbnail .image {
-    background: #eee;
+
+.frontpage .card-clickable:hover {
+    filter: brightness(85%);
+}
+
+.frontpage .advert-card:hover h3, p,
+.frontpage .card-clickable:hover h3, p {
```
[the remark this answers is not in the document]
saarasat: It should, did not notice that! Thanks!

course/views.py
```diff
@@ -33,17 +33,33 @@ def get_common_objects(self):
...
+                    self.my_instances.append(instance)
+        
+        all_open_instances = CourseInstance.objects.get_visible(user)\
+            .filter(ending_time__gte=end_threshold)
+        
+        for instance in all_open_instances:
+            if instance not in self.my_instances:
+                self.all_instances.append(instance)
+        
+        self.note("welcome_text", "internal_user_label", "external_user_label", \
```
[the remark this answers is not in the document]
saarasat: That's a good addition to the style-guide

assets/css/main.css
saarasat: Sure!

assets/css/main.css
```diff
@@ -246,46 +246,93 @@ html, body {
...
+}
+
+.frontpage .course-image {
+    color: #fff;
 }
-.course-thumbnails .advert-thumbnail .image .glyphicon,
-.course-thumbnails .course-thumbnail .image .glyphicon {
+
+.frontpage .course-image i,
+.frontpage .advert-image i {
```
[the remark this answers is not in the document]
saarasat: There is the benefit that this way the glyphicon in advert is blue, and in the course-image is white. The advert then pops out a bit more from the frontpage and doesn't get mixed with them so easily. This was also done previously, with a little bit different css-classes but the base-idea was the same. I thought it was worth saving :)

assets/css/main.css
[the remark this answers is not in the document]
saarasat: I also thought about that (and tried this option but did not like that sharp edge either). To my mind the hover on top of the title and image seems enough. I also used a bunch of news- and blog-sites as a reference, and hover on the text (darker or lighter) seemed to be a common way to indicate that a card-element is clickable. Most of them did not even have a hover effect on the image, just the headline.

course/views.py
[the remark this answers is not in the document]
raphendyr: > I am not sure is it a problem there as well? Maybe the caching is a discussion for another day :)

Yes, it should be done better. However, it isn't so worrisome as the cache entry is updated only very rarely. On the other hand, the front page is updated in every reload. Technically, we could use the cache to populate the front page too, but it's likely that the drop menu and the front page will have a bit different set of information or the details would require database query anycase.

assets/css/main.css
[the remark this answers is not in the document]
raphendyr: > There is the benefit that this way the glyphicon in advert is blue, and in the course-image is white. 

Of course! Forgot about that. I think you cleaned it up nicely :)

course/templates/course/index.html
```diff
@@ -8,56 +8,56 @@
...
 {% endif %}
 
 
 <p class="archive-link">
-    {% url 'archive' as url %}
-    {% blocktrans with url=url %}
-    Old courses can be found at the <a href="{{ url }}">course archive</a>.
-    {% endblocktrans %}
+	{% url 'archive' as url %}
+	{% blocktrans with url=url %}
```
raphendyr, line 58: I'm pretty sure this `blocktrans` was written originally before Django 1.7, as are many others too. Since then, there has been posibility to add `trimmed` argument, which makes the translation step a bit nicer, as it removes extra whitespace. It's pretty neat, I recommend to test it out :) 

```
{% blocktrans trimmed with url=url %}
```

course/views.py
```diff
@@ -37,29 +37,33 @@ def get_common_objects(self):
...
         end_threshold = timezone.now() - datetime.timedelta(days=30)
         user = self.request.user
+        my_instances = []
 
         if user and user.is_authenticated:
-            for course in user.userprofile.teaching_courses.all():
-                for instance in course.instances.all().filter(ending_time__gte=end_threshold):
-                    self.my_instances.append(instance)
+            for instance in CourseInstance.objects.filter(course__teachers=user.userprofile)\
+                .filter(ending_time__gte=end_threshold):
```
raphendyr, line 44: Typically, one should avout `\` when possible. One good place is `with` statement, where there is no better alternatives. Bot for here, you could consider formatting this like (key idea is to use brackets, mutliple ideas how to format it though):

```python
for instance in (CourseInstance.objects
                 .filter(course__teachers=user.userprofile,
                         ending_time__gte=end_threshold)
                 .all()):
    my_instances.append(instance)
```

or

```python
for instance in (CourseInstance.objects
                 .filter(
                     course__teachers=user.userprofile,
                     ending_time__gte=end_threshold,
                 )
                 .all()):
    my_instances.append(instance)
```

or

```python
my_instances.extend(
    CourseInstance.objects
    .filter(
        course__teachers=user.userprofile,
        ending_time__gte=end_threshold,
    )
    .all()
)
```

I guess the `.all()` could be left from the latter two, as it's not important to ensure that the Django won't use a cached queryset. In the first, it helps readability.

course/views.py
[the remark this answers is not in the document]
saarasat: This has now been fixed in yesterday's commits.

assets/css/main.css
saarasat: Here is the issue for this: https://github.com/apluslms/a-plus/issues/583

assets/css/main.css
```diff
@@ -246,46 +246,93 @@ html, body {
     list-style: none;
 }
 
-/* index thumbnails */
-.course-thumbnails .advert-thumbnail,
-.course-thumbnails .course-thumbnail {
-    height: 330px;
+/* index frontpage */
```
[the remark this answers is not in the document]
saarasat: This has now been changed in yesterda's commits. A very good point, again :)

assets/css/main.css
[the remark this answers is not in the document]
saarasat: Changed in yesterday's commits

assets/css/main.css
saarasat: This patch has been made now

course/templates/course/index.html
saarasat: Oh awesome, I have been wondering whether there is something like this. That part of the block-translation keeps bugging me every time. Will use this from now on!

course/templates/course/course_cards.html
saarasat: This was fixed now

saarasat: This was fixed yesterday as well

course/templates/course/index.html
saarasat: As talked elsewhere, we will leave this as it is now solved, since it is an improvement to earlier as it fixes the cases of single-language courses. This multilanguage-issue can be solved system-wide later on, to ensure that the solution is more re-usable and sustainable.

course/views.py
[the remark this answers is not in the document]
saarasat: Yes, from what I gather they require a bit different set of information. But in any case, good to know that about the cache and how important this issue there.

course/views.py
saarasat: This was also now fixed for the query with multiple filters. So all yesterday's review comments should have now been taken into consideration.

saarasat:
So to conclude here what was discussed, this has gotten a go-ahead from Matthew, and the final tweaks by Jaakko has been implemented. So this one is ready for a final checkup (by @Mankro or @jaguarfi ?) and then rebasing and merging.

saarasat:
> * Please test the cards in Chrome. It seems something is not working properly.

Hmm... I wonder what might be going wrong there. For me at least it works correctly in Chrome and Firefox, locally and in a container:

**Firefox**
image

**Chrome** 
image

I am not able to replicate this issue. I wonder if this is the same problem that there was when I tried Ella's fork initially. Do you get any error-messages in the terminal?

saarasat:
Those default messages might be a good idea! I think those could be worked in a separate issue, if everyone agrees they are worth doing. It is easier to work on and assess those once this has been implemented.

course/templates/course/index.html
```diff
@@ -8,89 +8,56 @@
...
-    {% endif %}
+{% if my_instances %}
+	<div class="frontpage-panel">
+		<h2 class="panel-heading">{% trans "My courses"%}</h2>
+		<div class="row frontpage">
+			{% include "course/_course_cards.html" with instances=my_instances %}
+		</div>
+	</div>
+{% endif %}
+
```
jaguarfi, line 23: I think you can add a hard-coded default message here or somewhere around here. There will be no need to create a new issue on this.

raphendyr:
> Should we keep the section "My courses" with a default message saying something like "You are not enrolled in any course". This might looks trivial, but I think it is important to tell students what is their current status. Also, a default message for the ongoing courses might be beneficial.

As long as it handles logged in and anonymous users nicely and isn't confusing if/when we hide old enrollments.

course/templates/course/_course_cards.html
```diff
@@ -0,0 +1,50 @@
...
+		<div class="card-footer">
+			{% if instance.enrollment_audience == 1 or instance.enrollment_audience == 3 %}
+				<span class="label label-success">{{ internal_user_label|safe }}</span>
+			{% endif %}
+			{% if instance.enrollment_audience == 2 or instance.enrollment_audience == 3 %}
+				<span class="label label-info">{{ external_user_label|safe }}</span>
+			{% endif %}
+		</div>
+	</div>
+</div>
```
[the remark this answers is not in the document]
saarasat: You are correct! I did not even notice that bs3 does not have cards. Those implementations worked better with the longer instance-names and seemed more flexible. Maybe because as you said, I seem to have actually customised them. In any case, they should follow the specs of bs4-cards pretty well. And bs4 does not seem to have the bs3-thumbnails used before, so with the migration there might actually be benefits in that sense. Probably there would have been collissions eitherway. With thumbnails or these customised cards.

course/templates/course/_course_cards.html
```diff
@@ -0,0 +1,50 @@
...
+				</div>
+			</a>
+		</div>
+		<div class="card-footer">
+			{% if instance.enrollment_audience == 1 or instance.enrollment_audience == 3 %}
+				<span class="label label-success">{{ internal_user_label|safe }}</span>
+			{% endif %}
+			{% if instance.enrollment_audience == 2 or instance.enrollment_audience == 3 %}
+				<span class="label label-info">{{ external_user_label|safe }}</span>
+			{% endif %}
```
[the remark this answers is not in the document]
saarasat: Mm.. That part I actually haven't touched. But those are the Aalto- and Mooc-badges. So they both might be there at once or neither or just one. So in that sense they are not mutually-exclusive, which `elif` would mean. If I got your point correctly :)

course/templates/course/index.html
saarasat: Alrighty, I will add those.

course/templates/course/_course_cards.html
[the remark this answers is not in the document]
jaguarfi: I was thinking of something like this 
image

In option A, the code avoids some useless validations. Again, this was just a thought, and maybe there are a hundred of approaches for this single line of code. Therefore, if the implementation already existed, you should keep it.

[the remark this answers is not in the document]
raphendyr: However, in option B, the label elements are duplicated. In theory, a bit wise operation would suite this well ;) , but I don't think they are supported in Django templates. In addition, they don't describe the case well. No need to spend time on this, as this should be generalized to support multiple different tags. (`if instance.enrollment_audience & 0x1:` and `if instance.enrollment_audience & 0x2:`).

course/templates/course/index.html
saarasat: These have now been added.

jaguarfi:
It looks good for me :+1:

course/views.py
```diff
@@ -24,6 +24,7 @@
 from .viewbase import CourseModuleBaseView, CourseInstanceMixin, EnrollableViewMixin
 
 
+
```
raphendyr, line 27: Is this new line added with a reaon or by accident?

course/views.py
```diff
@@ -59,12 +61,14 @@ def get_common_objects(self):
         
         self.all_instances = all_instances
         self.my_instances = my_instances
+        self.is_internal = is_internal
```
raphendyr, line 64: I'm not sure, but this could be seen as `INTERNAL` vs. `EXTERNAL` user thing and not as `internally stored and logged in user`. Could the variable name be something else?

saarasat: Hmm.. I tried to keep it short and describe that the user is internal (thus can have my courses). But what would you suggest instead?

course/views.py
saarasat: Probably by accident, when I import the logging and was logging stuff there quickly to see that everything works

course/views.py
saarasat: I am gonna change it to "is_logged_in". That describes that the user exists and is authenticated, thus can have my_courses

saarasat:
I changed the variable name for default messages and the extra line was removed. Everything should be squashed, fixed, checked, tested and ready to go.

Mankro:
@saarasat unfortunately, the django.po translation file now causes conflicts after I merged another PR.

@jaguarfi did you want to review this since Github still says you haven't done it? Or did you just forget to submit it in Github?

saarasat:
> @saarasat unfortunately, the django.po translation file now causes conflicts after I merged another PR.

Great timing :D Is it okay if I update the translations in a separate commit now that would be merged?

saarasat:
> @saarasat unfortunately, the django.po translation file now causes conflicts after I merged another PR.

Actually there only seems to be one like conflict, with the creation date of the django.po

Mankro:
If it's OK to you Saara, you could rebase and fix the small conflict in every commit that causes a conflict. The django.po file is mostly automatically generated and many lines easily cause conflicts, but fixing them is also easy.

Mankro:
@saarasat merging the master branch into your topic branch is not recommended. We get a cleaner history if you rebase. Well, I see you are doing these things at the same time.

saarasat:
Y

> @saarasat merging the master branch into your topic branch is not recommended. We get a cleaner history if you rebase. Well, I see you are doing these things at the same time.

Yeah, I tried to resolve the conflict on the github-interface, but that definitely did not work. Just removed that before you said this I guess :)

saarasat:
I'll rebase, just a sec

saarasat:
I rebased now

Mankro:
@saarasat Great work, thank you! And thanks to the reviewers. :+1:

PR #564, opened 2020-06-10
Improve extra login options on login page

### LINKED ISSUE #541: auth: hide maintenance login from login page
The maintenance login form shouldn't be visible by default, when there are other authentication boxes (Shibboleth or social).

So, hide the login div behind a button, a link or a collapse.

etanttila:
# Description

- The local user login (previously "maintenance login") is rarely used, yet was visible on the login page. This pull request hides that login option by default if there are normal login options and provides support for hiding other additional login options similarly. These hidden login options can be made visible by pressing a button.
- The "Maintenance login" was quite unclear and misleadingly named, the title is changed to a more fitting "Local users" and an instructive and descriptive text is added to explain who should use the login in question.
- The texts for the local user login used to be fetched from aplus/settings.py, this changes the implementation to hard coding by defining the login texts directly in the template.
- The texts in the login box headings and login buttons were updated to sentence case (from title case) to be consistent with the rest of the service.

Fixes #541 

# Testing

- All old basic tests except seemingly broken `test_uploading_and_viewing_file (exercise.tests.ExerciseTest)` work

Should be tested manually:
- When there are other login options, the local user login is hidden, there is an button with an chevron icon ("Show more login options"), and pressing that button displays the local user login (and possible other extra logins) and the button itself is hidden
- When there are no normal login options, the local user login (and other possible login options) are visible, and the "Show more login options" button is not.
- Navigating with keyboard works and the focus is changed to a visible input element after pressing the "Show more login options"
- The page and "Show more login options" is usable using a screen reader
- Logging in through the local user login is possible when JavaScript is disabled (both when there are other login options and when there aren't), and the "Show more login options" button should not be visible
- Everything is in Finnish when the browser's preferred language is set in Finnish (and likewise in English)
- Everything works fine even if using an older version of aplus/settings.py as well as when with a different brand name

- [ ] ~I (developer) have created unit tests and the tests pass~
- [ ] ~I (developer) have created functional tests (Selenium tests) if applicable~

### DIFF
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -3586,13 +3586,8 @@ msgid "Password"
 msgstr "Salasana"
 
 #: userprofile/templates/userprofile/login.html:94
-#, python-format
-msgid ""
-"\n"
-"  You may want to read our <a href=\"%(url)s\">privacy notice</a>.\n"
-"  "
+msgid "You may want to read our <a href=\"%(url)s\">privacy notice</a>."
 msgstr ""
-"\n"
 "Voit olla kiinnostunut <a href=\"%(url)s\">tietosuojailmoituksestamme</a>."
 
 #: userprofile/templates/userprofile/logout.html:4
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -985,7 +985,7 @@ msgstr "Avaa 0 arvosanan diplomi"
 
 #: diploma/templates/diploma/list.html:26
 #: edit_course/templates/edit_course/signin_as_user.html:14
-#: userprofile/templates/userprofile/login.html:78
+#: userprofile/templates/userprofile/login.html:84
 #: userprofile/templates/userprofile/profile.html:35
 msgid "Username"
 msgstr "Käyttäjätunnus"
@@ -3524,6 +3524,7 @@ msgid "Log out"
 msgstr "Kirjaudu ulos"
 
 #: templates/base.html:138 templates/base.html:174
+#: userprofile/templates/userprofile/login.html:90
 msgid "Log in"
 msgstr "Kirjaudu sisään"
 
@@ -3572,25 +3573,38 @@ msgstr "Vaativammat pisteet käytetään helpompien vaatimuksien täyttöön."
 msgid "{:d} points"
 msgstr "{:d} pisteet"
 
-#: userprofile/templates/userprofile/login.html:5
-#: userprofile/templates/userprofile/login.html:10
+#: userprofile/templates/userprofile/login.html:6
+#: userprofile/templates/userprofile/login.html:17
 msgid "Login"
 msgstr "Kirjaudu sisään"
 
-#: userprofile/templates/userprofile/login.html:54
-msgid "Login Using Google"
+#: userprofile/templates/userprofile/login.html:61
+msgid "Log in using Google"
 msgstr "Kirjaudu Google-tunnuksella"
 
-#: userprofile/templates/userprofile/login.html:81
+#: userprofile/templates/userprofile/login.html:72
+msgid "Local users"
+msgstr "Paikalliset käyttäjät"
+
+#: userprofile/templates/userprofile/login.html:75
+#, python-format
+msgid ""
+"If you have been provided with credentials specifically for %(brand_name)s, "
+"use this login."
+msgstr ""
+"Mikäli sinulla on tunnukset yksinomaan %(brand_name)s-palvelua varten, "
+"kirjaudu sisään tästä."
+
+#: userprofile/templates/userprofile/login.html:87
 msgid "Password"
 msgstr "Salasana"
 
+#: userprofile/templates/userprofile/login.html:98
 #: userprofile/templates/userprofile/login.html:99
-#: userprofile/templates/userprofile/login.html:100
 msgid "Show more login options"
 msgstr "Näytä lisää kirjautumisvaihtoehtoja"
 
-#: userprofile/templates/userprofile/login.html:108
+#: userprofile/templates/userprofile/login.html:107
 #, python-format
 msgid "You may want to read our <a href=\"%(url)s\">privacy notice</a>."
 msgstr ""
@@ -3686,6 +3700,6 @@ msgstr ""
 msgid "Forget automatic redirection"
 msgstr "Unohda automaattinen uudelleenohjaus."
 
-#: userprofile/views.py:80
+#: userprofile/views.py:78
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -3585,7 +3585,13 @@ msgstr "Kirjaudu Google-tunnuksella"
 msgid "Password"
 msgstr "Salasana"
 
-#: userprofile/templates/userprofile/login.html:94
+#: userprofile/templates/userprofile/login.html:99
+#: userprofile/templates/userprofile/login.html:100
+msgid "Show more login options"
+msgstr "Näytä lisää kirjautumisvaihtoehtoja"
+
+#: userprofile/templates/userprofile/login.html:108
+#, python-format
 msgid "You may want to read our <a href=\"%(url)s\">privacy notice</a>."
 msgstr ""
 "Voit olla kiinnostunut <a href=\"%(url)s\">tietosuojailmoituksestamme</a>."
--- a/userprofile/static/userprofile/extra_logins.js
+++ b/userprofile/static/userprofile/extra_logins.js
@@ -1,0 +1,24 @@
+(function ($) {
+
+  function showExtras(event) {
+    event.preventDefault();
+    const extras = $("#login-box-row").find(".extra-login")
+    extras.show();
+    $(".show-extra-login-btn").hide();
+    extras.find(":input:visible").first().focus() // focus for keyboard navigation
+  }
+
+  $(function () {
+    const row = $("#login-box-row")
+    const normal = row.find(".login-box").not(".extra-login")
+    if (normal.length > 0) {
+      row.find(".extra-login").hide();
+      $(".show-extra-login-btn").on("click", showExtras).show();
+
+      if (normal.length % 2 == 0) {
+        normal.last().after('<div class="clearfix visible-sm-block"></div>');
+      }
+    }
+  });
+
+}(jQuery));
--- a/userprofile/templates/userprofile/login.html
+++ b/userprofile/templates/userprofile/login.html
@@ -58,7 +58,7 @@
 							</p>
 						{% endif %}
 						<p>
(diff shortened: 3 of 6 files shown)


userprofile/templates/userprofile/login.html
```diff
@@ -1,99 +1,112 @@
...
-                </p>
-            </div>
-        </div>
-    </div>
-    {% endif %}
+		<div class="col-sm-6 col-md-3 login-box extra-login">
+			<div class="thumbnail">
+				<div class="caption">
+					<h3>
+						{% trans "Local users" %}
```
saarasat, line 72: In my opinion the "Local users" is actually a bit obscure, I wouldn't know what is meant by it. This might just be though, and the additional text explains the login-option better.

userprofile/templates/userprofile/login.html
```diff
@@ -1,99 +1,112 @@
...
+						{% endif %}
+						<p>
+							<a class="btn btn-primary btn-block" role="button" href="{% url 'shibboleth-login' %}?next={{ next|iriencode }}">
+								{{ shibboleth_button_text|safe }}
+							</a>
+						</p>
+					</div>
+				</div>
+			</div>
+		{% endif %}
```
saarasat, line 45: In general however, logging in as a teacher and student seem to work well with the currently available "Local user"-log in.

userprofile/templates/userprofile/login.html
```diff
@@ -1,99 +1,112 @@
...
-    <h1>{% trans "Login" %}</h1>
-</div>
+	<div class="page-header">
+		<h1>{% trans "Login" %}</h1>
+	</div>
+
+	<div id="login-box-row" class="row">
 
-<div class="row">
+		{% if shibboleth_login %}
```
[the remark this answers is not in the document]
raphendyr: > Is there a way to check these locally?

Uncomment/Add the shibboleth/social login to IǸSTALLED_APPS in the settings for example.

userprofile/templates/userprofile/login.html
etanttila: "Local users" certainly isn't perfect since the term unfamiliar to many users (such as myself), however, it's rather good since it's kind of exactly what it is, thus googling the term comes up with results that are the equivalent of this. We couldn't really come up with a better title. (And I don't think "A+ login" is any better, since at least my first thoughts to that are: "Don't the other login options log in to A+?") (If you have better ideas, please share them!) Yeah, the descriptive text is really, really important with an unclear, unfamiliar and jargon-y term

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
etanttila: That could be a good work-around. I just had `{% if shibboleth_login or True %}` :D

userprofile/templates/userprofile/login.html
```diff
@@ -1,99 +1,112 @@
...
-                    </div>
-                    <div class="form-group">
-                        <button type="submit" class="btn btn-default btn-block">{{ login_button_text|safe }}</button>
-                    </div>
-                </form>
-            </div>
-        </div>
-    </div>
-</div>
+		<button id="show-extra-logins" class="btn btn-default show-extra-login-btn"
```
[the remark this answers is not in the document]
etanttila: I removed this html-id as it wasn't entirely necessary. Good point! Thanks

userprofile/static/userprofile/extra_logins.js
```diff
@@ -0,0 +1,20 @@
...
+    $(".show-extra-login-btn").hide();
+    extras.find(":input:visible").first().focus() // focus for keyboard navigation
+  }
+
+  $(function () {
+    const row = $("#login-box-row")
+    if (row.find(".login-box").not(".extra-login").length > 0) {
+      row.find(".extra-login").hide();
+      $(".show-extra-login-btn").show();
+      $(".show-extra-login-btn").on("click", showExtras);
```
raphendyr, line 16: These lines can be joined, which would also mena that the query is not executed twice:

```js
$(".show-extra-login-btn").on("click", showExtras).show();
```

userprofile/templates/userprofile/login.html
```diff
@@ -1,99 +1,112 @@
 {% extends "base.html" %}
 {% load i18n %}
+{% load static %}
 {% load bootstrap %}
 
 {% block title %}{% trans "Login" %} | {{ block.super }}{% endblock %}
 
+{% block scripts %}
```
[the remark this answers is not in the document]
saarasat: Like we talked, this issue was resolved and these comments can be now skipped.

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
saarasat: Great, I was able to check these other options now :) On desktop everything seems to work well. But I was wondering about the view on mobile or tablet or some other screen widths even on desktop. At least to me, the arrow would go unnoticed, when the width gets smaller: 

**Mobile**

image

**Other screen widths**


image


image

userprofile/templates/userprofile/login.html
saarasat: Would "Admin users" do? Since in Finnish it is now "Ylläpidon kirjautuminen"? Just an idea :)

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
raphendyr: We talked about the arrow a bit with Ella earlier. For example, in mobile case, it should point downwards (my opinion), but I think doing that is a bit too much work for relatively small change. Same applies for the second screenshot, which is probably okeish for now. or is it not?

However, I think there should be bigger margin between the arrow and the text in the first and last image.

userprofile/templates/userprofile/login.html
raphendyr: Admins don't use that. Our best not-best word was "special login for special people". There are very few if no users wich use this login. Basically, if we need to grant someone access to the system and they can't use either of the normal logins.

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
etanttila: (As a note to outsiders: This problem was caused by something in Saara's settings, not by this code.)

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
etanttila: Now there is a larger margin above the privacy notice text, and the arrow points downwards when all the boxes are arranged vertically. The button (as well as the extra logins) should now be aligned with the left side even when the heights of the boxes differ. Hopefully these changes will allow the button to not go as easily unnoticed.

etanttila:
When you have time @saarasat, if you check that everything seems fine with the updated version, all problems discussed earlier are solved and no new ones have come up, I could clean up the git history.

userprofile/templates/userprofile/login.html
saarasat: Okay, seems reasonable :)

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
saarasat: This seems good now to me :) it is cleared on different devices.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3572,27 +3573,41 @@ msgstr "Vaativammat pisteet käytetään helpompien vaatimuksien täyttöön."
...
 msgid "Password"
 msgstr "Salasana"
 
-#: userprofile/templates/userprofile/login.html:94
+#: userprofile/templates/userprofile/login.html:98
+#: userprofile/templates/userprofile/login.html:99
+msgid "Show more login options"
+msgstr "Näytä lisää kirjautumisvaihtoehtoja"
+
+#: userprofile/templates/userprofile/login.html:107
```
saarasat, line 3607: DId you makemessages and compilemessages after yesterday's commit? 

For me the translations seem to be a bit off on the container and locally: 
image

And for instance this line is currently on a different line in the login.html, which would suggest that there has been changes in the translations after the previous message compiling.

userprofile/templates/userprofile/login.html
```diff
@@ -1,99 +1,115 @@
...
+						<div class="form-group">
+							<input class="form-control" type="text" id="id_username" name="username" maxlength="254" placeholder="{% trans 'Username' %}">
+						</div>
+						<div class="form-group">
+							<input class="form-control" type="password" id="id_password" name="password" placeholder="{% trans 'Password' %}">
+						</div>
+						<div class="form-group">
+							<button type="submit" class="btn btn-default btn-block">{% trans "Log in" %}</button>
+						</div>
+					</form>
```
saarasat, line 92: I did the "Matthew-accessibility-check" and seems like it says there are few errors here, mainly the form-labels are missing: 

image

locale/fi/LC_MESSAGES/django.po
```diff
@@ -985,7 +985,7 @@ msgstr "Avaa 0 arvosanan diplomi"
```
saarasat, line 985: In general the appearance, code and the UX is good! I also did not have those issues anymore with the static files not loading, so that is solved. Just a few things there might be worth checking that I mentioned below, after those, I consider this done.

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
saarasat: So can be resolved

locale/fi/LC_MESSAGES/django.po
raphendyr: The `compilemessages` must be run locally, as the binary files are not stored in the git.

userprofile/templates/userprofile/login.html
raphendyr: We talked that we leave this for the later. I finally created an issue about that #582. Thanks for testing the accessibility!

locale/fi/LC_MESSAGES/django.po
etanttila: I didn't update the translations, it seems like only the lines have changed on three of the translations (and the lines actually don't matter, they're just to help the life of the translators by helping to find the context, they are commented out). I could do one more fixup commit with that would update the line numbers. @raphendyr should I?

etanttila: Running A+ in the container should update the translations when you mount it to src (rather than srv).
I talked with Jaakko, he said not to update the translations (since it doesn't actually affect the content), and we talked that it would be good to update the translation-thingy so the comments don't include the line numbers (just the name of the file), since the line numbers can become rather annoying with git.

locale/fi/LC_MESSAGES/django.po
raphendyr: > after those, I consider this done.

Asuming those got resolved, can you verify and mark the review "accepted"?

locale/fi/LC_MESSAGES/django.po
saarasat: Sounds like a valid idea! The lines are a bit unnecessary in my opinion as well. And good to know that about the behavior of translations, I remembered something about src and srv working differently with translations, and apparently this was it. So this thing can be resolved :)

raphendyr:
Thanks! Very nice commits :)

Issue #559, opened 2020-06-10
Provide a link to skip over the navigation


mtthwhggns:
When navigating with the keyboard, users currently must tab through the brand link, the "select course" list, and profile link before reaching the page-specific content. After that, the course pages have a further menu which is present on every page. 

Criterion 2.4.1 of WCAG 2.1 (included in level A, the most basic) (https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks) requires that:

> A mechanism is available to bypass blocks of content that are repeated on multiple Web pages.

Most websites achieve this with [a link that's hidden until you tab into it](https://www.w3.org/WAI/WCAG21/Techniques/general/G1). At this point, it becomes visible. Clicking it jumps to the page content. Here's GitHub's example:

![GitHub skip link](https://user-images.githubusercontent.com/6674046/84235060-e7c33800-aafd-11ea-80be-35a06c1db945.png)

After activating it, focus is ready to enter the link "apluslms" visible in this screenshot - it has bypassed the whole top bar.

There's one which works exactly the same on Aalto.fi:

![Aalto.fi skip link](https://user-images.githubusercontent.com/6674046/84235658-e3e3e580-aafe-11ea-83c4-c1d2e04fbc3a.png)

I'd approach this with a similar link at the top of all pages, which is hidden by default. Activating the link takes focus to the first item after the top bar. I'd then consider a second link to skip the course navigation, which is only visible in the course navigation. This needs more testing though.

I'd not anticipate these will be noticed by users not using the keyboard to navigate. Users who navigate using the keyboard (whether related to disability or preference) probably see these everywhere and are used to them. 

The meta issue: #501

Mankro:
A very clear description. Your suggestion for the fix sounds good! :+1:

raphendyr:
> I'd not anticipate these will be noticed by users not using the keyboard to navigate.

I would agree. I didn't know this cool feature existed in the example sites. :+1:

PR #565, opened 2020-06-10
Create "skip links" for keyboard users

mtthwhggns:
# Description

Adds two skip links - one at the top of the page which jumps
to the page contents, and another at the top of the "course menu"
which jumps to the course content.

As described in https://github.com/apluslms/a-plus/issues/559 this is really useful for keyboard users.

It would be sensible to merge this after #547 , as it doesn't have it's own focus state.

Here are a couple of screenshots of how it looks on top of the focus state from #547:

Skip to content page link

Skip course menu link

Fixes https://github.com/apluslms/a-plus/issues/559

# Testing

I've not tried any changes to the automatic tests for keyboard-only stuff here.

It can be tested manually by tabbing through the pages, and especially pages where the course menu is visible. 

I don't think it's necessary to add any specific documentation here. 

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -228,6 +228,33 @@ html, body {
   padding-top: 70px;
 }*/
 
+/* Skip links */
+
+.skip-link { /* This one is equally applicable to all skip links */
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    margin: 0;
+    overflow: hidden;
+    clip: rect(0, 0, 0, 0);
+    display: block;
+}
+
+.skip-link:focus {
+    position: static;
+    width: 100%;
+    min-height: 50px;
+    height: auto;
+    margin: inherit;
+    overflow: visible;
+    clip: auto;
+    padding: 15px;
+}
+
+.page-skip-link:focus {
+    min-height: 24px;
+}
+
 /* menus */
 .course-menu .nav>li>a {
     padding: 2px 2px 2px 5px;
--- a/course/templates/course/_course_menu.html
+++ b/course/templates/course/_course_menu.html
@@ -7,6 +7,12 @@
 
 {% prepare_course_menu %}
 
+<li>
+	<a href="#course-content" class="skip-link page-skip-link">
+		{% trans "Skip course navigation" %}
+	</a>
+</li>
+
 {% if instance.language|first == "|" %}
 <li role="presentation" class="header">
 	<div class="row">
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-06-03 12:25+0300\n"
+"POT-Creation-Date: 2020-06-23 11:11+0300\n"
 "PO-Revision-Date: 2019-08-14 12:16+0200\n"
 "Last-Translator: Markku Riekkinen <markku.riekkinen@aalto.fi>\n"
 "Language-Team: Finnish <>\n"
@@ -260,7 +260,7 @@ msgstr "Valmis"
 msgid "Unlisted in table of contents"
 msgstr "Näkymätön sisällysluettelossa"
 
-#: course/models.py:680 course/models.py:806 templates/base.html:102
+#: course/models.py:680 course/models.py:806 templates/base.html:104
 msgid "Hidden"
 msgstr "Piilotettu"
 
@@ -382,44 +382,52 @@ msgstr "Seuraavat käyttäjät olivat jo ilmoittautuneet: {users}"
 msgid "Begin by enrolling in a course."
 msgstr "Aloita ilmoittautumalla jollekin kurssille."
 
-#: course/templates/course/_course_menu.html:14
-#: course/templates/course/_course_menu.html:38
+#: course/templates/course/_course_menu.html:11
+msgid "Course navigation"
+msgstr "Kurssivalikko"
+
+#: course/templates/course/_course_menu.html:12
+msgid "Skip course navigation"
+msgstr "Ohita kurssivalikko"
+
+#: course/templates/course/_course_menu.html:20
+#: course/templates/course/_course_menu.html:44
 #: edit_course/templates/edit_course/edit_course_base.html:26
 msgid "Course"
 msgstr "Kurssi"
 
-#: course/templates/course/_course_menu.html:18
+#: course/templates/course/_course_menu.html:24
 #: userprofile/templates/userprofile/profile.html:39
 msgid "Language"
 msgstr "Kieli"
 
-#: course/templates/course/_course_menu.html:51
+#: course/templates/course/_course_menu.html:57
 #: course/templates/course/_siblings.html:54
 #: exercise/templates/exercise/toc.html:7
 #: exercise/templates/exercise/toc.html:12
 msgid "Course materials"
 msgstr "Kurssimateriaali"
 
-#: course/templates/course/_course_menu.html:57
+#: course/templates/course/_course_menu.html:63
 #: exercise/templates/exercise/results.html:13
 msgid "Exercise results"
 msgstr "Pistetilanne"
 
-#: course/templates/course/_course_menu.html:66
+#: course/templates/course/_course_menu.html:72
 msgid "Form a group"
 msgstr "Luo uusi ryhmä"
 
-#: course/templates/course/_course_menu.html:94
+#: course/templates/course/_course_menu.html:100
 msgid "Apps"
 msgstr "Sovellukset"
 
-#: course/templates/course/_course_menu.html:108
+#: course/templates/course/_course_menu.html:114
 #: exercise/templates/exercise/_user_results.html:98
 #: exercise/templates/exercise/submission_plain.html:41
 msgid "Course staff"
 msgstr "Kurssihenkilökunta"
 
-#: course/templates/course/_course_menu.html:114
+#: course/templates/course/_course_menu.html:120
 #: course/templates/course/staff/enroll_students.html:16
 #: course/templates/course/staff/participants.html:6
 #: course/templates/course/staff/participants.html:11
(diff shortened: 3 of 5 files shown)


assets/css/main.css
```diff
@@ -447,3 +447,30 @@ table.results-table .category-row {
...
+    height: auto;
+    margin: inherit;
+    overflow: visible;
+    clip: auto;
+    padding: 15px;
+}
+
+.page-skip-link:focus {
+    min-height: 24px;
+}
```
raphendyr, line 476: It would be nice if there were a new line in the end of the file. I look into adding editorconfig, so this would be automatically configured for everyone's editor.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3688,3 +3698,1254 @@ msgstr "Unohda automaattinen uudelleenohjaus."
 #: userprofile/views.py:80
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
+#: venv/lib/python3.7/site-packages/colorfield/fields.py:10
```
raphendyr, line 3702: Please create messages with `./manage.py createmessages --ignore='venv/*' --no-obsolete`.

assets/css/main.css
```diff
@@ -447,3 +447,30 @@ table.results-table .category-row {
     margin-top: 4rem;
     width: 80%;
 }
+
+/* Skip link stuff */
```
raphendyr, line 451: Is the best location for these in the end of the file? I can be, but often people just add it to the end and then the content of the file will have no sorting after a while. To be fair, the file has a lot of that already, thus finding a good place is hard.

I typically see first generic stuff, then definitions matching elements in all pages and starting from the top of the page. Lastly, I would expect to find items on rare pages and often would hope those to be in different files. As this exists in few places in the page, would it be more suitable somewhere around lines 200-240 maybe?

locale/fi/LC_MESSAGES/django.po
mtthwhggns: I guess you mean `makemessages`? Or am I missing an undocumented command?

Additionally, can this be added to the development documentation somewhere?

assets/css/main.css
mtthwhggns: Without wishing to sound like a stuck record, with better tooling we could use one file per component and squish and minify them on deployment, which would be most logically organised and efficient.

I've never navigated big CSS files any other way than searching for selectors, so I'll yield to your view on this.

assets/css/main.css
mtthwhggns: Marking as resolved, as it's covered by https://github.com/apluslms/a-plus/pull/565#discussion_r438571411

locale/fi/LC_MESSAGES/django.po
raphendyr: > I guess you mean makemessages?

Yes, my bad.

p.s. how did I miss this response. I checked all the conversaionts from the notification. Sorry for that.

> Additionally, can this be added to the development documentation somewhere?

Yes. I think it should be. Specially that --no-obsolete. Ignore is only required if the virtual env is installed within the repository. I might add this to the styleguide (i.e. how translations should be handled with the git).

locale/fi/LC_MESSAGES/django.po
```diff
@@ -382,52 +382,56 @@ msgstr "Seuraavat käyttäjät olivat jo ilmoittautuneet: {users}"
 msgid "Begin by enrolling in a course."
 msgstr "Aloita ilmoittautumalla jollekin kurssille."
 
-#: course/templates/course/_course_menu.html:14
-#: course/templates/course/_course_menu.html:38
+#: course/templates/course/_course_menu.html:12
+msgid "Skip course menu"
+msgstr "Ohita kurssivalikko"
```
etanttila, line 387: Usually "kurssivalikko" refers to a listing of courses (or a menu consisting of courses) if I've understood correctly and have the correct impression. This seems to refer to the menu of the course. I'm not sure if the term can be used in that context. However, I'm not sure what would be a better term.

assets/css/main.css
raphendyr: > Without wishing to sound like a stuck record, with better tooling we could use one file per component and squish and minify them on deployment, which would be most logically organised and efficient.

Yes. This is the reason why I would hope we could to do that (relatively simply considering current deployment methods). For now, if the content applies to only few pages (e.g. login), then it's good to use a dedicated file. Due to HTTP/2 and 3, there isn't so much need for bundling the files anymore. Sadly, all minify tools seem to do both and configuring them to do only minifying is hard or complicated. I was planning on writing a new library to be run with `./manage.py collectstatic`, which would download remote files, minify local files and fix internal links, but I haven't had time for that yet.

> I've never navigated big CSS files any other way than searching for selectors, so I'll yield to your view on this.

I try to not to repeat past mistakes, so thank you. Personally, I consider cleaning up stuff to be one of the hardest things in the information technology.

raphendyr:
@Mankro I think this is fine for merging.

templates/base.html
```diff
@@ -73,6 +73,7 @@
         <meta name="msapplication-TileImage" content="{{ STATIC_URL }}mstile-144x144.png">
     </head>
     <body class="{% if profile %}{% if is_external_student %}external-student{% else %}internal-student{% endif %}{% endif %} lang-{{ request.LANGUAGE_CODE }}" data-view-tag="{% block view_tag %}{% endblock %}">
+        <a href="#content" class="skip-link">{% trans "Skip to content" %}</a>
```
etanttila, line 76: When clicking this link, it scrolls me down and moves me to the content (past the navigation up top) to this view:
image

However, when I press tab again, it scrolls me up a bit and displays the "Skip course menu" button. Can this unnecessary scrolling up and down be avoided?
image

locale/fi/LC_MESSAGES/django.po
raphendyr: I think, this could be called "Skip course navigation" in English, but that translates to "Ohita kurssinavigointi", which doesn't sound very good though.

templates/base.html
mtthwhggns: Hi @etanttila ,

I've tried to replicate this, and haven't been able to. It seems to work as expected (top bar is always visible, scrolling is to just the top of the content area) on Chrome, Firefox and Safari on macOS. I've had a bit of trouble getting a local version tested in Ubuntu so far.

raphendyr: I think the code works correctly for me too. Yesterday, I though the first link was supposed to take you to the course-content, but I was mistaken.

etanttila: The problem doesn't exist for me on Chrome (clicking on the link on pages with a course menu scrolls to just below the blue top bar so the blue top bar gets hidden), but does happen every time on Firefox (version 77.0.1 on Ubuntu) that there is a course menu (so not on the front page nor login page, on those pages the blue top bar stays up top). Strange, I wonder what causes this.

etanttila: I wonder if one of the other summer interns could check if they have the same problem or it it's just me. For example @saarasat

mtthwhggns:
I've taken on board @etanttila 's thoughts about the lack of clarity, and made a few changes:

- I've used the same phrasing for both links "skip X"
- I've described the nav bar as the "main navigation"
- I've stuck with "kurssivalikko" in Finnish for now - there doesn't seem to be consensus on a better option. 
- Both the main and the course navigation now have matching `aria-labels`, so "Skip course navigation" skips the area that's previously been described as "course navigation". This should help a little.

Please test it and see what you think now.

templates/base.html
```diff
@@ -74,6 +74,7 @@
         <meta name="msapplication-TileImage" content="{{ STATIC_URL }}mstile-144x144.png">
     </head>
     <body class="{% if profile %}{% if is_external_student %}external-student{% else %}internal-student{% endif %}{% endif %} lang-{{ request.LANGUAGE_CODE }}" data-view-tag="{% block view_tag %}{% endblock %}">
+        <a href="#content" class="skip-link" aria-label="{% trans "Main navigation" %}">{% trans "Skip main navigation" %}</a>
```
raphendyr, line 77: I think I'm just unknowledged about this, but should the `aria-label` be in the element that contains the labeled part, thus in this case in the `<nav -class="topbar"`? I assume it doesn't matter so much as the main importance is that the screen reader reads it at the correct time. Similarly, should this link be as the first thing within the topbar nav or just before that nav as a sibling element? I'm not familiar enough to know this, thus I'm mainly interested for how this location was selected assuming there are some reason for it.

mtthwhggns: Quite right @raphendyr , I copy/pasted the `aria-label` into the wrong element between testing it and fixing the translations. It's fixed now. 

Generally, the most important thing is that it's the first element that can receive keyboard focus, which is why it's not nested inside the main navigation. I also like the visual effect when it sticks to the top of the page and displaces the content a little bit.

raphendyr: > I also like the visual effect when it sticks to the top of the page and displaces all of the content a little bit.

I like it too. I didn't remember that, when I was thinking the location of the element.

course/templates/course/_course_menu.html
```diff
@@ -7,6 +7,12 @@
 
 {% prepare_course_menu %}
 
+<li>
+	<a href="#course-content" class="skip-link page-skip-link" aria-label="{% trans "Course navigation" %}">
```
raphendyr, line 11: Few thoughts that came into my mind (i.e. unimportant and very minor)

Should the `aria-label` be in the [parent nav element](https://github.com/apluslms/a-plus/blob/3368af1e4ca937d601730c925be9d29aa97e5a96/course/templates/course/course_base.html#L21)? For example, should we consider the advert box to be part of "course navigation" or not (I think it's part of the left panel, but I'm not 100% if all of that is course navigation)? The main navigation now has it in order where link comes before the aria-label, so is that better?

I feel that I'm unsure how to read `aria-label` in the source. Like, should I think that `aria-label` tells "this element contains thing X" or "thing X starts from here"?

If this `li` would be in `course_base.html` just before the include statement, then the link and destination would be in the same file. Though, everything else is in this partial file, so I'm not sure which is better. Additionally, it does not need to be under the `ul` element either, unless that is beneficial for the style (i.e. it would be fine to have a dedicated entry in the `nav` before `ul` just for this).

mtthwhggns: Oh dear. This is the same copy/paste error I fixed yesterday, just in the other component. Will fix it next.

mtthwhggns: Fixed this now @raphendyr

Mankro:
@mtthwhggns if you wanted to squash commits, you could do it now.

mtthwhggns:
@Mankro Squashed

Issue #554, opened 2020-06-10
automatic notification for user about arriving on pages of a past course


annirytkonen:
A+ knows that a course has closed since course instances have opening and closing times. Any closed course instance could show a warning easily. Adding an automatic link to the newest course instance would be easy too.

[Teacher needs spring 2020](https://wiki.aalto.fi/display/EDIT/Teacher+needs+spring+2020): nr 12

To be allocated to one of the summer interns

PR #570, opened 2020-06-11
Improve appearance and functioning of course news

### LINKED ISSUE #427: News and information order on the course front page: redesign the UX
disable (completely or as an option) the accordion and folding in the news  section of courses' front pages

saarasat:
# Description

Previously the news-section of the course frontpage could be cluttered and make the news go unnoticed by students. It also hid the content of all but the first news item by default. This PR works on that. 

This change:
- Adds the possibility for a teacher to pin news items on the top of the course news. For a teacher the items show with a pin, for students not. This is to direct teachers to notice that items can be pinned, but does not emphasize it too much for students.
- News items are ordered by: 1. pinning and 2. date of publishing
- Removes the previously cluttering option for alerts. **Removal has been done for the database field as well.**
- Removes the folding of the news content, making all the news content instantly visible.
- Boldens and removes the border of a news item -headline.
- Improves the appearance and accessibility of the news box in general.

This change does NOT:
- Change the way the content is ordered on the front page.
- Change the number or logic behind how many items are shown unfolded, as changing that would require more investigation
and analysis from the database of the way course news are used in general. 

Fixes #427 

# Testing

I ran the existing unit- and selenium-tests and they all passed. There were no previous Selenium-tests for the news section, the testing was limited to manual tests. 

I also checked the site with wave browser extension to check that no new accessibility issues were introduced and this removed some of the old ones. 

Below are images of the situation of course news before and after the changes were implemented with 8 different news items: 

**Before**

image
image

**After for the teacher**

image

**After for the student**

image

If these changes are implemented, there might be some needs to change the documentation for the teacher. 


raphendyr:
If I interpreted #427 correctly, I tink the request wanted for news items to be uncollapsed too. I.e. the collapse feature should be removed and the news items should be listed like blog posts. Though, if there were content below them, then that content would become hard to access.

saarasat:
I actually thought about that, but then concluded, that might make the news-section unnecessarily large. And the content hard to access. The issue/wish was explained like this: 

"The course front page now has an accordion component that displays course news. **Only a very small number of news items are visible by default. I'd prefer all news items to be visible,** not just a handful of most recent ones, because sometimes there are things further down on the list that students should notice." 

And I feel like this change fixes the issue of not all news items being visible, not just the most recent three. I did not find any specific mention that news items and their content should be visible.  

Additionally there was a comment about not liking the hiding of course info, but I thought that was a general comment made in the context, which just explained the need for showing more news items.

raphendyr:
Pinging @jsorva, do you feel that above images show the result you are looking for?

jsorva:
> 
> 
> @jsorva do you feel that above images show the result you are looking for?

I like it that all the items are listed. (No more "Show more".)

I'd prefer each news item to also be unfolded; i.e, with the entire text visible. Yes, it takes more space, but that seems relatively harmless here; I'm more worried about students not clicking the items open and missing out on something.

raphendyr:
We could use local storage and JS to highlight news which are "new", i.e., student has not been on that page for some seconds after the items have been added. Though, that would be an additional enchament.

saarasat:
Regarding this issue and pull request, what do you think @raphendyr , should I 
a) remove the folding from the content too and this PR can be then closed.  
b) remove the folding AND work on highlighting the news items that the user has not seen in this same pull request. And if this option is chosen, what kind of GDPR-stuff needs to be taken into consideration and how much is this issue worth using time?
c) remove the folding of the content AND then we can close this pull request AND then I open a new issue for the highlighting of the unseen news.
d) Something else, what?

raphendyr:
> I open a new issue for the highlighting of the unseen news.

I think the highlight should be a separate issue, so we cabn prioritize it. I.e. we have more important stuff for now (takes some time to make it work nicely).

I'm not sure what was the difference in points A and C, but I think you could remove the collapse feature of the news items.

saarasat:
The difference between A and C was just whether we do the highlighting, thus whether an issue should be created out of that :) But great, I'll remove the folding for content too, and then create a new issue for the non-urgent highlighting.

news/templates/news/user_news.html
```diff
@@ -2,37 +2,44 @@
...
+      	<div class="list-group-item">
+        	<h4 class="list-group-item-heading">
+				{{ item.title|safe }} <small>{{ item.publish }}</small>
+				{% if not item|is_published:now %}
+					<small><span class="label label-default">{% trans "Appears later" %}</span></small>
+				{% endif %}
+				{% if is_course_staff %}
+					<small><span class="label label-default">{{ item.audience|news_audience }}</span></small>
+				{% endif %}
+        	</h4>
```
raphendyr, line 37: This line has spaces.

news/templates/news/user_news.html
```diff
@@ -2,37 +2,44 @@
...
-        {% if is_course_staff %}<small><span class="label label-default">{{ item.audience|news_audience }}</span></small>{% endif %}
-      </h4>
-    </a>
-    <div class="list-group-item-text collapse{% if not item.collapsed %} in{% endif %}" id="news{{ item.id }}">
-      {{ item.body|safe }}
+	<div class="panel panel-default news-panel">
+    	<div class="panel-heading">
+      		<h3 class="news-panel panel-title">{% trans "Course news" %}</h3>
+    	</div>
+    	<div class="list-group">
```
raphendyr, line 26: This and above has spaces, 24 has mix of spaces and tabs.

news/templates/news/user_news.html
saarasat: You are correct, that has been fixed. Those two indendation-commits can be squashed into one once this is mergeable.

news/templates/news/user_news.html
saarasat: And here as well, this has now been fixed.

raphendyr:
I think this looks good. @jsorva does the last screenshot look good to you? I think we should validate this from couple more teachers, @Mankro what do you think?

jsorva:
> 
> 
> I think this looks good. @jsorva does the last screenshot look good to you? I think we should validate this from couple more teachers, @Mankro what do you think?

Looks good --- Thank you!

jaguarfi:
> I think this looks good. @jsorva does the last screenshot look good to you? I think we should validate this from couple more teachers, @Mankro what do you think?

I think we should give the user the opportunity to collapse the whole news section. I agree, that we should not make every news item collapsible, but we can have one collapse element for all the news. (Something between the last two images)
On the other hand, if there are really important news and the teacher want to keep them visible, the Pin option can be selected. (see the red box on Saara's screenshots)

raphendyr:
> I think we should give the user the opportunity to collapse the whole news section. I agree, that we should not make every news item collapsible, but we can have one collapse element for all the news. (Something between the last two images)

I would prefer the news to be open all the time. I think the news could be below toc and results (I don't actually like to have all results here). In addition, I think we should send the user notification when course news update (specially from the important/pinned ones). I think this notification system should be designed in a different issue.

news/templatetags/news.py
```diff
@@ -32,15 +32,12 @@ def user_news(context, num, more=0):
     for item in news:
         i += 1
         item['collapsed'] = i > num
-        if more > 0 and i == more:
-            item['begin_more'] = True
```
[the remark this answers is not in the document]
saarasat: That `item['collapsed']` was still in use after the first commit of removing the "more"-folding. So only the `more` was removed. But actually you are correct, it is no longer needed after the content-folding has been removed too. The whole loop has now been removed with this [commit](https://github.com/apluslms/a-plus/pull/570/commits/e2e75a86fe4ac799a7f2d7404989150975f04511). Good pick!

assets/css/main.css
```diff
@@ -300,11 +300,18 @@ html, body {
 .module-panel table {
     margin: 0;
 }
-.news-panel .list-group-item-text {
-  padding: 10px;
+
+.news-panel .panel-title {
+    font-size: 18px;
```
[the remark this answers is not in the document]
saarasat: I thought about that, using the rem-units. However, pixels seemed to be the more common way in the current main.css to define the (font)sizes and constraints, so I thought I use that too to be more in line with what had been done before. 

But I think your point is valid, and probably something that could be checked in a separate issue in general? Or maybe the style-guide could have some reference for this for the future :)

saarasat:
Regarding the questions and points @jaguarfi you raised above:
https://github.com/apluslms/a-plus/pull/570#pullrequestreview-432342142

1. That is a good point to be made. I personally think that "Alert" might be removed when the item is unpinned. However, it was a bit beyond the scope of this issue, but I'll write that down as an idea if/when we are working with the highlighting the newsitems more. 
2. Markku will probably get back to that
3. Great! 
4. Haha, yes, probably no need. 
5. We talked about this with @raphendyr, and for now I don't think there still was consensus whether the tab-size is 2 or 4 spaces? The current html-files seem to have spaces and tabs combined and they vary quite a lot. The indendation has now been changed anyway so that it uses tabs, which was a good point from Jaakko. So when opening it in editor, whoever is working on it, has a unified look to it. That can be easily then later on changed to eg. so that the tab-width is 2 spaces.
6. I noticed that the tags are problematic also from the accessibility -point of view. But similarly as with item nro 1., it was out of scope here, so I left them for now. But definitely something that could be addressed later on!

saarasat:
> I would prefer the news to be open all the time. I think the news could be below toc and results (I don't actually like to have all results here). In addition, I think we should send the user notification when course news update (specially from the important/pinned ones). I think this notification system should be designed in a different issue.

Regarding this, I am actually a bit worried that if the news are moved below the toc and results, will it beat the purpose of them being more easily noticed by students? For instance, when I changed the order now for the default-course, the user has to scroll very far to see the news. From what I gather, most of the courses at the moment don't have news (of course my knowledge is limited to the once that are currently on A+), or have only a couple of them. So it would not be a huge wall of text even when it is above the toc, and the teachers can always remove the ones which are unnecessary or outdated. 

The actual amount could be of course be verified by checking the amount of news items linked to course instances from the database. 


**News below toc**

image


For user results, at least with only few results, it does not seem so problematic, so that probably could be changed: 

**News below user results**


image

There is also the third option of "last visited", which could be changed as well and a little improvement done on the "last visited"-text in general at the same time, to make it a bit more noticeable:

image







 


The notification-system would definitely be the best solution. In my mind there would be a link on the menu on the left-hand side with "course-news" and some kind of indication when there is a new news item. And/or only the unread ones showing on the course front page. If this has a high enough priority, I am glad to work on this later on :)

raphendyr:
> I wonder if changing the indentation style of the file user_news.html should be reflected on the other HTML templates. Maybe @raphendyr can comment on this.

@jaguarfi one by one. If the change is big enough, then it's good to update the indentation in the file first. I prefer tabs for html, as that language doesn't benefit from spaces.

raphendyr:
Hmm... Should we actually return a bit of collapsing? We could always show news from past 2 weeks and put rest in a collapsed box. That way, we probably would always have only few important things there before anything else.

jaguarfi:
To summarise, perhaps we will need to create some new issues based on the discussions. @raphendyr and @Mankro  can decide if those issues are relevant.
1. Django `BaseFormView` with `models.TextFields` uses HTML editors as UI element. However, the UI does not inform about it. It might be beneficial to point out the possibility of using HTML inside the Text field UI element. We can find some suggestions [here](https://stackoverflow.com/questions/51915026/django-text-field-with-options)
2. Font sizes should be handle in a different manner. [Some explanation about the topic](https://css-tricks.com/accessible-font-sizing-explained/)
3. When the user creates a new "news" element, there is a selector called Alert, which asks for the Alert style. However, the selected style is not applied to the unpinned items.
4. Define style guides for Django Templates and CSS files.
5. The tags used in the news might need to be highlighted.

jaguarfi:
@saarasat  The following course https://plus.cs.aalto.fi/o1/2019/ can give you an idea on how many news a course could potentially have.

saarasat:
> @saarasat The following course https://plus.cs.aalto.fi/o1/2019/ can give you an idea on how many news a course could potentially have.

Good, this helps to get the idea! Knowing this, I would vote for the option were course news remain at the top of the page for all the different index-options, showing the content like it is now with this PR and return a bit of the folding of news items. Maybe showing the most recent news of 1 month?

etanttila:
As a student, I've quite liked the current system, it doesn't take much space, I can quickly skim the most recent news and see if there are any I haven't read yet or that apply to life right then. I would be very annoyed if all the new would be visible, because it would cause so much need to scroll.

I like the idea of indicating new news, I think having a tag such as "New" or "Unread" next to a item of news that a student hasn't seen yet (this could be determined by either (1) it's been published since the student has previously been on the page OR (2) the student hasn't collapsed the news manually) would be good and draw the attention to that item of news. If collapsing isn't set by default, I think the student should have the opportunity to collapse news items, which would also stay collapsed for that student unless she/he decides to uncollapse it.

> Hmm... Should we actually return a bit of collapsing? We could always show news from past 2 weeks and put rest in a collapsed box. That way, we probably would always have only few important things there before anything else.

> Good, this helps to get the idea! Knowing this, I would vote for the option were course news remain at the top of the page for all the different index-options, showing the content like it is now with this PR and return a bit of the folding of news items. Maybe showing the most recent news of 1 month? At least I don't visit course pages necessarily every two weeks as a student myself :D

I definitely think there should be some collapsing. I like the idea of seeing just recent news and unread ones. Since on some courses, I go to the A+ page 3 times a week, I wouldn't want it to necessarily show by default much over two weeks if there are a lot of course news. Of course for some courses, people access the course infrequently like @saarasat said, so that should be taken into consideration. I think it would be good if the course news would show the following pieces of news (in time order of the most recent up top):
1. all the unread items of news
2. all the news for the last two weeks
3 at least three most recent items of news (so in case the criteria listed above don't bring the number of news items up to three, the most recent pieces of news are shown until the number is three. If the criteria cause the number to be at least three, no additional news need to be shown)

I think the unread items of news (criterion 1) should be uncollapsed and able to be read right away (with preferably a "New" or "Unread" tag). I'm not sure about the other news for the last two weeks (criterion 2), they could be uncollapsed or collapsed, or just show the beginning of the body of text (or it all if it's short) with a "... show more".
I believe the other visible news (criterion 3) should display just the title and date (be collapsed like they currently are).

I definitely think the news should stay up top, otherwise I'd at least miss them.

jsorva:
Having the news at the top is important to me, too. 

Regarding folding, I'm not completely opposed to folding under all circumstances, but if it is used, it would be great if it wasn't too aggressive (i.e., nothing gets folded that might be still relevant to students). If the selection of what to fold is automatic, I think time (e.g. N weeks) sounds like a better option than the number of items (currently 3 items irrespective of time stamps, I think).

I would also be perfectly happy with manually marking which news items at the top are still relevant and which ones can be folded (in the course-staff menu).

etanttila:
> I would also be perfectly happy with manually marking which news items at the top are still relevant and which ones can be folded (in the course-staff menu).

@jsorva How about if when publishing a news item, one could specify how long (or until what date) the item is relevant? There could be a default of say, 2 weeks. Of course it would be good to manually change it later on as well, but the timing could help in minimizing the extra effort to teachers.

jsorva:
> 
> 
> > I would also be perfectly happy with manually marking which news items at the top are still relevant and which ones can be folded (in the course-staff menu).
> 
> @jsorva How about if when publishing a news item, one could specify how long (or until what date) the item is relevant? There could be a default of say, 2 weeks. Of course it would be good to manually change it later on as well, but the timing could help in minimizing the extra effort to teachers.

Sounds like it could work. I could use that. OTOH, it might be confusing someone uses it so that the "expiration dates" of the news items aren’t in order. Some other way of marking "show until here and fold the rest" in the news-item list could also work.

saarasat:
Great discussion! I think @jsorva that would actually be a great idea. To let the teachers have say in which are unfolded. That would also satisfy the need for students to receive the most relevant news, without the front page becoming a wall of text and the other content to become unreachable. And also possibly solve the thing that @jaguarfi said about the alerts. I have been thinking about this and the proposed changes and...

Would this be a good compromise to the issue and all of the discussion above:

1. Add the possibility for the teachers to pin items, and all the unpinned, over 1-month-old items would be unfolded under a tag "older news" or "show more". So even if the teacher does not pin any items, there is some kind of a limit so that the front page does not become just a news-site.
2. Possibly adding a small label, that can be set there, if the content is super-important as the image. The text of the badge could be for instance "important" or something similar. Or alternatively highlighting the background with a color. Which on the otherhand might be bad accessibility-wise, so I would prefer the badge. (@jaguarfi this would solve the alert-problem)
3. Possibly adding the pin-image to the ones that have been pinned. Another way to indicate to teachers that by pinning the items, they can be raised to the top of the list. 

**This presents one pinned item and the newest two items**

image


**This is how the feature would look like from the teacher's point of view**

image


4. The current "alert"-option would be removed, at least to my eye, it seems a bit excessive and does not really fall anymore under the "course-news". Also it provides information through color which I think is not good. 

image

image

5. This way the course news could stay on the top of all the three index-options, the amount wouldn't become excessive for students but remain informative. 

5. Additionally an issue could be made for adding the highlighting of the unread news, and that could be worked later on.

etanttila:
>     1. Add the possibility for the teachers to pin items, and all the unpinned, over 1-month-old items would be unfolded under a tag "older news" or "show more". So even if the teacher does not pin any items, there is some kind of a limit so that the front page does not become just a news-site.

@saarasat Could it be that the teacher could decide how long the unpinned items are unfolded (and of course there would be a default value)? A month is a really long time for a course like Programming 1 (where you have deadlines weekly and people visit the site usually several times a week to do problems), but might be good for some other courses.

saarasat:
Yes, that could be set for sure. A good point!

raphendyr:
1. The pinning feature is nice, let's keep that. However, let's ensure pinnedi tems are first in the list.
2. I don't see need for the color alert or label feature. I think it can be dropped. Pinning should be good enough. If news are very important, then email / notification should be send via other means (e.g. mycourses). System alerts (e.g. maintenance) is handled with the side wide alert already. If an exercise is broken, then that should be visible on top of that exercise and not only in the course front page.
3. The pinning should be visible to the student. Doesn't need to emphasised though. In addition, pinned items can have a bottom border before normal items (see twitter, I think they have done it quite nicely).
4. I don't see point for the old alert concept. In my opinion, the visual style doesn't even look like it's part of the news. My eyes just ignore it.
5. Yes, on top of other items is good.
6. on unread news:

If you implement the hiding so that, without JavaScript all news items are rendered, and then the JS would hide those that are too old. That way, it would be easy to extend the script later with an user option to say "I have read this", which would fold that and older news. The information for that can be stored in a local storage, or we could allocated field in the `userprofile`. However, I would prefer that we don't add new fields in the database, unless really needed.

Furthermore, I would prefer a hard coded value for the default visibility time frame. I don't like adding new configurations to A+, as those require two configuration channels currently: via the course repo and the web UI. Though, this decision would require information how news have been used so far (from teachers or analyzing the database).

saarasat:
@raphendyr Great notes, I will move forward with those, and let you know when the changes are done.

news/templates/news/user_news.html
```diff
@@ -1,46 +1,38 @@
...
-        {% if not item|is_published:now %}<small><span class="label label-default">{% trans "Appears later" %}</span></small>{% endif %}
-        {% if is_course_staff %}<small><span class="label label-default">{{ item.audience|news_audience }}</span></small>{% endif %}
+        {{ item.title|safe }}
+        {% if not item|is_published:now %}
+          <small class="future-instance">{% trans "Published on" %}: {{ item.publish }}</small>
+        {% else %}
+          <small class="current-instance">{{ item.publish }}</small>
+        {% endif %}
+        {% if is_course_staff %}<small><span class="label label-primary">{{ item.audience|news_audience }}</span></small>{% endif %}
+        {% if is_course_staff and item.pin %}<small><span class="glyphicon glyphicon-pushpin"></span></small>{% endif %}
```
raphendyr, line 21: Should the glyph be visible for all users? I think it could be nice for students too.

news/templates/news/user_news.html
```diff
@@ -1,46 +1,38 @@
...
-    <div class="list-group-item-text collapse{% if not item.collapsed %} in{% endif %}" id="news{{ item.id }}">
-      {{ item.body|safe }}
+      <div class="list-group-item-text">
+        {{ item.body|safe }}
+      </div>
     </div>
     {% if item.begin_more %}
-    <a class="list-group-item" href="#more-news" data-toggle="collapse" role="button" aria-controls="more-news" aria-expanded="false">
-      {% trans "Show more" %}
+    <a class="folding-list-group-item collapsed" href="#more-news" data-toggle="collapse" role="button" aria-controls="more-news" aria-expanded="false">
```
raphendyr, line 28: I think this solution is fine for now. If this would be reworked, then having two lists "visible" and "hidden" and two for loops for the collapsed area nd for the visible. Both for loops would use a partial for the news div. Sadly, this pattern is inefficient with django templates. Would be fine with Jinja. Hence, let's keep this complicated solution.

news/templatetags/news.py
```diff
@@ -20,25 +20,23 @@ def user_news(context, num, more=0):
     news = context['course_news']
 
     if context['is_course_staff']:
-        alerts,news = news.for_staff()
+        news = news.for_staff()
     else:
         user = context['request'].user
-        alerts,news = news.for_user(
+        news = news.for_user(
```
raphendyr, line 26: > Therei s a change from (a, b) to (a), i.e., from tuple to a single item. If this is supposed to return only a single value and not a tuple, then you should remove the brackets.

As the data is read corectly here, you should remove brackets from the cached/*.py file.

news/templatetags/news.py
```diff
@@ -20,25 +20,23 @@ def user_news(context, num, more=0):
...
             not user.is_authenticated
             or user.userprofile.is_external
         )
 
     i = 0
     for item in news:
```
raphendyr, line 32: This should use `enumerate`, but as this loop is useless now, it could be removed. The other line could be written as

```python
if more > 0 and more < len(news):
    news[more]['begin_more'] = True
```

news/cache.py
```diff
@@ -50,12 +44,10 @@ def filter_news(items, audiences):
             ]
         if is_external:
             return (
-                filter_news(self.data['alerts'], EXTERNAL),
-                filter_news(self.data['news'], EXTERNAL),
+                filter_news(self.data['news'], EXTERNAL)
```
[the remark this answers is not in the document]
saarasat: This has now been changed, https://github.com/apluslms/a-plus/pull/570/commits/458e4b11841b3b55a1042dedba903d6dd22508f3

news/cache.py
```diff
@@ -50,12 +44,10 @@ def filter_news(items, audiences):
...
             )
         return (
-            filter_news(self.data['alerts'], INTERNAL),
-            filter_news(self.data['news'], INTERNAL),
+            filter_news(self.data['news'], INTERNAL)
```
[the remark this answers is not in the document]
saarasat: Same here: https://github.com/apluslms/a-plus/pull/570/commits/458e4b11841b3b55a1042dedba903d6dd22508f3

news/cache.py
```diff
@@ -23,19 +22,14 @@ def _generate_data(self, instance, data=None):
                 'title': item.title,
                 'body': item.body,
                 'pin': item.pin,
-                'alert': item.alert,
             }
-            if item.pin and item.alert:
-                alerts.append(entry)
-            else:
-                news.append(entry)
+            news.append(entry)
```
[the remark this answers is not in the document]
saarasat: Has been reshaped: https://github.com/apluslms/a-plus/pull/570/commits/458e4b11841b3b55a1042dedba903d6dd22508f3

news/templates/news/user_news.html
```diff
@@ -1,46 +1,38 @@
...
 {% if news %}
 <div class="panel panel-default news-panel">
   <div class="panel-heading">
-    <h4 class="panel-title">{% trans "Course news" %}</h4>
+    <h3 class="panel-title">{% trans "Course news" %}</h3>
   </div>
   <div class="list-group">
     {% for item in news %}
-    <a class="list-group-item{% if item.collapsed %} collapsed{% endif %}" href="#news{{ item.id }}" data-toggle="collapse" role="button" aria-controls="news{{ item.id }}" aria-expanded="{% if item.collapsed %}false{% else %}true{% endif %}">
+    <div class="{% if item.pin %}pinned-{% endif %}list-group-item">
```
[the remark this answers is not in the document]
saarasat: I think that since also the pinned items fall under the bootstrap-list-group, it is logical that the pinned items are list-group-items as well. The pinned items share a lot of visual aspects with the unpinned items, so now they just require a few lines of css to be separated. It has also the benefit, that it does not add a lot of code in the html-template.

news/templates/news/user_news.html
saarasat: I thought about that but since you actually said that there is no need to emphasize the pinning for students, I left the pin out for them :D But I added it now for students as well, I think it is nice for them too :)

news/templates/news/user_news.html
saarasat: I tried to simplify this a bit. The  this version allowed me to remove also the for-loop from news.py and still works for all the different course index -options:
https://github.com/apluslms/a-plus/pull/570/commits/03b9cecbba023f49826f94f3460bcaab1e449c2c

news/templatetags/news.py
saarasat: This was also simplified here: https://github.com/apluslms/a-plus/pull/570/commits/03b9cecbba023f49826f94f3460bcaab1e449c2c.

news/templatetags/news.py
saarasat: Aaand done: 
https://github.com/apluslms/a-plus/pull/570/commits/458e4b11841b3b55a1042dedba903d6dd22508f3

saarasat:
Did the minor changes @raphendyr proposed and fixed the indendation of the user_news.html now.

news/templates/news/user_news.html
```diff
@@ -3,37 +3,39 @@
...
-      </div>
-    </div>
-    {% if item.begin_more %}
-    <a class="folding-list-group-item collapsed" href="#more-news" data-toggle="collapse" role="button" aria-controls="more-news" aria-expanded="false">
-        {% trans "Show older" %} <span class="caret"></span>
-    </a>
-    <div class="collapse" id="more-news">
-    {% endif %}
-    {% endfor %}
-    {% if more > 0 and news|length > more %}
```
raphendyr, line 42: I think this was lost. Was it supposed to be dropped? Does the new layout produce correct divs, i.e. the closing tag too?

saarasat: Yes. Actually I noticed that there was an error before I touched that logic, when there was exactly 3 news items. The code produced the "see more"-button, but due to the structure of the for-loop, then there were no news to be folded and the closing div-tag was not produced at all. So this should change should remove that old error and additionally simplify the logic. I tried with all the three different index-options and with 2, 3 or 4 items, and everything looked good :)

news/templates/news/user_news.html
```diff
@@ -1,47 +1,41 @@
...
+					</h4>
+					<div class="list-group-item-text">
+						{{ item.body|safe }}
+					</div>
+				</div>
+				{% if forloop.counter == more and news|length > more %}
+					<a class="folding-list-group-item collapsed" href="#more-news" data-toggle="collapse" role="button" aria-controls="more-news" aria-expanded="false">
+						{% trans "Show older" %} <span class="caret"></span>
+					</a>
+					<div class="collapse" id="more-news">
```
raphendyr, line 35: Could you add comments for this start tag and for the matching end tag, so the reader can find these tags. For example `{# start of div.collapse for news #}` and `</div>{# end of div.collapse for news #}`.

saarasat: Yes, a very good idea! Since that part was not very self-evident. Those have now been added.

news/templates/news/user_news.html
```diff
@@ -1,47 +1,41 @@
...
+		</div>
+		<div class="list-group">
+			{% for item in news %}
+				<div class="{% if item.pin %}pinned-{% endif %}list-group-item">
+					<h4 class="list-group-item-heading">
+						{{ item.title|safe }}
+						{% if not item|is_published:now %}
+							<small class="future-instance">{% trans "Published on" %}: {{ item.publish }}</small>
+						{% else %}
+							<small class="current-instance">{{ item.publish }}</small>
```
mtthwhggns, line 18: Would it make sense to use the slightly more semantic `<date>` element here?

news/templates/news/user_news.html
```diff
@@ -1,47 +1,41 @@
...
+						{% endif %}
+						{% if is_course_staff %}
+							<small><span class="label label-primary">{{ item.audience|news_audience }}</span></small>
+						{% endif %}
+						{% if item.pin %}
+							<small><span class="glyphicon glyphicon-pushpin"></span></small>
```
mtthwhggns, line 24: Since this is the top (and implied, most recent) item, I'm concerned that there's no way for screen reader users to know it's pinned. I can imagine someone listening to the title/date of the pinned item, and thinking "I've seen this before and it's the newest, so I can skip past the entire news block".

How about adding a screen reader specific note alongside the pin icon?

mtthwhggns, line 24: Adding a separate comment to the same line to the discussions get tracked separately.

VoiceOver tries to read the icon - (assuming the other point is taken care of) should it be hidden from screen reader users? 

Screenshot 2020-06-29 at 15 44 53

news/templates/news/user_news.html
```diff
@@ -1,47 +1,41 @@
...
-  </div>
-</div>
+	<div class="panel panel-default news-panel">
+		<div class="panel-heading">
+			<h3 class="panel-title">{% trans "Course news" %}</h3>
+		</div>
+		<div class="list-group">
+			{% for item in news %}
+				<div class="{% if item.pin %}pinned-{% endif %}list-group-item">
+					<h4 class="list-group-item-heading">
```
mtthwhggns, line 13: I'm on the fence about whether it makes sense for the `h4` element to include both the headline and the date. If you do the thing mentioned on line 18, maybe you could try moving it outside of the heading element?

Mankro:
There were old questions about the **HTML text fields** and using a WYSIWYG editor instead of a plain text field. There are many other HTML text fields that teachers see in different forms. Improving the usability of those HTML fields is another issue, **not part of this PR**. However, I don't think that teachers are concerned about writing HTML into plain text fields since they do it very rarely or the content is very short. Thus, it is a very minor issue and we don't need to prioritize it now. The biggest concern is probably the manual assistant feedback in the inspect submission page, but the usability of writing manual feedback in A+ is awful for many other reasons as well (it is slow to open submissions individually and hard to find the most relevant submission for giving feedback).

saarasat:
> Hi @saarasat, I did a review specifically testing keyboard accessibility and using VoiceOver on the area.
> 
> Generally, this is a big improvement.
> 
> I've just flagged a couple of small things you might want to consider before this gets merged. Feel free to ask me if you need any more explanation.

Thanks for all these comments. I think they were valid and doable to improve the experience from the screen reader -point of view :)

news/templates/news/user_news.html
saarasat: Done

news/templates/news/user_news.html
saarasat: I added a sr-only-comment there, actually already before the date, so that it should be clear that the item is pinned. I have not added those sr-onlys before, so if there is something off with it, please let me know :)

saarasat: And done

news/templates/news/user_news.html
saarasat: I moved it outside the h4-tag, while trying to keep the css minimum and classes still explanatory

saarasat:
@jaguarfi I guess you said on Friday, that you are ok with these changes now in the pr? If so, could you check it approved? And if there is something you'd like me to check, just please let me know!

jaguarfi:
@saarasat is it only me or is this happening to you?
image

news/templates/news/user_news.html
```diff
@@ -1,47 +1,48 @@
...
+					</p>
+				</div>
+				{% if forloop.counter == more and news|length > more %}
+					<a class="folding-list-group-item collapsed" href="#more-news" data-toggle="collapse" role="button" aria-controls="more-news" aria-expanded="false">
+						{% trans "Show older" %} <span class="caret"></span>
+					</a>
+					<div class="collapse" id="more-news">{# start of div.collapse for news #}
+				{% endif %}
+			{% endfor %}
+			</div>{# end of div.collapse for news #}
```
raphendyr, line 45: When adding of `<div>` is condiftional, then `</div>` has to be too or otherwise the html layout breaks. This is basically the reason why these should be two different lists so things like this wouldn't happen. @jaguarfi's screenshot should be result of this.

saarasat: Weird that it did not come up earlier, since Matthew also did his screenreader test after that change and neither of us noticed anything. I wonder if I made a mistake somewhere yesterday making those screen reader -fixes. Let me check and go back in time a bit. Sorry for the extra hassle.

saarasat: So looks like there was some error earlier in the simplifying commit https://github.com/apluslms/a-plus/pull/570/commits/cb16f33f109e176150f16461f98b67c277c4fc68. 
I am not sure what happened, why Matthew or I did not spot it, maybe my testing-course did not have enough other content (and long enough toc for it to skip on the right-hand-side). But never the less, of course it needs to be fixed.
 
Here in the current master there is the conditional div which I mentioned earlier, that also breaks the html if there are exactly three news items. https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/news/templates/news/user_news.html#L35

This is how three news items looks with the current master:

image

What do you think, should I do revert to time before that simplifying commit and implement the two separate lists then? even though like @raphendyr said that they are slow? Personally I also think that for readibility that might be better. 

Or should I remove that https://github.com/apluslms/a-plus/commit/cb16f33f109e176150f16461f98b67c277c4fc68 -commit and just try to find a work around for the situation where there are exactly three items?

raphendyr: I think the first if requires just `forloop.counter == more` and second div should be fine with `news|length > more`. Just need to check that the index go right, but I think it should.

saarasat: That did not actually quite do it. Since if the "more" was set to 0, as is the case for instance with the "last visited content"-index option, no folding is present. Also had to take care that the ending-div wasn't printed with each round but only for the final item. But thanks for the tip, it gave me another idea :) 

The fix has now been implemented in the final commit https://github.com/apluslms/a-plus/pull/570/commits/f2eada6ca0039192cda99e754e9daf2ccabe2076, it is probably not the prettiest, but I commented and tested it and at least it fixes the three-item-thing at the same time:

**More than three items**
image

**And three items**
image

news/templates/news/user_news.html
```diff
@@ -35,14 +35,17 @@ <h4 class="list-group-item-title">
...
-				{% if forloop.counter == more and news|length > more %}
+				{# If a limit for folding news (more) is set, this creates the folding section #}
+				{% if forloop.counter == more and news|length > more and more > 0 %}
 					<a class="folding-list-group-item collapsed" href="#more-news" data-toggle="collapse" role="button" aria-controls="more-news" aria-expanded="false">
 						{% trans "Show older" %} <span class="caret"></span>
 					</a>
-					<div class="collapse" id="more-news">{# start of div.collapse for news #}
+					<div class="collapse" id="more-news">{# start of collapse.div #}
+				{% endif %}
+				{% if forloop.counter == news|length and news|length > more and more > 0 %}
```
raphendyr, line 45: First part might be easier to comprehend as `if forloop.last and ...`

news/migrations/0002_remove_news_alert.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('news', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
raphendyr, line 13: Should all the old news which are alerts be marked as pinned? Or should we handle alerts someway in the migration?

news/templates/news/user_news.html
raphendyr: The old code didn't check if the entry which starts more is the last entry in the list. New code includes `news|length > more`, which should do that.

news/templates/news/user_news.html
```diff
@@ -1,47 +1,51 @@
...
+								<span class="glyphicon glyphicon-pushpin" aria-hidden="true"></span>
+							{% endif %}
+						</div>
+					</div>
+					<p class="list-group-item-text">
+						{{ item.body|safe }}
+					</p>
+				</div>
+				{# If a limit for folding news (more) is set, this creates the folding section #}
+				{% if forloop.counter == more and news|length > more and more > 0 %}
```
raphendyr, line 39: I think this code shows 4 news items when more is set to 3. I think this element should be before the news item. Thus, when index (`forloop.counter`) is 3, then there has been indexes 0, 1, and 2 (i.e. 3 news items). Technically it doesn't matter so much, as more can be selected correctly. Though, I didn't yet test this and I'm not sure how moving this works with the `news|lenght > more` test.

assets/css/main.css
```diff
@@ -403,11 +403,72 @@ html, body {
...
 }
-.news-panel .list-group-item-text {
-  padding: 10px;
+
+/* course news */
+.news-panel {
+    margin-bottom: 30px;
+}
+
+.future-instance {
```
raphendyr, line 412: Should this be `.news-panel .future-instance {` to ensure it doesn't affect wrong places (future instance is quite generic name). In addition, it would make it more clear, where this class typically exists.

assets/css/main.css
```diff
@@ -403,11 +403,72 @@ html, body {
...
+    color: #c0322d;
+    background-color: #fff;
+}
+
+.current-instance {
```
raphendyr, line 417: Same `.news-panel .current-instance` thing as above.

news/templates/news/user_news.html
saarasat: Hmm.. I am not quite sure what you mean? Currently the more is set to 3 in the course.html (that has not been changed) and above the folding three items are shown.

So in that sense the limit for folding works like before, just without the for-loop in the news.

news/migrations/0002_remove_news_alert.py
saarasat: Currently, if the news item has the "alert-field" set but "pin" is false, the appearance on the course front page is not changed. It was actually already referred to earlier: 

> 1. When the user creates a new "news" element, there is a selector called Alert, which asks for the Alert style. However, the selected style is not applied to the unpinned items. Do you think we should apply the class also to unpinned news? or maybe remove the selector when the "Pin" checkbox is unselected?
>    image

So I guess those that actually are wanted to show as more noticeable for students via the current alert, already have that "pin" luckily in place :) 

Maybe we could check the database how many courses have news that would be affected by this? I think that would help in deciding if any further actions are needed, if the "pin" is already in place for those that actually use the alerts.

assets/css/main.css
saarasat: I added these for clarification

assets/css/main.css
saarasat: Same as above

news/templates/news/user_news.html
raphendyr: I see.. I always forget that `forloop.counter` starts from 1, thus this works correctly. Alternatively, the test could be before the news item and be `forloop.counter0 == more and more > 0` (no need to check length of the news as that would be the first entry whcih goes within the collapsed section. Current solution is totally fine :)

news/migrations/0002_remove_news_alert.py
raphendyr: > So I guess those that actually are wanted to show as more noticeable for students via the current alert, already have that "pin" luckily in place :)

Makes sense, thus this should be fine. I forgot/missed that part earlier, sorry.

Currently pinned + alerted news are:

```
test/vartiah4/
    2017-03-06 12:44:12+00:00 Diibadaaba
a1141/2018/
    2019-09-03 06:04:18+00:00 Kurssi päättyy
a1141/2019/
    2020-06-08 10:25:35+00:00 Kurssi on päättynyt
a1141/2020
    2020-06-08 10:26:33+00:00 Tietorakenteet ja algoritmit Y
o1/2016/
    2016-12-21 10:16:04+00:00 Epävirallinen suoritustodistus
o1/2017/
    2018-08-24 10:53:30+00:00 Tämä on vuoden 2017 kurssimateriaali / This is an old copy of the course material
    2018-01-09 16:55:03+00:00 Epävirallinen suoritustodistus
o1/2018/
    2019-09-09 10:57:09+00:00 Tämä on vanha versio kurssimateriaalista / This is an old copy of the course material
tietotekniikka-sovelluksissa/2020/
    2020-03-15 17:30:21+00:00 Harjoitustilaisuudet 16.3. alkaen
wsd/2017-2018/
    2018-03-12 10:25:16+00:00 Lecturer is on sick leave - Team Demos on Monday and Tuesday have to be rescheduled
    2018-03-08 08:02:07+00:00 Project demo times available for reservation
```

I think all of these are fine as pinned, which they are when alert is dropped. Looks good :+1:

news/templates/news/user_news.html
saarasat: :D

news/templates/news/user_news.html
```diff
@@ -1,47 +1,51 @@
...
+							{{ item.title|safe }}
+						</h4>
+						<div class="list-group-item-details">
+							{% if not item|is_published:now %}
+								<span class="future-instance">{% trans "Published on" %}: <time>{{ item.publish }}</time></span>
+							{% else %}
+								<time class="current-instance">{{ item.publish }}</time>
+							{% endif %}
+							{% if is_course_staff %}
+								<span class="label label-primary">{{ item.audience|news_audience }}</span>
```
raphendyr, line 27: I'm not sure which is better `primary` or `default` for the label color. I guess this is fine.

jaguarfi:
I think the changes have been iterated enough and therefore these are quite good and well thought out. However, since you are removing the A+ alert completely, I think these changes will affect some courses. So, is there any way to make these changes compatible with the reverse? But, I think @raphendyr and @Mankro should tell you if there is a need to make these changes backwards compatible.

Mankro:
I won't look into this PR in detail due to the holidays. You can merge this without me. I think @raphendyr already said that very few courses in production have used the news with the alert option. I didn't participate in the discussion when you decided to remove the alert feature from the news, but I assume you have good reasons and I am not against it. The news feature has been underutilized and not all teachers really know about it. To my understanding, this PR has taken the good parts out of Juha's wishes and added some more too. The pinning of the important news sounds very good.

Since hardly anyone has used the alert news, I doubt anybody cares that the feature is removed and I guess the alerts are replaced by pinning, which sounds good.

saarasat:
> I think the changes have been iterated enough and therefore these are quite good and well thought out. However, since you are removing the A+ alert completely, I think these changes will affect some courses. So, is there any way to make these changes compatible with the reverse? But, I think @raphendyr and @Mankro should tell you if there is a need to make these changes backwards compatible.

Good that you made sure. There is only a handful of those cases where the alert is being used and they were discussed above, on Friday the 3rd, with @raphendyr :

> I think all of these are fine as pinned, which they are when alert is dropped.

And many of those seem to be alerts that the course has ended and/or the material is of an old version of the course. That notification will be automated with this issue in any case :) 
https://github.com/apluslms/a-plus/issues/554

raphendyr:
> However, since you are removing the A+ alert completely, I think these changes will affect some courses. So, is there any way to make these changes compatible with the reverse? But, I think @raphendyr and @Mankro should tell you if there is a need to make these changes backwards compatible.

The alerts, which are currently used in plus.cs.aalto.fi, are listed in https://github.com/apluslms/a-plus/pull/570#discussion_r448977811. Based on those, I don't see any reason to warn teachers. Alerts are mostly used to alert about past courses, which requires a better solution (see #554). Those and rest of the alerts should be fine as pinned news.

saarasat:
Just let me know then if the commits can be squashed for merging, or something else is needed here :)

jaguarfi:
Of course, the previous implementation of the news did not consider any aspect of accessibility. Nevertheless, (accidentally) it was possible to navigate each news item because they were collapsible elements.

saarasat:
> However, I have one more concern related to keyboard accessibility. Previous to your changes, all the news titles were accessible through the keyboard, but now, the news have been left out. You may have to fix that before merging this PR.

Regarding that, if I am not mistaken, the keyboard accessibility is an issue of the interactive elements of a webpage, thus links, buttons, text fields? The parts that user navigates with. As least this would suggest so: https://webaim.org/techniques/keyboard/#intro. Since those news items no longer are interactive, and you cannot click them via mouse and cursor either, do they need to be accessible with keyboard?

news/cache.py
```diff
@@ -13,29 +13,23 @@ def __init__(self, course_instance):
...
                 'title': item.title,
                 'body': item.body,
                 'pin': item.pin,
-                'alert': item.alert,
             }
-            if item.pin and item.alert:
-                alerts.append(entry)
-            else:
-                news.append(entry)
+            for item in instance.news.all()
```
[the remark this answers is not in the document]
saarasat: The way the items were set was actually written a little bit different before by the previous developer, but I changed it to this for-format after the discussion above: https://github.com/apluslms/a-plus/pull/570#discussion_r446785968

So the previous version worked, but this should be a little more concise way to write it :)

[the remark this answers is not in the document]
jaguarfi: :thinking: I see, that is a clever way of doing it :)

jaguarfi:
yes, you are right about the keyboard navigation.

news/migrations/0002_remove_news_alert.py
jaguarfi: Not many courses are affected by the removal of the alerts, but in any case, those courses will stop working because Django won't find the field `alert `in the table news. am I right? or how are those courses going to respond?

I understand that once this PR is merged those courses will be "past courses", and not many people will use them, but I wonder if we want to have those broken courses around.

raphendyr: > those courses will stop working because Django won't find the field alert in the table news. am I right? or how are those courses going to respond?

Incorrect, the "alert" news will be just pinned news for them. The migration just drops the field but the code changes at the same time. all courses use the same django code.

jaguarfi:
@saarasat can you squash the commits?. Feel free to choose the most logical and relevant commit messages.

saarasat:
Commits squashed, tests run, and changes checked in a container and locally after squasing.

saarasat:
_saarasat force-pushed the saarasat:course-news branch from 8b7648f to 4d18f93 8 minutes ago_

Reworded some commit-messages per request. Nothing else in commits changed.

PR #575, opened 2020-06-12
Compress HTML pages in cache

### LINKED ISSUE #510: Compress HTML pages in the cache
Exercise and chapter pages are stored in the cache and only update if the backend reports that there is a change. Some pages might be large (e.g. 1M), but do compress relatively well. Memcached API supports compression on the fly, but that is not usable over django API.

Thus, we should at least compress HTML content manually. Alternatively, we can specialize `CachedAbstract` for memcached, which would also allow us to use `cas` operation.

Relevant files:
* `lib/cache/cached.py`
* `exercise/cache/exercise.py` (`content` in `_generate_data(...)` and `content()`)

veeraalt:
# Description

- Exercise and chapter pages are stored in the cache and only update if the backend reports that there is a change. Some pages might be too large to fit in the cache (e.g. 1M), but do compress relatively well. Memcached API supports compression on the fly, but that is not usable over django API. This pull request adds lz4 compressor (https://pypi.org/project/lz4/) and zlib as backup (https://docs.python.org/3/library/zlib.html) to compress HTML exercise and chapter pages so that they can be stored in the cache, and decompress them before viewing.
- Compress and decompress functions can be changed easily by modifying the corresponding global variables in exercise.py.

Fixes #510 

# Testing

- The existing unit tests pass, I have not added new unit tests.
- Testing can be done manually by checking that the exercise and chapter pages load correctly, and by making these changes to settings.py file:
  - Move CACHES block into end of the file
  - Add into LOGGING/logger: 
```
  'aplus.cached': {
        'level': 'DEBUG',
    },
```
   Log shows now cache events, and it shouldn't say anywhere "Failed to store a value to the cache. It might be too big!".

- I don't think that documentation needs to be updated.

### DIFF
--- a/exercise/cache/exercise.py
+++ b/exercise/cache/exercise.py
@@ -1,10 +1,24 @@
+import logging
 import time
+
 from django.conf import settings
 from django.db.models.signals import post_save, post_delete
 
 from lib.cache import CachedAbstract
 from lib.remote_page import RemotePageNotModified
 from ..protocol.aplus import load_exercise_page
+
+logger = logging.getLogger('aplus.cached')
+
+try:
+    from lz4.block import compress as _compress, decompress
+    def compress(data):
+        return _compress(data, compression=1)
+except ImportError:
+    logger.warning("Unable to import lz4, using a slower zlib instead")
+    from zlib import compress as _compress, decompress
+    def compress(data):
+        return _compress(data, level=1)
 
 
 class ExerciseCache(CachedAbstract):
@@ -26,9 +40,12 @@ class ExerciseCache(CachedAbstract):
                 *self.load_args,
                 last_modified=data['last_modified'] if data else None
             )
+
+            content = compress(page.content.encode('utf-8'))
+
             return {
                 'head': page.head,
-                'content': page.content,
+                'content': content,
                 'last_modified': page.last_modified,
                 'expires': page.expires if page.is_loaded else 0,
             }
@@ -41,7 +58,8 @@ class ExerciseCache(CachedAbstract):
         return self.data['head']
 
     def content(self):
-        return self.data['content']
+        content = decompress(self.data['content']).decode('utf-8')
+        return content
 
 
 def invalidate_instance(instance):
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,6 +13,7 @@ git+https://github.com/Aalto-LeTech/django-colortag.git@2.1.1#egg=django-colorta
 feedparser~=5.2.1
 html5lib
 lxml ~= 4.3.2
+#lz4==3.1.0
 cachetools~=3.1.0
 icalendar~=4.0.3
 mimeparse~=0.1.3
--- a/doc/DEPLOYMENT.md
+++ b/doc/DEPLOYMENT.md
@@ -43,7 +43,7 @@ Common system configuration
           postgresql memcached \
           uwsgi-core uwsgi-plugin-python3 \
           python3-virtualenv \
-          python3-psycopg2 python3-certifi python3-reportlab python3-reportlab-accel
+          python3-certifi python3-lz4 python3-psycopg2 python3-reportlab python3-reportlab-accel
 
  1. Create a new user for a-plus
 


veeraalt:
@raphendyr Any thoughts about needed unit tests for this change?

jaguarfi:
Do you have examples on how much the size was reduced?

veeraalt:
> Do you have examples on how much the size was reduced?

Yes, for example some exercise pages with sys.getsizeof()
 
size_ch2_6_1
size_ch2_6_3
size_ch2_6_2

I think these are in bytes

exercise/cache/exercise.py
```diff
@@ -1,4 +1,4 @@
-import time
+import time, lzma
```
raphendyr, line 1: Could you split these to their own lines, see https://www.python.org/dev/peps/pep-0008/#imports

raphendyr:
> Yes, for example some exercise pages with sys.getsizeof()

Can you test the sizes with python module gzip too. In addition, if you can find a way to compare the compression speed between the libs, that would be cool too.

exercise/cache/exercise.py
```diff
@@ -26,6 +28,9 @@ def _generate_data(self, exercise, data=None):
                 *self.load_args,
                 last_modified=data['last_modified'] if data else None
             )
+            
+            page.content = self.compressor(page.content.encode('utf-8'))
```
raphendyr, line 32: I don't think it's a good idea to overwrite the value in the `page` object. As the object is destroyed in the end of this block, I think it would be fine to just use temporary variable, e.g., just `content = ..`.

exercise/cache/exercise.py
```diff
@@ -14,6 +14,8 @@ class ExerciseCache(CachedAbstract):
     def __init__(self, exercise, language, request, students, url_name):
         self.exercise = exercise
         self.load_args = [language, request, students, url_name]
+        self.compressor = lzma.compress
```
raphendyr, line 17: Because these variables shouldn't be defined for specific exercise, I would move them to the module globals or I would set them as class attributes. For the former, you could do `from lzma import compress as compressor, decompress as decompressor`. For the latter:

```
class ExerciseCache(CachedAbstract):
  compressor = lzma.compress
  decompressor = lzma.decompress

  def content(self):
    return self.decompressor(self.data['content']).decode('utf-8')
```

raphendyr:
> Any thoughts about needed unit tests for this change?

Can you spend a day next week to look into if you can come up with ideas how to write test for these. We can then discuss how huge effort it would be to actually implement something that would matter.

veeraalt:
> Can you test the sizes with python module gzip too. In addition, if you can find a way to compare the compression speed between the libs, that would be cool too.

Same pages with gzip, not much difference:
gzip_ch2_6_1
gzip_ch2_6_3
gzip_ch2_6_2

I will add the speed comparisons later.

veeraalt:
> Can you spend a day next week to look into if you can come up with ideas how to write test for these. We can then discuss how huge effort it would be to actually implement something that would matter.

Sure!

exercise/cache/exercise.py
```diff
@@ -1,10 +1,11 @@
-import time, lzma
+import time
 from django.conf import settings
 from django.db.models.signals import post_save, post_delete
 
 from lib.cache import CachedAbstract
 from lib.remote_page import RemotePageNotModified
 from ..protocol.aplus import load_exercise_page
+from lzma import compress as compressor, decompress as decompressor
```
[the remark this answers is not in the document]
veeraalt: Good to know, thanks! I'll change that.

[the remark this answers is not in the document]
veeraalt: So to my understanding the line should be after ```import time``` and there should also be a blank line before django imports because django is not a standard module, is that correct?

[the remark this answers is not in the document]
raphendyr: Yes. The correct order and newlines are not used in many places, but that should be fixed when we edit the imports. I also forgot to mention that all groups should be sorted alphabeticall e.g.:

```
import os
from lzma import ..
from time import time
from zlib import ...
```

veeraalt:
>  Can you test the sizes with python module gzip/zlib too. In addition, if you can find a way to compare the compression speed between the libs, that would be cool too.

> In addition, you could try few different preset values (lower value is faster, but compresses less, though that could be fine for this usecase).

I had some problems with installing pylzma and snappy but here is some comparisons of included libs and lz4:

```
             size           time (s)          speed (MB/s)
original     217800
preset: 1
lzma         23765          0.008538          22.72
gzip         29989          0.001942          96.7
zlib         29977          0.001711          109.8
bz2          21900          0.011883          16.49
lz4          38551          0.000357          501.55
preset: 3
lzma         23097          0.011997          16.23
gzip         27720          0.001515          125.43
zlib         27708          0.001469          129.43
bz2          20666          0.009463          20.83
lz4          30380          0.001176          159.35
preset: 6
lzma         21997          0.030825          6.35
gzip         24673          0.002795          69.09
zlib         24661          0.002967          65.1
bz2          20666          0.009844          20.03
lz4          29602          0.001747          107.73
preset: 9
lzma         21997          0.054189          3.61
gzip         24464          0.003807          50.79
zlib         24452          0.004276          45.22
bz2          20666          0.009616          20.5
lz4          29525          0.002219          84.84
```

raphendyr:
@veeraalt that looks nice. Thanks!

I think we should use lz4 with preset 1. It's good enough and wouldn't slow the service down. As it's harder to install, the code could fallback to zlib if lz4 is missing. To do that:

```python
import logging
# other imports...

logger = logging.getLogger('aplus.' + __name__)

try:
  from lz4.blocks import compress, decompress
except ImportError:
  logger.warning("Unable to import lz4, using a slower zlib instead")
  from zlib import compress, decompress
```

veeraalt:
> I think we should use lz4 with preset 1.

@raphendyr Sounds good. Should the lz4 be added to ```requirements```?

raphendyr:
> @raphendyr Sounds good. Should the lz4 be added to requirements.txt?

No, because it's an optional requirement. However, it must be documented "somewhere".

I think you should:

* add ``python3-lz4` to the line containing `python3-psycopg2 python3-certifi python3-reportlab python3-reportlab-accel` (line 46) in https://github.com/apluslms/a-plus/blob/master/doc/DEPLOYMENT.md  (sort the words inside the line, so they are in order)
Installation quide would be the best.
* you could add a commented entry in the requirements.txt, including the version that you have testetd this with
* can you create an issue in https://github.com/apluslms/run-aplus-front/issues so we remember to add this to the test container too.

I think those are the best locations for the information currently. I would prefer few other things to be optional requirements too (specially things that require binary files like lz4).

veeraalt:
@raphendyr I did those changes, but

> Installation quide would be the best.

Do you mean I should write some instructions on how to install lz4, in DEPLOYMENT.md or where?

> can you create an issue in https://github.com/apluslms/run-aplus-front/issues 

Will do that too

exercise/cache/exercise.py
```diff
@@ -28,8 +36,11 @@ def _generate_data(self, exercise, data=None):
                 *self.load_args,
                 last_modified=data['last_modified'] if data else None
             )
-            
-            content = compressor(page.content.encode('utf-8'))
+
+            try:
+                content = compress(page.content.encode('utf-8'), compression=1)
```
raphendyr, line 41: Hmm, apparently they have a different APIs... That should be handled when defining the functions. Currently, it's not clear if compress raises a TypeError for some other reason.

You can use `functools.partial` or define the functions your self. E.g.:

```python
try:
    from lz4.block import compress as _compress, decompress
    def compress(data):
        return _compress(data, compression=1)
except ImportError:
    logger.warning("Unable to import lz4, using a slower zlib instead")
    from zlib import compress as _compress, decompress
    def compress(data):
        return _compress(data, level=1)
```

requirements.txt
```diff
@@ -13,6 +13,7 @@ git+https://github.com/Aalto-LeTech/django-colortag.git@2.1.1#egg=django-colorta
 feedparser~=5.2.1
 html5lib
 lxml ~= 4.3.2
+#lz4~=3.1.0
```
raphendyr, line 16: Spaces make these a bit more readable (see lxml above) and make it more clear that it's not exacpt match. Apparently many others don't have space either (which is probably the reason I hadn't noitced those before). Sadly, I didn't find any discussion about this in the internet.

I'll mark this to be added to the style guide too.

raphendyr:
> Do you mean I should write some instructions on how to install lz4, in DEPLOYMENT.md or where?

I think it was left there from editing the comment.. The adding of lz4 to deployment.md was that part.

requirements.txt
veeraalt: Oops, it should be == and not ~=, is spaces used in that case also?

raphendyr: The `~=` is pretty good in our kind of use. Requirements file is use more for requirements and less as a lock (see package-lock.json with JavaScript npm and Gemfile.lock with Ruby bundler). Because it's a comment, both version requirements are fine.

Most of the requirements have `==` and no spaces, as the file has been created with `pip freeze`, I assume.

Issue #431, opened 2020-01-17
LTI: use launch return url for reauthentication


raphendyr:
Goal: we want to be able to redirect from tool provider urls back to A+ for re-authentication (session expired or link in telegram). Tool provider will then forward to presentation return url with some information, where A+ should forward the authenticated request.

e.g.

1. open tool provider `tp/item1` -> `a+/return_url?lti_next=tp/item1`
2. login in A+, `a+/return_url` -> `tp/lti_login?lti_next?tp/item1`
3. forward to `tp/item1`

PasiSa:
Removing high priority label for no activity in more than a year, and tagged for discussion

PR #576, opened 2020-06-15
Replace get_host() with BASE_URL

### LINKED ISSUE #422: LTI: remove request.host from LTI parameters
Remove `request.get_host()` from all LTI parameter calculations. Use `settings.BASE_URL` instead. `request.get_host` comes from the browser, thus user can use it to manipulate the data send as part of the post.

These related issues can be fixed in connection or should be created as another issues:

- [ ] `tool_consumer_instance_guid` this field doesn't seem to include relevant information. it should be created from settings.BASE_URL
- [ ] `launch_presentation_return_url` this should include link to the page in A+, which will render the launch button (basically to the page this link is rendered at).
- [ ] Interesting second part is to include or keep query parameters, so we can request reauthentication by the tool provider (e.g. koodisäilö), which would be sending browser to `{launch_presentation_return_url}?tc_return_url=<ulr>` or such. TODO: raphendyr needs to write down better explanation (i.e. create an issue).

HenrikSundell:
# Description

Remove all uses of `request.get_host()` and use `settings.BASE_URL` instead. 
This is safer because BASE_URL is defined in the settings while get_host() comes from the users' browser which allows it to be manipulated by the user.

Fixes #422

# Testing
All the URLs that are now calculated using `BASE_URL` have been checked to render the same result as before with `get_host()`.

Not relevant

### DIFF
--- a/external_services/lti.py
+++ b/external_services/lti.py
@@ -1,4 +1,6 @@
 from hashlib import md5
+from urllib.parse import urlsplit, urljoin
+
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.utils.translation import get_language
@@ -19,8 +21,9 @@ class LTIRequest(object):
     def __init__(self, service, user, instance, request, title, context_id=None, link_id=None, add=None, exercise=None):
         self.service = service
         course = instance.course
+        base_url_parts = urlsplit(settings.BASE_URL)
         # Context and resource parameters.
-        context_id = context_id or (request.get_host() + instance.get_absolute_url())
+        context_id = context_id or (base_url_parts.netloc + instance.get_absolute_url())
         link_id = link_id or "aplus{:d}".format(service.pk)
         title = title or link_id
 
@@ -60,9 +63,9 @@ class LTIRequest(object):
             "launch_presentation_locale": get_language(),
             "launch_presentation_document_target":
                 "iframe" if exercise and exercise.open_in_iframe else "window",
-            "launch_presentation_return_url": request.scheme + '://' + request.get_host() + instance.get_absolute_url(),
+            "launch_presentation_return_url": urljoin(settings.BASE_URL, instance.get_absolute_url()),
 
-            "tool_consumer_instance_guid": request.get_host() + "/aplus",
+            "tool_consumer_instance_guid": base_url_parts.netloc + "/aplus",
             "tool_consumer_instance_name": "A+ LMS",
         })
 


external_services/lti.py
```diff
@@ -8,6 +8,7 @@
 from oauthlib.oauth1 import Client, SIGNATURE_HMAC, SIGNATURE_TYPE_BODY, \
     SIGNATURE_TYPE_QUERY
 import json
+from urllib.parse import urlparse
```
raphendyr, line 11: The other method, `urlsplit`, is a bit more preferred one.

> This is similar to urlparse(), but does not split the params from the URL. This should generally be used instead of urlparse() if --

- src: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit

(You can see from [the source](https://github.com/python/cpython/blob/master/Lib/urllib/parse.py#L369) that urlparse uses urlsplit and just does more, which we don't need).

external_services/lti.py
```diff
@@ -19,8 +20,9 @@ class LTIRequest(object):
     def __init__(self, service, user, instance, request, title, context_id=None, link_id=None, add=None, exercise=None):
         self.service = service
         course = instance.course
+        parsed_url = urlparse(settings.BASE_URL)
         # Context and resource parameters.
-        context_id = context_id or (request.get_host() + instance.get_absolute_url())
+        context_id = context_id or (parsed_url.hostname + ":" + str(parsed_url.port) + instance.get_absolute_url())
```
raphendyr, line 25: I think `parsed_url.netloc` should be ok here? I think we can assume that ports 80 and 443 (defaults for http and https) always share the same site. However, are you ommitting a sub path (e.g. `http://example.com/aplus/`), or was that handled by the get_absolute_url()?

(again, nice source ref: https://github.com/python/cpython/blob/master/Lib/urllib/parse.py#L199)

external_services/lti.py
```diff
@@ -19,8 +20,9 @@ class LTIRequest(object):
     def __init__(self, service, user, instance, request, title, context_id=None, link_id=None, add=None, exercise=None):
         self.service = service
         course = instance.course
+        parsed_url = urlparse(settings.BASE_URL)
```
raphendyr, line 23: Name `parsed_url` doesn't indicate what this variable contains. For example, `parsed_base_url` or `base_url_parts` would indicate information about the content.

external_services/lti.py
```diff
@@ -60,9 +62,9 @@ def __init__(self, service, user, instance, request, title, context_id=None, lin
             "launch_presentation_locale": get_language(),
             "launch_presentation_document_target":
                 "iframe" if exercise and exercise.open_in_iframe else "window",
-            "launch_presentation_return_url": request.scheme + '://' + request.get_host() + instance.get_absolute_url(),
+            "launch_presentation_return_url": settings.BASE_URL + instance.get_absolute_url(),
```
raphendyr, line 65: You should consider using `urllib.parse.urljoin` for this or `parsed_url._replace(path=absolute_url)`. Both might yiedl more robust results (e.g. if there is or sin't slash in end of `BASE_URL`.

external_services/lti.py
```diff
@@ -60,9 +62,9 @@ def __init__(self, service, user, instance, request, title, context_id=None, lin
...
 
-            "tool_consumer_instance_guid": request.get_host() + "/aplus",
+            "tool_consumer_instance_guid": parsed_url.hostname + ":"+ str(parsed_url.port) + "/aplus",
```
raphendyr, line 67: This `parsed_url.hostname + ":"+ str(parsed_url.port)` (or alternative, if changed) should be calculated only once in the function.

lib/viewbase.py
```diff
@@ -74,7 +76,7 @@ def redirect_kwarg(self, kw, backup=None):
         return self.redirect(to, backup)
 
     def redirect(self, to, backup=None):
-        if not is_safe_url(url=to, allowed_hosts={self.request.get_host()}):
+        if not is_safe_url(url=to, allowed_hosts={urlparse(settings.BASE_URL).hostname + ":" + str(urlparse(settings.BASE_URL).port)}):
```
raphendyr, line 79: Actually, this redirect check _should_ use request.get_host(), as this is a string we are returning to the browser. We are just cheking that the browser is not going to leave the instances. Consider that the person is using url `https://aplus` to acccess `plus.cs.aalto.fi`, which is totally valid as long as they have `.cs.aalto.fi` in the dns search path.

shibboleth_login/views.py
```diff
@@ -61,8 +61,9 @@ def login(request):
         logger.debug("Shibboleth login: {}".format(user.username))
         
         redirect_to = request.GET.get(REDIRECT_FIELD_NAME, '')
+        allowed = urlparse(settings.BASE_URL).hostname + ":" + str(urlparse(settings.BASE_URL).port)
         if not is_safe_url(url=redirect_to,
-                           allowed_hosts={request.get_host()},
+                           allowed_hosts={allowed},
```
raphendyr, line 66: Same as in lib/viewbase.py

raphendyr:
I created continuation in #577 , I recommend to check it out. It can be worked on here or separately.

external_services/lti.py
HenrikSundell: Yeah, netloc seems very ok, I didn't notice such an option. About the subpath, I'm pretty sure it's handled by get_absolute_url() (works ok at least locally), so it should be good unless I'm missing something.

HenrikSundell:
@raphendyr  Everything you mentioned should be ok now, please let me know if there is still something to improve.

external_services/lti.py
raphendyr: Hmm.. Yeah. It (get_absolute_url) appears to handle this correctly as long as the system admin installs the application correctly.

external_services/lti.py
```diff
@@ -62,9 +62,9 @@ def __init__(self, service, user, instance, request, title, context_id=None, lin
             "launch_presentation_locale": get_language(),
             "launch_presentation_document_target":
                 "iframe" if exercise and exercise.open_in_iframe else "window",
-            "launch_presentation_return_url": settings.BASE_URL + instance.get_absolute_url(),
+            "launch_presentation_return_url": urljoin(settings.BASE_URL, instance.get_absolute_url()),
```
raphendyr, line 65: This is basically the same as the `netloc + absolute_url`, but it's fine as is. This will be worked on in #431

external_services/lti.py
```diff
@@ -8,6 +8,7 @@
 from oauthlib.oauth1 import Client, SIGNATURE_HMAC, SIGNATURE_TYPE_BODY, \
     SIGNATURE_TYPE_QUERY
 import json
+from urllib.parse import urlsplit, urljoin
```
raphendyr, line 11: Didn't notice before, but these should use the import ordering set in PEP 8: standard lib import, standard lib import from, 3rd party libs (import + from), project imports. Without fixing everything, you could add the new line in the start:

```python
from hashlib import md5
from urllib.parse import urlsplit, urljoin

from django.core.exceptions import PermissionDenied
from django.utils.translation import get_language
from rest_framework.reverse import reverse
from rest_framework.settings import api_settings
from oauthlib.common import urldecode
from oauthlib.oauth1 import Client, SIGNATURE_HMAC, SIGNATURE_TYPE_BODY, \
    SIGNATURE_TYPE_QUERY
import json
```

external_services/lti.py
HenrikSundell: Yes, except the scheme. I wanted to keep the scheme there since it was previously there.

raphendyr: > Yes, except the scheme. I wanted to keep the scheme there since it was previously there

Good point!

Issue #584, opened 2020-06-17
Upgrade the CSS framework


raphendyr:
Because bootstrap 3 classes are basically part of A+ assessment protocol, we can't just replaces it with another CSS framework. Instead, we need to first fix our first problem, which was not having prefixes for bootstrap 3 classes.

Second huge problem is, that we can't have bootstrap 3 and 4 classes at the same time. It might be enough, that we leave existing classes as is and just prefix the new framework. However, I think it would still be beneficial to cleanup the old classes first, so we clearly know, where they are used.

One challenge is that may of the classes are used in the feedback html stored in the database and there is no methods to migrate those. Thus, before we drop the old classes, we need to wait so long that none of the classes are in the database anymore.

Possible migration plan:

- [ ] Start using SASS for A+ and replace CDN bootstrap 3 with a locally build version.
- [ ] Create a second set of bootstrap 3 classes with a prefix (e.g. `bs3-` or `aplus1-` to match the assessment tool protocol)
- [ ] Migrate all assessment tools to use these prefixed classes (will take probably 1-2 years)

  - [ ] Update assessment protocol documentation
  - [ ] [MOOC-Grader](https://github.com/apluslms/mooc-grader)
  - [ ] [MOOC-Jutut](https://github.com/aalto-letech/mooc-jutut)
  - [ ] [Rubyric](https://github.com/aalto-letech/rubyric) (check if it uses any bootstrap classes)
  - [ ] [ACOS](https://github.com/acos-server/acos-server) (check if it uses bootstrap classes at all)
  - [ ] [Peer Review Platform](https://github.com/piehei/prplatform/)
  - [ ] [Grading Helper](https://github.com/eliisav/gradinghelper/)
  - TODO: other systems?

- [ ] Wait for a long time for all courses to rotate, so there isn't any old classes in the database (e.g. 3 years from last assessment tool update)
- [ ] Drop the default class names
- [ ] Add bs4 prefixed classes (assessment tools can start using bs4 cllasses now)

Adapt this plan, when better solution are drafted. For example, if bootstrap 3 and 4 can coexist better than I assume, then we can start using bs4- classes earlier (adapt the plan then).

Any case, starting to use SASS would be priority 1 to start this long process.

Mankro:
Is it necessary to start anything in this issue during the summer 2020? This is a huge task. Once we have new long-term employees (hopefully in the autumn), they could deal with these matters from start to finish.

raphendyr:
[...]

PR #585, opened 2020-06-22
Add labels for local user login form

### LINKED ISSUE #582: Maintenance/Local login does not have labels for username and password
Currently, the form is missing label fields, but that information is currently delivered via placeholders. However, that is not enough.

Solutions ideas:

* just add labels there. maybe placeholders should be removed?
* implement something like material design, and move the label away when field is focused (see [Material Design / Text Fields](https://material.io/components/text-fields) and [Material Design for Bootstrap / labels](https://fezvrasta.github.io/bootstrap-material-design/docs/4.0/material-design/labels/))

Image of the current state (note the warnings about missing labels):

screenshot of accessibility issues on login form

etanttila:
# Description
Add labels to local user login form to support accessibility and fulfill accessibility standards. By switching the login form to use the Django form, also error messages and error handling are provided automatically.
Also change color of the login button to make it look more like a button.

Fixes #582 

# Testing

All existing unit tests pass (except the one broken one)

Manual tests:
- Works with keyboard navigation
- Works both in English and Finnish
- Works with a screen reader

- [ ] ~I (developer) have created unit tests and the tests pass~
- [ ] ~I (developer) have created functional tests (Selenium tests) if applicable~

### DIFF
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -999,7 +999,6 @@ msgstr "Avaa 0 arvosanan diplomi"
 
 #: diploma/templates/diploma/list.html:26
 #: edit_course/templates/edit_course/signin_as_user.html:14
-#: userprofile/templates/userprofile/login.html:84
 #: userprofile/templates/userprofile/profile.html:35
 msgid "Username"
 msgstr "Käyttäjätunnus"
@@ -3617,12 +3616,8 @@ msgstr ""
 "Mikäli sinulla on tunnukset yksinomaan %(brand_name)s-palvelua varten, "
 "kirjaudu sisään tästä."
 
-#: userprofile/templates/userprofile/login.html:87
-msgid "Password"
-msgstr "Salasana"
-
-#: userprofile/templates/userprofile/login.html:99
-#: userprofile/templates/userprofile/login.html:100
+#: userprofile/templates/userprofile/login.html:93
+#: userprofile/templates/userprofile/login.html:94
 msgid "Show more login options"
 msgstr "Näytä lisää kirjautumisvaihtoehtoja"
 
--- a/userprofile/static/userprofile/extra_logins.js
+++ b/userprofile/static/userprofile/extra_logins.js
@@ -2,16 +2,18 @@
 
   function showExtras(event) {
     event.preventDefault();
-    const extras = $("#login-box-row").find(".extra-login")
+    const extras = $("#login-box-row").find(".extra-login");
     extras.show();
     $(".show-extra-login-btn").hide();
-    extras.find(":input:visible").first().focus() // focus for keyboard navigation
+    extras.find(":input:visible").first().focus(); // focus for keyboard navigation
   }
 
   $(function () {
-    const row = $("#login-box-row")
-    const normal = row.find(".login-box").not(".extra-login")
-    if (normal.length > 0) {
+    const row = $("#login-box-row");
+    const normal = row.find(".login-box").not(".extra-login");
+    const errors = row.find(".extra-login").find(".alert");
+    // if any of the extra logins have errors, none of them are hidden
+    if (normal.length > 0 && errors.length == 0) {
       row.find(".extra-login").hide();
       $(".show-extra-login-btn").on("click", showExtras).show();
 
--- a/userprofile/templates/userprofile/login.html
+++ b/userprofile/templates/userprofile/login.html
@@ -79,15 +79,9 @@
 					</p>
 					<form method="post" role="form">
 						{% csrf_token %}
-						{#{ form|bootstrap }#}
+						{{ form|bootstrap }}
 						<div class="form-group">
-							<input class="form-control" type="text" id="id_username" name="username" maxlength="254" placeholder="{% trans 'Username' %}">
-						</div>
-						<div class="form-group">
-							<input class="form-control" type="password" id="id_password" name="password" placeholder="{% trans 'Password' %}">
-						</div>
-						<div class="form-group">
-							<button type="submit" class="btn btn-default btn-block">{% trans "Log in" %}</button>
+							<button type="submit" class="btn btn-info btn-block">{% trans "Log in" %}</button>
 						</div>
 					</form>
 				</div>


userprofile/templates/userprofile/login.html
```diff
@@ -81,10 +81,12 @@ <h3>
 						{% csrf_token %}
 						{#{ form|bootstrap }#}
 						<div class="form-group">
-							<input class="form-control" type="text" id="id_username" name="username" maxlength="254" placeholder="{% trans 'Username' %}">
+							<label for="id_username">{% trans "Username" %}</label>
+							<input class="form-control" type="text" id="id_username" name="username" maxlength="254">
```
mtthwhggns, line 85: These `id`s look a little strange, but I can see they've not been changed in this PR, so it's fine to leave them alone here!

raphendyr: The form field id's could have a prefix related to the form: https://docs.djangoproject.com/en/3.0/ref/forms/api/#django.forms.Form.prefix

Not important for this page, but would make sense.

jaguarfi:
Since this PR intend to fix the accessibility issues of the login form. Should the login button have a different color?. I think we extend the scope of this PR and make the login button to look more like a button. What do you think @raphendyr  and @mtthwhggns 
image

mtthwhggns:
> Since this PR intend to fix the accessibility issues of the login form. Should the login button have a different color?. I think we could extend the scope of this PR and make the login button to look more like a button. What do you think @raphendyr and @mtthwhggns

I agree that the button should look more consistent with our other buttons. 

From a usability viewpoint, I think that previously, it was nice to have this one more muted, so that when it appeared alongside "Aalto Login" and "Google Login", it was less attractive, to "nudge" the users the correct way. Now that the whole section is hidden, I think this would be a good change.

However, I have no strong feelings on whether this should be done by expanding the scope of this PR or separately. I'll be creating a new issue shortly about inconsistencies elsewhere in our interfaces, so I'd be equally happy if this was taken care of at the same time as other buttons.

raphendyr:
The button can be fixed here too, but the PR title should be updated then to reflect what the PR does. @mtthwhggns explained well why it was as it was.

etanttila:
So should I fix this here or should it be left to be taken care of at the same time as the other inconsistencies with buttons, @raphendyr and @mtthwhggns? Either way is fine for me, I just wish to know which we end up with (and if we decide that I should fix it here, what color the button should be).

mtthwhggns:
The consensus seems to be to do it here.

I think adding `.btn-primary` should suffice, and make it match the other main buttons we have.

etanttila:
> I think adding `.btn-primary` should suffice, and make it match the other main buttons we have.

I personally think it would normally look kind of strange, as the "Aalto login" button is dark blue (btn-primary) and the "Login using Google" button is red (btn-danger), so it would be weird having two buttons the same color and one a different one. I would at least wonder why the Google login button is a different color than the other two. It would be be a consistency issue. I personally think either (1) all the buttons should be different colors or (2) they should all be the same color or (3) there should be a clear reasoning why some are the same color and others aren't, and that reasoning should be clear to the users as well.

raphendyr:
Actually, all the buttons could include a glyph based on the login type, but let's not do that now. I think you could use `.btn-info` for now.

I would prefer all the buttons to have the same color and glyphs to differentiate, but it's too much effort before there is more than 3 login types.

etanttila:
> I strongly advise against `.btn-info`. It has a contrast ratio of 2.09 (we should be aiming for at least 3.5).

Yeah, the info button does have a terrible contrast. If we were using Bootstrap 4, the secondary button could be good, but we aren't :/

raphendyr:
> I strongly advise against .btn-info. It has a contrast ratio of 2.09 (we should be aiming for at least 3.5). 

The color of the info has to be fixed later anycase (hopefully before next release), so I don't see this as an argument here. We can't just ignore one of the themed color variants, just becouse it's broken currently. It's used for some things any case, so let's fix all of them.

Semantically, the `info` is bit nicer than e.g. `warning`. Though, semantically `danger` is pretty bad for google.

mtthwhggns:
> > I strongly advise against .btn-info. It has a contrast ratio of 2.09 (we should be aiming for at least 3.5).
> 
> The color of the info has to be fixed later anycase (hopefully before next release), so I don't see this as an argument here. We can't just ignore one of the themed color variants, just becouse it's broken currently. It's used for some things any case, so let's fix all of them.

The benefit is that we avoid the following situation: we hit a delay fixing the colour and this gets deployed deployed in January with a new accessibility issue added which wasn't there before.

Since there isn't a schedule for fixing the colour at all yet, I believe it is a valid point.

etanttila:
If we want something that would work already now and not be too uncontrasting, we could use "btn-success active", which would have a contrast of 3.41, but sematically we might not want it to be green.

jaguarfi:
You could add some CSS. If the contrast is a problem here, it might be a problem everywhere else. So, you could overwrite the `background-color` and `border-color`   of the `.btn-info` and `.btn-info:hover`. Maybe, it is a drastic solution, but it might work :)

raphendyr:
>: we hit a delay fixing the colour 

I'm pretty confident we can fix it before next release. The is the prefixing bootstrap classes step in #584, which should also be done for that, thus we need to prebuild some bootstrap LESS files any case. Though, I would look more into this after @mtthwhggns analysis, so we know what else should be fixed at the same time.

Any case, the local log in can be any other color than danger from my point of view.

> You could add some CSS. If the contrast is a problem here, it might be a problem everywhere else. So, you could overwrite the background-color and border-color of the .btn-info and .btn-info:hover. Maybe, it is a drastic solution, but it might work :)

I don't personally like hacking a single button, because the whole theme is broken. It's as costly to fix all of the info buttons than just one, so why not to fix all of them.

mtthwhggns:
I don't want to push this discussion, on a PR which was meant to fix a really tiny issue, much deeper.

This needs fixing, it's been on my list since week one, but maybe the workload is growing too much to make sense to do it in this PR. I trust @etanttila 's judgement to choose the right solution, I'll raise the accessibility-specific stuff elsewhere.

raphendyr:
Before bikeshedding too much, @etanttila should just pick any color, where that info or anything else, and we'll go with that. As we don't deploy directly to production, it's fine to make suboptimal, but working, solutions inside a PR/commit and fix them a bit later (as long that the raised issues are tracked). The situation would be different, if we would use continuous deployment to production, as then we couldn't merge any new defects.

In addition, I prefer not to add hacks, just so a single thing works _now_, as they are never cleaned afterwards and contribute to technical debt. Thus, do this feature as it's supposed to work and from this issues point of view. Leave the rest of the problems for the another issue to handle. That limits the scope here and let's us focus in the issue at hand. Specially for this PR, when we know accessibility fixes have relatively high priority, e.g., we should assume the color for the info will be fixed. Sadly, we don't yet have the issues to milestone them and make this visible.

etanttila:
> I don't personally like hacking a single button, because the whole theme is broken. It's as costly to fix all of the info buttons than just one, so why not to fix all of them.

@raphendyr, I believe @jaguarfi meant that the problem with info buttons probably exists everywhere, so the CSS could be edited to overwrite the btn-info and btn-info:hover in all of A+. I believe that could be a valid fix if we don't switch from Bootstrap 3 in the near future, although a drastic solution to do as a part of this PR as Jhosmimar mentioned. I think there are several options here:

1. The login button gets left as a btn-default for now and gets fixed when all the other button inconsistancies and contrast problems are confronted.
PROS: no hacks, consistant.
CONS: needs to be done later, might be forgotten; might go into production as white.

2. The login button is changed to a btn-info and the btn-info color gets fixed in the near future.
PROS: long-term solution, this PR doesn't expand too much.
CONS: someone needs to change the btn-info at some point; if the btn-info doesn't get fixed early enough, might go into production with horrid contrast.

3. The login button is changed to a btn-info and I fix the btn-info color for all of A+ at the same time.
PROS: The info-buttons are changed in all of A+ right away; no risk of a bad login button going into production.
CONS: this PR expands (or a new PR is created); I may not have the expertise to choose a good color that takes colorblindess into enough consideration.

4. The login button is changed to a btn-primary.
PROS: enough contrast, no need for hacking.
CONS: Inconsistancy of having some login buttons one color and one a different one.

etanttila:
> Before bikeshedding too much, @etanttila should just pick any color, where that info or anything else, and we'll go with that. As we don't deploy directly to production, it's fine to make suboptimal, but working, solutions inside a PR/commit and fix them a bit later (as long that the raised issues are tracked). The situation would be different, if we would use continuous deployment to production, as then we couldn't merge any new defects.
> 
> In addition, I prefer not to add hacks, just so a single thing works _now_, as they are never cleaned afterwards and contribute to technical debt. Thus, do this feature as it's supposed to work and from this issues point of view. Leave the rest of the problems for the another issue to handle. That limits the scope here and let's us focus in the issue at hand. Specially for this PR, when we know accessibility fixes have relatively high priority, e.g., we should assume the color for the info will be fixed. Sadly, we don't yet have the issues to milestone them and make this visible.

I guess this supports option 2. Should I reference the issue linked above in the commit message somehow to indicate that that issue should be fixed before the particular commit is ready for production?

raphendyr:
> Should I reference the issue linked above in the commit message somehow to indicate that that issue should be fixed before the particular commit is ready for production?

We'll track this in the new issue. @mtthwhggns has list of issues to write, but if this wasn't part of those, then I'll create an issue too. We had a chat and some planning, and he will write info on the issues about the plans. I guess the best way to say "should be done before a release" is to mark the issue for the milestone. I'm just not sure how to make it clear that the milestone shouldn't be postponed (e.g. should we have a label? would it be enough?).

Nevertheless, from this PR's point of view, no need to worry how or when the info is fixed (if you choose to use it).

jaguarfi:
Everything looks great now :). One last thing, please squash all commits to one. Maybe something like this, but not strictly this :).
```
Change UI of the local users login form

* Add labels to improve accessibility
* Change UI of the login button
```

Mankro:
Should the login form error message be fixed in this PR too since it is a small change? 
https://github.com/apluslms/a-plus/issues/591
(That is to say, if the login fails, there should be an error message.)

mtthwhggns:
@Mankro In the spirit of keeping PRs on topic relative to the issue, I'd suggest not. 

We don't have any consistent error styling yet, so that is potentially a design task as well as a login tweak. A separate PR would be a good place to test and discuss, for example, the wording of the error message, which is likely not trivial.

Mankro:
> @Mankro In the spirit of keeping PRs on topic relative to the issue, I'd suggest not.
> 
> We don't have any consistent error styling yet, so that is potentially a design task as well as a login tweak. A separate PR would be a good place to test and discuss, for example, the wording of the error message, which is likely not trivial.

Sounds good. The error message shall be left for a new PR.

raphendyr:
I'm not sure if [the earlier implementation](https://github.com/apluslms/a-plus/commit/a86bf5fe02eb3682c2b7e44d12d00443cdefcec2#diff-b6bf26a00455d2704aadb6efec0a06c5L60) would be "good enough", specially as the button string is now hardcoded (not in settings). Just pointing it out as it's another way to add labels. Interestingly, that commented out bootstrap version still exists in the file...

Alternatively, one could hard code the fields, like in [another old version](https://github.com/apluslms/a-plus/commit/1687145ed3db8af1058e7c554c79cac395fb7f40#diff-b6bf26a00455d2704aadb6efec0a06c5L22) (will take a while to load).

However, I think the translation part is bit harder and requires replacing the used django login form, so I guess just adding the `{{ form.username.errors }}` might be ok, asuming Django has translated all errors. Asuming the first simple solutions don't work due to missing translations, then the more complicated solutions should be done in another PR. Asuming that a different solution to add labels work, then it can be done here.

(If django forms must be edited for labels/errors, then [this comment](https://stackoverflow.com/a/43112491) might help.)

etanttila:
@raphendyr the Django form seems to otherwise work very well (even with translations), but the "Please fill in this form" message is not translated:
image

mtthwhggns:
Is that one actually coming form the browser? It looks like the HTML5 validation.

etanttila:
> Is that one actually coming form the browser? It looks like the HTML5 validation.

Good point! It probably is! Can browser validation be translated / does it need to be?

mtthwhggns:
> > Is that one actually coming form the browser? It looks like the HTML5 validation.
> 
> Good point! It probably is! Can browser validation be translated / does it need to be?

I think it's possible, but maybe unneccesary - we can probably trust that the user can read the language they use their browser in, even if it doesn't match the current interface langauge.

Mankro:
Yes, exactly. That looks like the browser validation message and we do not need to worry about its language.

etanttila:
I just pushed the commit with the change. It seems to work well with both Finnish and English, both on its own and with a normal logins, and with both Firefox and Chrome. It also seemed to work with the screen reader. The error message seemed alright as well. If this new implementation seems good, I'd next probably squash the commits and rebase them on top of the current master.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3703,3 +3698,6 @@ msgstr "Unohda automaattinen uudelleenohjaus."
 #: userprofile/views.py:78
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
+#~ msgid "Password"
```
raphendyr, line 3702: Can you run `makemessages` with option `--no-obsolete` thanks. Sorry, this is missing from the guides..

userprofile/static/userprofile/extra_logins.js
```diff
@@ -2,16 +2,17 @@
...
   }
 
   $(function () {
-    const row = $("#login-box-row")
-    const normal = row.find(".login-box").not(".extra-login")
-    if (normal.length > 0) {
+    const row = $("#login-box-row");
+    const normal = row.find(".login-box").not(".extra-login");
+    const errors = row.find(".extra-login").find(".alert");
+    if (normal.length > 0 && errors.length == 0) {
```
raphendyr, line 15: This is fine, but maybe there should be a comment to mention that this now doesn't hide any of the extra logins if any of them have errors. Alternatively, I think you could do `row.find('.extra-login').filter(function(b) { return !b.find('.alert'); }).hide();`.

locale/fi/LC_MESSAGES/django.po
etanttila: Would that mean that the translation is lost if in the future one would add to add something with the text "Password"?

userprofile/static/userprofile/extra_logins.js
etanttila: I think it's better if none of the extra logins are hidden, so the view stays the same (so if there were several, logins stay in the same spot and none disappear). Otherwise that could have been implemented in the template with the extra-login class being conditional. But yeah, I'll comment that!

locale/fi/LC_MESSAGES/django.po
raphendyr: Yes, but we have git so they could find the translation from the history.

userprofile/static/userprofile/extra_logins.js
raphendyr: A good argument!

userprofile/static/userprofile/extra_logins.js
```diff
@@ -2,16 +2,18 @@
 
   function showExtras(event) {
     event.preventDefault();
-    const extras = $("#login-box-row").find(".extra-login")
+    const extras = $("#login-box-row").find(".extra-login");
     extras.show();
     $(".show-extra-login-btn").hide();
     extras.find(":input:visible").first().focus() // focus for keyboard navigation
```
jaguarfi, line 8: You added the missing semicolons, so maybe you can add the missing one here.

etanttila: Thanks, I added the semicolon, squashed and rebased this!

PR #588, opened 2020-06-25
Allow language to be set in API submission

### LINKED ISSUE #581: The grading language can not be set in the API submission endpoint
New submissions for grading can be sent to the API via the endpoint `/api/v2/exercises/ID/submissions/submit/`. The language can not be set via the API. The language affects the grading since it is one of the parameters sent to the grader in the A+ grader protocol. The API must be changed so that it accepts an optional language parameter and uses it in the grader protocol. If no language is given to the API, then it uses the (course) default language like it seems to do now.

The name of the POST parameter for the language in the API could be `__lang`. Underscores try to avoid collisions with real parameters used by exercises. We need to think how this relates to the existing hacky submission data parameter `__grader_lang` or should we use that one for this case as well.

Edit: see the discussion for better ideas about the parameter names.

A+ API code:
submit endpoint: https://github.com/apluslms/a-plus/blob/cd0cfb593656340dbcdab9078a7781d0d96599cd/exercise/api/views.py#L235
grade function: https://github.com/apluslms/a-plus/blob/cd0cfb593656340dbcdab9078a7781d0d96599cd/exercise/exercise_models.py#L693
language is taken from the Django settings, but the language can not be set via the API yet: https://github.com/apluslms/a-plus/blob/cd0cfb593656340dbcdab9078a7781d0d96599cd/exercise/exercise_models.py#L697

HenrikSundell:
# Allow language to be set in API submission

*Provide an overview of the changes and their purpose: what, why, how.*
Allow the grading language to be set when submitting through the API.
Previously the language was always retrieved from django settings,
which would not allow the user to choose a submission language when
submitting using the API. 
The language can be added to the submission
POST data using the __aplus__  namespace as a JSON string with the key
"lang". For example to set the grading language to English:
`__aplus__={"lang": "en"}`

Fixes #581

# Testing

Changes have been tested using curl. No testing has been done
on graders that might break from somehow broken or missing language codes.

Some stuff in the wiki should probably be updated when this is final. 
At least:
https://wiki.aalto.fi/pages/viewpage.action?pageId=165137732


exercise/exercise_models.py
```diff
@@ -694,7 +695,16 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
...
         Loads the exercise feedback page.
         """
-        language = get_language()
+        # Get the language from the submission
+        try:
+            aplus_data_index = [i[0] for i in submission.submission_data].index("__aplus__")
+            aplus_data_string = submission.submission_data[aplus_data_index][-1]
+            aplus_data = json.loads(aplus_data_string)
+            language = aplus_data.get("lang",  get_language())
+        except ValueError:
```
Mankro, line 704: Does this exception catch all cases: 

* submission has no submission data
* submission data does not contain the `__aplus__` key
* the `__aplus__` object does not contain the `lang` key

exercise/exercise_models.py
```diff
@@ -694,7 +695,16 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
         """
         Loads the exercise feedback page.
         """
-        language = get_language()
+        # Get the language from the submission
+        try:
+            aplus_data_index = [i[0] for i in submission.submission_data].index("__aplus__")
```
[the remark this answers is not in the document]
Mankro: `_aplus_group` should then also be moved to the submission meta data from the normal submission data. At least the new submit API should support it before the new API is widely used. The old ExerciseView could be changed later.

The old hack `__grader_lang` possibly can't be changed now, or at least it would also require changes in the MOOC-Grader too.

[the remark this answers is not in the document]
raphendyr: Once A+ handles the language correctly, then MOOC-Grader can possibly drop that field. In other words, leave the `__grader_lang` to be MOOC-Grader's problem.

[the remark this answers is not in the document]
raphendyr: > What kind of data would be included in the meta? Only the language?

For now, nothing else than lang and group id. It makes it easy place to start adding values.

exercise/exercise_models.py
```diff
@@ -694,7 +695,16 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
...
+            aplus_data_string = submission.submission_data[aplus_data_index][-1]
+            aplus_data = json.loads(aplus_data_string)
+            language = aplus_data.get("lang",  get_language())
```
raphendyr, line 703: For information. It's beneficial to think when things are evaluated. Here, the `get_language` is evaluated always before `aplus_data.get`. It should be cheap, but as it's a function call we can't be sure when reading the code (function calls are expected to do processing while attribute resolution should be "cheap").

Hence, things like this might be beneficial to write `language = aplus_data.get('lang') or get_language()`. The `or` makes it so that the first part is evaluated and only if that is false (None or empty string), then the latter part is evaluated.

In addition, try to use `'` for for dictionary key strings (`'lang'` vs. `"lang"`). That wey the quotes can be used to indicate the strings purpose/type for the reader. There is a bit more in the style guide about this, once we get that done.

exercise/exercise_models.py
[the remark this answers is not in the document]
HenrikSundell: Good idea might be a bit less hacky than this:), should the submission_data stay as it is even it means there would be some duplicate data

exercise/exercise_models.py
HenrikSundell: Good to know I thought it wouldn't be evaluated if lang is found, but I guess it's logical that the arguments are evaluated before the function is called.

exercise/exercise_models.py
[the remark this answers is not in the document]
Mankro: No, we don't want to have the new `__aplus__` field in the submission data. Its contents shall be stored in the new submission meta data field in the Submission model. The database field does not need to contain the word `__aplus__` as it is enough to store the actual JSON object. This object is stored as an object so no need to convert it into a list of 2-tuples like has previously been done with the submission data.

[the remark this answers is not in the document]
Mankro: About moving `_aplus_group` into the new submission meta data: this line in `check_submission_allowed` uses the old way without the new `__aplus__` object: https://github.com/apluslms/a-plus/blob/2195e36b86db9eb77c6695fdb37c5af7bba4dad9/exercise/exercise_models.py#L578
The function is called by the normal web submission (`class ExerciseView`) and the new submit API endpoint. If we don't modify the web submissions now, then we need to support both ways. The API should prefer the new method with `__aplus__` since the API itself is new. Note that `check_submission_allowed` does not have any `submission` object since it is called before creating a new submission, but then we can assume that the `request.POST` contains the values.

exercise/submission_models.py
```diff
@@ -30,7 +31,8 @@ def get_queryset(self):
     def create_from_post(self, exercise, submitters, request):
         new_submission = Submission.objects.create(
             exercise=exercise,
-            submission_data=query_dict_to_list_of_tuples(request.POST)
+            submission_data=query_dict_to_list_of_tuples(request.POST),
+            meta_data=json.loads(request.POST.get('__aplus__','{}'))
```
raphendyr, line 35: Please add the comma so the next change doesn't need to do the same as this one did, i.e, edit the line above just to add the comma.

Also, there should be space between arguments, please check [spaces section of PEP 8](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements).

Lastly, this should handle invalid JSON some way too, mostly so that the user doesn't get 500 for invalid json in `__aplus__`.

exercise/submission_models.py
```diff
@@ -31,7 +31,7 @@ def get_queryset(self):
     def create_from_post(self, exercise, submitters, request):
         new_submission = Submission.objects.create(
             exercise=exercise,
-            submission_data=query_dict_to_list_of_tuples(request.POST),
+            submission_data=[i for i in query_dict_to_list_of_tuples(request.POST) if not '__aplus__' == i[0]],
```
raphendyr, line 34: This lines becomes pretty long, so It could be a good idea to use a temporary variable. In addition, `i` is often used to mean `index` (sometimes `item`, like here, but that is maybe a bit more rare). As the value is always tuple of size 2, you could unpack and pack it again. That is questionable, as then there is repacking and then it's unclear if the order of the values change.

```python
well_named_variable = [
  (name, value)
  for (name, value) in query_dict_to_list_of_tuples(request.POST)
  if name != '__aplus__'
]
new_submission = Submission.objects.create(
￼            exercise=exercise,
￼            submission_data=query_dict_to_list_of_tuples(request.POST),
￼            submission_data=well_named_variable,
```

Something note too, is that typically the condiftion is more readable if it's in order `x not something y` instead of `not x someting y`, even though they are logically the same.

exercise/exercise_models.py
```diff
@@ -576,7 +576,8 @@ def _check_submission_allowed(self, profile, request=None):
 
         # Support group id from post or currently selected group.
         group = None
-        group_id = request.POST.get("_aplus_group") if request else None
+        group_id = json.loads(request.POST.get('__aplus__','{}')).get('group') or request.POST.get("_aplus_group") if request else None
```
raphendyr, line 579: This is a pretty long line and more importantly, it includes a lot of different things (i.e. hard to follow). In addition, now the condifition is really far on the right, so it's easy to miss. Might better to use traditional `if cond: <then>; else: group_id = None`

Further, you need to handle JSON parsin errors. For an example, see another parto f the code: https://github.com/apluslms/a-plus/blob/8c7c5a19669cdcf32efd32d2cafc617cead490f8/edit_course/operations/batch.py#L15

exercise/submission_models.py
HenrikSundell: Should meta_data just be set to an empty dict continue grading or should there be some message returned  if there is invalid json

exercise/exercise_models.py
```diff
@@ -694,7 +696,9 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
         """
         Loads the exercise feedback page.
         """
-        language = get_language()
+        # Get the language from the submission
+        language = submission.meta_data.get('lang', None) or get_language()
```
raphendyr, line 700: You could consider if the `submission` should expose the language via a property. Might be nice to use `submission.lang` here.

In addition, should we store the `get_language()` information in the database, thus making resubmissions consistent too? (did we have an issue about that?).

exercise/submission_models.py
Mankro: About potential JSON syntax errors in the meta data: ignoring errors silently makes debugging harder, so we could at least log an error. Since the meta data is not critical for grading, it could be skipped and the submission could still be sent to the grader, but it is also suspicious. Is there another problem in the submission if the client sends broken JSON with it? Well, what does Jaakko think?

exercise/exercise_models.py
Mankro: That's interesting. Indeed, we could store the language in the meta data even if the original submission did not contain any language meta data. It would make sense for regrading submissions. I don't remember if we have had issues with that, but it makes sense that regrading can only use data from the submission. It should not take the language from the teacher's web session when the regrading is activated.

Mankro:
Thus far this is progressing nicely. Remember that you will need to generate migrations in the end because you have modified the Submission model.

exercise/submission_models.py
HenrikSundell: @raphendyr

raphendyr: At least the API should return error on invalid POST. Web side requires some UI to handle that though. Luckily, that shouldn't happen there.

exercise/exercise_models.py
raphendyr: For this PR, I could accept a TODO/FIXME comment here, marking that the `get_language` should be moved elsewhere (i.e. so that it's done when storing the submission to db and not when sending to grader).

exercise/submission_models.py
```diff
@@ -290,6 +303,10 @@ def set_error(self):
     def is_graded(self):
         return self.status in (self.STATUS.READY, self.STATUS.UNOFFICIAL)
 
+    @property
+    def lang(self):
+        return self.meta_data.get('lang', None)
```
Mankro, line 308: What if the `meta_data` is empty? Is it a dictionary then or does it become `None` or an empty string? In other words, does the `get` method always exist? You have modified the `Submission.objects.create_from_post` method, but sometimes new submissions are created without using that method. Additionally, when this update is deployed to production, all old existing submissions have empty meta data.

exercise/views.py
```diff
@@ -144,29 +145,40 @@ def post(self, request, *args, **kwargs):
...
-                        ("?wait=1" if page.is_wait else ""))
+            try:
+                new_submission = Submission.objects.create_from_post(
+                    self.exercise, students, request)
+            except ValueError as error:
+                messages.error(
+                    request,
+                    format_lazy(
+                       _("The submission could not be saved for some reason. "
+                         "The submission was not registered. error: {}"),
```
Mankro, line 156: Typo: error -> Error

Did you try this in practice? Does the formatting look good when the Exception is included in the string? Was there a specific reason for using `format_lazy` since I thought that the normal `.format` method works?

exercise/exercise_models.py
```diff
@@ -575,7 +576,15 @@ def _check_submission_allowed(self, profile, request=None):
 
         # Support group id from post or currently selected group.
         group = None
-        group_id = request.POST.get("_aplus_group") if request else None
+        group_id = None
+        if request:
+            try:
+                group_id = json.loads(request.POST.get('__aplus__','{}')).get('group')
```
Mankro, line 582: This was said earlier too. There should be a space between function parameters: `.get('__aplus__', '{}')`

exercise/views.py
HenrikSundell: It looks fine, no specific reason for lazy_format just blindly following examples. I changed it to .format, which looks fine too.

exercise/submission_models.py
```diff
@@ -39,7 +51,7 @@ def create_from_post(self, exercise, submitters, request):
             logger.exception("Failed to save submitted files: %s %s",
                 request.user.username, exercise);
             new_submission.delete()
-            return None
+            raise DatabaseError
```
Mankro, line 54: I have a feeling that this instantiates a new DatabaseError instead or re-raising the caught DatabaseError. It's better to re-use the existing error. It contains all the details added by Django when it raised the error.

HenrikSundell: Ok, should the details be used somewhere or are the current messages fine?

exercise/submission_models.py
```diff
@@ -28,18 +29,29 @@ def get_queryset(self):
...
+        submission_data_list = [
+                (key, value) for (key, value) in query_dict_to_list_of_tuples(request.POST)
+                if key != '__aplus__'
+        ]
+        try:
+            meta_data_dict = json.loads(request.POST.get('__aplus__','{}'))
+        except json.JSONDecodeError:
+            raise ValueError("invalid json for __aplus__")
+
         new_submission = Submission.objects.create(
```
Mankro, line 42: Now I wonder why no exception is caught from `Submission.objects.create`. I guess somebody wanted to particularly log errors if the submission is created, but saving the submitted files fails after it.

exercise/submission_models.py
Mankro: I think that `raise DatabaseError from error` also instantiates a new DatabaseError object. When you have the error variable referring the original exception, you can `raise error` and then it certainly does not create a new exception object.

I don't know if these errors have such messages that we could show them directly to the users. If they are descriptive, they could be shown to the user. Just saying "some error" is not helpful even though the old message is like that. It is also important to log errors so that sysadmins can inspect them afterwards. If the `Submission.objects.create` method fails, the current code does not seem to even log the error, so that could be added.

raphendyr: just `raise` or `raise NewException(message) from error` would be good. The Latter if the type or the message should be changed.

exercise/views.py
```diff
@@ -148,33 +149,30 @@ def post(self, request, *args, **kwargs):
                 new_submission = Submission.objects.create_from_post(
                     self.exercise, students, request)
             except ValueError as error:
-                messages.error(
-                    request,
-                       _("The submission could not be saved for some reason. " 
-                         "The submission was not registered. Error: {}").format(error)
+                messages.error(request,
+                    _("The submission could not be saved for some reason. "
```
raphendyr, line 153: This could use a different string "due to malformed post data" or such to indicate it's the users fault.

exercise/views.py
```diff
@@ -148,33 +149,30 @@ def post(self, request, *args, **kwargs):
...
+                      "The submission was not registered. Error: {}").format(error)
```
raphendyr, line 154: this should use `format_lazy`, even though this code most likely works ok.

exercise/api/views.py
```diff
@@ -250,32 +251,36 @@ def submit(self, request, *args, **kwargs):
...
                 }
                 if page.errors:
-                    data = {'errors': page.errors}
-            else:
-                data = {
-                    'detail': _("The submission could not be saved for some reason. "
-                                "The submission was not registered."),
-                }
-                status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
+                        data = {'errors': page.errors}
```
raphendyr, line 283: This seems to have a wrong indentation.

exercise/submission_models.py
```diff
@@ -28,18 +29,29 @@ def get_queryset(self):
             .prefetch_related('submitters')
 
     def create_from_post(self, exercise, submitters, request):
+
+        submission_data_list = [
+                (key, value) for (key, value) in query_dict_to_list_of_tuples(request.POST)
+                if key != '__aplus__'
+        ]
+        try:
+            meta_data_dict = json.loads(request.POST.get('__aplus__','{}'))
```
raphendyr, line 38: The space thingie.. :)

exercise/submission_models.py
```diff
@@ -28,18 +29,29 @@ def get_queryset(self):
...
+        except json.JSONDecodeError:
+            raise ValueError("invalid json for __aplus__")
```
raphendyr, line 40: I think `in` is a better preposition here, or is it? Or `The content of the field __aplus__ is not valid json`?

exercise/submission_models.py
```diff
@@ -28,18 +29,29 @@ def get_queryset(self):
...
         new_submission.submitters.set(submitters)
         try:
             new_submission.add_files(request.FILES)
-        except DatabaseError:
+        except DatabaseError as error:
             logger.exception("Failed to save submitted files: %s %s",
                 request.user.username, exercise);
             new_submission.delete()
-            return None
+            raise DatabaseError from error
```
raphendyr, line 54: Interesting. I have been wrong. The `from error` is not actually must in python 3, however it should be prefered due to semantic difference (also recommended in PEP 8). See the different message in following examples (during vs. direct cause):

```python
>>> try:
...   raise ValueError('foo')
... except ValueError:
...   raise RuntimeError()
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
ValueError: foo

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError
```

compared to 

```python
>>> try:
...   raise ValueError('foo')
... except ValueError as error:
...   raise RuntimeError from error
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
ValueError: foo

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError
```

To remove the chaining information would require syntax:

```python
try:
  ...
except Exception:
  raise AnyException from None
```

To conclude, this line is now ok.

exercise/submission_models.py
```diff
@@ -290,6 +303,13 @@ def set_error(self):
     def is_graded(self):
         return self.status in (self.STATUS.READY, self.STATUS.UNOFFICIAL)
 
+    @property
+    def lang(self):
+        try:
+            return self.meta_data.get('lang', None)
+        except AttributeError:
```
raphendyr, line 310: What would raise `AttributeError`?

exercise/views.py
```diff
@@ -144,29 +145,34 @@ def post(self, request, *args, **kwargs):
...
+                messages.error(request,
+                    _("The submission could not be saved for some reason. "
+                      "The submission was not registered.")
+                )
+            else:
                 page = self.exercise.grade(request, new_submission,
                     url_name=self.post_url_name)
                 for error in page.errors:
                     messages.error(request, error)
-
```
raphendyr, line 154: Was removing these lines imporant or an accident (indent + unindent)?

exercise/submission_models.py
HenrikSundell: .get if meta_data == None for submissions created before these changes

exercise/views.py
raphendyr: > Was there a specific reason for using format_lazy since I thought that the normal .format method works?

_() returns a lazy string and using `.format` on it retrieves the translation at that spot. If for some reason the code is executed somewhere where language is not yet defined correctly, then the wrong translation code is retrieved. In other words. It's always good to use `format_lazy` with `_`. (aka. `ugettext_lazy`)

exercise/submission_models.py
raphendyr: I see. Was commented here https://github.com/apluslms/a-plus/pull/588#discussion_r446917317  :)

exercise/submission_models.py
raphendyr: Hmm. So the old interface actually was value, None or exception as `.create` can raise IntegrityError, which is a subclass of DatabaseError. Thus, we are actually fixing a bug here too :)

exercise/submission_models.py
raphendyr: This is pretty good too. Maybe a comment saying "to handle cases where database includes null or non dictionary json"

```python
>>> json.loads('{}').get('foo', '<not found>')
'<not found>'
>>> json.loads('true').get('foo', '<not found>')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'bool' object has no attribute 'get'
```

Alternative would be checking that the `meta_data` is actually a dict, [like here](https://github.com/apluslms/a-plus/blob/master/exercise/templatetags/exercise.py#L241).

exercise/submission_models.py
Mankro: Is it OK? At the moment, I see `raise DatabaseError from error` and doesn't that instantiate a new exception even though the same exception could be re-raised?

exercise/views.py
```diff
@@ -144,9 +146,23 @@ def post(self, request, *args, **kwargs):
...
-            new_submission = Submission.objects.create_from_post(
-                self.exercise, students, request)
-            if new_submission:
+            try:
+                new_submission = Submission.objects.create_from_post(
+                    self.exercise, students, request)
+            except ValueError as error:
+                messages.error(request,
+                    format_lazy(
+                    _("The submission could not be saved due to malformed post data. "
```
Mankro, line 155: You have modified or added some translated strings, so don't forget to add Finnish translations. `manage.py makemessages`, update the file `locale/fi/LC_MESSAGES/django.po` and `manage.py compilemessages` etc.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -111,7 +111,7 @@ msgstr "Käyttäjät"
 msgid "Input an URL identifier for this course."
 msgstr "Syötä URL-osoitteessa käytettävä tunnus kurssille."
 
-#: course/models.py:68 course/models.py:729 exercise/exercise_models.py:127
+#: course/models.py:68 course/models.py:729 exercise/exercise_models.py:128
```
Mankro, line 114: We will have less git conflicts across different branches and pull requests if these line number comments are not included in the git commit. With `git add -p`, you can add only the relevant lines from one file into the staging area and commit.

raphendyr: It's good to include them time to time, but in cases like these where you add a single new translation string, then it's better to include only that part. When there are a lot of other changes too, then it's a good place to update the numbers.

Though, I was supposed to thing this in the style guide, as there is an option not to inlcude line numbers (but the filenames would stay). Should help with this.

exercise/submission_models.py
raphendyr: This should be good now

exercise/submission_models.py
```diff
@@ -28,18 +29,33 @@ def get_queryset(self):
...
+            new_submission = Submission.objects.create(
+                exercise=exercise,
+                submission_data=submission_data_list,
+                meta_data=meta_data_dict,
+            )
+            new_submission.submitters.set(submitters)
+        except DatabaseError as error:
+            logger.exception("Failed to create submission: %s %s",
+                request.user.username, exercise);
+            raise DatabaseError from error
```
Mankro, line 51: I'm not sure if messages got lost due to the long discussion chains, but I tried to say a couple of times that I think `raise DatabaseError from error` could be simply `raise error` or just `raise` here so that it re-raises the same error instance without creating new instances.

exercise/views.py
Mankro, line 155: Shouldn't the parameters of `format_lazy` be indented here?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2889,7 +2889,15 @@ msgstr "Tehtävässä on huoltokatko, ja sisältö on piilotettu opiskelijoilta.
 msgid "Unfortunately this exercise is currently under maintenance."
 msgstr "Valitettavasti tässä tehtävässä on juuri nyt huoltokatko."
 
-#: exercise/views.py:181
+#: exercise/views.py:155
+msgid ""
```
Mankro, line 2893: I think `python manage.py makemessages` is supposed to add a comment `#, python-brace-format` here because the string contains the `{error}` format variable.

exercise/exercise_models.py
```diff
@@ -575,7 +576,15 @@ def _check_submission_allowed(self, profile, request=None):
 
         # Support group id from post or currently selected group.
         group = None
-        group_id = request.POST.get("_aplus_group") if request else None
+        group_id = None
+        if request:
+            try:
+                group_id = json.loads(request.POST.get('__aplus__', '{}')).get('group')
+            except json.JSONDecodeError:
+                pass
```
Mankro, line 584: The JSON syntax errors are ignored here in the function `_check_submission_allowed` that is called before any submission is created (in `Submission.objects.create_from_post` that raises an error for JSON syntax errors). This can hide JSON errors in the request. A JSON syntax error would drop the group from the submission and then the submission may be denied due to a required group, so `create_from_post` is never called and the user gets an error about wrong group size or wrong group members, but really there was a syntax error in the JSON of the request.

exercise/submission_models.py
raphendyr: This is also fine. Benefit is that the top most exception doesn't include the detailed info, which might be beneficial when the exception is rendered to string for API.

exercise/views.py
raphendyr: They should.

exercise/exercise_models.py
raphendyr: #613

Issue #512, opened 2020-03-27
Student is able to enroll on a course multiple times


oseppala:
A student can in some cases enroll on the course multiple times. An example can be found in the WSD 2019-2020 course on the Aalto production server.

Parts of the system which are or might be affected (mark if fixed or checked to be not affected):
- [ ] API / student points: `/api/v2/courses/<course_id>/points/<user_id>/`

raphendyr:
Problem:
* Enrollment model doesn't enforce unique pair of course and user. Code first check if there exists a enrollment, if not, it creates one in [course/models.py](https://github.com/apluslms/a-plus/blob/master/course/models.py#L522) using django `get_or_create`, which is [not atomic](https://github.com/django/django/blob/master/django/db/models/query.py#L559).

How to fix:
* Add unique constraint to Enrollment model (course, user) -pair.
* Create migration to merge duplicates to a single enrollment in the existing database

raphendyr:
I raise the priority, as it seems be affecting two Universities.

Mankro:
We got a report that double enrollments possibly cause a crash (HTTP 500) in the API: `/api/v2/courses/COURSEID/points/ID/`

I feel like increasing the priority even further.

PR #599, opened 2020-06-26
Delete and prevent duplicate course enrollments

veeraalt:
# Description

Previously Enrollment model didn't enforce unique pair of user and course instance. As a result, users were in some cases able to enroll on a course multiple times.

This change:
* Deletes existing duplicate course enrollments from the database (the first enrollment is saved)
* Prevents future duplicate enrollments by adding unique_together constraint for Enrollment model (course_instance, user_profile) -pair. 

Fixes https://github.com/apluslms/a-plus/issues/512

# Testing

* The existing unit tests pass

* Testing has been done manually by testing these use cases:
   * Student has been enrolled on the same course instance two or more times
   -> All other than the first enrollment is deleted from database
   * Student has been enrolled on a course only once
   -> Nothing is changed
   * Student has been enrolled multiple times on the same course with different instances (once for each)
   -> Nothing is changed
   * User tries to enroll a student on a course instance that the student has already been enrolled on
   -> Error message appears at the top of the page: "Enrollment with this Course instance and User profile already exists."
   * User tries to enroll a student on a course instance that the student has not previously been enrolled on
   -> Student is enrolled succesfully

  *Applies to and has been tested with all user types, not only students*

* No, not relevant


Mankro:
At a quick glance, this looks good and your manual test cases cover everything that I can think of straight away. If a manager has time, it would be good to test the migration again manually and verify it is OK. 

When a teacher tries to enroll a student twice in the manual enrollment page under the course participants page, what kind of error does the teacher see? Is it an error message at the top of the page or does the app crash with HTTP 500?

veeraalt:
> When a teacher tries to enroll a student twice in the manual enrollment page under the course participants page, what kind of error does the teacher see? Is it an error message at the top of the page or does the app crash with HTTP 500?

It's an error message at the top of the page that says

`Please correct the error below.`
`Enrollment with this Course instance and User profile already exists.`

course/migrations/0047_delete_duplicate_enrollments.py
```diff
@@ -0,0 +1,41 @@
...
+def delete_duplicates(apps, schema_editor):
+
+    unique_fields = ['course_instance', 'user_profile']
+
+    duplicates = (
+        Enrollment.objects.values(*unique_fields)
+        .order_by()
+        .annotate(min_id=Min('id'), count_id=Count('id'))
+        .filter(count_id__gt=1)
+        )
```
raphendyr, line 17: I wanted to say here, that I'm considering which style should be suggested for code like this. Based ond PEP 8, current layout is fine. Though, I typically prefer closing bracket to be in the same column as the starting line (e.g. here at the start of the line).

For discussion, here are relevant layout posibilities:

```python
# pros: clean and compact, and subject of the action is in the first line
# cons: problematic when adding a new method
duplicates = (Enrollment.objects
    .values(*unique_fields)
    .order_by()
    .annotate(min_id=Min('id'), count_id=Count('id'))
    .filter(count_id__gt=1))

# pros: closing bracket is in a right place, and subject of the action is in the first line
# cons: starting and closing brackets don't feel to be in balance
duplicates = (Enrollment.objects
    .values(*unique_fields)
    .order_by()
    .annotate(min_id=Min('id'), count_id=Count('id'))
    .filter(count_id__gt=1)
)

# pros: newline after the starting bracket
# cons: no separation between object and chained methods
duplicates = (
    Enrollment.objects
    .values(*unique_fields)
    .order_by()
    .annotate(min_id=Min('id'), count_id=Count('id'))
    .filter(count_id__gt=1)
)

# pros: newline in start, and methods are indented to separate from the object
# cons: takes a lot of space
duplicates = (
    Enrollment.objects
        .values(*unique_fields)
        .order_by()
        .annotate(min_id=Min('id'), count_id=Count('id'))
        .filter(count_id__gt=1)
)
```

As comparison, I would consider the following example to match Airbnb's JavaScript style guide.

```javascript
const selected = (
  $('#items')
    .find('.open')
      .highlight()
      .end()
    .find(':selected')
      .first()
)
```

This would give merit for the last python layout. Anyone, feel free to write opinions here.

Mankro: I like the enrollment options 3 and 4.

etanttila: I like options 3 and 4 the most as well. 4 is my favourite and I think it's the easiest to read.

raphendyr: I think the last one would be the one to recommend.

veeraalt: Let's go with the last one then.

course/migrations/0047_delete_duplicate_enrollments.py
```diff
@@ -10,18 +10,19 @@ def delete_duplicates(apps, schema_editor):
...
-        .order_by()
-        .annotate(min_id=Min('id'), count_id=Count('id'))
-        .filter(count_id__gt=1)
-        )
+        Enrollment.objects
+            .values(*unique_fields)
+            .order_by()
+            .annotate(min_id=Min('id'), count_id=Count('id'))
+            .filter(count_id__gt=1)
+    )
```
raphendyr, line 18: What do you personally think about this douple deindentation? This could be worse than the other option too... I'm personally really not sure anymore :disappointed: 

A good counter example:

```python
bar = (
    foo
        .something()
        .another()
)
```

veeraalt: Your counter example looks a bit strange and there I would probably use just single indentation. I'm not sure about the double indentation, I think it looks good in that case (probably only because there is Enrollment.objects and not just Enrollment on the first line). We could go with the option 3.

veeraalt:
I think this is ready for rebasing and merging but if @raphendyr or @jaguarfi (or anyone else) have change requests or comments, please tell.

raphendyr:
It's nice. Could you rebase and squash, thanks.

veeraalt:
Thank you! Rebased and squashed now.

PR #607, opened 2020-07-02
Show all modules to course teachers and assistants in API endpoint

### LINKED ISSUE #549: API does not show course exercises to the teacher in a hidden course
If a course instance is hidden from students (referring to the "visible to students" checkbox in the course settings), then the teacher does not see the exercises at all in the API endpoint `/api/v2/courses/ID/exercises/`. The teacher should see the exercises since teachers are of course allowed to see their own hidden courses.

Check also if the API hides exercises from teachers in modules that haven't opened yet or have closed. The teacher should see all exercises despite the module opening times.

Edit: see comments below. The "visible to students" checkbox was not a problem, but teachers can not see modules that have not opened yet.

Hints:
* The API view for the course exercises: https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/course/api/views.py#L64
* The filtering method that is called from the API view: https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/course/models.py#L657
* The filter builder that seems to think that "visible to students" should apply to all users: https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/course/models.py#L308

HenrikSundell:
# Show unopened modules to teachers

Show all modules to course teachers and assistants in API endpoint
`/api/v2/courses/ID/exercises/`even if not open yet.

Previously unopened modules were hidden from everyone instead of only
students.

Fixes #549 

# Testing
API endpoint has been tested for all user types.

Not relevant

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -305,13 +305,10 @@ class UserTagging(models.Model):
         ordering = ['tag']
 
 
-def get_course_visibility_filter(user, prefix=None):
+def get_course_staff_visibility_filter(user, prefix=None):
     class OR(Q):
         default = Q.OR
-
-    filters = (
-        ('visible_to_students', True),
-    )
+    filters = ()
     if isinstance(user, User):
         user = user.userprofile
         filters += (
@@ -341,7 +338,10 @@ class CourseInstanceManager(models.Manager):
         if not user or not user.is_authenticated:
             return self.filter(visible_to_students=True)
         if not user.is_superuser:
-            return self.filter(get_course_visibility_filter(user)).distinct()
+            return self.filter(
+                get_course_staff_visibility_filter(user)
+                | Q(visible_to_students=True),
+            ).distinct()
         return self.all()
 
 
@@ -662,8 +662,8 @@ class CourseModuleManager(models.Manager):
             )
         if not user.is_superuser:
             return self.filter(
-                get_course_visibility_filter(user, 'course_instance__'),
-                opening_time__lte=timezone.now(),
+                get_course_staff_visibility_filter(user, 'course_instance__')
+                | Q(course_instance__visible_to_students=True, opening_time__lte=timezone.now()),
             ).distinct()
         return self.all()
 


course/models.py
```diff
@@ -654,12 +654,16 @@ def get_queryset(self):
         return super().get_queryset().select_related(
             'course_instance', 'course_instance__course')
 
-    def get_visible(self, user=None):
+    def get_visible(self, user=None, is_course_staff=False):
```
Mankro, line 657: The `get_visible` method already has the user argument. A CourseModule always belongs to a CourseInstance (that defines assistants) and CourseInstance belongs to a Course (that defines teachers). The `is_course_staff` argument is not needed since the staff status can be queried from the database. The caller of the function already supplies the user and it should not be necessary to supply more info about the same user.

In addition, the CourseModuleManager is not tied to a single course. It queries all CourseModules from the database. The user is not a teacher on all courses. I think the `is_course_staff` argument makes the interface of this function feel very wrong. What do you think?

HenrikSundell: I didn't really like this solution, but couldn't get any other way to work :)

Mankro: Did you try similar tricks that are used in the `get_course_visibility_filter` function here: https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/course/models.py#L318

course/models.py
```diff
@@ -654,12 +654,16 @@ def get_queryset(self):
...
         if not user or not user.is_authenticated:
             return self.filter(
                 course_instance__visible_to_students=True,
                 opening_time__lte=timezone.now(),
             )
+        if not user.is_superuser and is_course_staff:
```
raphendyr, line 663: I think you could just drop the opening time restriction from the branch below. Alternatively, the opening time limitation could probably be moved into that function (i.e. the filter would become `(visible to students && opening time < now) || is_teacher || is_assistant)`)

course/models.py
```diff
@@ -341,7 +338,7 @@ def get_visible(self, user=None):
         if not user or not user.is_authenticated:
             return self.filter(visible_to_students=True)
         if not user.is_superuser:
-            return self.filter(get_course_visibility_filter(user)).distinct()
+            return self.filter(get_course_visibility_filter(user) | Q(visible_to_students=True)).distinct()
```
raphendyr, line 341: This lines might be a bit long. Could to add some newlines and try to keep line lengths within 100 or 80 characters. Operator would go in the start of line, thus

```python
return self.filter(
    get_course_visibility()
    | Q(....)
).distinct()
```

Not the perfect. You can also wrap everything in brackets and add new lines accordingly

course/models.py
```diff
@@ -662,8 +659,7 @@ def get_visible(self, user=None):
             )
         if not user.is_superuser:
             return self.filter(
-                get_course_visibility_filter(user, 'course_instance__'),
-                opening_time__lte=timezone.now(),
+                Q(course_instance__visible_to_students=True, opening_time__lte=timezone.now()) | get_course_visibility_filter(user, 'course_instance__'),
```
raphendyr, line 662: Same here, some newlines to make it more readable.

course/models.py
HenrikSundell: Sure, I wasn't sure how to indent it correctly so left it like that :)

course/models.py
```diff
@@ -341,7 +338,10 @@ def get_visible(self, user=None):
         if not user or not user.is_authenticated:
             return self.filter(visible_to_students=True)
         if not user.is_superuser:
-            return self.filter(get_course_visibility_filter(user)).distinct()
+            return self.filter(
+                get_course_visibility_filter(user)
```
raphendyr, line 342: I think the function name could be updated to reflect the "new" role (i.e. doesn't take care of the visible to student anymore).

course/models.py
```diff
@@ -662,8 +662,8 @@ def get_visible(self, user=None):
             )
         if not user.is_superuser:
             return self.filter(
-                get_course_visibility_filter(user, 'course_instance__'),
-                opening_time__lte=timezone.now(),
+                Q(course_instance__visible_to_students=True, opening_time__lte=timezone.now())
```
raphendyr, line 665: These Q + visibility_fitler parts are in different order in these two locations. Might be easier to read of they would have the same order always. Doesn't affect the logic though.

PR #608, opened 2020-07-03
Front page flex-box

### LINKED ISSUE #583: Implementing a flex-box for front page course-sections
The course-items on the front page have a fixed height at the moment which might cause problems in case the course-name or course instance -name are very long. Or very short. 

Therefore it would be good to organize the items with a flexbox-container instead on the frontpage, so that the height of each item on the same course-row would grow if one needs extra space, and the grid would look clean and present all the course headlines clearly.

saarasat:
# Description

In order to better accommodate longer course and course instance -names, the bootstrap3 row-column-grid is removed from front page sections and flex-box is applied instead. This enables the course cards to grow vertically.

Four media-breakpoints have been applied, so that the width of the cards would look good with different screens.

The second commit 57e54963a is only to fix the indendation of the adverts, so that the changes made in the first commit can be more easily spotted.

Fixes #583

**Desktop width**

Screenshot from 2020-07-03 10-17-35

**Tablet width**

Screenshot from 2020-07-03 10-18-48

**Mobile width**

Screenshot from 2020-07-03 10-19-14

# Testing

- The appearance after changes has been tested on Chrome and Firefox and different screen widths.
- The appearance has also been tested with Lambdatest on Edge and IE. 
- It has also been tested for very long and very short course and course instance names
- Appearance has been tested when there are very few course instances and very many instances.
- The appearance has been tested also in case there is a site-wide advert present

Not relevant

### DIFF
--- a/assets/css/main.css
+++ b/assets/css/main.css
@@ -311,12 +311,21 @@ html, body {
     font-size: 18px;
 }
 
+.frontpage-section {
+    display: flex;
+    flex-flow: row wrap;
+    justify-content: flex-start;
+    width: 100%;
+}
+
 .frontpage .card {
-    margin-bottom: 20px;
+    display: flex;
+    flex-direction: column;
+    width: 400px;
     border: solid 1px #ccc;
-    position: relative;
-    height: 355px;
     border-radius: 4px;
+    margin: 8px;
+    margin-bottom: 20px;
     padding: 4px;
 }
 
@@ -325,8 +334,36 @@ html, body {
     border-color: #bce8f1;
 }
 
+@media (min-width: 1200px) and (max-width: 1800px) {
+    .frontpage .card {
+        width: 23%;
+    }
+}
+
+@media (min-width: 960px) and (max-width: 1199px) {
+    .frontpage .card {
+        width: 30%;
+    }
+}
+
+@media (min-width: 670px) and (max-width: 959px) {
+    .frontpage .card {
+        width: 47%;
+    }
+}
+
+@media (max-width: 669px) {
+    .frontpage .card {
+        width: 100%;
+    }
+}
+
 .frontpage a:hover {
     text-decoration: none;
+}
+
+.frontpage .card-clickable {
+    flex-grow: 1;
 }
 
 .frontpage .card-clickable:hover {
@@ -347,8 +384,12 @@ html, body {
     background-color:#eeeeee;
 }
 
+.frontpage .advert-image,
 .frontpage .course-image {
     position: relative;
+}
+
+.frontpage .course-image {
     color: #fff;
 }
 
@@ -375,11 +416,9 @@ html, body {
 }
 
 .frontpage .card-footer {
-    position: absolute;
-    bottom: 0px;
-    left: 0px;
+    padding: 10px 10px 14px 14px;
     width: 100%;
-    padding: 14px;
+    flex-shrink: 0;
 }
 
 .archive-link {
--- a/course/templates/course/_course_cards.html
+++ b/course/templates/course/_course_cards.html
@@ -3,8 +3,7 @@
 
 {% for instance in instances %}
 {% with instance_url=instance|url %}
-<div class="col-sm-6 col-md-4 col-lg-3">
-	<div class="frontpage-course card">
+	<div class="frontpage card">
 		<div class="card-clickable">
 			<a href="{{ instance_url }}">
 				<div class="course-image" aria-hidden="true">
@@ -45,6 +44,5 @@
 			{% endif %}
 		</div>
 	</div>
-</div>
 {% endwith %}
 {% endfor %}
--- a/course/templates/course/index.html
+++ b/course/templates/course/index.html
@@ -36,19 +36,19 @@
 		{% if advert %}
 			<div class="frontpage advert-card card">
 				<a href="{{ advert.href }}" target="_blank">
-						<div class="advert-image" aria-hidden="true">
-							{% if advert.image %}
-								<img class="card-img-top" src="{{ advert.image }}" alt="" />
-							{% else %}
-								<div class="card-img-top">
-									<i class="glyphicon glyphicon-comment"></i>
-								</div>
-							{% endif %}
-						</div>
-						<div class="card-body">
-							<h3 class="card-title">{{ advert.title }}</h3>
-							<p class="card-text">{{ advert.text|safe }}</p>
-						</div>
+					<div class="advert-image" aria-hidden="true">
+						{% if advert.image %}
+							<img class="card-img-top" src="{{ advert.image }}" alt="" />
+						{% else %}
+							<div class="card-img-top">
+								<i class="glyphicon glyphicon-comment"></i>
+							</div>
+						{% endif %}
+					</div>
+					<div class="card-body">
+						<h3 class="card-title">{{ advert.title }}</h3>
+						<p class="card-text">{{ advert.text|safe }}</p>
+					</div>
 				</a>
 			</div>
 		{% endif %}
--- a/course/templates/course/index.html
+++ b/course/templates/course/index.html
@@ -16,9 +16,9 @@
 	<div class="frontpage-panel">
 		<h2 class="panel-heading">{% trans "My courses" %}</h2>
 		{% if my_instances %}
-			<div class="row frontpage">
+			<section class="frontpage frontpage-section">
 				{% include "course/_course_cards.html" with instances=my_instances %}
-			</div>
+			</section>
 		{% else %}
 			<div class="panel-body">
 				<p class="panel-default-text">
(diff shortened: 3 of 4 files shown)


assets/css/main.css
```diff
@@ -325,10 +334,38 @@ html, body {
     border-color: #bce8f1;
 }
 
+@media (min-width: 1200px) and (max-width: 1800px) {
+    .frontpage .card {
+        width: 23%;
+    }
+}
+
+@media (min-width: 960px) and (max-width: 1200px) {
```
mtthwhggns, line 343: I wonder if it would make sense if these matched the Bootstrap ones which affect our other layouts - https://getbootstrap.com/docs/3.4/css/#grid-media-queries

assets/css/main.css
```diff
@@ -325,10 +334,38 @@ html, body {
...
+        width: 100%;
+    }
+}
+
 .frontpage a:hover {
     text-decoration: none;
 }
 
+.frontpage .card-clickable {
+    flex-grow: 1;
```
mtthwhggns, line 366: What happens if this isn't applied?

course/templates/course/index.html
```diff
@@ -16,7 +16,7 @@ <h1>{{ welcome_text|safe }}</h1>
 	<div class="frontpage-panel">
 		<h2 class="panel-heading">{% trans "My courses" %}</h2>
 		{% if my_instances %}
-			<div class="row frontpage">
+			<div class="frontpage frontpage-section">
```
mtthwhggns, line 19: I wonder if [`<section>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) would make more sense semantically here?

assets/css/main.css
saarasat: I thought about this too and actually tried those. I also tried to find information of if there are any newer, "standard" breakpoints, but they seemed to vary among different sources. The breakpoints I used looked better for these course-cards, so that is the reason I went for those. 

I guess this is about whether we want to standardize the breakpoints throughout, with a bit worse appearance on the front page or customize these here.

assets/css/main.css
saarasat: This makes sure that the footer will stick to the bottom of the card, and the clickable part is the one that grows, if there is need for the cards to vertically grow.

course/templates/course/index.html
saarasat: Good point, did not thought of that. Though the `<section>`-tag is not used elsewhere in the code... What do you think of using different tags even though the rest of the code is not using them? If it is ok, I could apply it after I hear I reviews and possible change-ideas of others.

assets/css/main.css
mtthwhggns: Ah, interesting! Thanks

course/templates/course/index.html
raphendyr: > What do you think of using different tags even though the rest of the code is not using them?

If they make the site better, feel free to make the site better :)

mtthwhggns: That's fine!

raphendyr:
If Aalto VDI has IE 11, this change could be checked there. I'm not sure how, but I think the front page should be usable to some extend with that browser.

course/templates/course/index.html
Mankro: Like Matthew has suggested earlier, there are small accessibility issues that can be fixed when someone is working on the relevant code. For example, fixing the semantic markup for sections fits here.

assets/css/main.css
```diff
@@ -325,10 +334,38 @@ html, body {
...
+        width: 47%;
+    }
+}
+
+@media (max-width: 670px) {
+    .frontpage .card {
+        width: 100%;
+    }
+}
+
```
etanttila, line 360: These min-widths and max-widths overlap. In border cases, it seems to use the latest applicable definition. This probably isn't a big problem but it might be better to use restrictions with a difference of a pixel (eg. `max-width: 1199px` and of another:` min-width: 1200` px (or 1200px and 1201px)).

locale/fi/LC_MESSAGES/django.po
```diff
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-06-24 15:50+0300\n"
+"POT-Creation-Date: 2020-07-03 10:03+0300\n"
 "PO-Revision-Date: 2019-08-14 12:16+0200\n"
 "Last-Translator: Saara Satokangas <saara.satokangas@aalto.fi>\n"
```
etanttila, line 11: To my understanding, it's usually unnecessary to update/commit these (as we use git) and their updation easily causes merge-conflicts

course/templates/course/_course_cards.html
```diff
@@ -3,8 +3,7 @@
 
 {% for instance in instances %}
 {% with instance_url=instance|url %}
-<div class="col-sm-6 col-md-4 col-lg-3">
-	<div class="frontpage-course card">
+	<div class="frontpage course-card card">
```
etanttila, line 6: Is this course-card class used for anything? Or what is its intended purpose?

assets/css/main.css
saarasat: That is a good point! Probably won't change much, but definitely should be logical limits. Those have now been changed at: https://github.com/apluslms/a-plus/pull/608/commits/96a174516db506c94d5314fc0d29acccb0d11a0c

locale/fi/LC_MESSAGES/django.po
saarasat: You mean the POT-Creation-Date changes or the translation changes? 

I like to add the changes in translations commit that change them, as I think that the creation-date and who made the changes to translations is important if the state of the branch would be reverted to some specific point (commit) in time. And I also like to ensure that they work before I commit/create a pull request. I guess they could be separated into their own commit at the end of a PR. Though then the branch would not necessarily work after each commit ... hmm.. not sure what is the best way. 

But for example, with my previous pull request, some new translations appeared from other earlier pull requests, since those files had not been committed after their final changes in the related PRs. It looked a bit weird as I hadn't wrote those translations, and at least I try to prevent that from happening.

course/templates/course/_course_cards.html
saarasat: That has become obsolete, you are correct. I removed that.

saarasat:
> Aalto VDI has IE 11, this change could be checked there. I'm not sure how, but I think the front page should be usable to some extend with that browser.

Here are the results of IE 11, and Edge as well. Both seem to work. 

Screenshot from 2020-07-07 09-29-47

image

locale/fi/LC_MESSAGES/django.po
etanttila: I mean the POT-Creation-Date and Last-Translator changes. The translations themselves should definitely be part of the same commit as the things that are changed.
I agree that it's kind of weird that there are new changes that aren't mentioned in the meta-data (and I think it's even more weird that the revision date is before the creation date), but when I've asked @raphendyr about whether to contain those POT-Creation-Dates and Last-Translator changes, he's always instructed to leave them out. He's probably able to give more input on it.

etanttila: I asked him whether to include them or not: "saa, ei oo pakko. pitää keksiä style guideen jotain"

saarasat: Alrighty, that clears it up, thanks :)

saarasat:
@raphendyr @mtthwhggns @etanttila 
Since you all reviewed this PR, would you like confirm if it is okay or is there something you'd like me to add :)?

course/templates/course/index.html
```diff
@@ -31,33 +31,31 @@ <h2 class="panel-heading">{% trans "My courses" %}</h2>
...
+							<img class="card-img-top" src="{{ advert.image }}" alt="" />
+						{% else %}
+							<div class="card-img-top">
+								<i class="glyphicon glyphicon-comment"></i>
+							</div>
+						{% endif %}
+					</div>
+					<div class="card-body">
+						<h3 class="card-title">{{ advert.title }}</h3>
+						<p class="card-text">{{ advert.text|safe }}</p>
```
jaguarfi, line 50: Does this behave differently than the other courses? Why?. Btw, how do you mark a course as  `advert`?. I was trying to find this option, but I could not find it.

saarasat: You mean the advert? That whether that behaves differently? Yeah, it was a bit hard to find at first for me too. But there is a sidewide advertisement that you can place from settings.py and it shows like this on the frontpage. So acts like as one of the course-cards:

image

Teachers can't set that for any course basically, it is only for the site-managers to set.

jaguarfi: Great, I did not know about that option :smile: . And yes, I was asking about the advert card.

saarasat:
Commits squashed, and changes tested one more time just in case :)

PR #615, opened 2020-07-06
Add landmark roles to the template

### LINKED ISSUE #595: Page template is missing a `main` landmark
Specifying a `main` element would be important for accessibility. 

Reference: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/main.html
Related: https://apluslms.github.io/accessibility-audit/patterns/fixable-a/semantic-markup-and-sequences/

mtthwhggns:
# Description

The previous audit highlighted that we have very few landmark roles in use. These are useful for giving semantic strucutre to the page and some assistive technology users will use them to "jump" from element to element around the page (similar to the skip links). 

One particular challenge here was that we need `main` to include the content. On all non-course pages, that's everything after the main navigation, but on course pages it starts after the course navigation. For this reason, there's an optional block which can be defined to override the usual `<main>` tag, for example, with a semantic-free `<div>` or any other element as needed. This is currently only defined for the course template. 

Fixes #595 

# Testing

If you're not familiar with landmarks, I suggest starting with [W3Cs example pages](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/index.html), which can highlight them and also explains how they should be used, then WAVE can show you where they are, and shows them in the "Structure" tab. 

I also suggest reviewers test this manually with a screen reader. 

It might be useful to make a note of how to override the `<main>` somewhere, but we don't yet have a good place for that. 

### DIFF
--- a/course/templates/course/_course_menu.html
+++ b/course/templates/course/_course_menu.html
@@ -17,7 +17,7 @@
 <li role="presentation" class="header">
 	<div class="row">
     <div class="col-sm-10">
-			<h4>{% trans "Course" %}</h4>
+			<h4 id="course-menu-heading">{% trans "Course" %}</h4>
     </div>
     <div class="col-sm">
 			<div class="dropdown h4">
@@ -41,7 +41,7 @@
 </li>
 {% else %}
 <li role="presentation" class="header">
-	<h4>{% trans "Course" %}</h4>
+	<h4 id="course-menu-heading">{% trans "Course" %}</h4>
 </li>
 {% endif %}
 
--- a/course/templates/course/course_base.html
+++ b/course/templates/course/course_base.html
@@ -13,12 +13,20 @@
 {% endfor %}
 {% endblock %}
 
+{% block container_opening_tag %}
+<div class="site-content container-fluid" id="content">                
+{% endblock %}
+
+{% block container_closing_tag %}
+</div>                
+{% endblock %}
+
 {% block title %}{{ course.name|parse_localization }} | {{ block.super }} {% endblock %}
 
 {% block content %}
 <div data-taggings="{{ get_taggings|join:' ' }}"  class="row">
     <div class="col-sm-2 hidden-xs">
-        <nav class="course-menu" id="main-course-menu" aria-label="{% trans "Course navigation" %}">
+        <nav class="course-menu" id="main-course-menu" aria-labelledby="course-menu-heading">
             <ul class="nav nav-pills nav-stacked">
                 {% include "course/_course_menu.html" %}
             </ul>
@@ -60,7 +68,7 @@
             {% endblock %}
         </ol>
         {% endblock %}
-        <div class="row">
+        <main class="row">
             {% block columns %}
             <div class="col-lg-9">
                 {% block coursecontent %}
@@ -71,7 +79,7 @@
                 {% endblock %}
             </div>
             {% endblock %}
-        </div>
+        </main>
         {% block siblings_bottom %}{% endblock %}
     </div>
 </div>
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-07-03 10:03+0300\n"
+"POT-Creation-Date: 2020-07-14 12:07+0300\n"
 "PO-Revision-Date: 2019-08-14 12:16+0200\n"
 "Last-Translator: Saara Satokangas <saara.satokangas@aalto.fi>\n"
 "Language-Team: Finnish <>\n"
@@ -576,6 +576,10 @@ msgstr ""
 "          Palauta ryhmässä: %(collaborators)s\n"
 "          "
 
+#: course/templates/course/_siblings.html:4
+msgid "Pagination"
+msgstr "Sivuilla siirtyminen"
+
 #: course/templates/course/_siblings.html:22
 #: course/templates/course/_siblings.html:45
 #: exercise/templates/exercise/_children.html:28
@@ -590,11 +594,7 @@ msgstr "Aikainen pääsy"
 msgid "Course archive"
 msgstr "Kurssiarkisto"
 
-#: course/templates/course/course_base.html:21
-msgid "Course navigation"
-msgstr "Kurssivalikko"
-
-#: course/templates/course/course_base.html:34
+#: course/templates/course/course_base.html:42
 msgid "Read more"
 msgstr "Lue lisää"
 
@@ -3504,8 +3504,8 @@ msgstr ""
 msgid "Skip main navigation"
 msgstr "Ohita päävalikko"
 
-#: templates/base.html:79
-msgid "Main navigation"
+#: templates/base.html:80
+msgid "Main"
 msgstr "Päävalikko"
 
 #: templates/base.html:83
--- a/templates/base.html
+++ b/templates/base.html
@@ -76,131 +76,139 @@
     <body class="{% if profile %}{% if is_external_student %}external-student{% else %}internal-student{% endif %}{% endif %} lang-{{ request.LANGUAGE_CODE }}" data-view-tag="{% block view_tag %}{% endblock %}">
         <a href="#content" class="skip-link">{% trans "Skip main navigation" %}</a>
         <div class="page-wrap">
-            <nav class="topbar navbar navbar-inverse navbar-static-top" aria-label="{% trans "Main navigation" %}">
-                <div class="container-fluid">
-                    <div class="navbar-header">
-                        <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#bs-navbar-collapse" aria-expanded="false">
-                            <span class="sr-only">{% trans "Toggle navigation" %}</span>
-                            {% notification_count as n_count %}
-                            {% if n_count > 0 %}
-                            <span class="badge badge-danger pull-right">{{ n_count }}</span>
-                            {% endif %}
-                            <span class="icon-bar"></span>
-                            <span class="icon-bar"></span>
-                            <span class="icon-bar"></span>
-                        </button>
-                        <a class="navbar-brand hidden-xs" href="{% url 'home' %}">{% brand_name %}</a>
-                        <span class="navbar-brand visible-xs">
-                        {% brand_name %}{% if instance %} &nbsp; {{ course.name|parse_localization }}{% endif %}
-                        </span>
-                    </div>
-                    <div class="collapse navbar-collapse" id="bs-navbar-collapse">
-                        <ul class="nav navbar-nav hidden-xs">
-                            <li role="presentation" class="dropdown">
-                                <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">
-                                    {% if instance %}
-                                    {{ course|parse_localization }}
-                                    {% if not instance.visible_to_students %}
-                                    <span class="label label-danger">{% trans "Hidden" %}</span>
-                                    {% endif %}
-                                    {% else %}
-                                    {% trans "Select course" %}
-                                    {% endif %}
-                                    <span class="caret"></span>
-                                </a>
-                                {% course_menu %}
-                            </li>
-                        </ul>
-                        <ul class="user-menu nav navbar-nav navbar-right hidden-xs">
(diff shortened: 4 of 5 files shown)


locale/fi/LC_MESSAGES/django.po
```diff
@@ -3529,7 +3529,11 @@ msgstr "Päävalikko"
 msgid "Skip main navigation"
 msgstr "Ohita päävalikko"
 
-#: templates/base.html:83
+#: templates/base.html:80
+msgid "Main"
+msgstr "Pää"
```
raphendyr, line 3534: I think these are bit too short strings to be clear. Specially the Finnish one is confusing. As such, it just means "head" as body part. To mean "main" it usually requires a context. I'm not sure if the location of the aria-label gives enough context, but it's possible.

templates/base.html
```diff
@@ -76,131 +76,139 @@
...
-                                    {% if not instance.visible_to_students %}
-                                    <span class="label label-danger">{% trans "Hidden" %}</span>
-                                    {% endif %}
-                                    {% else %}
-                                    {% trans "Select course" %}
-                                    {% endif %}
-                                    <span class="caret"></span>
-                                </a>
+            <header>
+                <nav class="topbar navbar navbar-inverse navbar-static-top" aria-label="{% trans "Main" %}">
```
raphendyr, line 80: I assume the block is just indented and there are no other changes within, or is there?

templates/base.html
```diff
@@ -76,131 +76,139 @@
...
+                                        {% trans "Log in" %}
+                                    </a>
+                                </li>
+                                {% endif %}
+                            </ul>
+                        </div>
                     </div>
-                </div>
-            </nav>
+                </nav>
```
raphendyr, line 184: However, you can clean the spaces at the end of the line though (probably left from before).

locale/fi/LC_MESSAGES/django.po
mtthwhggns: I'd be inclined to disagree. 

These are only used by assistive technology, and the technology that I've tested always present them in the format `[NAME] [ROLE]`. I didn't test this enough when I implemented these names to start with, so hadn't noticed that we were actually getting things like this before:

Screenshot of the VoiceOver Rotor showing "Main navigation navigation" and some more landmarks with odd names.

So dropping the "navigation" part and leaving only the name seems a little more correct and understandable.

Screenshot of the VoiceOver Rotor showing "Main navigation" and some more landmarks with correct names.

Also, I'd planned on adding an explicit note about this, but the PR is still marked as a draft, so I wasn't expecting any reviews yet.

mtthwhggns: And I double-checked how this happens in Finnish...

Näyttökuva 2020-7-6 kello 13 29 13

And it becomes:

Näyttökuva 2020-7-6 kello 13 31 32

etanttila: Those spaces in Finnish cause me to cringe really, really badly. It looks like someone is horrid at spelling, since those could very well be compound words mostly, but grammatically could not be written like that.

mtthwhggns: I'd assume that things like this are familiar enough to Finnish-speaking regular users of assistive technology though. The second word is the role, so this is the expected behaviour. 

Or do you think we should actually go with `Päänavigointi navigointi` here?

etanttila: (The second screenshot ("pää navigointi" and "kurssi navigointi") is actually giving me serious anxiety ("pää navigointi" and "kurssi navigointi"). The first Finnish version doesn't seem nearly as bad somehow, I guess it's easier to consider the things as separate entities rather than butchered spellings. Of course, if this would never be seen, just read aloud by something such as a screen reader, the second version isn't bad.)

mtthwhggns: Here's vero.fi in Finnish

Näyttökuva 2020-7-6 kello 14 18 35

If you think it's much better, I don't mind going with something like this, but it does feel as though it contradicts the W3C guidelines:

> Avoid using the landmark role as part of the label. For example, a navigation landmark with a label "Site Navigation" will be announced by a screen reader as "Site Navigation Navigation". The label should simply be "Site".

What do you think @etanttila ?

etanttila: I really don't know. I have so little experience with that kind of stuff (I don't even know how to view landmarks on a site). Why can't there be specific guidelines for this in Finnish (and possibly other languages that have compound words)? :/ It would be interesting to hear the thoughts of a person or people who are Finnish speaking and are used to using these landmarks. My gut feeling is that the latter one is quite a bit better (especially in written form), but I assume that what people are actually used to affects which is better.

mtthwhggns: I'll resolve this conversation now. There's potentially more to discuss, but so many things have changed, let's start the conversation again at the appropriate time.

templates/base.html
mtthwhggns: As they were wrapped in `<header>`, the main change here was the indentation.

mtthwhggns:
Following some early feedback and @etanttila 's advice, I've re-written these a little.

Note that we can't find any authoritative guidelines on doing this in Finnish, but took lots of inspiration from the Näkövammaisten liitto website, as such, it now looks like this in Finnish: 

Näyttökuva 2020-7-7 kello 11 30 32

And in English (the Finnish words come from my browser, I was just too lazy to restart it into a different language):

Näyttökuva 2020-7-7 kello 11 34 18

This is ready for a proper review now, and I suggest that @etanttila starts, as she has the freshest experience exploring the guidelines.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -580,6 +576,10 @@ msgstr ""
 "          Palauta ryhmässä: %(collaborators)s\n"
 "          "
 
+#: course/templates/course/_siblings.html:4
+msgid "Pagination"
+msgstr "Sivuilla siirtyminen"
```
raphendyr, line 581: This probably should be iterated at some point, but I guess it's fine. A bit problematic that we add translation for a single word "Pagination", which _could_ be used somewhere else too. For now it is not, thus fine.

etanttila: Could the context be provided?

mtthwhggns: @etanttila do you mean add it as a comment in the locale file?

raphendyr: Technically it could be possible to use actual id as the reference and not English translation, but that requires that there would be another file with the English string too.

etanttila: @mtthwhggns I meant the PO msctxt context, but wasn't very familiar with it. But if it requires that the English string appears somewhere else as well like @raphendyr says, I guess it's "sen ajan murhe".

templates/base.html
```diff
@@ -76,131 +76,139 @@
...
                 {% include "_messages.html" %}
                 {% block content %}
                     <div class="error">
                         No content
                     </div>
                 {% endblock %}
-            </div>
+            {% block container_closing_tag %}
+                </main>                
+            {% endblock %}
```
etanttila, line 199: There seems to be a variety of indentation styles in this file alone (not indenting the content of template blocks vs. indenting). Since most of the changes in this file are indenting the the things put in new tags, you could fix indentation as well if you wish. The guidelines that are hoped for are.
- indent the content of template blocks
- use tabs to indent html-files and django templates

etanttila:
The format that ORCA uses to list the landmarks makes it easier to consider the label and row as separate entities.

kuva

templates/base.html
mtthwhggns: Also, remember that on GitHub's web interface you can append `?w=1` to skip over changes that are only whitespace, like most of the lines here.

course/templates/course/_course_menu.html
```diff
@@ -41,7 +41,7 @@ <h4>{% trans "Course" %}</h4>
 </li>
 {% else %}
 <li role="presentation" class="header">
-	<h4>{% trans "Course" %}</h4>
+	<h4 id="course-menu-heading">{% trans "Course" %}</h4>
```
jaguarfi, line 44: Could you tell me why are you using the same id for all those three elements?. I did not understand these changes.

course/templates/course/_course_menu.html
```diff
@@ -169,7 +169,7 @@ <h4>{% trans "Course staff" %}</h4>
 {% for group in course_menu.staff_link_groups %}
 {% if group.label %}
 <li role="presentation" class="header">
-	<h4>{{ group.label|parse_localization }}</h4>
+	<h4 id="course-menu-heading">{{ group.label|parse_localization }}</h4>
```
etanttila, line 172: Since this is in a for-loop, is it possible that this id would appear several times on the resulting page?

course/templates/course/_course_menu.html
mtthwhggns: If you look at the file as a whole, there is an `if` on line 16 which has the part including lines 20 and the part including line 44 as two options. It renders two different things depending on whether the course is multilingual or not. As a result, the the template only renders one of these options.

The on included on line 172 was added mistakenly, and has now been removed.

course/templates/course/_course_menu.html
mtthwhggns: This was added my mistake, thanks for spotting!

course/templates/course/_course_menu.html
jaguarfi: mmm, this comment was published before I finished the whole review. Anycase, it looks fine now.

course/templates/course/course_base.html
```diff
@@ -13,12 +13,20 @@
...
 {% endblock %}
 
+{% block container_opening_tag %}
+<div class="site-content container-fluid" id="content">                
+{% endblock %}
+
+{% block container_closing_tag %}
+</div>                
+{% endblock %}
+
```
jaguarfi, line 23: Why are you using one block for the opening tag and another for the closing tag? is not it better to use only one block?

mtthwhggns: I couldn't see how to make them overridable in Django's templating language while preserving whatever children they have. It's probably possible, but this was the only way I managed to get it to work after a couple of hours of Googling.

Can you provide a minimal example of your approach?

templates/base.html
```diff
@@ -76,131 +76,139 @@
...
-                </div>
-            </nav>
+                </nav>
+            </header>
 
-            <div class="site-content container-fluid" id="content">
-                {% site_alert %}
+            {% block container_opening_tag %}
+            <main class="site-content container-fluid" id="content">                
+            {% endblock %}
```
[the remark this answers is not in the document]
mtthwhggns: You seem to have addressed this in your next comment.

course/templates/course/course_base.html
jaguarfi: I do not have an example, I was only curious about what you wanted to accomplish with this opening and closing blocks. Now, it is more clear, and I think it is fine if you do it in this way.

jaguarfi:
**Question** for @mtthwhggns: Now, that you have added the `<main>` tag to the base template, and after reading the websites you linked to this PR. It seems that these changes will improve the "semantic structure" of the course page. But, it also makes me think that at some point the templates will need to be completely restructured in order to improve the accessibility. Do you think we need to create an issue for restructuring the templates or is it fine to do this type of small fixes all around. What would you suggest?

mtthwhggns:
@jaguarfi it is my hope that we can meet all the basic requirements with lots of small changes that are achievable this summer. 

In the long term, it would probably be easier to maintain (especially in accessibility terms) if we had a slightly different template structure overall, but I'm not currently sure what that structure would be. It's probably worth considering when it's time to replace a bigger part of the platform.

jaguarfi:
Yes, I understand the time limitation we have now. But I only wanted to hear your opinion about this issue. So I think you can fix the conflicts and squash the commits

mtthwhggns:
Conflicts resolved.

jaguarfi:
- typo in your commit message - **elswhere**
  ```
  Pages didn't have a main landmark, and others were labelled
  incorrectly or unclearly.

  This:
  - Adds a main landmark - with a way to override it if you want to place
    it elswhere
  - Labels all "navigation" landmarks
  - Swaps the footer to a "contentinfo"
  ```
* There is still a conflict in this PR `locale/fi/LC_MESSAGES/django.po`

mtthwhggns:
@jaguarfi Rebased again and fixed the typo. 

For the record, there wasn't _still_ a conflict, there was a _new_ conflict after PR #608 got merged.

Issue #518, opened 2020-04-16
In A+ it should be possible to set student's default language in a specific course


jaguarfi:
> **Migrated from Trello**

- [x] Add language to the userprofile - #617
- [ ] When user is viewing material with different language, make it clear that it's not their default, e.g., show a warning.

    Additionally, buttons (links) should be shown to set that language as their preferred language (to enrollment object) and to switch back to their preferred language.

    This is more about the language set via url (query parameter), so that when the user is directed to a material in a different language than in their enrollment object (or profile), then that would be pointed out to them.

- [x] Store language in the Enrollment object. Think about the case where system has 3 languages sv, en, and fi, user has set the language in profile to sv and course is provided in en and fi. In that case, A+ should be able to remember the language the user likes to read that course and not force to set profile language. - #632

----
**Comments**

>**Jaakko Kantojärvi:**
At least for a profile, so when student logs in, the language is returned to original one.

Mankro:
The UserProfile already has a `lang` field: https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/userprofile/models.py#L37-L38
Is it used anywhere? 

Like the issue says, when the student comes to a multilingual course, the course should be by default displayed in the language set in the userprofile. It would be good to check which language is now used in the old code. Is it random, or is it based on the HTTP language headers?

If the user views the course in other language than what is set in the profile, we need to careful about making too obnoxious warnings about it. If the user really wants to view the other language, then they should not be constantly bombarded with warnings about the language not being their default.

This issue can be done after https://github.com/apluslms/a-plus/issues/520 (different language URLs for multilingual courses).

raphendyr:
> It would be good to check which language is now used in the old code. Is it random, or is it based on the HTTP language headers?

Django session (or actually a separate cookie) contains the language information. It's initialized with the HTTP language. I didn't remember we had DB reservation for the language, but I don't think it is used yet at all.

Current behavriour is that the user can have multiple sessions with different languages. We should replace this with the user linked language. The language selection in profile page would make more sense then.

veeraalt:
> This issue can be done after #520 (different language URLs for multilingual courses).

I think this issue needs to be done before #520 because #520 affects how the language is selected, and now the language is stored in the wrong place (in a cookie). The lang field in UserProfile is not in use now, so it would make sense to fix this first.

Mankro:
[...]

PR #617, opened 2020-07-09
Save language to userprofile

veeraalt:
# Description

Previously, when a user changed the language on their profile or on a course page, language selection was saved in a cookie and `lang` field in `UserProfile` object was not in use. This PR changes this logic so that when logged-in user changes language in their profile or on a course page, it is saved to the `language` field in their `UserProfile` object, and used as default language for displaying A+ content, instead of saving the language to a cookie. For users who are not logged in, language is still saved to a cookie when changed on course page.
This is implemented by:
  * Creating a new middleware class `LocaleMiddleware` for overriding Django's default `LocaleMiddleware` ([`django.middleware.locale.LocaleMiddleware`](https://docs.djangoproject.com/en/1.11/_modules/django/middleware/locale/)).
  * Creating `set_user_language` function that overrides `set_language` function in [`django.views.i18n`](https://docs.djangoproject.com/en/2.2/_modules/django/views/i18n/)
  * Deleting `lang` field and adding `language` field in `UserProfile` and setting it blank as default. Previously default was `en_US` which didn't make a lot sense.

1/3 of issue https://github.com/apluslms/a-plus/issues/518

# Testing

* Existing unit tests pass

* Testing has been done manually by checking that the selected language is saved to user's UserProfile object when changed on profile page or on course page (this is changed in a later PR so that language selection on course page changes language of Enrollment object instead of UserProfile object, see https://github.com/apluslms/a-plus/issues/518), and that A+ content is correctly displayed in the selected language. It has also been tested that the selected language is saved in a cookie and content displayed correctly when user is not logged in.

* No, I don't think it's relevant.


aplus/middleware.py
```diff
@@ -0,0 +1,23 @@
+from django.conf import settings
```
raphendyr, line 1: The `lib/` is the location for files like these, even though the folder is badly named.

userprofile/views.py
```diff
@@ -1,10 +1,17 @@
...
 from django.contrib.auth.views import LoginView
 from django.core.cache import cache
 from django.core.cache.utils import make_template_fragment_key
+from django.http import HttpResponse, HttpResponseRedirect
 from django.template.loader import TemplateDoesNotExist, get_template
-from django.utils.translation import get_language
+from django.urls import translate_url
+from django.utils.http import is_safe_url
+from django.utils.translation import (
+    LANGUAGE_SESSION_KEY, check_for_language, get_language
```
raphendyr, line 13: Please, add a line break after each comma.

aplus/settings.py
```diff
@@ -174,6 +174,7 @@
     'social_django.middleware.SocialAuthExceptionMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'aplus.middleware.LocaleMiddleware',
```
raphendyr, line 177: I think this should be as early as possible (i.e. just after we have request.user). In addition, I think you should drop the LocaleMiddleware, so only our version is used.

veeraalt: > In addition, I think you should drop the LocaleMiddleware, so only our version is used.

Is the `process_response` function of django's  `LocaleMiddleware` (https://docs.djangoproject.com/en/1.11/_modules/django/middleware/locale/) not in use or what should we do to it?

raphendyr: I think the part

```python
patch_vary_headers(response, ('Accept-Language',)) # no path stuff
if 'Content-Language' not in response:
    response['Content-Language'] = language
```

would be relevant.

veeraalt: Ok, I'll check what is needed from that and add it to our custom middleware.

lib/middleware.py
```diff
@@ -0,0 +1,31 @@
...
+from django.utils.deprecation import MiddlewareMixin
+
+
+class LocaleMiddleware(MiddlewareMixin):
+
+    response_redirect_class = HttpResponseRedirect
+
+    def process_request(self, request):
+        urlconf = getattr(request, 'urlconf', settings.ROOT_URLCONF)
+        i18n_patterns_used, prefixed_default_language = is_language_prefix_patterns_used(urlconf)
```
raphendyr, line 15: I think you can drop the language prefix / path stuff from the process_request too

lib/middleware.py
```diff
@@ -11,15 +10,11 @@ class LocaleMiddleware(MiddlewareMixin):
...
-        i18n_patterns_used, prefixed_default_language = is_language_prefix_patterns_used(urlconf)
-        language = translation.get_language_from_request(request, check_path=i18n_patterns_used)
-        language_from_path = translation.get_language_from_path(request.path_info)
-        if not language_from_path:
-            if request.user.is_authenticated:
-                language = request.user.userprofile.lang
-            elif not prefixed_default_language and i18n_patterns_used: 
-                language = settings.LANGUAGE_CODE
+        language = translation.get_language_from_request(request)
+        if request.user.is_authenticated:
```
raphendyr, line 13: Maybe also check that `request.user` exists, might be. I guess it would make sense to check how django code middlewares handle that internally. (i.e. do they raise an exception if there isn't one or do they have a check `hasattr(reqyest, 'user')`)

userprofile/views.py
```diff
@@ -47,6 +56,48 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class LanguageView():
+    """"This language view class overrides set_language function from 
+    django.views.i18n."""
+        
+    def set_language(request):
+        LANGUAGE_QUERY_PARAMETER = 'language'
```
raphendyr, line 64: I think it would be a good idea to rename this so that there isn't word `QUERY` in it. It's likely that when removing the cookie, the language is set as a query parameter (e.g. `url + '&hl=' + lang`).

lib/middleware.py
veeraalt: I think that `request.user` always exists when there is an authentication system, it's just `AnonymousUser` when not logged in.

https://docs.djangoproject.com/en/3.0/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated

veeraalt:
@raphendyr What do you think about fixing this part in this pull request? I tested it briefly and it would be a pretty easy and quick fix (and maybe harder to fix later when the `userprofile.lang` field is actually in use) but I'm not sure if it's in this PR's scope.

I think it would make sense to change the field name from `lang` to `language` and set it blank as default, or at least the default could come from `settings.py`, doesn't make sense to use `en_US` as default.

https://github.com/apluslms/a-plus/blob/d24046b2acfa9366155d9821b610be89546b7657/userprofile/models.py#L37-L38

lib/middleware.py
raphendyr: I was thinking django would do something like this https://github.com/django/django/blob/master/django/contrib/auth/middleware.py#L17

raphendyr:
> I think it would make sense to change the field name from lang to language and set it blank as default

Sounds reasonable and the argument that it's harder later is valid. Assuming the field isn't used anywhere for now, you can just drop the old field and add a new, thus purging the `en_US` from the database.

veeraalt:
Thanks, I'll do that.

lib/middleware.py
veeraalt: So I should add that also to my `process_request`?

raphendyr: I guess it would be a bit nicer way to crash, than an AttributeError

userprofile/urls.py
```diff
@@ -12,5 +12,6 @@
         name="privacy_notice"),
     url(r'^accounts/$', views.ProfileView.as_view(),
         name="profile"),
+    url(r'^i18n/setlang', views.LanguageView.set_language),
     url(r'^i18n/', include('django.conf.urls.i18n')),
```
raphendyr, line 16: I think you can drop this entyr, as it only adds the same url as above https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/conf/urls/i18n.py#L38

veeraalt: I tried dropping it but the request seems to do nothing after that, because the request can't be connected to the right view i guess.

raphendyr: Remember the `name` parameter as seen in the django source.

veeraalt: Do you mean to drop the last url, because I don't understand how it works if I drop the `/setlang` url, like combine it with the `/accounts` url above?

This seems to work:
`url(r'^setlang', views.LanguageView.set_language,
     name="set_language")`

or this:
`url(r'^i18n/', views.LanguageView.set_language, name="set_language")`

raphendyr: This `include('django.conf.urls.i18n')` shouldn't be needed at it only added `/accounts/i18n/setlang`, which you overwrote in the line above it, hence making the include line irrelevant (except it did add name for that url).

veeraalt: Ok now I got it!

userprofile/urls.py
```diff
@@ -12,6 +12,5 @@
         name="privacy_notice"),
     url(r'^accounts/$', views.ProfileView.as_view(),
         name="profile"),
-    url(r'^i18n/setlang', views.LanguageView.set_language),
-    url(r'^i18n/', include('django.conf.urls.i18n')),
+    url(r'^setlang', views.LanguageView.set_language, name="set_language"),
```
raphendyr, line 15: These could be named `set_user_language` and `set_enroll_language`, so it would be always clear on which action is supposed to happen. In the future, `set_language` will be a bit ambiguous.

userprofile/urls.py
```diff
@@ -12,5 +12,5 @@
         name="privacy_notice"),
     url(r'^accounts/$', views.ProfileView.as_view(),
         name="profile"),
-    url(r'^setlang', views.LanguageView.set_language, name="set_language"),
+    url(r'^setlang', views.set_user_language, name="set_user_language"),
```
raphendyr: Could you still change the name to `set-user-language`, so the url name format would match all other url names in the code base.

veeraalt: I did, check the last commit.

raphendyr: Sorry, github notifications got me on an old page.. :disappointed:

userprofile/models.py
```diff
@@ -34,8 +34,7 @@ def get_by_request(cls, request):
         raise RuntimeError("Seeking user profile without authenticated user.")
 
     user = models.OneToOneField(User, on_delete=models.CASCADE)
-    # FIXME: refactor lang to selected_language which by default is blank
-    lang = models.CharField(max_length=5, default="en_US")
```
raphendyr, line 38: How cool, this was added when the initial code was added to git in 2012... now after 8 years it actually got implemented :smiley:

raphendyr:
I guess I forgot this PR. @veeraalt is there any changes from later development, which would make sense to merged as part of this PR?

veeraalt:
I don't think there is. This should be fine for merging as is.

PR #618, opened 2020-07-13
Questionnaire feedback - changes in a-plus

### LINKED ISSUE #490: Better visual highlighting (and/or other UX) of the feedback that students receive for  their answers to a questionnaire
> **Migrated from Trello**

saarasat:
# Description

Related pull request in mooc-grader: https://github.com/apluslms/mooc-grader/pull/76

**1. Separates the css of a questionnaire**
-  The questionnaire can be targeted more easily from aplus or by customization of an individual course

**2. Improves the appearance of hints**
- Using similar styling as the bootstrap help-block-class to show the questionnaire hints instead of unordered list 

**3. Connect the hints to checkboxes** 

- Links hints of a checkbox-question to each related option 
- Shows common hints (that are shown when some answer is NOT selected) below the checkbox-question 
- If some hint is already shown below an option, that is not duplicated below to minimize confusion 

**4. Separates the handling of validation errors and wrong answers by students** 

- Validation errors now use the “has-error” class of bootstrap and have been separated into their own file, which can be used consistently later on 
- Validation errors are shown with bolded form-control, question-label, the related validation error message and a red color, since yellow was already in use for the focus state. 
- If the answer is wrong (apart from checkboxes), the wrong answer is indicated by a red cross and the possible hint.  
- If the answer is correct, it is indicated by a green check mark and the possible hints.  
- Messages for correct and wrong answers have been added: "Correct!", "Incorrect" 
- The chosen options have been highlighted in the result-form. This helps the student to see which options were chosen with the previous result.  

**5. Cleans up the way correct answers are shown after max-submissions have been reached**
- There is a directive `:reveal-model-at-max-submissions: true` which can be set for questionnaires
- This enables showing the correct answers related to the questionnaire
- This pull request cleans the way those are shown and combines them with the other feedback (hints, default messages for correctness)
- The correct answers are only shown at the moment when max-submissions are reached (attribute data-aplus-once="yes" in the html-element), and not visible anymore after the page has been refreshed. This logic has not been changed with the pull request.

**6. Adds helpful comments to the html-template of a questionnaire to ease up the future development** 

**This pull request does NOT:** 

- Yet remove any other additional feedback that the questionnaire provides. Next step would be to remove the alert indicating the points and the mainly empty feedback-modal. These could be done now that the feedback otherwise has been improved. Changes in this PR still work alone as well.  
- Touch the point-badges, those too would be taken another look afterwords  
- Remove anything from the python-side, in order to try to ensure that old yaml-confs which cannot be now tested will still work. 
- Affect the way a questionnaire is graded, the changes are done only to improve the indication of correct and wrong answers and their hints better 
- YET fix the indentation of the graded form, so that changes are more noticeable. Indentation will be fixed once the pr becomes mergeable.    

Fixes #490 

# Testing

- The automatic tests have been run in aplus and mooc-grader, they both work.  
- The outlook has been tested in Chrome and Firefox 
- Testing requires both, the changes in aplus and mooc-grader to be mounted in a container. 

**Testcases:**

1. Checkbox-question with no hints, got wrong and right 
2. Checkbox-question with common hints and option-related hints, got wrong and right 
3. Checkbox-question with validation errors, field not answered even when required 
4. Radio-question with no hints, got wrong and right 
5. Radio-question with hints, got wrong and right
7. Drop-down, with no hints, wrong and right 
8. Text-type-question with hints, got wrong and right 
9. Text-type-question with no hints, got wrong and right 
10. Text-type-question with validation errors, field has a string when int is needed and a case where the required answer is missing 
11. Model-answer 
12. Individual submission 
13. :reveal-model-at-max-submissions: true

### Hints:

Before-multiple-choice-wrong-answers

### Validation errors:

Before-free-text-A-simple-multiple-choice-questionnaire-validation-error

### Correct answers:

image

### Hints and correctness-feedback for checkboxes and radio-types

image

### Text-boxes showing hints

1b-text-with-hints

### Validation errors

1a-Validation-text

image

### Correct answers at max-submissions

1b-reveal-checkbox

1d-Displaying-the-correct

## Test cases

<code>
.. questionnaire:: questionnaire_demo 70
  :title: A simple multiple-choice questionnaire
  :submissions: 3
  :points-to-pass: 20

  .. pick-any:: 10

    Subdirective ``pick-any`` defines a multiple-choice question.
    When :math:`(x + 1)^2 = 16`, what is :math:`x`?

    a. 4
    *b. an integer
    *c. 3
    d. an irrational number
    e. -3
    *f. -5

  .. pick-any:: 10

    Subdirective ``pick-any`` defines a multiple-choice question.
    When :math:`(x + 1)^2 = 16`, what is :math:`x`?

    a. 4
    *b. an integer
    *c. 3
    d. an irrational number
    e. -3
    *f. -5

    a § Rather close. Remember that you can add or subtract the same number to the both sides of the equation.
    !b § This is a common hint that is shown when -5, 3, or integer is not chosen
    !c § This is a common hint that is shown when -5, 3, or integer is not chosen
    d § No. This equation has a nice and easy solution.
    e § Remember that :math:`x^2 = q \leftrightarrow x = \pm \sqrt{q}`
    !f § This is a common hint that is shown when -5, 3, or integer is not chosen
  
  .. pick-any:: 10

    Subdirective ``pick-any`` defines a multiple-choice question.
    When :math:`(x + 1)^2 = 16`, what is :math:`x`?

    a. 4
    *b. an integer
    *c. 3
    d. an irrational number
    e. -3
    *f. -5
  
  .. pick-any:: 10
    :required:

    Subdirective ``pick-any`` defines a multiple-choice question.
    When :math:`(x + 1)^2 = 16`, what is :math:`x`?

    a. 4
    *b. an integer
    *c. 3
    d. an irrational number
    e. -3
    *f. -5

    a § Same option has multiple hints
    a § So this is the second one

  .. pick-one:: 10
    :dropdown:
  
    Subdirective ``pick-one`` defines a single-choice question.
    When :math:`(x + 1)^3 = 27`, what is :math:`x`?

    a. 9
    *b. 2
    c. 3

  .. pick-one:: 10
  
    Subdirective ``pick-one`` defines a single-choice question.
    When :math:`(x + 1)^3 = 27`, what is :math:`x`?

    a. 9
    *b. 2
    c. 3

    a § Not quite. Remember the cube root.
    b § Correct!
    c § Rather close. Remember that you can add or subtract the same number to the both sides of the equation.
    !b § Are you sure?
  
  .. pick-one:: 10
    :required:
  
    Subdirective ``pick-one`` defines a single-choice question.
    When :math:`(x + 1)^3 = 27`, what is :math:`x`?

    a. 9
    *b. 2
    c. 3

    a § Not quite. Remember the cube root.
    b § Overriding the default correct-message
    c § Rather close. Remember that you can add or subtract the same number to the both sides of the equation.

</code>

<code>
.. questionnaire:: questionnaire_text_demo 20
  :title: A simple multiple-choice questionnaire
  :submissions: 3

  .. freetext:: 5
    :length: 10

    This is the most basic free text questionnaire. The correct answer is
    ``test``. You can write at most 10 characters into the box.

    test § This is a message overriding the correct message
    !test § Follow the instruction.

  .. freetext:: 5 int
    :length: 7

    The answer can be a number, an integer. What is :math:`3 + 8`?

    11

  .. freetext:: 5 float
    :length: 7

    The answer can also be a decimal number (floating point number).
    What is :math:`3 / 8` in decimal? (When the question uses the float type,
    the grader accepts also answers that slightly differ from the model solution.)

    0.378
    !0.378 § Hint: the answer is between 0 and 1. Use the decimal point and write three first decimals, for example, ``0.924``.

  .. freetext:: 5 
    :length: 7
    :height: 4

    The text-area can be larger as well. Here the correct answer is "text-area"

    text-area This is a message for correct answers, overriding the default
    !0.378 § Hint: the answer is between 0 and 1. Use the decimal point and write three first decimals, for example, ``0.924``.
</code>

Course-templates-repo will need some updating if this pull request is merged. 


assets/css/main.css
```diff
@@ -568,6 +568,56 @@ td.submissions-dropdown {
...
+.exercise .aplus-quiz1 .has-error .form-control {
+    border-width: 2px;
+}
+
+.exercise .aplus-quiz1 .has-error .radio,
+.exercise .aplus-quiz1 .has-error .checkbox {
+    color: #333;
+}
+
+.exercise .aplus-quiz1 .help-block {
```
mtthwhggns, line 596: I'm concerned about the same style being applied to both the option-specific feedback and the whole-question outcome in a multiple choice (multiple correct answers) situation.

localhost_8000_def_current_m01_introduction_07_questionnaires_ (2)

I'm wondering if this is confusing, as the bottom of the list of options includes both "Correct" and "not correct" and they're given equal prominence.

Can you think about how to distinguish them? One idea might be adding a margin and border to the left of the option-specific feedback, so that it appears indented underneath the option, but that's not the only possible solution. 

localhost_8000_def_current_m01_introduction_07_questionnaires_ (3)

saarasat: I actually tried precisely what you proposed there :D But if there was a lot of option-specific hints, it became quite messy with so many different icons. So I tried to keep them to the minimum.  

And I think you do not have the current styles from aplus running, as there the feedback for options, and about the question correctness have been at least somewhat separated :) For wrong answers the cross is red. And for correct answers the check-mark is green, and the feedback-text (eg. "Correct!") is darker, exactly for that purpose. 

Now that you mentioned this, I added the darkness for the feedback-text for wrong answers as well. So the final looks like this:

image

raphendyr: I think it looks pretty nice and clear. However, is there enough separation between what is the question/answer and what is the grade or feedback. In contrast, on a pen and paper your answer is in black and the feedback/grading in red, thus creating super clear mental model what is what. At least for me, the hints look like a help text for the option, i.e, something that exists before selecting anything. Some of your drafts had a dark background for these, but that is probably too much.

In addition, is the best location for correct/wrong after the question or should it be closer to the points badge to explain that (we currently use only color for the correctness there)? It would be even possible to render the checkmark over the whole question with x20 bigger font, so it would look like it was done with a pen. Probably not a good idea though.

Is a text "your answer to this question is not correct" a bit long way to explain "incorrect"? Not a focus of the PR, so no need to address here. Just raised a thought.

assets/css/main.css
```diff
@@ -568,17 +568,60 @@ td.submissions-dropdown {
...
+.exercise .aplus-quiz1 .help-block-checkbox {
+    padding-left: 20px;
+}
+
+.exercise .aplus-quiz1 .has-error .help-block {
+    font-style: bold;
+    margin-bottom: 0px;
+}
+
+.exercise .aplus-quiz1 .glyphicon-ok {
```
raphendyr, line 609: Could we copy the glyphicon definition here and name this to something else like `.exercise-icon-correct` and another `.exercise-icon-incorrect` or something in those lines. All of these classes are going to be part of the aplus protocol API, so they will stick and we need to keep supporting them in the future, so I would prefer names that require less changing later.

saarasat:
> Can you check all the class names that they would best describe what they are and if they don't fit fully, you can copy more of the bootstrap etc. stuff so we can use a different name. For example, help-block vs. hint and so on.
> 
> All of the classes here will be part of the API, thus once introduced, they can't be renamed (at least easily).

Yes, sure! I was not aware of this need, but of course I can do it. Is there are part in the main.css or somewhere that this has been done well, so I could maybe follow that example / naming convention?

raphendyr:
> Yes, sure! I was not aware of this need, but of course I can do it. Is there are part in the main.css or somewhere that this has been done well, so I could maybe follow that example / naming convention?

You get to come up with some good names, as this hasn't been done before. No need to address everything here. Also, the `aplus-quiz1` is part of this thing, so everything tagged with it is inteteresting for now. You can consider bootstrap3 to be part of the protocol (I called this API before, but that isn't exactly a good term), however I would prefer it not to be in a long run. Hence, when bootstrap names don't make sense, then feel free to rename those.

I don't expect this to go perfectly or all the classes to be named, but if we get 40% in the right direction, I'm happy.

assets/css/main.css
saarasat: Hmm.. That is (those are) valid points to think about.

I tried those darker backgrounds, but they became very problematic with longer hints :/ Then there were just gray lines all over the place. I also tried that darkened background with just the bottom-part, but then I could not create a consistent look between checkbox-questions and other types, so I gave that up. It was also very hard to then refer to options with very long labels from that general-feedback-part. This should support longer hints and longer option-labels.  

I also would hinder from adding too much and too big extra-elements here. I think the questionnaire was quite cluttered as it is, so my instinct would be to remove extra information and not add too much extra styles there. And to try to keep it simple and suitable for different cases. Some questionnaire have very long option texts, some have very short ones, some have a lot of options, some have only few options, some have hints, some have no hints, some have only common hints, and some have only option-specific hints, some have neutral options, some have partial points, some have model-answers. and there the added complexity of the old yaml-confs that I am not able to fully test... I also tried that version were the text about correctness was next to the question, below the description or above it. But if there was an image related to the question, or the question text itself was very long, then that looked funny, and the text kind of melted into the description.

BUT, I have made these improvements now that should somehow try to answer to both of your concerns :)  

1. The crosses and checkmarks are no longer used with hints. They are there just to indicate whether the answer is correct or false, with the text "Correct!" or "Incorrect", for a consistent look and with the aim for that mental model. I think this combination of text and the clear icon, would suit for accessibility purposes as well. 

2. Hints are presented with the lighter gray color, and are in italics. This has been done to separate them from the outcome of the questionnaire. 

3. If there is a correct model answer, it is shown instead of hints. No point in showing hints if there is already the model answer showing.  

4. The classes have been checked and should reflect what is the actual content rather than the bootstrap-class. Basically anything related to feedback can now be hidden or changed if wanted, by accessing the related css-class: option-related hints, general hints and general feedback of the questionnaire. 

5. Coloring of glyphicons has been done like they were done before. So if the icon is changed to something else in the mooc-grader-end, it can be done without touching the aplus-side.

assets/css/main.css
```diff
@@ -568,17 +568,63 @@ td.submissions-dropdown {
...
+.exercise .aplus-quiz1 .has-error .radio,
+.exercise .aplus-quiz1 .has-error .checkbox {
+    color: #333;
+}
+
+.exercise .aplus-quiz1 .option-checked {
+    font-weight: bold;
 }
+
+.exercise .aplus-quiz1 .hint {
```
raphendyr, line 600: These are really good names. However, could the `hint` be more precise like `question-hint` or such?

assets/css/main.css
```diff
@@ -568,17 +568,63 @@ td.submissions-dropdown {
...
+    margin-bottom: 5px;
+    color: #737373;
+    font-style: italic;
+}
+
+.exercise .aplus-quiz1 .hint-checkbox {
+    padding-left: 20px;
+}
+
+.exercise .aplus-quiz1 .icon-correct {
```
raphendyr, line 612: I would prefer these classes to contain all the details, so that the exercise service could just write `<i class="icon-correct" aria-hidden="true"></i>`. Hence, something like:

```css
.icon-correct {
    color: #5cb85c;
    /* glyphicon */
    position: relative;
    top: 1px;
    display: inline-block;
    font-family: 'Glyphicons Halflings';
    font-style: normal;
    font-weight: 400;
    line-height: 1;
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
}
.icon-correct:before {
    content: "\e013";
}

Comment (raphendyr) [reply assets/css/main.css:612]: Note that the glyphicon part would be easy to include with `@import` (or `@use`, I'm not sure right now) in SASS.

Comment (raphendyr) [reply assets/css/main.css:612]: Hmm.. Also, should these be `question-icon-ok` or `quiz1-icon-ok` :thinking:

File: assets/css/main.css (hunk 1 above)
Comment (raphendyr) [reply assets/css/main.css]: > and are in italics

I think this was simple and very effective. At least for me hints are now clearly separate from the other things.

Thanks. I think you found a really good solutions for the concerns and the questions now look pretty clear and cool.

File: assets/css/main.css (hunk 3 above)
Comment (saarasat) [reply assets/css/main.css:600]: Sure, I can change that!

File: assets/css/main.css (hunk 4 above)
Comment (raphendyr) [reply assets/css/main.css:612]: The generic stuff like form-group, form-control makes sense to support, glyphicons are something that might be interesting to replace with fontawesome at some point.

With the prefixes, I was thinking that they might easily collide with some other classes somewhere else. It would be wery possible to add classes like `icon-*` for some totally different purpose, and then those would effect these too. Basically, it would be beneficial if _all_ the classes would have `aplus-quiz1-` prefix, but that becomes very annoying.

Comment (raphendyr) [reply assets/css/main.css:612]: This doesn't seem to be well understoond problem in the internet. Bootstrap ignored to problem and materialui has an open issue about it. Some small and unknown stackoverflow about this https://softwareengineering.stackexchange.com/a/351554 .

Basically, every library whih adds css classes, should namespace those. In our case, that means the protocol support should namespace all classes. However, I assume that not so common words also work for the purpose well enough.

Comment (saarasat) [reply assets/css/main.css]: This is interesting, and now I definitely understand the issue better. I changed those, and added the quiz1-prefix for the icons.

**Comment by saarasat** [comment]
@raphendyr Just let me know then if there is something here with these two pull requests, that you would like me to take a look at or test, or whether these could be rebased and commit messages re-written :)

**Comment by raphendyr** [comment]
@saarasat I like to merge #630 first, as that allows implementing css in this PR better (due to SASS). Only things left are the string `hint`, which is probably too generic and I would prefer only a single class for the icons (i.e. not `icon icon-ok` but just `icon-ok`), which is easier to do with SASS include.

**Comment by saarasat** [comment]
> @saarasat I like to merge #630 first, as that allows implementing css in this PR better (due to SASS). Only things left are the string `hint`, which is probably too generic and I would prefer only a single class for the icons (i.e. not `icon icon-ok` but just `icon-ok`), which is easier to do with SASS include.

Okay great! I talked with Matthew about that SASS-implementation. I will then move around the css if necessary once that one has been merged.

And actually I already changed the `hint` --> `question-hint` a couple of days ago here: https://github.com/apluslms/a-plus/pull/618/commits/cea184c091fac987e1bfb374c54931d894b3e5e2 . I also added then the quiz1-prefixes to the icons like you suggested, so that they are not too general in case the icon-class would be used elsewhere. 

I now changed the icons to have only one class, it creates quite a bunch of copy-paste with the glyphicons, but if it is better for the protocol and SASS, then it is quite understandable :) https://github.com/apluslms/a-plus/pull/618/commits/485a2a641e3e3873cae57255c376ce09cd56597c

**Comment by mtthwhggns** [comment]
> I now changed the icons to have only one class, it creates quite a bunch of copy-paste with the glyphicons, but if it is better for the protocol and SASS, then it is quite understandable :) 485a2a6

To clarify, I think the idea here was that this could be achieved more effectively with a SASS mixin, eliminating this copy/paste stuff. You could add all of the repeatedly used styles to a mixing, then include them in multiple other classes, take a look at [this example from the SASS. documentation](https://sass-lang.com/documentation/at-rules/mixin).

**Comment by raphendyr** [comment]
Now that #633 is merged, the main.css changes here should be converted to sass variant. Here is an idea:

```patch
diff --git a/assets/sass/main.scss b/assets/sass/main.scss
index 5204a67..6bea5c3 100644
--- a/assets/sass/main.scss
+++ b/assets/sass/main.scss
@@ -22,3 +22,4 @@ $brand-primary: #005EB8;
 /*!
  * A+ Styles
  */
+@import 'protocols/quiz1';
diff --git a/assets/sass/protocols/_quiz1.scss b/assets/sass/protocols/_quiz1.scss
new file mode 100644
index 0000000..257f76d
--- /dev/null
+++ b/assets/sass/protocols/_quiz1.scss
@@ -0,0 +1,92 @@
+// FIXME: find a better place for this
+
+@mixin glyphicon {
+  position: relative;
+  top: 1px;
+  display: inline-block;
+  font-family: "Glyphicons Halflings";
+  font-style: normal;
+  font-weight: 400;
+  line-height: 1;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+
+/* Mooc-grader questionnaire */
+
+.exercise .aplus-quiz1 {
+  .question-area {
+    margin-top: 20px;
+    margin-bottom: 20px;
+  }
+
+  .question-description {
+    margin-bottom: 5px;
+  }
+
+  .question-feedback {
+    margin-top: 5px;
+  }
+
+  .question-hint {
+    display: block;
+    margin-top: 5px;
+    margin-bottom: 5px;
+    color: #737373;
+    font-style: italic;
+  }
+
+  .error-text {
+    margin-top: 5px;
+    margin-bottom: 5px;
+    color: #a94442;
+    font-weight: bold;
+    margin-bottom: 0px;
+  }
+
+  .has-error .radio,
+  .has-error .checkbox {
+    color: #333;
+  }
+
+  .option-checked {
+    font-weight: bold;
+  }
+
+  .hint-checkbox {
+    padding-left: 20px;
+  }
+
+  .quiz1-icon-correct {
+    @include glyphicon;
+    color: #5cb85c;
+
+    &:before {
+      content: "\e013";
+    }
+  }
+
+  .quiz1-icon-neutral {
+    @include glyphicon;
+    color: #f5bc09;
+
+    &:before {
+      content:"\e085";
+    }
+  }
+
+  .quiz1-icon-incorrect {
+    @include glyphicon;
+    color: #d9534f;
+
+    &:before {
+      content: "\e014";
+    }
+  }
+
+  .btn-group button:first-child {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+}
```

saarasat:
Great, I implemented now your idea, thank you for it! I also ran all the tests, and checked with different questionnaire question combinations, everything seems to work. 

I guess for now the css-files should be still included in git? 

I checked from [Sass-guidelines](https://sass-guidelin.es/#architecture), that the reusable components could be places in `components`-folder. I named the file `icons.scss` and commented them as bootstrap glyphicons. This structure should then work later on, if the switch is made to font-awesome etc. The downside is that now I had to import the icons within the `quiz1.scss`. In any case, this was just an idea, so do let me know if you think that glyphicon should be moved somewhere else or kept still within the `quiz1.scss` -file :)

assets/css/main.css
```diff
@@ -1,9 +1,8150 @@
-﻿/*!
```
raphendyr, line 1: You didn't run `drone exec` before adding this file. It creates compressed version instead of this more readable one created by the watch thing (which is easier to debug).

assets/sass/components/icons.scss
```diff
@@ -0,0 +1,14 @@
+
+/* Bootstrap glyphicons */
```
raphendyr, line 2: You should use underscore (_) in the file name to mark it as a partial.

assets/sass/protocols/_quiz1.scss
```diff
@@ -0,0 +1,74 @@
+@import '../components/icons.scss';
```
raphendyr, line 1: The `@use` might suit this better https://sass-lang.com/documentation/at-rules/use

assets/css/components/icons.css
```diff
@@ -0,0 +1,3 @@
+/* Bootstrap glyphicons */
```
raphendyr, line 1: FYI: if you rename the `icons.scss` to `_icons.scss`, then this file is not created anymore.

assets/css/protocols/quiz1.css.map
```diff
@@ -0,0 +1 @@
+{"version":3,"sourceRoot":"","sources":["../../sass/protocols/quiz1.scss","../../sass/components/icons.scss"],"names":[],"mappings":";AAEA;AAIE;EACE;EACH;;AAGC;EACE;;AAGF;EACE;;AAGF;EACE;EACA;EACA;EACA;EACA;;AAGF;EACE;EACA;EACA;EACA;EACA;;AAGF;AAAA;EAEE;;AAGF;EACE;;AAGF;EACE;;AAGF;EC9CD;EACA;EACA;EACA;EACA;EACA;EACA;EACA;EACA;EDwCG;;AAEA;EACE;;AAIJ;ECvDD;EACA;EACA;EACA;EACA;EACA;EACA;EACA;EACA;EDiDG;;AAEA;EACE;;AAIJ;EChED;EACA;EACA;EACA;EACA;EACA;EACA;EACA;EACA;ED0DG;;AAEA;EACE","file":"quiz1.css"}
```
raphendyr, line 1: I think this file was not supposed to be included.

saarasat: Just to clarify, which files are to be included in git after changes in styles or new .scss-files? I noticed that some of the `.map` are there, so just to be sure.

saarasat:
Good to learn at least a tiny bit of Sass too, all new to me :) I think I covered @raphendyr your earlier points. If you could just clarify the inclusion of those css.map -files: which to include here and why, then I'll know that for future as well.

assets/css/protocols/quiz1.css.map
raphendyr: All `.map` discribe where the source lines can be found for minimised files. E.g. `main.css` is unreadable as is, but `main.css.map` tells where every line is from, thus when you open browser dev tools, you get to read the actual sources and not just unreadable files.

In short, every compressed/minified/generated *.css or *.js file should have a *.map file too. You can notice that the *.css file includes a comment at the end of the file, which points to the *.map file (see `/*# sourceMappingURL=main.css.map */` in the `main.css`).

saarasat: That clears it up well, thanks :)

assets/sass/components/_icons.scss
```diff
@@ -0,0 +1,14 @@
+
+/* Bootstrap glyphicons */
+
+@mixin glyphicon {
```
raphendyr, line 4: I was thinking the filename of this file. Currently, it only contains a mixing, which doesn't output any css as itself, hence it could be in a folder called `abstracts` or `utilities`. However, it defines glyphicon, which could be used as is and this component could someday extended to define a more icons, which would be used as is.

I think a good layout we could adapt is described in [this article](https://itnext.io/structuring-your-sass-projects-c8d41fa55ed4).

However, alternative would be using [@extend](https://sass-lang.com/documentation/at-rules/extend), which I didn't remember before. I'll add another comment about that in relevant location.

assets/sass/protocols/_quiz1.scss
```diff
@@ -0,0 +1,74 @@
...
+		margin-bottom: 0px;
+	}
+
+	.has-error .radio,
+	.has-error .checkbox {
+		color: #333;
+	}
+
+	.quiz1-icon-correct {
+		@include icons.glyphicon;
```
raphendyr, line 49: I think it could be possible to use bootstrap3 classes here with [@extend](https://sass-lang.com/documentation/at-rules/extend). It could be possible to just write `@extend .glyphicon`.

assets/sass/components/_icons.scss
saarasat: Hmm.. That is an interesting point. I think our glyphicon indeed falls right in between the two categories "abstracts" and "components/icons". I myself prefer the name "icons", in the sense, that readily tells you, what to expect at this point. But that is beside the actual arguments, just my prefence.  

It really could be either way I think at this point of development.

**Within abstracts:**
- If were to have a lot of mixins, it would make sense to have it here. However, we don't yet have that much.
- It is only a mixin, should be in mixins. Like you said, doesn't output any CSS when compiled.

**Within components:**
- I think Matthew's original plan was to develop a library of re-usable components, so in that sense (future) icons fit well within the components-folder.
- Like you said, the glyphicon could be used as is. And I think this mixin is only (?) used for icons, wouldn't be used for anything else. Thus, it kinda makes sense to have it alongside the icons if more of them would be defined here. 
- If in the future there are other icons to be used or the change is made to fontawesome, this structure works with those cases.

However, this could always be changed, if we decide after working with Sass a bit more, that this is a poor place for it. So I think it works for now, and doesn't have to be set in stone :)

raphendyr: I think we can keep the mixin here, if the `@extend` way is not better.

assets/sass/protocols/_quiz1.scss
raphendyr: This would yield smaller CSS, but current solution is better base to start changing to font awesome.

PR #620, opened 2020-07-13
Save language in submission even if not provided

### LINKED ISSUE #611: Store the language meta data for all submissions
Now that https://github.com/apluslms/a-plus/issues/581 has been done and submissions store a meta data field for at least the language and the group ID, the code should be changed so that the language is always stored in the meta data even if the HTTP POST request does not include the `__aplus__={"lang": "en"}` parameter. If the request does not define the language, then the language is taken from the user's session as usual (basically, Django's `get_language()` unless there is a better alternative). In the API, the user's default language from the user profile could maybe be used, though, other language issues may change the userprofile.

Storing the language is important so that it can be used for regrading. Otherwise, I think regrading might read the language from the teacher's session when he/she starts the regrading in the inspect submission page and that is not sensible at all.

The exercise grade method 
https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/exercise/exercise_models.py#L693
should always take the language from the submission meta data.

HenrikSundell:
# Description

Save language to submission `meta_data` even if not provided. This way the language 
is consistent when regrading a submission where language data isn't given.

Fixes #611

# Testing

Changes have been tested manually.

Not relevant.

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -474,6 +474,14 @@ class CourseInstance(UrlMixin, models.Model):
     def is_valid_language(self, lang):
         return lang == "" or lang in [key for key,name in settings.LANGUAGES]
 
+    @property
+    def default_language(self):
+        language = self.language
+        language_code = language.lstrip('|').split('|', 1)[0]
+        if language_code:
+            return language_code
+        return settings.LANGUAGE_CODE.split('-', 1)[0]
+
     def save(self, *args, **kwargs):
         """
         Saves the model.
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -705,7 +705,7 @@ class BaseExercise(LearningObject):
         Loads the exercise feedback page.
         """
         # Get the language from the submission
-        language = submission.lang or get_language()
+        language = submission.lang or self.course_module.course_instance.default_language
 
         submission_url = update_url_params(
             api_reverse("submission-grader", kwargs={
--- a/exercise/submission_models.py
+++ b/exercise/submission_models.py
@@ -8,7 +8,7 @@ from django.core.files.storage import default_storage
 from django.db import models, DatabaseError
 from django.db.models.signals import post_delete
 from django.utils import timezone
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import get_language, ugettext_lazy as _
 from mimetypes import guess_type
 
 from lib.fields import JSONField, PercentField
@@ -38,6 +38,9 @@ class SubmissionManager(models.Manager):
             meta_data_dict = json.loads(request.POST.get('__aplus__', '{}'))
         except json.JSONDecodeError:
             raise ValueError("The content of the field __aplus__ is not valid json")
+        if 'lang' not in meta_data_dict:
+            meta_data_dict['lang'] = get_language()
+
         try:
             new_submission = Submission.objects.create(
                 exercise=exercise,


exercise/exercise_models.py
```diff
@@ -704,7 +704,9 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
         Loads the exercise feedback page.
         """
         # Get the language from the submission
-        language = submission.lang or get_language()
+        language = submission.lang
+        if language is None:
+            language = submission.meta_data['lang'] = get_language()
```
raphendyr, line 709: It would make more sense to set this in the location, which stores `meta_data` for the first time. Even if this function is executed as part of the original request, it's likely that it will change.

HenrikSundell: I thought it would be safer to do this way, do you mean it would be enough to only do this in `create_from_post`?
https://github.com/apluslms/a-plus/blob/73aea9db45f186da052957024500d44ea6932274/exercise/submission_models.py#L38-L45

raphendyr: Yeah. I think it should be part of that function. Though, that function shouldn't be part of the query manager object, but we'll go with that for now.

exercise/submission_models.py
```diff
@@ -36,6 +36,8 @@ def create_from_post(self, exercise, submitters, request):
         ]
         try:
             meta_data_dict = json.loads(request.POST.get('__aplus__', '{}'))
+            if 'lang' not in meta_data_dict:
```
raphendyr, line 39: Typically, the try-except shouldn't include only as little code as possible. In other words, this part could be moved out of the block.

As a side note, try-except-else is nice pattern for this kind of things, specially when except doesn't raise/return. Here that wouldn't be so useful.

HenrikSundell:
> Nice and small change :) I guess the hardest part is to verify or to be sure enough that nothing breaks..

I tried to check that nothing breaks, but yeah that's why I tried to cheat and do it in the grade function :)

exercise/exercise_models.py
```diff
@@ -705,7 +705,7 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
         Loads the exercise feedback page.
         """
         # Get the language from the submission
-        language = submission.lang or get_language()
+        language = submission.lang
```
raphendyr, line 708: I just noticed that the `submission.lang` can be empty, e.g., for all the currently existing submissions. Rest of the function doesn't work perfectly in that case. I think the `_build_service_url` should be edited to handle empty language values and such. Though, that would require a version of `pick_localized`, which would also return the selected language (it sometimes selects a different than in the language parameter, e.g., when that is not a valid language).

Hmm.. I'm not 100% sure what how to fix this well, but feel free to come up with some ideas.

HenrikSundell: Do you still think it would be bad to save it to meta_data here if None, seems like an ok enough solution here. If it shouldn't be saved is there a reason why changing `_build_servise_url` would be better than `or get_language()` here. Seems like it would make it more complicated for basically the same functionality.

raphendyr: Assuming this gets into a async task, then the `get_language()` and `request` are not available, thus the error handling should a) fail the grade or b) find a solution without above methods. In addition, `_build_service_url` already does misbehave with invalid language data, or more exactly, it adds `lang=` query argument which differs from the url it selects as `pick_localized` does pick something (and doesn't tell what) if the requested (invalid) language is not available.

HenrikSundell: I like the simplicity of failing the grade. This will mean that we can't regrade old submissions created before these changes. Is this something we can accept? I'm guessing this problem wouldn't be encountered many times.

raphendyr: I would like @Mankro's input for that decision.

Mankro: If `submission.lang` is empty, could we just use the main language of the course? The submission always belongs exactly one course (references are about like this: `submission.exercise.course_module.course_instance`). If the language of the course instance is also empty, then it could default to English. I think it would make sense that the `grade` method selects the language to use in all cases. I don't want to fail grading just because the language is unclear, particularly when course instances usually define one or two languages (course instance language really shouldn't be empty in production even though the database allows empty values there).

I think `_build_service_url` should always receive a valid language parameter.

exercise/exercise_models.py
```diff
@@ -706,6 +706,10 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
         """
         # Get the language from the submission
         language = submission.lang
+        if language is None:
+            language = submission.exercise.course_module.course_instance.language
```
raphendyr, line 710: As `self` is the exercise, then `self.course_module.course_instance.language` should be enough. However, the language could be `|en|fi|`, thus you need to take the first. As I think this might be relatively common, I think we should add a property to the course instance `default_language`, which would contain the logic for that (so, `self.course_module.course_instance.default_language`).

exercise/exercise_models.py
```diff
@@ -705,7 +705,11 @@ def grade(self, request, submission, no_penalties=False, url_name="exercise"):
         Loads the exercise feedback page.
         """
         # Get the language from the submission
-        language = submission.lang or get_language()
+        language = submission.lang
+        if language is None:
+            language = submission.exercise.course_module.course_instance.language
+        if language is None or language == "":
```
Mankro, line 711:
```suggestion
        if not language:
```

course/models.py
```diff
@@ -474,6 +474,14 @@ def clean(self):
     def is_valid_language(self, lang):
         return lang == "" or lang in [key for key,name in settings.LANGUAGES]
 
+    @property
+    def default_language(self):
+        language = self.language
+        language_list = [x for x in language.split("|") if x]
+        if language_list:
+            return language_list[0]
+        return "en"
```
raphendyr, line 483: I think `settings.LANGUAGE_CODE.split('-', 1)[0]` might be better last fallback.

course/models.py
```diff
@@ -474,6 +474,14 @@ def clean(self):
     def is_valid_language(self, lang):
         return lang == "" or lang in [key for key,name in settings.LANGUAGES]
 
+    @property
+    def default_language(self):
+        language = self.language
+        language_list = [x for x in language.split("|") if x]
```
raphendyr, line 480: Just `language.lstrip('|').split('|', 1)[0]` might be nice. No need to build a list

```
>>> timeit("foo = [x for x in l.split('|') if x]", "l = '|en|fi|'")
0.2993435220005267
>>> timeit("foo = [x for x in l.split('|') if x]", "l = 'fi'")
0.20665564800037828
>>> timeit("foo = l.lstrip('|').split('|', 1)[0]", "l = '|en|fi|'")
0.17270740999992995
>>> timeit("foo = l.lstrip('|').split('|', 1)[0]", "l = 'fi'")
0.10890936499981763
```

course/models.py
```diff
@@ -477,10 +477,10 @@ def is_valid_language(self, lang):
     @property
     def default_language(self):
         language = self.language
-        language_list = [x for x in language.split("|") if x]
+        language_list = language.strip('|').split('|')
```
raphendyr, line 480: lstrip and strip are different, and `split(str, number)` limits how many times the split is executed, we need to find only the first.

raphendyr: Interestingly. the `split('|')` is faster with 2 languages, but slower with more (where `split('|', 1)` is faster :thinking:

Mankro: Now this code does not work with an empty `self.language`:
`"".strip("|").split("|")  -> ['']`
It is a list with one element, the empty string, so the list evaluates to True in the if condition and an empty language would be returned.

raphendyr: It's missing the `[0]`
```python3
>>> ''.lstrip('|').split('|', 1)[0]
''
>>> 'fi'.lstrip('|').split('|', 1)[0]
'fi'
>>> '|en|fi|'.lstrip('|').split('|', 1)[0]
'en'
```

Mankro:
@HenrikSundell you can squash again, please. The new `default_language` should work with empty values, a single language, and multiple languages. It also passes Jaakko's performance review. So this is really finished!

course/models.py
```diff
@@ -474,6 +474,14 @@ def clean(self):
     def is_valid_language(self, lang):
         return lang == "" or lang in [key for key,name in settings.LANGUAGES]
 
+    @property
+    def default_language(self):
+        language = self.language
+        language_code = language.lstrip('|').split('|', 1)[0]
+        if language_code:
+            return language_code
+        return settings.LANGUAGE_CODE.split('-', 1)[0]
```
raphendyr, line 483: Actually,

```python3
language = self.language
if language and language[0]:
  return language[1:language.find('|', 1)]
return language
```

Seems to be the fastest :/

```
>>> lgs = ('', 'en', '|en|fi|', '|en|fi|sv|')
>>> timeit("for lg in lgs: l = lg.lstrip('|').split('|', 1)[0]", "from __main__ import lgs")
0.5716783219831996
>>> timeit("for lg in lgs: l = lg[1:lg.find('|', 1)] if lg and lg[0] == '|' else lg", "from __main__ import lgs")
0.47175573103595525
```

Mankro: ugh, no need to change because of a small difference. Let's not spend more time on micro-optimization for performance.

Mankro:
This PR was actually assigned to @raphendyr, but I thought it would be appropriate that I merge it. It would probably have been assigned to me if I had been here in July. Thanks Henrik for the work!

PR #621, opened 2020-07-15
Fix the broken selenium-test-locator

saarasat:
# Description

One of the items on BasePageLocators referenced to a `<nav>`, which
had been replaced by a `<div>` in https://github.com/apluslms/a-plus/commit/426a0c727c5b718754c0e48d8d4aecdc1028bd45. It broke the selenium-tests. This fixes that.

# Testing

All selenium-tests and unit-tests pass now.

# Have you updated the README or other relevant documentation?

Not relevant 

# Is it [Done](https://wiki.aalto.fi/display/EDIT/Definition+of+Done)?

- [ ] I (developer) have created unit tests and the tests pass
- [ ] I (developer) have created functional tests (Selenium tests) if applicable
- [x ] I (developer) have tested the changes manually
- [ ] Reviewer has finished the code review
- [ ] After the review, the developer has made changes accordingly
- [ ] Customer/Teacher has accepted the implementation of the feature

PR #622, opened 2020-07-15
Past instance notification

### LINKED ISSUE #554: automatic notification for user about arriving on pages of a past course

saarasat:
# Description

- If the course instance has ended and enrolling has already closed, the page visitor will be told that the course has already ended. 
- If there is an instance of the same course, that has a later ending time than now, and that instance is visible for that user, they will be given a link to that instance. 
- If there are several instances that fit this description, the latest is given.
- Even if the course has already ended, but the enrolling would for some reason be open, the enroll_form is shown.  
- Front page and enroll-page of the course show the link to the latest instance of the course. 
- Other course-pages show only the text "This course has already ended" to save some database-queries
(If the link is wanted for other pages as well, most likely the best option would be to move the alert to the course_base.html, that seems to work as well and is easy to move around. That however will cause an extra query for old instances. If this issue is considered more important and worth using extra time, a cache entry could be probably implemented.)

Fixes #554

**Front page of the course**:

image

**Content-pages of the course**

image

# Testing

Unit-tests and selenium-tests pass once this has been merged: https://github.com/apluslms/a-plus/pull/621

The changes have been tested as a logged in user, a student and a staff member with past courses that have no future instances and with courses that have future instances. Both hidden ones and ones visible to students. Also the case where a future instance exist and where the enrolling has not yet started has been tested. 

The changes do not change the way the content is shown to the user, only limit the showing of the enroll_form to courses where the enrolling is still .  
 
o1 has other implementations of this, which would become unnecessary now: 
https://plus.cs.aalto.fi/o1/2018/

Not relevant here.

### DIFF
--- a/course/templates/course/_enroll_form.html
+++ b/course/templates/course/_enroll_form.html
@@ -2,42 +2,42 @@
 {% load course %}
 
 {% if not is_student and not instance.is_past and not instance.has_enrollment_closed %}
-<div class="alert alert-info">
-{% if enrollable or is_course_staff %}
-  {% if instance.is_enrollment_open %}
-  <form method="post" action="{{ instance|url:'enroll' }}">
-    {% csrf_token %}
-    {% if instance.view_content_to > 1 %}
-    {% trans "In order to submit exercises you must enroll in the course." %}
-    {% else %}
-    {% trans "In order to submit exercises and see material, you must enroll in the course." %}
-    {% endif %}
-    <input type="submit" value="{% trans 'Enroll' %}" class="btn btn-info">
-    {% if instance.enrollment_ending_time %}
-    <div>
-      <small>
-        {% blocktrans with end=instance.enrollment_ending_time %}
-        The course enrollment is open until {{ end }}.
-        {% endblocktrans %}
-      </small>
-    </div>
-    {% endif %}
-  </form>
-  {% elif instance.has_enrollment_closed %}
-  {% trans "Unfortunately, enrolling in the course has ended." %}
-  {% else %}
-  {% blocktrans with start=instance.enrollment_start end=instance.enrollment_end %}
-  The course opens {{ start }}.
-  {% endblocktrans %}
-  {% endif %}
-{% elif profile %}
-  {% if instance.view_content_to < 3 %}
-  {% trans "Unfortunately, you may not enroll in this course and you may not explore the contents either." %}
-  {% else %}
-  {% trans "Unfortunately, you cannot enroll in this course, but you may explore the course material." %}
-  {% endif %}
-{% else %}
-  {% trans "Login is required to submit exercises, but you may anonymously explore the course material." %}
-{% endif %}
+	<div class="alert alert-info">
+	{% if enrollable or is_course_staff %}
+		{% if instance.is_enrollment_open %}
+			<form method="post" action="{{ instance|url:'enroll' }}">
+				{% csrf_token %}
+				{% if instance.view_content_to > 1 %}
+					{% trans "In order to submit exercises you must enroll in the course." %}
+				{% else %}
+					{% trans "In order to submit exercises and see material, you must enroll in the course." %}
+				{% endif %}
+				<input type="submit" value="{% trans 'Enroll' %}" class="btn btn-info">
+				{% if instance.enrollment_ending_time %}
+					<div>
+						<small>
+							{% blocktrans with end=instance.enrollment_ending_time %}
+							The course enrollment is open until {{ end }}.
+							{% endblocktrans %}
+						</small>
+					</div>
+				{% endif %}
+			</form>
+		{% elif instance.has_enrollment_closed %}
+			{% trans "Unfortunately, enrolling in the course has ended." %}
+		{% else %}
+			{% blocktrans with start=instance.enrollment_start end=instance.enrollment_end %}
+				The course opens {{ start }}.
+			{% endblocktrans %}
+		{% endif %}
+	{% elif profile %}
+		{% if instance.view_content_to < 3 %}
+			{% trans "Unfortunately, you may not enroll in this course and you may not explore the contents either." %}
+		{% else %}
+			{% trans "Unfortunately, you cannot enroll in this course, but you may explore the course material." %}
+		{% endif %}
+	{% else %}
+		{% trans "Login is required to submit exercises, but you may anonymously explore the course material." %}
+	{% endif %}
 </div>
 {% endif %}
--- a/course/templates/course/_enroll_form.html
+++ b/course/templates/course/_enroll_form.html
@@ -1,7 +1,7 @@
 {% load i18n %}
 {% load course %}
 
-{% if not is_student %}
+{% if not is_student and not instance.is_past and not instance.has_enrollment_closed %}
 <div class="alert alert-info">
 {% if enrollable or is_course_staff %}
   {% if instance.is_enrollment_open %}
--- a/course/templates/course/_enroll_form.html
+++ b/course/templates/course/_enroll_form.html
@@ -16,7 +16,7 @@
 				{% if instance.enrollment_ending_time %}
 					<div>
 						<small>
-							{% blocktrans with end=instance.enrollment_ending_time %}
+							{% blocktrans trimmed with end=instance.enrollment_ending_time %}
 							The course enrollment is open until {{ end }}.
 							{% endblocktrans %}
 						</small>
@@ -26,7 +26,7 @@
 		{% elif instance.has_enrollment_closed %}
 			{% trans "Unfortunately, enrolling in the course has ended." %}
 		{% else %}
-			{% blocktrans with start=instance.enrollment_start end=instance.enrollment_end %}
+			{% blocktrans trimmed with start=instance.enrollment_start end=instance.enrollment_end %}
 				The course opens {{ start }}.
 			{% endblocktrans %}
 		{% endif %}
--- a/course/templates/course/course_base.html
+++ b/course/templates/course/course_base.html
@@ -60,6 +60,21 @@
         {% endcomment %}
     </div>
     <div class="col-sm-10" id="course-content">
+        {% if instance.is_past and instance.has_enrollment_closed %}
+            <div class="alert sticky-alert alert-default">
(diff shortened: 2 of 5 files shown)


course/views.py
```diff
@@ -106,8 +106,23 @@ def get(self, request, *args, **kwargs):
...
+
+        if self.instance.is_past:
+            course = self.instance.course
+            now = timezone.now()
+            try:
+                later_instance = CourseInstance.objects.get_visible(request.user).filter(
+                    ending_time__gte=now,
+                    course=course
+                ).latest('starting_time')
+            except ObjectDoesNotExist:
```
raphendyr, line 119: You can use the specific exception `CourseInstance.DoesNotExist`.

course/views.py
```diff
@@ -106,8 +106,23 @@ def get(self, request, *args, **kwargs):
...
+                pass
+
+        return super().get(request,
+                            *args,
+                            **kwargs,
+                            later_instance=later_instance)
```
raphendyr, line 125: If the `later_instance` exists in the `kwargs`, then this line will crash. It might be a good idea to always overwrite the field using line `kwargs['later_instance'] = later_instance`.

course/views.py
```diff
@@ -106,8 +106,23 @@ def get(self, request, *args, **kwargs):
...
-
+        later_instance = None
+
+        if self.instance.is_past:
+            course = self.instance.course
+            now = timezone.now()
+            try:
+                later_instance = CourseInstance.objects.get_visible(request.user).filter(
+                    ending_time__gte=now,
+                    course=course
```
raphendyr, line 117: The line is missing the comma at the end.

raphendyr: In addition, as this chains multiple methods, it might be readable if every methods starts a new line. Hence, maybe:

```python
later_instance = (
    CourseInstance.objects
        .get_visible(request.user)
        .filter(course=course, ending_time__gte=now)
        .latest('starting_time')
)
```

saarasat:
> I'm wondering about the appearance.
> 
> First, I wonder if this is prominent enough. The text matches the size and style of nearby text, so maybe there is a risk it gets skipped over.
> 
> My other concern is that if someone follows a link to, for example, heading or an exercise further down the page, they completely miss this even though they might benefit from it. Have you considered trying a sticky version?

That is a valid point. Adding stickiness and pinning it eg. on the top or the bottom of the page might work. I'll try that!

saarasat:
I now made the improvements to the python as @raphendyr suggested and changed the alert to a sticky one like @mtthwhggns proposed. I changed the coloring of the alert, as when it was blue, it got mixed with the colors of exercises when scrolling down. Now it should be noticeable but not too over-powering or annoying. There is no default-alert in bs3, but I added styles here for that. In case the switch is done to bs4 at some point, then things should work :)

course/templates/course/course_base.html
```diff
@@ -60,6 +60,21 @@ <h4>{% trans "Calendar" %}</h4>
...
     <div class="col-sm-10" id="course-content">
+        {% if instance.is_past and instance.has_enrollment_closed %}
+            <div class="alert sticky-alert alert-default">
+                {% trans "This course has already ended. " %}
+                {% if later_instance %}
+                    <p>
+                        {% trans "The latest version of the course can be found at: " %}
+                        <a class="alert-link" href="{{ later_instance|url }}">
+                            <span lang="{{ later_instance.language }}">
+                                {{ later_instance.course.name|parse_localization }}: {{ later_instance.instance_name|parse_localization }}
```
raphendyr, line 71: Hmmm. I'm not sure if colon is the best separator for combining name and instance name. However, it appears to be the one that is used in those few locations where these are combined. I used format `name (instance)` in MOOC-Jutut, but I'm not sure if it would be better or not.

If `parse_localization` would handle multiple substitution, then we could say something like `{{ later_instance|parse_localization }}`, which would use `CourseInstance.__str__` (which uses format `name: instance`). If we did something like that, then the format could be changed in a single location.

course/templates/course/course_base.html
```diff
@@ -60,6 +60,21 @@ <h4>{% trans "Calendar" %}</h4>
         {% endcomment %}
     </div>
     <div class="col-sm-10" id="course-content">
+        {% if instance.is_past and instance.has_enrollment_closed %}
+            <div class="alert sticky-alert alert-default">
+                {% trans "This course has already ended. " %}
```
raphendyr, line 65: The space at the end of the string doesn't seem to be importart. The new line will tell browsers to use a space between text and the next element.

course/templates/course/course_base.html
```diff
@@ -60,6 +60,21 @@ <h4>{% trans "Calendar" %}</h4>
...
+                {% if later_instance %}
+                    <p>
+                        {% trans "The latest version of the course can be found at: " %}
```
raphendyr, line 68: > The space at the end of the string doesn't seem to be importart. The new line will tell browsers to use a space between text and the next element.

course/views.py
```diff
@@ -106,8 +106,24 @@ def get(self, request, *args, **kwargs):
         elif lti_msg:
             messages.info(request, html.escape(lti_msg))
 
-        return super().get(request, *args, **kwargs)
+        later_instance = None
+
+        if self.instance.is_past:
+            course = self.instance.course
```
raphendyr, line 112: I think there is also `self.course`. If so, then I think this "alias" might not be so important. Same consideration can be used for the `now`, which is used only once.

course/templates/course/course_base.html
saarasat: I copied this format from the top nav, there the instance is listed as "course: instance". Though the thing that is missing here is the course code, but I thought that with that one it becomes a bit messy looking. The course code is already listed three times on the page as it is. 

image

course/templates/course/course_base.html
saarasat: Oh nice, good to know. The space has now been removed.

course/templates/course/course_base.html
saarasat: Same here

course/views.py
saarasat: There seemed to be that alias, so I cleared that now too

course/templates/course/course_base.html
raphendyr: Hmm.. I think we should visit the format if we touch that menu. In addition, I guess Jutut should use the same format too...

raphendyr:
I think it's good too. You could rebase and clean the history now, thanks.

saarasat:
Rebased, squashed, commit message checked, tests run and the code checked locally one more time.

course/templates/course/_enroll_form.html
```diff
@@ -1,43 +1,43 @@
...
+					{% trans "In order to submit exercises you must enroll in the course." %}
+				{% else %}
+					{% trans "In order to submit exercises and see material, you must enroll in the course." %}
+				{% endif %}
+				<input type="submit" value="{% trans 'Enroll' %}" class="btn btn-info">
+				{% if instance.enrollment_ending_time %}
+					<div>
+						<small>
+							{% blocktrans with end=instance.enrollment_ending_time %}
+							The course enrollment is open until {{ end }}.
```
raphendyr, line 20: Did these block trans edits trigger a change in the translation file? I think you could add the `trimmed` option and make sure the translations still work.

saarasat: Great pick. Did not realise that fixing indentation might break old translations where the trimmed-attribute has not been added. 

I now added trimmed to both, compiled the messages and checked that they worked

course/views.py
```diff
@@ -106,8 +106,22 @@ def get(self, request, *args, **kwargs):
...
             messages.info(request, html.escape(lti_msg))
 
-        return super().get(request, *args, **kwargs)
+        later_instance = None
+
+        if self.instance.is_past:
+            try:
+                later_instance = (
+                    CourseInstance.objects
+                        .get_visible(request.user)
```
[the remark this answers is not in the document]
saarasat: Since `request` is an argument of the function, it can be accessed straightaway, without using the `self`. Works both ways though, with the `self` as well.

course/templates/course/course.html
```diff
@@ -3,6 +3,7 @@
 {% load exercise %}
 {% load news %}
 {% load apps %}
+{% load course %}
```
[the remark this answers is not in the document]
saarasat: Oh, I think that was left there by accident, now that the implementation was different from what I originally tried. I removed that bit, good catch :)

saarasat:
> Do you think we should limit the enrolment ending time according to the ending time of the course? @raphendyr Maybe you can explain why this behaviour?

At least its seems to be possible have enrolling open even though the course has ended, with the current way the instances are set. So I thought I should cover that case as well, thus that the enroll_form is shown, if the enrolling is still actually used even though the course has already ended. How much this option is used though, I am not sure.

raphendyr:
> Do you think we should limit the enrolment ending time according to the ending time of the course? @raphendyr Maybe you can explain why this behaviour?

Of top of my head, it soulds stranges to be able to enroll after the course has ended. It's possible that there has just been assumption that it's always configured so. @Mankro did revisit those time fields and has a better knowledge how they probably should work.

saarasat:
@jaguarfi Just let me know if there is something here then that you'd like me to check or revisit, or whether this can be merged :)

jaguarfi:
@saarasat could you create a new issue to verify that the enrollment time behaves according to the duration of the course? So that we can eventually verify whether or not this is the expected behaviour.

saarasat:
> @saarasat could you create a new issue to verify that the enrollment time behaves according to the duration of the course? So that we can eventually verify whether or not this is the expected behaviour.

And done https://github.com/apluslms/a-plus/issues/636

saarasat:
And just rebased on top of yesterday's latest commit.

jaguarfi:
Terrific :100:

PR #625, opened 2020-07-21
Remove old _aplus_group parameter

### LINKED ISSUE #612: Refactor submission handlers
- [ ] Submission made in the web page shoud use the new `__aplus__={"group": ID}` POST parameter for selecting the group. Then we can remove the old `_aplus_group` parameter completely. Note for example these: https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/exercise/static/exercise/group.js#L49 and https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/exercise/exercise_models.py#L578 (issue https://github.com/apluslms/a-plus/issues/581 introduced the `__aplus__` parameter).
- [ ] Refactor submission code in the ExerciseView (https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/exercise/views.py#L135) and the submit API (https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/exercise/api/views.py#L235) so that they use the same function for the common parts. Now some code is duplicated in those two places and we want to remove duplication.
- [ ] The API submit function (https://github.com/apluslms/a-plus/blob/0cf8c3126d6011f08201764c3878b930ec4f3e31/exercise/api/views.py#L238) has an outdated commect referring to the `_aplus_group` parameter, but `__aplus__={"group": ID}` should be used nowadays.

HenrikSundell:
# Description

Change web submissions to use the new POST parameter
`__aplus__={"group": ID}` to set submission group.
Remove all uses of the old `_aplus_group` POST parameter.

Fixes #612

# Testing

Changes have been manually changed, to work the same way as before using the old parameter.

Wiki still mentions `_aplus_group` as the old way, but I guess that's OK. Might make it cleared for 
someone who has used the old parameter before.

### DIFF
--- a/exercise/static/exercise/group.js
+++ b/exercise/static/exercise/group.js
@@ -52,7 +52,9 @@
         var list = this.ui.find('select');
         for (var i = 0; i < this.groups.length; i++) {
           var group = this.groups[i];
-          list.append($('<option value="{&quot;group&quot;:' + group.id + '}" data-group-size="' + group.size + '">' + group.text + '</option>'));
+          var meta_data = JSON.stringify({group: parseInt(group.id)})
+          /* TODO: add lang to meta_data*/
+          list.append(($('<option value=\''+ meta_data + '\'data-group-size="' + group.size + '">' + group.text + '</option>')));
         }
       }
 		},
--- a/exercise/static/exercise/group.js
+++ b/exercise/static/exercise/group.js
@@ -46,13 +46,13 @@
         });
 
         this.ui = $('<div class="submit-group-selector input-group col-md-6">'
-          + '<select name="_aplus_group" class="form-control"></select>'
+          + '<select name="__aplus__" class="form-control"></select>'
           + '<span class="input-group-btn"></span>'
           + '</div>');
         var list = this.ui.find('select');
         for (var i = 0; i < this.groups.length; i++) {
           var group = this.groups[i];
-          list.append($('<option value="' + group.id + '" data-group-size="' + group.size + '">' + group.text + '</option>'));
+          list.append($('<option value="{&quot;group&quot;:' + group.id + '}" data-group-size="' + group.size + '">' + group.text + '</option>'));
         }
       }
 		},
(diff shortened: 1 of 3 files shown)


exercise/static/exercise/group.js
```diff
@@ -46,13 +46,13 @@
...
         this.ui = $('<div class="submit-group-selector input-group col-md-6">'
-          + '<select name="_aplus_group" class="form-control"></select>'
+          + '<select name="__aplus__" class="form-control"></select>'
           + '<span class="input-group-btn"></span>'
           + '</div>');
         var list = this.ui.find('select');
         for (var i = 0; i < this.groups.length; i++) {
           var group = this.groups[i];
-          list.append($('<option value="' + group.id + '" data-group-size="' + group.size + '">' + group.text + '</option>'));
+          list.append($('<option value="{&quot;group&quot;:' + group.id + '}" data-group-size="' + group.size + '">' + group.text + '</option>'));
```
raphendyr, line 55: I think this is probably a bit too simplified change. I think you need to build the `__aplus__` in the javascript so we can bundle other things like the language in there too.

HenrikSundell: Yeah that would make sense, but there isn't any language yet for the submission,  I don't know if there is a point in setting the language in the js, when it's going to be done when the submission is created. It would look nicer if  `__aplus__` is built but wouldn't it still be only the group in there? (at least for now)

raphendyr: The language should be kept as part of the data somehow, as the user language might change between GET and POST, If the language changes between GET and POST, then the submission is graded with different language than what was shown to the user. This is the problem the `__grader_lang` tries to solve for programming exercises.

raphendyr:
We drop this implementation and do a different one on a later date.

Mankro:
@raphendyr it is unclear to me what was wrong with this PR. Is it very problematic to include the language in the POST data?

raphendyr:
The `chapter.js` submission logic should collect the information for the `__aplus__` and include it in the post data, instead of adding a dropdown with pregenerated JSON values.

Issue #594, opened 2020-06-25
Explore methods to change the colours of the Bootstrap theme


mtthwhggns:
As discussed with @raphendyr and in https://github.com/apluslms/a-plus/pull/585, we should twask some of the Bootstrap colours. 

Let's explore an asset pipeline that doesn't require any major work at deployment-time or any new server dependencies.

raphendyr:
This is bit related to first steps of #584. In short, I think we should pre build some SASS files and include the resulting CSS files in the repo. We could make a custom bootstrap 3 theme and fix issues in it and then build that. We should also add prefixed version too to support #584 .

jaguarfi:
Theming bootstrap is relatively easy. You only have to overwrite the variables and elements you want to modify. That is why come rules have the `!default` property. Of course, you can use Sass preprocessor to overwrite some components as well. I will link some websites with relevant information:

* https://bootstrap.themes.guide/how-to-customize-bootstrap.html
* https://bootstrap.themes.guide/how-to-create-bootstrap-themes.html
* Check how the Bootstrap team does with their own website https://github.com/twbs/bootstrap/tree/v4.5.3/site/docs/4.5/assets

However, I think that the approach that was taken during the summer was to add some aplus colours that are used instead of the default colours.  Therefore, I think we can close this issue and focus more on the issue #584, and also in the https://github.com/apluslms/a-plus-design-system. Designing a proper Sass architecture may be extremely beneficial.

PR #630, opened 2020-07-22
Use SASS to build and customise Bootstrap

mtthwhggns:
# Description

As per Jaakko's suggestion, this splits #606 into parts.

This part adds the generated files, SASS source files, fonts and entries to .gitignore to allow local use of SASS for anyone editing styles, while still working normally for anyone who doesn't need to. 

Starts (but doesn't finish) https://github.com/apluslms/a-plus/issues/594

# Testing

This can be tested manually by running and checking the new stylesheet is used (the default colour is made a tiny bit darker which should make this easy to spot). 

If you need to modify the styles before the rest of #594 is done, take a look at draft PR #606 for code to do that, it's quite simple, and the `.gitignore` of this PR means you can do that freely. 


assets/sass/components/_navbar.scss
```diff
@@ -0,0 +1,128 @@
...
+  border-right: none;
+  padding: 0.2em 1em;
+}
+.topbar ul.visible-xs .header h4 {
+  font-size: 10px;
+  font-weight: bolder;
+  color: #75A5D0;
+  margin: 0;
+}
+.navbar-nav>li.menu-groups>a.dropdown-toggle {
```
raphendyr, line 71: Should these use nested SASS format?

assets/sass/components/_navbar.scss
```diff
@@ -0,0 +1,128 @@
...
+.navbar-nav>li.menu-groups>a.dropdown-toggle:hover,
+.navbar-nav>li.menu-groups.open>a.dropdown-toggle {
+  background: rgba(255, 0, 0, 0.4);
+}
+.navbar-nav>li.menu-groups .profile-avatars {
+  display: inline-block;
+  margin: -6px 3px;
+}
+.dropdown-menu>li button {
+display: block;
```
raphendyr, line 83: This file doesn't seem to have consistent formatting. Can you fix the format a bit? On that note, some preferences for css and scss fileS? 2 spaces etc?

assets/sass/components/_navbar.scss
```diff
@@ -0,0 +1,128 @@
...
+0%,10%,20%,25%,35%,40%,50%,55%,65%,70%,80%,90%,100% { top:15px; transform:scale(1); }
+  5%,15%,    30%,    45%,    60%,    75%,    95% { top:50px; transform:scale(2); }
+}
+@keyframes bounce {
+0%,10%,20%,25%,35%,40%,50%,55%,65%,70%,80%,90%,100% { top:15px; transform:scale(1); }
+  5%,15%,    30%,    45%,    60%,    75%,    95% { top:50px; transform:scale(2); }
+}
+/* for navbar-fixed-top: body {
+padding-top: 70px;
+}*/
```
raphendyr, line 128: For some reason, new line at the end of file is nice..

assets/sass/components/_navbar.scss
mtthwhggns: I didn't re-write the whole file yet, only added the variables. I'll re-write it before this gets merged.

assets/sass/components/_navbar.scss
mtthwhggns: I've now formatted the file using the defaults in [Sasslint](https://github.com/sasstools/sass-lint). They're probably a good starting point - no need to reinvent the wheel.

assets/sass/components/_navbar.scss
mtthwhggns: The linter took care of this when I installed it.

assets/sass/components/_navbar.scss
mtthwhggns: Maybe I can just remove this file for now. There are bigger problems with the navbar which need to be addressed, so I'm afraid that a lot of these styles may become redundant.

mtthwhggns:
Thanks for the review @raphendyr 

## Llama and d3stream

Moved these back now. The original hope was that these could be folded into one SASS file, hence moving them into the SASS source directory, but that's not time-sensitive, and can be done later if it's a good idea.

## Navbar

There's lots of code here which could be refactored or totally removed. The only change I made so far was to use the template variables. 

I don't believe it's a very productive use of time to refactor big files which likely need replacing soon, so I think maybe the best compromise is that this PR not change the navbar at all, and this be refactored next time it's touched (likely by me, quite soon), so that it can be considered in parts. 

## `.gitignore`

I would disagree about the .gitignore changes. In my opinion, they're critical to the functioning of the temporary SASS development process, and therefore to this PR. Skipping them causes quite a bit of extra repetitive work. 

The two options are:

1. The developer gets the extra files from the other branch, runs the script to download the `node_modules`, runs the container locally to build their SASS, and when they're done, either delete the ~80 extra files (and start over next time) or go file-by-file through their unstaged changes with lots of scope for mistakes. Either way adds several minutes of work to every CSS change. 
2. These lines are added, the developer obtains the files as previously mentioned once and has a clean working tree - the files mentioned are ignored and ready-to-use next time. 

I really don't see what harm having an extra three lines in `.gitignore` does. They can, and should be replaced as soon as the next stage of this implementation is done. As you've said before, it's fine to have suboptimal, but working solutions in a commit and fix them later.

raphendyr:
It seems `llama.min.css` got dropped now.

templates/base.html
```diff
@@ -76,139 +75,131 @@
     <body class="{% if profile %}{% if is_external_student %}external-student{% else %}internal-student{% endif %}{% endif %} lang-{{ request.LANGUAGE_CODE }}" data-view-tag="{% block view_tag %}{% endblock %}">
         <a href="#content" class="skip-link">{% trans "Skip main navigation" %}</a>
         <div class="page-wrap">
-            <header>
```
raphendyr, line 79: These html changes are not documented (i.e. they are just part of the sass commit). Do I read correctly, that you just removed `<header>` and changed `<footer>` to `<nav>`? Where those part of some other PR too, or do I just remember wrong?

mtthwhggns: You're quite right! I'd missed these, they were previously introduced in the landmarks PR, and I was using an old version of `base.html` here. 

I've updated it now, so the only changes to this file should be removing the CDN Bootstrap CSS, which is part of the SASS changes as the SASS version of the CSS will include all of the Bootstrap styles.

mtthwhggns:
@raphendyr I've handled both of your issues now.

mtthwhggns:
@raphendyr Squashed now

Issue #626, opened 2020-07-21
Small accessibility issues in course archive


mtthwhggns:
The course archive contains some small accessibility issues which need to be addressed:

- `h3`s are used as a direct descendent of `a`, leading to an unclear role (it's a heading and it's a link, which does it behave like if you can't see it?)
- The dates have too low contrast

In addition to just fixing these, I propose addressing these by re-writing the course archive to share the core structure of the front page, like so:

![localhost_8000_](https://user-images.githubusercontent.com/6674046/88054113-857e3e00-cb65-11ea-8635-452c2bc6e0e5.png)

This is a relatively small amount of work, and gets some consistency between the views. I don't imagine it will be much longer, but if it is, that will be compensated by the subheadings rather than "one flat long list".

Issue #596, opened 2020-06-25
Course archive uses heading elements as list items


mtthwhggns:
Generally, headings should be used to break down content into shorter blocks.

On the course archive, every list item is an `h3` tag, which isn't very semantic. 

Related: https://apluslms.github.io/accessibility-audit/patterns/fixable-a/semantic-markup-and-sequences/

Issue #578, opened 2020-06-16
Course-page containing all the instances of a course


saarasat:
Currently there is no list of the instances of the same course. A page collecting these different instances would needed, as discussed in https://github.com/apluslms/a-plus/issues/425

The page should contain the past, current and future instances of the same course. 

Details about the style, navigation and possible limitations by user-type (teacher, student, assistant) can be discussed here. 

This old issue is related https://github.com/apluslms/a-plus/issues/11

raphendyr:
> O1 needs

The specific Programming 1 need is the ability to link to the newest version of the course. See the part about query string `?code=cs-a1110&when=current&next=/foo/bar&forward=true` in #522

Mankro:
When/if this issue is taken under development, it is important to think carefully of the use cases and URLs. There are different ways to organize the URLs for the list of course instances and linking to the latest instance.

From #746:

> Perhaps there could be one URL for the list of instances (e.g., plus.cs.aalto.fi/o1) and another for the most recent instance (e.g., plus.cs.aalto.fi/o1/current).
>
> Alternatively plus.cs.aalto.fi/o1 for the recent instance and plus.cs.aalto.fi/o1/instances or similar for the list.

PR #631, opened 2020-07-23
Use card layout on course archive page

mtthwhggns:
# Description

This switches the course archive to use a layout similar to the front
page, and adapts the frontpage `card` component to include a "condensed"
version.

This leads to more consistency and resolves the small accessibility
issues found in the page.

Note that the page doesn't actually get longer on desktop size screens (on smaller ones it does a little). 

Fixes https://github.com/apluslms/a-plus/issues/626
Fixes https://github.com/apluslms/a-plus/issues/596

Screenshot:
Grouped card-based archive

# Testing


raphendyr:
I guess this layout is totally fine. I was considering if it would take too much space, but I presume that as long as we like to list "all old courses", there aren't so many that it would take too much space.

However, I think this kind of layout could be loaded with ajax and appended at the end of the front page. What do you think?

mtthwhggns:
Good points @raphendyr .

As I mentioned in the notes, on desktop, this page actually takes less space, but on mobile it does get longer. It's worth noting that not many people visit the archive, as their own old courses are accessible from the main navigation more easily.

I'm sceptical about loading unlimited courses on the front page. As we've discussed before, I think the front page could potentially show additional information on the user's current own courses (next deadline, points situation, etc.), but I'm not sure what the benefit would be of allowing it to show a list including 8 year old courses would be. It's maybe one to test in the future. Other things to consider might be SEO (this page currently allows easy, deep crawling) and what the role of this page would be after the "all instances" course page (https://github.com/apluslms/a-plus/issues/578) gets implemented. 

Regarding this page, there were two main motivations:

1. Fix the accessibility and semantic elements issues that came up on this page.
2. Reduce the maintenance overhead by re-using the front page components and styles - the bonus of this is that it gives a really consistent experience. 

I think this solution suits those well - it might not be the best long-term solution, but I'm quite pleased with it for about an hour's work!

raphendyr:
On the ajax topiz, I was thinking that the current "see archive" would just replace the link whit the content of archive page instead of loading a new page. It would then work like "load more". But I guess it's bit too much work for no real benefits, so let's just ignore that idea.

mtthwhggns:
> The only thing I was thinking about is that, now the link is still “old courses” and the name of the page "archive" but the list shows all courses until today. And for teachers the future courses as well. Should we limit what shows here, so that anything that isn’t shown timewise on the front page would be shown in the archive page? Just a thought. Should be pretty simple to change the query in views.py.

Interesting point @saarasat 

I've been thinking about it since yesterday, and I'm not sure this is necessarily an issue. It seems to be convention that on the web, "archive" usually means "all"; here's [the first example I found with the word archive in the title](https://www.pori.fi/en/uutisarkisto?field_theme_tid=All).

The wording on the home page isn't exclusive - "Old courses can be found at the course archive" doesn't suggest (in my opinion) that you can't also find current and future courses there, just that you definitely can find the old ones there.

I like having a sitemap-style page that shows _all_ courses, so rather than change how this page works, I'd suggest renaming it to "all courses" or something like that, if we are decided that the title is confusing.

saarasat:
> It seems to be convention that on the web, "archive" usually means "all"; here's [the first example I found with the word archive in the title](https://www.pori.fi/en/uutisarkisto?field_theme_tid=All).

Now that you put it like that, it makes total sense having a page showing all courses. 

I think that the first time I was browsing through A+, this user experience was what confused me a bit: 
1. front page showed the most recent courses (makes sense), 
2. clicked through the link that says "old courses can be found at the course archive" (at this point I assumed I would land to a page with just the old courses that do not fit the front page) 
3. arrived in a page that was named archive which has also the most recent courses that I just saw on the front page as well.

So maybe indeed the link text or the name of the page could be changed. Something like "all courses can be found at the course archive" or "browse all courses at the course archive" or "check all courses at the course archive" or "browse rest of the courses at the course archive" or "you can see a list of all courses at the course archive" and/or changing the title to "all courses".  

I kinda like the user experience here: https://courses.aalto.fi/

If we are undecided here, maybe we can take a quick vote at the daily to see whether I am the only one who finds this experience confusing :)

mtthwhggns:
@saarasat how would you feel about something like this, instead of the current sentence-based link?

localhost_8000_ (3)

saarasat:
> @saarasat how would you feel about something like this, instead of the current sentence-based link?

Yes, at least to me that looks good! And logical. I would go with that

mtthwhggns:
As per @saarasat 's suggestion, I've now added a new link on the front page. 

localhost_8000_ (4)

assets/css/main.css
```diff
@@ -293,29 +293,38 @@ html, body {
```
saarasat, line 293: I just ran this locally on my machine, and is this happening to you as well, or is just me?

image

course/templates/course/index.html
```diff
@@ -13,63 +13,63 @@ <h1>{{ welcome_text|safe }}</h1>
...
-					<div class="card-body">
-						<h3 class="card-title">{{ advert.title }}</h3>
-						<p class="card-text">{{ advert.text|safe }}</p>
-					</div>
-				</a>
+		<div class="section-heading">
+			<h2>{% trans "All ongoing courses" %}</h2>
+			<div class="action">
+				{% url 'archive' as url %}
+				<a href="{{ url }}">{% trans "All courses" %}</a>
```
saarasat, line 39: And a tiny thing, but should the link be in line with the width of the card-section? I think the panel heading -line was overflowing before a bit. It didn't matter as much when there wasn't a link, but now it seems just maybe a tad off? 

image

mtthwhggns: I don't think this is overflowing, it's aligned with the `.cards` container (outlined in red here)
Näyttökuva 2020-7-24 kello 11 20 35

The problem is that we show cards which don't stretch to fill the available space, and have a fixed width of `30%`, which naturally leaves some space. I'm not sure of the right way to address this, as the window size can be anything - @saarasat what are your thoughts?

mtthwhggns:
> I just re-ran this locally. Is this happening just to me?

@saarasat No, I changed lots of the CSS for the front page and didn't go back to fix the archive yet until we were in agreement on the front page changes. It's a 30-second task though, just need to update a couple of classes to more generic names.

course/templates/course/index.html
saarasat: Hmm.. the flexible card-sizes do present a problem with that alignment, that's true. I can't think of a way to align it better either, maybe it can be as it is.

saarasat: Did you remove the word "see" from the link text on purpose? I thought it was helpful and suits the link well, maybe that could be added back? Then I think we are all set on my behalf :)

mtthwhggns: I never added it, actually. It was only in my quick devtools mockup, then I forgot about it. 

But I think it makes sense, so I'll add it now.

mtthwhggns:
I've updated this now, taking into account all of the comments so far.

mtthwhggns:
This has been rebased and rewritten in SCSS, so that the "fontpage cards" become their own imported component.

assets/sass/components/_frontpage-cards.scss
```diff
@@ -0,0 +1,129 @@
+
+.frontpage {
+  &.frontpage-section {
+    margin: 1%;
```
raphendyr, line 4: These use indentation of 2 spaces, while rest of the file tabs (set in .editorconfig).

mtthwhggns: Thanks, I'd not got that far in getting up-to-speed with the changes. I think I updatd this now.

assets/sass/components/_frontpage-cards.scss
```diff
@@ -0,0 +1,128 @@
...
+				flex-direction: row;
+				justify-content: center;
+				align-items: center;
+				font-size: 18px;
+				margin-top: 20px;
+				margin-bottom: 10px;
+			}
+		}
+	}
+	.panel-default-text {
```
raphendyr, line 23: I was thinking the style guide, but should there be empty line between `}` and the next style selector? Would it help readability?

Same with the nested stuff like

```scss
.foo {
  color: black;
  > .bar {
    color: white;
  }
}
// vs.
.foo {
  color: black;

  > .bar {
    color: white;
  }
}
```

Wanna share some opinions?

mtthwhggns: I don't have much preference. 

At the moment, I've used the SCSS defaults that ship with [Prettier](https://prettier.io/docs/en/why-prettier.html) in vscode, because I like not having to think too much about the styles - if we're going to make some rules, it would be nice to find rules which can be applied automatically like this.

raphendyr: For sure, there should be configurations for different formatters and linters. My idea was to write down them first as text and developers could configure their preferred tools and contribute the configs to the repo.

I didn't find the default configs to look for those rules. I guess what ever formatting is fine for now.

Btw. `options.editorconfig` might be handy for the indentation.

mtthwhggns: I guess I mis-described it, [Prettier isn't a linter](https://prettier.io/docs/en/comparison.html) with a list of style rules, as such. It reprints the code consistently - we couldn't, for example, just tweak a setting to get the blank lines.

If we want those, I can switch to a tool like stylelint. Since we didn't have any guidelines yet, I preferred going with the simpler Prettier approach. 

With relatively limited time this summer, I wouldn't caution against putting much time into writing guidelines yet if it's going to block things. With the tools available, files can easily be re-formatted if we get such guidelines in the future.

raphendyr: Yes, prettier / auto formatter / beautifier is not a linter, but both families of tools often have configuration.

You can edit configuration, if you like, but I think it's not required. I would prefer to use as much defaults as possible, but I personally prefer there to be a bit of extra space (as empty lines).

Nevertheless, this PR can be accept as it is now.

assets/sass/main.scss
```diff
@@ -22,4 +22,5 @@ $brand-primary: #005EB8;
 /*!
  * A+ Styles
  */
-@import 'protocols/quiz1';
+ @import 'components/frontpage-cards';
+ @import 'protocols/quiz1';
```
Mankro, line 26: Is there a reason for starting the `@import` lines with one space?

Mankro:
Looks like this is now ready, reviewed, rebased and waiting for merge, right?

I was quickly looking at the code and saw some "condensed" course cards. Does the front page show cards with full info and image headers while the archive lists all courses in the system with smaller, condensed cards?

mtthwhggns:
> Looks like this is now ready, reviewed, rebased and waiting for merge, right?
> 
> 
> 
> I was quickly looking at the code and saw some "condensed" course cards. Does the front page show cards with full info and image headers while the archive lists all courses in the system with smaller, condensed cards?

That's exactly how it works, yes!

Mankro:
@mtthwhggns I merged the button PR first and now this PR has git conflicts in the CSS files. I will merge this once you have rebased again. :+1:

mtthwhggns:
> @mtthwhggns I merged the button PR first and now this PR has git conflicts in the CSS files. I will merge this once you have rebased again. 👍

Done now.

PR #632, opened 2020-07-27
Save language to enrollment

veeraalt:
# Description

Previously, when user changed the language on a course page, language selection was saved in a cookie or in their `UserProfile` object. This PR changes this logic so that
* When a logged-in student changes language in a course page, it is saved to the new `language` field in their `Enrollment` object, and used as default language for displaying content in course pages.
* For course staff (teachers and assistants), selected language is saved to their `UserProfile` and used for displaying course content as well as other A+ pages.
* For users who are not logged in, language is still saved to a cookie when changed on course page.

2/3 of issue https://github.com/apluslms/a-plus/issues/518

# Testing
* Existing unit tests pass

* Testing has been done manually by checking that the selected language is saved to user's `Enrollment` or `UserProfile` object or in a cookie when changed on a course page, depending of the type of user, and that course content is correctly displayed in the selected language. For students, changing the course language doesn't affect the language in which other A+ content is displayed.

* No, I don't think it's relevant.


course/views.py
```diff
@@ -224,3 +229,49 @@ def form_valid(self, form):
...
+        response = HttpResponseRedirect(next) if next else HttpResponse(status=204)
+        if request.method == 'POST':
+            lang_code = request.POST.get(LANGUAGE_PARAMETER)
+            if lang_code and check_for_language(lang_code):
+                if next:
+                    next_trans = translate_url(next, lang_code)
+                    if next_trans != next:
+                        response = HttpResponseRedirect(next_trans)
+                    if request.user.is_authenticated:
+                        enrollment = self.instance.get_enrollment_for(self.request.user)
```
raphendyr, line 259: Because this is a derived from `CourseInstanceMixin`, you can just use `self.user_in_course`

course/viewbase.py
```diff
@@ -57,31 +57,36 @@ def get_resource_objects(self):
...
-                    translation.activate(lang)
+                instance_languages = self.instance.language
+
+                languages = []
+                if self.user_in_course and self.user_in_course.language:
+                    languages.append(self.user_in_course.language)
+                if user.is_authenticated and user.userprofile.language:
+                    languages.append(user.userprofile.language)
+                languages.append(get_language())
+                language = languages[0].split("-")[0]
```
raphendyr, line 84: I think this would benefit from a loop, i.e., it would try to find a language that is defined for the course. If not, then the above list could be replaced with just a `if .. elif .. else`.

veeraalt: Thanks, I'll try to make it use a loop. 

Is it needed to check that the language is defined for the course? So for example if the course is only available in English, should the course page be in English when user has set their profile language to be Finnish? If the course is not available in the given language, it only affects the side bar and dates etc but not the actual content, like in this pic:

518-course-language

raphendyr: Idea is that all content in the page is in the same language, hence we try to find a language that is defined for the course.

veeraalt: Ok, that clears it :+1:

course/viewbase.py
```diff
@@ -57,31 +57,43 @@ def get_resource_objects(self):
...
+
+                languages = []
+                if self.user_in_course and self.user_in_course.language:
+                    languages.append(self.user_in_course.language)
+                if user.is_authenticated and user.userprofile.language:
+                    languages.append(user.userprofile.language)
+                languages.append(get_language())
+
+                for lang in languages:
+                    lang = lang.split("-")[0]
```
raphendyr, line 87: If `languages = ['en-fi', 'fi']` and ` instance_languages = |en|sv|`, then this code will activate language `translation.activate('en')`, which would mean that formats strored in `formats/en-fi/formats.py` would not be used.

Thus, I think you should do:

```python3
instance_languages = self.instance.language.strip('|').split('|')
instance_def_language = instance_languages[0]
instance_languages = set(instance_languages) # find (i.e. `in`) is fast with sets

for lang in languages:
  if lang.split('-', 1)[0] in instance_languages:
    language = lang
    break
```

However, this requires to fix all instances of `get_language()` in rest of the code for `en-fi` etc. to work correctly, which of course would be a problem for another issue/PR.

raphendyr: This assumes that courses can use only the language part of the locale string (i.e. they can't use `en-us`).

course/viewbase.py
```diff
@@ -57,31 +57,43 @@ def get_resource_objects(self):
...
+                    if lang in instance_languages:
+                        language = lang
+                        break
                 else:
-                    translation.activate(lang)
+                    if instance_languages.startswith("|"):
+                        language = instance_languages[1:instance_languages.find("|", 1)]
```
raphendyr, line 93: This is good. Alternative would be using `split`, but both are relatively readable, so this is fine. But just for your info:

```
>>> timeit("lg = lgs[1:].partition('|')[0]", "lgs = '|en|fi|sv|ru|'")
0.15018817299278453
>>> timeit("lg = lgs[1:lgs.find('|', 1)]", "lgs = '|en|fi|sv|ru|'")
0.14492674503708258
>>> timeit("lg = lgs.split('|', 2)[1]", "lgs = '|en|fi|sv|ru|'")
0.11502113298047334
```

course/viewbase.py
```diff
@@ -57,31 +57,43 @@ def get_resource_objects(self):
             self.instance = instance
             self.course = self.instance.course
             self.content = CachedContent(self.instance)
+            self.user_in_course = None
```
raphendyr, line 60: Let's go with `user_course_data` for this.

course/viewbase.py
raphendyr: Btw. #620 will introduce `instance.default_language`, which could be used here, once that is merged.

course/viewbase.py
veeraalt: Good points regarding the `formats/en-fi/formats.py` implementation.

> This assumes that courses can use only the language part of the locale string (i.e. they can't use en-us).

Should we try to handle this somehow or is it okay to assume that the courses can now use only the language part?

course/viewbase.py
veeraalt: Thanks!
> #620 will introduce instance.default_language, which could be used here, once that is merged.

Nice, I'll use that when it's merged.

Mankro: #620 is now merged, but it is not yet in the master branch since it is part of the hotfixes that we want to deploy to production in August.

course/viewbase.py
raphendyr: > -- is it okay to assume that the courses can now use only the language part?

Yes, I think we can do that assumption.

p.s. sorry, missed the comment when handling notifications.

course/viewbase.py
raphendyr: Apparently, we didn't need that here after all.

veeraalt:
Squashed and rebased!

PR #637, opened 2020-08-01
Questionnaire feedback summary  - changes in a-plus

saarasat:
Related to the changes in https://github.com/apluslms/mooc-grader/pull/78


assets/sass/components/_alerts.scss
```diff
@@ -0,0 +1,19 @@
+
+.error-summary-outer {
```
raphendyr, line 2: Can we assume these to be part of `aplus-quiz1` protocol, so should they be wrapped into that too?

saarasat: Mine and Matthew's original idea was that these could be used elsewhere, for other types of validation error summaries or error summaries within A+ too. After this proof-of-concept would be agreed on. So I left the aplus-quiz1 out for that reason. What is the correct way for introducing new components that could be used within the quiz and outside of it? I am not quite sure

raphendyr: I guess it could then be included in both.

e.g.

```sass
.error-summary {
  color: black;
}

.aplus-quiz1 .error-summary {
  @extend .error-summary;
}
```

This would make it clear that it's also part of the protocol, specially if it's in the same file as the rest of the quiz1 stuff.

However, it could be enough just to mention the class names in the quiz1.scss file and here, so if the class is removed, then the developer remembers to move it instead.

saarasat: First I added also the `@extend` to the quiz1-file, but realised that if the summary is used for other exercises as well within the mooc-grader, then html-template would have that unnecessary "aplus-quiz1" mention. So I mentioned it in the quiz1-file, as a reminder. The file will obviously then make more sense once the questionnaire feedback PR is merged :D

assets/sass/components/_alerts.scss
```diff
@@ -0,0 +1,19 @@
+
```
mtthwhggns, line 1: I wonder if this file should be named to match the component, so `_error_summary.scss`, or something similar?

assets/sass/components/_alerts.scss
mtthwhggns, line 2: Could we try wrapping these styles inside a container with a single class? Take a look at #631 for the pattern. 

It's commonly used in methodologies like [BEM (Block, element modifier)](http://getbem.com/introduction/), and SASS makes it easy with nesting. If every new "component" has this style of container, there's much less change of styles overwriting each other by accident.

assets/sass/components/_alerts.scss
```diff
@@ -0,0 +1,19 @@
...
+.error-summary-outer {
+	width: 100%;
+	background: repeating-linear-gradient(
+		45deg,
+		#a94442,
```
mtthwhggns, line 6: Thinking in terms of "magic numbers", could these be replaced with a variable at the top of the file?

saarasat: This was something I was meant to ask you about now that Sass has been added. Where would be the correct location for an "error-color"? And what's the correct way to declare it? The red color defined there is the same one, bs3 uses for alert texts, that is why I chose it. And then again, the white color there is the general background color of A+.

mtthwhggns: Good question.

Our SASS migration isn't finished. In PR #645 I've moved to new universal colours. If that is merged first, I'd use `$aplus-danger` as the red, `$aplus-background` as white. Ideally no components will include colour definitions.

Another option is to use the Bootstrap-specific variables. This works, but I don't want to do this in the longer term, when we have custom variables with more consistent use.

If this needs to be merged time-sensitively, then it might be easiest to just move these to variables at the top of the file, then replace them later. 

> the white color there is the general background color of A+

For now. we might want a "dark mode" one day, so I'd still recommend reading this from a variable.

assets/sass/components/_alerts.scss
saarasat: These were very good points, your input was very welcomed here! I implemented them now. I was trying to think of a way to give descriptive name for the different classes. Let me know, if you think these are ok, or whether you had something else in mind.

assets/sass/components/_alerts.scss
saarasat: I also removed the use of magic numbers, and used the same shade of red you used with #645. If your button-thing is merged before, I can then move these. If not, at least there is now the continuity with colors.

assets/sass/components/_alerts.scss
mtthwhggns: This looks good 👍

assets/sass/components/_alerts.scss
mtthwhggns: This is good, now we just have to remember to remove these when we can!

raphendyr:
Blocked by #645 (let's merge it first)

assets/sass/components/_error_summary.scss
```diff
@@ -0,0 +1,27 @@
+$aplus-danger: #A50000;
+$aplus-background: #FFF;
+$width: 100%;
+$padding-amount: 5px;  
+$margin: 20px;
+
+
+.error-summary {
+	> .background {
```
raphendyr, line 9: These are pretty short selectors. It's very possible that there is css class that targets `.background`. So maybe `&__background` (i.e. `.error-summary__background`).

assets/sass/components/_error_summary.scss
```diff
@@ -0,0 +1,27 @@
...
+			45deg,
+			$aplus-danger,
+			$aplus-danger $padding-amount,
+			$aplus-background $padding-amount,
+			$aplus-background $padding-amount * 2
+		);
+		padding: $padding-amount;
+		color: $aplus-danger;
+		margin: $margin 0px; 	
+		> .text-container {
```
raphendyr, line 20: These are pretty short selectors. It's very possible that there is css class that targets `.text-container`. So maybe `&__text` (i.e. `.error-summary__background__text`).

saarasat:
I now modified these classes based on the comments here and in the other repo, to account all of the above :)

assets/sass/main.scss
```diff
@@ -33,6 +33,7 @@ $brand-danger: $aplus-danger;
 /*!
  * A+ Styles
  */
-@import 'protocols/quiz1';
 @import 'components/button';
+@import 'components/error_summary';
 @import 'components/frontpage-cards';
+@import 'protocols/quiz1';
```
raphendyr, line 39: These could be the last thing in the `main.scss` and could even have similar block as above with "A+ Protocols" (note, the format including `!` is important).

PR #639, opened 2020-08-04
Jwt token

### LINKED ISSUE #514: Use JWT token to encode the current course instance for API authentication
Some API endpoints, such as `/api/v2/users`, are not tied to any course instance. In order to allow teachers to use the API, we need to know the course so that we can check the teachers of the course and if the user is one of the teachers, It is inefficient to check all courses in the database. The course instance can be encoded in a JWT token so that the course instance can be parsed from the request without using URL parameters for that.

This is related to https://github.com/apluslms/a-plus/issues/476 and its PR https://github.com/apluslms/a-plus/pull/479. The comment (in #479) mentions the permission `IsTeacherOrAdminOrSelf` https://github.com/apluslms/a-plus/pull/479#issuecomment-607310883

HenrikSundell:
# Description

*Provide an overview of the changes and their purpose: what, why, how.*

Fixes #514

# Testing


templates/base.html
```diff
@@ -49,7 +52,9 @@
         <link rel="stylesheet" href="{% static 'django_colortag.css' %}" />
         <script src="{% static 'js/polyfill-find.js' %}"></script>
         <script src="{% static 'js/focus-visible-polyfill.min.js' %}"></script>
-        <script src="{% static 'js/aplus.js' %}"></script>
+        <script 
+            src="{% static 'js/aplus.js' %}"
+            course-code="{{ course.code }}"></script>
```
ttsirkia, line 57: This could be a separate `meta` tag containing the course code. Would also remove the IE11 hack.

assets/js/aplus.js
```diff
@@ -600,6 +604,47 @@ $(function() {
...
+                var header = {
+                "alg": "HS256",
+                "typ": "JWT"
+                };
+  
+                var data = {
+                    "course": course_code
+                };
+  
+                var secret = "extremly secret:)";
```
raphendyr, line 618: If you think about it, this isn't very secret :)

In other words, the JWT token is supposed to be generated where it is secret, i.e., in the server.

HenrikSundell: Yeah thought so, not really sure how/where it should e done

lib/api/authentication/api_authentication.py
```diff
@@ -21,12 +31,15 @@ def authenticate(self, request):
...
             return None
-        
+
         self.enforce_csrf(request)
         
         if 'Authorization' in request.headers:
             token = request.headers['Authorization'].split()[1]
-            return (user, token)
+            payload = jwt.decode(token, settings.SECRET_KEY) #TODO check if valid and handle bad signature
+            web_token = WebToken(payload['uid'], payload['course'])
```
raphendyr: You could move the decode to the token class:

```python
class WebToken:
    @classmethod
    def from_jwt(cls, token, secret=None):
        if not secret:
            secret = settings.SECRET_KEY
        payload = jwt.decode(token, secret)
        return cls(payload)

    def __init__(self, payload=None):
        self._payload = payload
        self.user_id = payload.get('uid')
        self.course_id = payload.get('course')

# usage:
web_token = WebToken.from_jwt(token)
```

btw. you could also wrap encoding to the class (though, then you should the class to another file)

```python
class WebToken:
    def __init__(self, payload=None):
        self._payload = payload or {}

    @property
    def user_id(self):
        return self._payload.get('uid')

    @user_id.setter
    def user_id(self, value):
        self._payload['uid'] = value

    def _user_set(self, user):
        if user:
            self.user_id = user.id
    user = property(None, _user_set) # getter might not be so relevant

    def encode(self, secret=None):
        if not secret:
            secret = settings.SECRET_KEY
        return jwt.encode(self._payload, secret, 'HS256')

# usage:
wt = WebToken()
wt.user = user
wt.course = course
token = wt.encode()
```

Remember, these are just some ideas for you to think about if you like or not.

lib/api/authentication/web_authentication.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Enforce CSRF validation for session based authentication.
+        """
+        check = CSRFCheck()
+        # populates request.META['CSRF_COOKIE'], which is used in process_view()
+        check.process_request(request)
+        reason = check.process_view(request, None, (), {})
+        if reason:
+            # CSRF failed, bail with explicit error message
+            raise exceptions.PermissionDenied('CSRF Failed: %s' % reason)
```
raphendyr, line 57: The `exceptions` is not defined (i.e. imported).

lib/api/authentication/web_authentication.py
```diff
@@ -0,0 +1,57 @@
...
+        user = getattr(request._request, 'user', None)
+
+        # Unauthenticated, CSRF validation not required
+        if not user or not user.is_active:
+            return None
+
+        self.enforce_csrf(request)
+        
+        if 'Authorization' in request.headers:
+            token = request.headers['Authorization'].split()[1]
```
raphendyr, line 38: Remember to check that the first word is `Bearer`.

HenrikSundell: Is it ok to handle somehow faulty tokens as no token, eg missing bearer or invalid jwt, if not what would be better?

ttsirkia: And especially that there are two parts after splitting, otherwise the server simply crash when submitting a request with nice headers.

Mankro:
Should the PyJWT library (pip package) be added to the requirements.txt file?

With these new JWT tokens implemented, how would the permission `IsTeacherOrAdminOrSelf` https://github.com/apluslms/a-plus/blob/9e595a0a902d19bcadeeaff8b3160873b0265f43/userprofile/permissions.py#L31 be implemented more efficiently? Does the `request` object or `request.user` get a new attribute for the course instance? That permission class actually doesn't get the request in parameters.

raphendyr:
> Does the request object or request.user get a new attribute for the course instance? That permission class actually doesn't get the request in parameters.

`request.auth.course` would be likely the location for the info. The permission check can be made better, once the normal TokenAuthentication is replaced with JWT based solution too, i.e., when the permission can disallow access if the course info is missing.

lib/api/authentication/web_authentication.py
```diff
@@ -35,9 +43,13 @@ def authenticate(self, request):
...
-            token = request.headers['Authorization'].split()[1]
-            payload = jwt.decode(token, settings.SECRET_KEY) #TODO check if valid and handle bad signature
-            web_token = WebToken(payload['uid'], payload['course'])
+            try:
+                scheme, token = request.headers['Authorization'].split()
+                if scheme != 'Bearer':
+                    raise ValueError
+            except ValueError:
+                return (user, None)
+            web_token = WebToken.from_jwt(token)
```
raphendyr, line 53: You should capture the verification failure, which can be raised by the `jwt.decode`.

lib/api/authentication/web_authentication.py
```diff
@@ -48,7 +48,7 @@ def authenticate(self, request):
                 scheme, token = request.headers['Authorization'].split()
                 if scheme != 'Bearer':
                     raise ValueError
-            except ValueError:
+            except (ValueError, jwt.exceptions.InvalidSignatureError, jwt.exceptions.DecodeError):
```
raphendyr, line 51: Not sure where the jwt errors would come?

HenrikSundell: One the line below ;) but yeah it's wrong

lib/api/authentication/web_authentication.py
```diff
@@ -48,9 +48,9 @@ def authenticate(self, request):
                 scheme, token = request.headers['Authorization'].split()
                 if scheme != 'Bearer':
                     raise ValueError
+                web_token = WebToken.from_jwt(token)
```
raphendyr, line 51: Separate try..except blocks would be better. Basically, a single block should contain only the code that is required. A+ code has a lot of bad practices related to this.

course/templatetags/course.py
```diff
@@ -124,6 +125,15 @@ def tags(profile, instance):
     tags = UserTagging.objects.get_all(profile, instance)
     return mark_safe(' '.join(tag.html_label for tag in tags))
 
+@register.simple_tag
+def api_token(profile, instance):
+    payload = {}
+    if profile:
+        payload['uid'] = profile.id
```
raphendyr, line 132: There is no mitch match. You add `user.userprofile.id` to `uid` and compare that later to `user.id`, which are different.

templates/base.html
```diff
@@ -24,6 +24,9 @@
 
         <script src="https://cdnjs.cloudflare.com/ajax/libs/clipboard.js/1.5.12/clipboard.min.js"></script>
 
+        <script src="//cdnjs.cloudflare.com/ajax/libs/crypto-js/3.1.2/rollups/hmac-sha256.js"></script>
```
raphendyr, line 27: These are not used anymore.

assets/js/aplus.js
```diff
@@ -600,6 +604,9 @@ $(function() {
             if (!csrfSafeMethod(settings.type) && !this.crossDomain) {
                 xhr.setRequestHeader("X-CSRFToken", getCookie('csrftoken'));
             }
+            if (true) {
```
raphendyr, line 607: You need to check at least `!this.crossDomain`, but also cheking that the url starts with `/api/` would be nice too.

course/templatetags/course.py
raphendyr: Would be better to store the `user.id`, which can be get from profile with `profile.user.id`.

assets/js/aplus.js
```diff
@@ -595,11 +599,25 @@ $(function() {
         // these HTTP methods do not require CSRF protection
         return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
     }
+    function isApiCall(url){
+        if (url.indexOf("/api/") == 0){
+            return true;
+        }
+        try {
+            var pathname = new URL(url).pathname;
```
raphendyr, line 607: This doesn't work in the IE 11, thus this requries something like

```js
const temp = document.createElement('a');
temp.href = url;
if (temp.path.indexOf(...))
```

lib/api/authentication/web_authentication.py
```diff
@@ -0,0 +1,73 @@
...
+                scheme, token = request.headers['Authorization'].split()
+                if scheme != 'Bearer':
+                    raise ValueError
+            except ValueError:
+                return (user, None)
+            try:
+                web_token = WebToken.from_jwt(token)
+            except (jwt.exceptions.InvalidSignatureError, jwt.exceptions.DecodeError):
+                return (user, None)
+            if web_token.verify_user(user):
```
raphendyr, line 57: If the user doesn't match, we should also fail. I.e. if there is a token, then it should be correct.

Issue #493, opened 2020-03-03
A+ shouldn't say "katso palaute/go to feedback" in the popup after a questionnaire's been submitted unless there's actually some written feedback to see


jaguarfi:
> **Migrated from Trello**

Related issues: https://github.com/apluslms/a-plus/issues/492, https://github.com/apluslms/a-plus/issues/490

Mankro:
I guess the problem here is that it is confusing to see an almost empty modal dialogue after submitting a questionnaire. It is also confusing to see the phrase "go to feedback" in the dialogue if there is no feedback to see besides the points.

Is it a good idea to completely remove the modal dialogue for questionnaires? 

A+ does not understand the content inside the feedback HTML. It is only a chunk of HTML from A+'s perspective, thus A+ can not easily know whether the feedback contains any useful comments besides the points. In questionnaires, the feedback HTML contains the whole form with the questions, so it is not feasible to parse "the real feedback" from it. A+ does not know if any "real feedback" with useful comments is included in the HTML.

In very distant future, we could rewrite how A+ handles and renders feedback and then A+ could understand what kind of content is included there. Until then, it is not feasible to recognize whether the feedback contains any "real feedback".

@jsorva do you want to comment? Would you like to simply remove the almost empty modal dialogue for questionnaires?

PR #640, opened 2020-08-04
Remove questionnaire feedback modal

saarasat:
Preferably not to be merged before: https://github.com/apluslms/a-plus/pull/618

Fixes https://github.com/apluslms/a-plus/issues/493

Previously the modal indicated the status of a submission, whether it was accepted, rejected, there was an error or it the submission was waiting. It also included a feedback-button going nowhere and the points gained. This information is duplicate to the information presented already within the exercise-frame, so the modal was removed. 

**Examples of the feedback modal:**
Sub-initialized-modal

Feedback after wrong answer

**Currently**:

- Submission accepted: points, hints, and feedback on correctness is shown and improved in https://github.com/apluslms/a-plus/pull/618

- Submission rejected: User generated errors, such as validation errors when data is in the wrong format are handled now in the mooc-grader, and error message has been improved in https://github.com/apluslms/a-plus/pull/618 and https://github.com/apluslms/a-plus/pull/637. 

- Error where there is no form (or feedback) to be shown are already presented within the frame as well.

- Posting submission: The main thing missing from within the frame, was the indication of posting submission and the possible error at the aplus-side. That has been added as a similar progress-bar as presented elsewhere and placed in the exercise frame. The overlay is opened and closed with JS. 

**While posting:**

image

**After posting, if there is an error at the aplus-side. The content of the form will not be deleted**
Screenshot from 2020-08-19 13-38-25

**After posting and if there is no feedback to be presented**:

image

**For other types of exercises, the modal remains the same. This form of presenting the submission on an overlay and displaying the error message could be implemented elsewhere as well** 

# Testing

- The implementation has been tested with working and broken mooc-grader questionnaires.
- The "broken form -situation" and a longer submission time can be tested eg. by mounting both the mooc-grader and a-plus-code in a container, and adding an indentation error somewhere in the GradedForm-class within the mooc-grader.
- Has been tested by raising ValueErrors in exercise/views.py
- Has also been tested when there are validation errors 
- Has been tested in a case where there are randomized checkbox-questions. (So that the removal of the modal does not interfere with that process and feedback).
- Translations have been checked, and they work.

Not relevant


raphendyr:
How does this handle/show errors from the post request (e.g. takes too much time or A+ doesn't respond after being shutdown)?

saarasat:
> How does this handle/show errors from the post request (e.g. takes too much time or A+ doesn't respond after being shutdown)?

> Error with submission, connection errors etc. where there is no form (or feedback) to be shown are already presented within the frame as well. 

The errors are handled in so many ways and places that it took me a while to figure this one out, but I think the modal does not present any new valuable information about the error in any situation, and the error situation is already handled within the exercise-frame, in any case in here: https://github.com/apluslms/a-plus/blob/master/exercise/protocol/aplus.py 

Basically what happens in an communication error with the mooc-grader situation now **with the modal,** is that it loads for a very long time, the modal pops up, shows the progress bar with the text "posting submission" and after the submission fails, this type of not-very-informative modal is given:
image

The badge-text is set according to the submission status, and does not really tell that much. Also the title-text can also be "Submission received" coming from the quiz-success-selector, even though there actually is an error with the submission, so that is not really helpful. 

And then afterwards when page is loaded, and there is no feedback the alert within the exercise-frame is shown here: https://github.com/apluslms/a-plus/blob/master/exercise/templates/exercise/submission_plain.html.

The other possible post error message -situation is set like this:
1. open general page modal on submission:
https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L564
2. Time fail of submission on the aplus-side:
https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L427
3. Set the modal to show the "error version" of the progress bar:
https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L443
https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L18
4. And present the text "Communication error with the exercise" and the red progress bar:
https://github.com/apluslms/a-plus/blob/master/exercise/templates/exercise/exercise.html#L64

So the modal never presents any specific error messages in any case. 

What happens after I removed the modal is that if the submission takes longer than 1,5 seconds, the progress bar and the text "posting submission" is displayed. The submit-button remains disabled for the duration of submission. If then, the post fails for any reason, the exercise frame is filled with the error message coming from the RemotePageException in any case (right?), so the almost empty modal is not needed: https://github.com/apluslms/a-plus/blob/master/exercise/protocol/aplus.py

image

And later on when, the page is reloaded and there is no feedback to be presented, everything happens according to this:
https://github.com/apluslms/a-plus/blob/master/exercise/templates/exercise/submission_plain.html

Can you think of an convenient way of testing those kinds of errors in aplus? Or the possible scenarios of submission?

raphendyr:
Thanks for the details. As you noticed, there are many different types of error cases. I think cases where the exercise is left in a broken state (no feedback, hence no questionnaire) can be left for later (I think there was an issue about it).

For now, I was mostly interested about the communication error, which changes the text in the progress bar to "communication error", but so far that hasn't been very visible. I think that could be done better here. For example, the progress bar could be removed, the submission button enabled, and the error message rendered there (which is removed if the person presses submit again).

> Can you think of an convenient way of testing those kinds of errors in aplus? Or the possible scenarios of submission?

One way is to add "raise RuntimeError" to the code that handles submissions and such. Sadly, testing errors is typically hard.

saarasat:
1. You mean this one information presented in the modal previously?
https://github.com/apluslms/a-plus/blob/master/exercise/templates/exercise/_exercise_wait.html

2. Is it also triggered here, by setting `page.is_wait`?
https://github.com/apluslms/a-plus/blob/master/exercise/protocol/aplus.py

3. And if yes to both questions above, am I interpreting this correctly, that when the "page.is_wait" for the questionnaire, the `quiz-submit-progress`-bar should be removed, the submit-button enabled, the form shown with whatever values are inserted in the fields at the time, and an error message presented? 

4. And is everything else currently presented within the frame as it was before submission, once the modal with the "communication error" -bar is closed? 

5. What is the status of the submission at that point?
Cause I wonder if the submit-button is already handled here in the js in those cases:
https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L865

And if I am completely off tracks here, or even if it is just tedious to reply to all these questions here :D , would you have time today or tomorrow to have a quick call about this?

raphendyr:
> would you have time today or tomorrow to have a quick call about this?

That is a good idea, let's talk after next daily.

exercise/static/exercise/chapter.js
```diff
@@ -213,6 +192,9 @@
 
 			this.loader = this.chapter.cloneLoader();
 
+			if (this.quiz) {
+				this.element.append(this.settings.quiz_progress_selector);
```
raphendyr, line 196: This doesn't look right. Assuming `this.element` is an element, then append would expect a HTML element and not a string containing `.quiz-submit-progress`.

exercise/static/exercise/chapter.js
```diff
@@ -561,19 +543,28 @@
 					});
 				});
 			} else {
-				chapter.openModal(chapter.messages.submit);
 				var exercise = this;
+				var submission_complete = false;
+				if (exercise.quiz) {
+					setTimeout(function(){
```
raphendyr, line 549: What happens if the submission is not completed after 1.5s? Timers are typically the wrong solution for everything (sometimes they are the right, but typically not). You can do same in the jqeury ajax `.always` callback.

Additionally, global variables, or variables which are conceptually polled (like here), do not fit in callback/event/future based programs (see https://en.wikipedia.org/wiki/Event-driven_programming for concepts).

As a rule of thumb, if you use `var` in JavaScript, then there is basically always a better way to do it. Bascally, all "variables" should be stored in `const`.

saarasat: As I (tried at least) to explain above is the process what happens here:

1. User clicks submit, and the "clock" on 1,5 s is started, submitAjax is called
2. For the first 1,5 s, everything in the form is shown as it is, except the submit-button is disabled
3. After 1,5 s has passed, and the process is stuck in these retrys https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L430 the progress-bar of the images above is shown, it says "posting submission".
4a. If the submitAjax fails, the "clock" is stopped, the progress bar hidden an this current message is shown:
image.
4b. Alternatively, if the submitAjax succeeds on some of the retrys, the progress bar is hidden and the result form is shown.

So the progress bar was there just to indicate to the user that "hold on, this submission is taking a bit longer than expected". And since the quiz submissions tend to be fast, I thought it does not make sense to show the user just a quick flash of a progress bar, hence the 1,5 seconds before the progress bar is changed. That was at least I was trying to accomplish here. 

What do you think? What would you differently here?

exercise/static/exercise/chapter.js
saarasat: You are right! I removed it

exercise/static/exercise/chapter.js
raphendyr: What happens when submission takes 10 seconds? Let's assume the user press "send" and after 1.5s they change some values and press send again (or don't). After 10 seconds, the JS would replace the form with a new feedback and all the students edits would be deleted?

I think this was the reason why I originally were thinking of an overlay to block the user from editing the fields while the data is under javascript control and might be replaced.

saarasat: These are valid points! 

However, originally [in the issue](https://github.com/apluslms/a-plus/issues/493), there was only talking about removing the empty modal. So the overlay-idea is new to me, at least I am not able to recall a situation where it would have come up. I can only recall talking about presenting the content of the modal within the exercise frame. It also differs from this idea:

> For now, I was mostly interested about the communication error, which changes the text in the progress bar to "communication error", but so far that hasn't been very visible. I think that could be done better here. For example, the progress bar could be removed, the submission button enabled, and the error message rendered there (which is removed if the person presses submit again).

But. I still have some time to implement this and I would like to get it right :) 

So could you clarify what would you like to be implemented here regarding the overlay? Is what you mean by the "data being under javascript control" that the overlay would be launched when the student hits "submit" and disappears when one of the events "exercise-submission-failure", "submission-finished" or "submission-aborted" is introduced? So similar kind of flow as with the submit-button now: 
https://github.com/apluslms/a-plus/blob/master/exercise/static/exercise/chapter.js#L857

Just so I know on how to continue.

raphendyr: Basically that. We can chat about it on monday!

exercise/static/exercise/chapter.js
```diff
@@ -436,18 +417,23 @@
...
+					exercise.quizSubmission("error");
+				} else if (exercise.active_element) {
 					// active elements don't use loadbar so the error message must be shown
 					// in the element container
 					var feedback = $("<div>");
 					feedback.attr('id', 'feedback');
 					feedback.append(exercise.chapter.messages.error);
 					exercise.updateOutput(feedback);
+				} else {
+					exercise.dom_element.dispatchEvent(
```
raphendyr, line 433: Can you look if you can reformat this logic, so that the failure event is only written once and not copied to two if-elif-else branches. E.g. by nesting if statements. This might make the flow of the code easier to follow.

exercise/static/exercise/chapter.js
Mankro: Since the `setTimeout` and `var submission_complete` are still here, does it mean that you ran out of time for doing this without them? Just to clarify, no panic if that's the case :)

saarasat: Yes, you are correct :) I didn't come up with a good way to replace them without using `setTimeout,` since the status of the submission does not actually change (I think). And it also looked a bit weird to have the overlay flashing there for a while with fast submissions. If you have any other ideas how that could be accomplished, or if you'd like 1s-delay to be removed all in all and the overlay to flash there, even for a split-second, just do let me know! That can still easily be changed of course :)

exercise/static/exercise/chapter.js
saarasat: Sure! I'll take a look at it later today

exercise/static/exercise/chapter.js
raphendyr: You could use a transparent overlay. I think the CSS can be used to change the transparency e.g. in 0.4s, thus if it flashes, then it's not actually visible to the user, but does still block any input.

saarasat: Uuu, that's a very nice idea! Thanks a lot!! I will try that today :)

saarasat: I tried the transition from transparent background to the darked background, it was a very good idea. But then the progress bar looked a bit weird, "hanging in the air" if someone happens to look at the top part. 

I also tried to just change the opacity of the background, but the child-elements also have then the same opacity and so the progress bar ended up looking weird. And there was no good way around that child-element-problem.

But talking about the transparency gave me another idea, and I implemented that. So I added the progress-bar within the exercise-frame, but in the upper part and the overlay as a a shadowy gray. 

It might not before the perfect solution, since there is that slight twitch of showing the progress bar, when the submission is fast. However, I guess it is also somehow expected behavior from a form, that the elements might move around slightly upon submission. And at least the input is now blocked by the overlay, so students don't go writing a lot of stuff there and lose it if the connection fails. 

And in any case, I think now this already significantly better than the modal. 

The JS is not the prettiest still cause there are those branches for different exercises. But this could be set as a base for removing the modal in the future from other exercises, and then those branches wouldn't be no longer needed.

exercise/static/exercise/chapter.js
```diff
@@ -158,6 +137,9 @@
 		content_element: '<div class="exercise-content"></div>',
 		content_selector: '.exercise-content',
 		exercise_selector: '#exercise-all',
+		quiz_progress_selector: '#quiz-submit-progress',
```
raphendyr, line 140: I think, thise should be classes as there can be multiple of these on a single page.

exercise/static/exercise/chapter.js
```diff
@@ -561,8 +563,12 @@
 					});
 				});
 			} else {
-				chapter.openModal(chapter.messages.submit);
 				var exercise = this;
+				if (exercise.quiz) {
+					exercise.quizSubmission("submit");
+				} else {
+					chapter.openModal(chapter.messages.submit);
```
raphendyr, line 570: These could include a comment `/* TODO: migrate to overlay */` as you are adding a new branch here (same for the other instances of the same thing too).

exercise/static/exercise/chapter.js
```diff
@@ -506,6 +492,22 @@
...
 
+		quizSubmission: function(message) {
+			var exercise = this;
+			if (message == "submit") {
+				exercise.element.find(exercise.settings.quiz_overlay_selector).removeClass("hide");
+				exercise.element.find(exercise.settings.quiz_progress_selector).removeClass("hide");
+			} else if (message == "error") {
+				exercise.element.find(exercise.settings.quiz_overlay_selector).hide();
+				exercise.element.find(exercise.settings.quiz_progress_selector).hide();
+				exercise.element.find(exercise.settings.quiz_error_selector).removeClass("hide");
```
raphendyr, line 503: Is this `removeClass("hide")` same as `show()`?

raphendyr (commented):
This is really good. However, could the progress bar be at the end of the exercise or both, in the start and the end? Or, could it be a sticky element? Would that work? My concern is that when the form is long, the progress bar witll be out of the view.

exercise/static/exercise/chapter.js
saarasat: I changed these now

exercise/static/exercise/chapter.js
saarasat: And commented the branches as well

exercise/static/exercise/chapter.js
saarasat: According to this, no actually: https://api.jquery.com/show/

Regarding this:

> However, could the progress bar be at the end of the exercise or both, in the start and the end? Or, could it be a sticky element? Would that work? My concern is that when the form is long, the progress bar witll be out of the view.

I added the stickiness, it actually works nicely! It was a good idea. And I think it was the best option of these, the duplicate bars looked funny I think. I added a mention of moving the sticky-definition to a helper-file. For future reference, would the `sass/helpers` be the correct place for something like this? 

Everything has also now been rebased and translations checked as well. I also run the tests now, and seems that selenium-tests now also all pass, after Veera's fix :)

exercise/static/exercise/chapter.js
```diff
@@ -571,9 +579,8 @@
 					//exercise.hideLoader();
 					var input = $(data);
 					if (exercise.quiz) {
-						var badge = input.find('.badge').eq(2).clone();
```
raphendyr, line 574: Do you have idea what this `.badge` refers to? Is it just points/status or is it some achievement thing?

exercise/templates/exercise/_submit_progress.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+
+<div class="quiz-submit-overlay submission-overlay hide"></div>
```
raphendyr, line 3: This element is missing when chapter is leaded with an existing feedback (i.e. first post a solution, reload page, then try to post a solution again). Could this be created in the JS instead?

exercise/static/exercise/chapter.js
saarasat: It was the badge on the modal for the questionnaire _(the 0/20 on the picture)_. It was given to the `modalSuccess` there.

image

exercise/templates/exercise/_submit_progress.html
saarasat: Hmm.. Probably. Is there an example somewhere? 

I mean, I have like 45 minutes left before I need to go (:D) and still some feedback questionnaires to do, I might not be able to get back to this now. But if you have a quick idea, I could try to fix it :)

exercise/static/exercise/chapter.js
raphendyr: This is probably not repeatable, as `hide` sets `display=none` and `removeClass("hide")` doesn't work for that, only for the class added in the template. `show()` checks that if the element has effective `display==none`, then it will set element to `display=<default display for the element>`, thus forcing it to be visible.

exercise/templates/exercise/_submit_progress.html
Mankro: Sounds like a very severe bug. We can't merge this yet. "Somebody" (maybe me) needs to spend some time working on this in the autumn.

Mankro (approved):
The code looks great! I will now fix the remaining issues myself.

exercise/templates/exercise/_submit_progress.html
Mankro: If the problem is simply that these overlay divs are missing from the page in some situations, then this small template could be always included in exercise.html.

Mankro: or maybe submission_plain.html

Mankro (approved):

exercise/templates/exercise/_submit_progress.html
Mankro: I added the overlay divs to to submission_plain.html and then I had to fix active elements in chapter.js since the code looks for alert divs in the feedback. The submission_plain.html now always contains a hidden alert div.

Issue #520, opened 2020-04-16
A+ should support different URLs for different langauges (e.g. same chapter in Finnish and English); both should be indexed by Google and be searchable with targeted search


jaguarfi:
>**Migrated from Trello**
----
Update: the support for different language URLs has been implemented (#518). The Google search indexing is missing. See the TODO list in the bottom.

**Conversation**

>**Markku Riekkinen:**
This sounds quite tricky. Currently, the language is stored in the user session, but this feature would override that language setting temporarily if the URL supplies it. The same chapter would also have three different URLs: 1) current normal URL without language 2) URL with English, and 3) URL with Finnish language code. Would it be adequate to add the new language code to GET parameters since those would not conflict with existing URL configuration?

>**Juha Sorva:**
**@markkuriekkinen** Not sure. How would that interact with Google indexing and SEO? In any case, I think the current URL configuration is fundamentally flawed in that it treats language versions of chapters as different variants of the same content (same URL) when they really should be treated as separate (even if related) documents. There's been some tentative discussion with **@raphendyr** about redesigning the URL scheme in a future version of A+.

>In any case, this is a very major issue IMHO: currently, students (and others) cannot find our content through web searches and our local Google search of ebook chapters doesn't work either.

>**Jaakko Kantojärvi:**
Due to how data is stored in the DB and stuff, I would say the query parameter that temporarily overwrites the user option for that page is the best option. Sadly, it will mostlikely mean that urls (e.g. navigation) will point to "normal" language and will drop the query parameter (Note: we can fix this with JS pretty nicely). Personally, I don't feel there is really good solution in any available ones.

>Any case, with query parameters we should be able to generate sitemap for google, so they would know all the available versions. Documentation from google about the topic and sitemap on this matter:

>https://support.google.com/webmasters/answer/182192
https://support.google.com/webmasters/answer/189077

>p.s. those google support pages use the same system. The page includes query parameter ``?hl=en`` when found using google search. Parameter ``?hl=fi`` changes the page to Finnish. e.g.:

>https://support.google.com/webmasters/answer/182192?hl=en
https://support.google.com/webmasters/answer/182192?hl=fi

>In addition, google support changes the locale parameter in your session after visiting the site. Thus, A+ could change the language from the query parameter for the current session, but keep it in user settings (#518 ). This of course, requires some UI to tell the user that they are viewing the course in a language not set in their settings.

TODO:
- [x] Change language based on language query parameter. This should not be saved in the django session (depends on #518)
[...]

PR #641, opened 2020-08-07
Change language based on query parameter

veeraalt:
# Description

This PR adds a possibility of linking to a specific page in A+ in a specific language or viewing content with different language without having to change and save it in profile or in course page, by using language query parameter at the end of the url. 
For example, adding ?hl=en at the end of the url changes the content to English if it's supported for the page. If not, an error is raised.

* If the language is supported for A+ but not for this course, a 404 html page is shown with a list of available languages and links for each.
520-error1

* If the language is not supported for A+ at all, a 404 page is shown with a link to the material in default language.
404-default

Language parameter stays in the url when navigating through page, except when user changes language either on course page
or on profile page.

This is achieved by adding an event listener to `assets/js/aplus.js` that listens to click events and copies query parameters from the current url to the next when a link is clicked. When user changes language either on course page or on profile page, the language query parameter is removed from the url.

1/3 of issue https://github.com/apluslms/a-plus/issues/520

# Testing

* Existing unit tests pass

* Testing has been done manually by navigating through pages and checking that the language parameter changes the content language and also stays in the url, and that the errors are raised properly.

* No, I'm not sure if it's needed.

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -629,5 +629,25 @@ $(function() {
     });
 })(jQuery, window, document);
 
+/*
+ * Listen to link click events and copy query parameters from current url
+ * to the next
+ * Note! Translation requires that at least 'hl' query parameter is copied
+ */
+(function(window) {
+    window.addEventListener('click', function(event) {
+        if (
+            event.target.nodeName === 'A'
+            && event.target.href
+            && event.target.protocol === window.location.protocol
+            && event.target.host === window.location.host // hostname:port
+            && (event.target.search === '' || event.target.search === '?') // no search
+        ) {
+            event.preventDefault(); // stop default redirect
+            event.target.search = window.location.search; // copy query params
+            window.location.href = event.target.href; // redirect
+        }
+    }, false);
+})(window);
 
 /* vim: set et ts=4 sw=4: */
--- a/course/viewbase.py
+++ b/course/viewbase.py
@@ -1,16 +1,18 @@
 from django.contrib import messages
 from django.core.exceptions import PermissionDenied
 from django.http import Http404
-from django.shortcuts import get_object_or_404
+from django.shortcuts import get_object_or_404, render_to_response
 from django.utils import translation
 from django.utils.translation import ugettext_lazy as _
-from django.utils.translation import get_language
+from django.utils.translation import get_language, get_language_info
 
 from authorization.permissions import ACCESS
 from exercise.cache.content import CachedContent
+from lib.helpers import remove_query_param_from_url, update_url_params
 from lib.viewbase import BaseTemplateView
 from userprofile.viewbase import UserProfileMixin
 from .cache.students import CachedStudent
+from .exceptions import TranslationNotFound
 from .permissions import (
     CourseVisiblePermission,
     CourseModulePermission,
@@ -75,24 +77,31 @@ class CourseInstanceBaseMixin(object):
             # Try to find a language that is defined for this course instance
             # and apply it
             if self.instance.language:
-
-                languages = []
-                if self.user_course_data and self.user_course_data.language:
-                    languages.append(self.user_course_data.language)
-                if is_real_user and user.userprofile.language:
-                    languages.append(user.userprofile.language)
-                languages.append(get_language())
-
                 instance_languages = self.instance.language.strip('|').split('|')
                 instance_def_language = instance_languages[0]
                 instance_languages = set(instance_languages)
 
-                for lang in languages:
-                    if lang.split('-', 1)[0] in instance_languages:
-                        language = lang
-                        break
+                query_language = self.request.GET.get('hl')
+                if query_language:
+                    if query_language[:2] in instance_languages:
+                        language = query_language
+                    else:
+                        raise TranslationNotFound
                 else:
-                    language = instance_def_language
+                    languages = []
+                    if self.user_course_data and self.user_course_data.language:
+                        languages.append(self.user_course_data.language)
+                    if is_real_user and user.userprofile.language:
+                        languages.append(user.userprofile.language)
+                    languages.append(get_language())
+
+                    for lang in languages:
+                        if lang[:2] in instance_languages:
+                            language = lang
+                            break
+                    else:
+                        language = instance_def_language
+
                 translation.activate(language)
 
     def get_access_mode(self):
@@ -116,6 +125,15 @@ class CourseInstanceMixin(CourseInstanceBaseMixin, UserProfileMixin):
             url=self.kwargs[self.instance_kw],
             course__url=self.kwargs[self.course_kw],
         )
+
+    def handle_exception(self, exc):
+        if isinstance(exc, TranslationNotFound):
+            instance_languages = self.instance.language.strip("|").split("|")
+            url = remove_query_param_from_url(self.request.get_full_path(), 'hl')
+            for i, lang in enumerate(instance_languages):
+                instance_languages[i] = {"name": get_language_info(lang)['name'], "url": update_url_params(url, {'hl' : lang})}
+            return render_to_response('404.html', {'error_msg': str(exc), 'languages': instance_languages}, status=404)
+        return super().handle_exception(exc)
 
 
 class CourseInstanceBaseView(CourseInstanceMixin, BaseTemplateView):
--- a/lib/middleware.py
+++ b/lib/middleware.py
@@ -3,6 +3,9 @@ from django.http import HttpResponseRedirect
 from django.utils import translation
 from django.utils.cache import patch_vary_headers
 from django.utils.deprecation import MiddlewareMixin
+from django.shortcuts import render_to_response
+
+from lib.helpers import remove_query_param_from_url
 
 
 class LocaleMiddleware(MiddlewareMixin):
@@ -10,7 +13,14 @@ class LocaleMiddleware(MiddlewareMixin):
     response_redirect_class = HttpResponseRedirect
 
     def process_request(self, request):
-        if hasattr(request, 'user') and request.user.is_authenticated:
+        query_language = request.GET.get('hl')
+        if query_language:
+            language = query_language
+            lang_codes = set(lang[0][:2] for lang in settings.LANGUAGES)
+            if language[:2] not in lang_codes:
+                url = remove_query_param_from_url(request.get_full_path(), 'hl')
+                return render_to_response('404.html', {'url_without_language': url}, status=404) 
+        elif hasattr(request, 'user') and request.user.is_authenticated and not request.user.is_anonymous:
             language = request.user.userprofile.language
         else:
             language = translation.get_language_from_request(request)
--- a/templates/404.html
+++ b/templates/404.html
@@ -16,12 +16,29 @@
                         <h1>404 Not found</h1>
                     </div>
                     {% include "_messages.html" %}
+                    {% if url_without_language %}
+                    <div class="alert alert-danger">
+                        {% trans "The page is not available in this language." %}
+                        <a href="{{ url_without_language }}">{% trans "Open the page in default language" %}</a>
+                    </div>
+                    {% elif languages %}
+                    <div class="alert alert-danger">
+                        {% blocktrans trimmed %}
+                        The course is not available in this language.
+                        You can view the course in these languages:
+                        {% endblocktrans %}
+                        {% for lang in languages %}
+                            <a href="{{ lang.url }}">{{ lang.name }}</a>
+                        {% endfor %}
+                    </div>
+                    {% else %}
                     <div class="alert alert-danger">
                         {% blocktrans %}
                         For an unknown reason you arrived somewhere that does not exist.
                         Please, try again.
                         {% endblocktrans %}
                     </div>
+                    {% endif %}
                 </div>
             </div>
         </div>
--- a/authorization/exceptions.py
+++ b/authorization/exceptions.py
@@ -1,4 +1,4 @@
 class ValidationFailed(Exception):
     def __init__(self, response):
-        super(ValidationFailed, self).__init__("HTTP Reqeust validation failed")
+        super(ValidationFailed, self).__init__("HTTP Request validation failed")
         self.response = response
--- a/course/exceptions.py
+++ b/course/exceptions.py
@@ -1,0 +1,2 @@
+class TranslationNotFound(Exception):
+    pass
--- a/course/views.py
+++ b/course/views.py
@@ -20,7 +20,7 @@ from django.utils.translation import ugettext_lazy as _
 from authorization.permissions import ACCESS
 from exercise.cache.hierarchy import NoSuchContent
 from exercise.models import LearningObject
-from lib.helpers import settings_text
+from lib.helpers import settings_text, remove_query_param_from_url
 from lib.viewbase import BaseTemplateView, BaseRedirectMixin, BaseFormView, BaseView
 from userprofile.viewbase import UserProfileView
 from .forms import GroupsForm, GroupSelectForm
@@ -251,12 +251,12 @@ class LanguageView(CourseInstanceMixin, BaseView):
     def post(self, request, *args, **kwargs):
         LANGUAGE_PARAMETER = 'language'
         
-        next = request.POST.get('next', request.GET.get('next'))
+        next = remove_query_param_from_url(request.POST.get('next', request.GET.get('next')), 'hl')
         if ((next or not request.is_ajax()) and
                 not is_safe_url(url=next,
                                 allowed_hosts={request.get_host()}, 
                                 require_https=request.is_secure())):
-            next = request.META.get('HTTP_REFERER')
+            next = remove_query_param_from_url(request.META.get('HTTP_REFERER'), 'hl')
             next = next and unquote(next)  # HTTP_REFERER may be encoded.
             if not is_safe_url(url=next,
                                allowed_hosts={request.get_host()},
--- a/lib/helpers.py
+++ b/lib/helpers.py
@@ -4,7 +4,7 @@ import functools
 import warnings
 from cachetools import cached, TTLCache
 from collections import OrderedDict
-from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
+from urllib.parse import parse_qs, parse_qsl, urlencode, urlsplit, urlunsplit
 from PIL import Image
 from django.conf import settings
 from django.utils.crypto import get_random_string as django_get_random_string
@@ -84,6 +84,16 @@ def url_with_query_in_data(url: str, data: dict = {}):
 def update_url_params(url, params):
     delimiter = "&" if "?" in url else "?"
     return url + delimiter + urlencode(params)
+
+
+def remove_query_param_from_url(url, param):
+    """
+    Take an url with (or without) query parameters. Return url without the selected query parameter.
+    """
+    url = urlsplit(url)
+    query = parse_qs(url.query, keep_blank_values=True)
+    query.pop(param, None)
+    return urlunsplit(url._replace(query=urlencode(query, True)))
 
 
 FILENAME_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ._-0123456789"
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-08-19 09:22+0300\n"
+"POT-Creation-Date: 2020-08-27 15:56+0300\n"
 "PO-Revision-Date: 2019-08-14 12:16+0200\n"
 "Last-Translator: Saara Satokangas <saara.satokangas@aalto.fi>\n"
 "Language-Team: Finnish <>\n"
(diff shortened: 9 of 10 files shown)


course/viewbase.py
```diff
@@ -117,6 +123,20 @@ def get_course_instance_object(self):
...
+        if isinstance(exc, TranslationNotFound):
+            instance_languages = self.instance.language
+            path_info = self.request.META.get('PATH_INFO')
+            languages = []
+            if instance_languages.startswith("|"):
+                languages = instance_languages.strip("|").split("|")
+            else:
+                languages = [instance_languages]
+            for i, lang in enumerate(languages):
+                languages[i] = {"name": get_language_info(lang)['name'], "url": path_info + '?hl=' + lang}
```
raphendyr, line 136: You could use `update_url_params` from `lib/helpers.py` for joining the url.

course/viewbase.py
```diff
@@ -117,6 +123,20 @@ def get_course_instance_object(self):
             course__url=self.kwargs[self.course_kw],
         )
 
+    def handle_exception(self, exc):
+        if isinstance(exc, TranslationNotFound):
+            instance_languages = self.instance.language
+            path_info = self.request.META.get('PATH_INFO')
```
raphendyr, line 129: Maybe a bit nicer would be calling `request.get_full_path()`. Ouh, and `PATH_INFO` doesn't contain the full url when the app is bound to a sub path (see `SCRIPT_NAME` and `PATH_INFO` [here](https://wsgi.readthedocs.io/en/latest/definitions.html)).

templates/404.html
```diff
@@ -16,12 +16,22 @@
                         <h1>404 Not found</h1>
                     </div>
                     {% include "_messages.html" %}
+                    {% if languages %}
+                    <div class="alert alert-danger">
+                        <p>The course is not available in this language.</p>
```
raphendyr, line 21: These are missing a translation block.

lib/middleware.py
```diff
@@ -1,19 +1,35 @@
...
 
 class LocaleMiddleware(MiddlewareMixin):
 
     response_redirect_class = HttpResponseRedirect
 
     def process_request(self, request):
-        if hasattr(request, 'user') and request.user.is_authenticated:
+        language = translation.get_language_from_request(request)
+        current_url = request.META.get('PATH_INFO')
+        last_url = request.META.get('HTTP_REFERER', '')
```
raphendyr, line 18: Yeah, this doesn't look very nice way to do this. I'll return on this after thinking a bit of ideas where to look a better solution.

templates/404.html
veeraalt: For loop doesn't work inside the translation block because "'blocktrans' doesn't allow other block tags (seen 'for lang in languages') inside it". Is it a good solution to wrap only the text part inside translation block?

raphendyr: You don't need to translate the for part, only the text before it

course/viewbase.py
```diff
@@ -74,15 +76,20 @@ def get_resource_objects(self):
             # Try to find a language that is defined for this course instance
             # and apply it
             if self.instance.language:
-
+                instance_languages = self.instance.language.strip('|').split('|')
                 languages = []
+                url_lang = self.request.GET.get('hl')
+                if url_lang:
+                    if url_lang in instance_languages:
```
raphendyr, line 83: This would benefit from the `instance_languages`, which is a set and not a list (i.e. you should have moved all 3 lines.

course/viewbase.py
```diff
@@ -116,6 +123,15 @@ def get_course_instance_object(self):
             course__url=self.kwargs[self.course_kw],
         )
 
+    def handle_exception(self, exc):
+        if isinstance(exc, TranslationNotFound):
+            instance_languages = self.instance.language.strip("|").split("|")
+            url = remove_query_param_from_url(self.request.get_full_path(), 'hl')
+            for i, lang in enumerate(instance_languages):
+                instance_languages[i] = {"name": get_language_info(lang)['name'], "url": update_url_params(url, {'hl' : lang})}
```
raphendyr, line 131: Hmm.. If A+ would be configured to use languages `en-fi` and `fi-fi` and the course would have `en` and `fi`, then this would generate links, which wouldn't use the correct value for the `hl` parameter. So, you probably need to iterate over `settings.LANGUAGES` and filter that, so that only those which have translation in the course are listed.

veeraalt: Can you clarify why is the `settings.LANGUAGES` needed? If the course has languages `en` and `fi`, then the links would have `?hl=en` and `?hl=fi`, aren't those the right ones?

veeraalt: Or should it be `?hl=en-fi` and `?hl=fi-fi`?

raphendyr: Your other part of the code requires that the `?hl=<lang>` is found from the `settings.LANGUAGES`, thus `?hl=en` should result 404 if there is only `en-us` and `fi`.

veeraalt: Maybe it's still quite unclear for me how the `en-fi` should work. If A+ has language `en-fi` and course is available in `en`, the course content should still surely be displayed in English? And I guess `hl` tag can only contain the first two characters of  the language code or how should it work with `en-us` for example?

raphendyr: As the tanslation.activate is called with the value of the `hl`, then the `hl` should contain the full language info (i.e. locale = language + location). The second code, in course instance, then shortens this locale to just tlanguage, which are used to select content from the course.

In other words, all Django stuff expect the language parameter to be fully qualified name (e.g. `en-us`), so it can select correct translation file and formats. However, the course content is translated only with 2 letter codes, so we need to shorten the `en-us` to just `en` in all course material related places (and assessment tools, like grading).

veeraalt: Thanks, that clears some things up but I'm somewhat confused about the formats still.

This is more a concern of https://github.com/apluslms/a-plus/pull/604, but that depends on how we solve the language locale/region issues here:

If we added the `formats/en-fi/formats.py` folder and it would be possible for the language to be `en`, `en-fi` and even `en-us` or `en-gb` all in the same session, would the date and time formats then vary between the ones defined in `formats/en/formats.py` and `formats/en-fi/formats.py`? That would be very confusing.

I don't see the point in supporting these different language formats (other than maybe for displaying dates and times) because it's not even possible for users to pick `en-fi`, `en-gb` or `en-us` as the preferred language, there is only `English`, so in which case `en-fi` would come up?

veeraalt: Maybe it would be simpler if the language could be only `fi` or `en` and there was a separate `date_format` field in `UserProfile` that could be changed from profile page, but I'm not sure how displaying the correct format could be implemented.

raphendyr: Let's write this down. Back in the day at times of UNIX, POSIX standard specified [locale syntax](https://en.wikipedia.org/wiki/Locale_(computer_software)) `[language[_territory][.codeset][@modifier]]`, e.g., `en_US.UTF-8` and `fi_FI.UTF-8` (the charset is now-a-days pretty useless as unicode contains it all, things were different before it). Another relevant info is how [HTTP specified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) `Accept-Language` header (e.g. `en`, `en-US`, `fi`) (see also [BCP 47 / IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag)). The problem which people has tried to solve is how to differentiate differences in the language.

A good question is why formats and languages are typically connected like this. Like why `fi_FI` defines formats used in Finalnd? Technically, that string contains all required information. The `en_FI` could be parsed as show texts with language English, but use geographical information related to Finland. For the latter, there is clear definition how numbers and such are represented in our location. Interestingly in Switzerland, the location part is not enough to define the used formats, thus both the language and the location are required to get the exact information.

Many distributions don't include all possible combinations for locales, e.g., `en_FI` doesn't exist in Debian or Ubuntu. However, it's part of the [Unicode Common Locale Data Repository](http://cldr.unicode.org/). This was the best reference to say that it's a valid locale string, which I could find.

So, languages _often_ requires the geographical location to represent the whole UI correctly (including formats).

However, we have so far went with the simpler language formats for course material. Though, I think we should start to plan that the course language code could include possibly include the location information too, at least to some extend. The full IETF standard would be very complicated though. So, for now we use only the two letter code to indicate the language for the material. As the locale to language map is always surjection (i.e. for every language, there can exist one or more locales). Thus, we can always map any locale to a language, but we can't map any language to one locale.

That said, Django should need to provide those formats files, as that information should be available via some library, but apparently that isn't the case yet. Thus, for the formats, I think we need to provide definitions for those locales, which we would like to use, but are not covered by the Django yet.

Next problem. The `hl` parameter (and userprofile.language and enrollment.language) requires the fully qualified locale name always, so the `translation.activate()` can operate correctly, as it needs to find the correct formats.py file too. If we think A+ as an application, which is used to read course material, then we can think that the application is always configured to show the UI with specific locale and not just with specific language. However, the application would select the best matching material from the database for that locale.

I guess, the biggest challenges are the cases when  the best information comes from `Accept-Language` and it doesn't include the location (e.g. just `en`). Maybe that is when we need to guess the best possible language, and provide way to change it.

Another annoyance would be that when viewing the material in English (US), we could provide button to change the language to English (UK), but as that would so the same content, but only change how the datetimes are represented, it could be very confusing to the user. I'm not sure what would be the best. Luckily, it's likely that there exist only one English definition in the settings, which would mean that typically locales and languages would have one to one mapping.

Maybe this contains some valuable information, maybe not. Have to say that things didn't become super clear for me either, even though I spend the evening reading about this stuff.

To summarise, the application (A+) should use the most accurate locale information (e.g. `en-us`, `en-fi`, `fi-fi`), when possible, and would select the best matching course material. If no version of the material is available for the locale, then A+ would show 404 with the list of possible versions (even if there would be multiple English variants).

lib/middleware.py
```diff
@@ -1,19 +1,27 @@
...
+        query_language = request.GET.get('hl')
+        if query_language:
+            language = query_language
+            lang_codes = [lang[0][:2] for lang in settings.LANGUAGES]
+            if language and (language[:2] not in lang_codes):
+                raise Http404('Page is not available in this language.')   
+        elif hasattr(request, 'user') and request.user.is_authenticated:
             language = request.user.userprofile.language
-        else:
-            language = translation.get_language_from_request(request)
```
raphendyr, line 16: This `else` branch could be restored as it would make more sense with the current code. Would remove the execution of `get_language_from_request` in case it comes from somewhere else.

course/viewbase.py
```diff
@@ -74,20 +76,25 @@ def get_resource_objects(self):
...
+                instance_languages = set(instance_languages)
 
                 languages = []
+                query_language = self.request.GET.get('hl')
+                if query_language:
+                    if query_language[:2] in instance_languages:
+                        languages.append(query_language)
+                    else:
+                        raise TranslationNotFound
                 if self.user_course_data and self.user_course_data.language:
```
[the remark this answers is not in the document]
veeraalt: Yes, the part where we look for enrollment and profile languages could be skipped, the last part should be still included because it checks that the course is available in the (query) language.

assets/js/aplus.js
```diff
@@ -629,5 +629,26 @@ $(function() {
...
+            event.target.nodeName === 'A'
+            && event.target.href
+            && event.target.protocol === window.location.protocol
+            && event.target.host === window.location.host // hostname:port
+            && (event.target.search === '' || event.target.search === '?') // no search
+        ) {
+            event.preventDefault(); // stop default redirect
+            event.target.search = window.location.search; // copy query params
+            window.location.href = event.target.href; // redirect
+            event.target.hreflang = window.location.hreflang
```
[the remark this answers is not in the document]
veeraalt: So the last line might be unnecessary?

course/viewbase.py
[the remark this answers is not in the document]
raphendyr: Isn't that done already above? `if query_language[:2] in instance_languages:`

assets/js/aplus.js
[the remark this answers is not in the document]
raphendyr: To me understanding, it has no effect (except removes hreflang if that exists).

course/viewbase.py
veeraalt: Thank you for the clarification, I still have a few questions about the logic/implementation:
1. Do we want to change the `LANGUAGES` in settings.py to contain only `('en-fi', 'English (FI)'), ('fi-fi', 'Finnish (FI)')` (and maybe also 'en-gb' and 'en-us') or just add those two and keep `('en', 'English'), ('fi', 'Finnish')` there too? If we replace the shorter formats completely in this PR, the `formats.py` files probably need to be replaced here also.
2. A course is only available in English, and user tries to access it in Finnish, what languages are there in the 404 error page, i.e. is there only `English` as the course language does not contain information about location or is there `English (FI)`, `English (UK)` etc?
3. When using only `?hl=en`, do we want to show the error page or direct the user to `?hl=en-fi`? I think it would be better to not show the error but the logic might get complicated if we need to guess the longer language code from the shorter one and change it every time. 
4. How about with Finnish? When we probably have only `fi-fi`, it would make sense to accept both `fi` and `fi-fi` as language codes and not show the error page when query parameter is `?hl=fi`.

There's so many unclear things about this that we need to find some compromise solution in scope of this PR if we want to get this somehow done and merged.

course/viewbase.py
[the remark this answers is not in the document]
veeraalt: Oh that is true, don't know how I didn't notice that.

course/viewbase.py
veeraalt: I think that `('en', 'English'), ('fi', 'Finnish')` has to be there as long as course code can be only 2 characters. Then we might be left with this kind of error message:
404-english

raphendyr: > Do we want to change the LANGUAGES in settings.py to contain only ('en-fi', 'English (FI)'), ('fi-fi', 'Finnish (FI)') (and maybe also 'en-gb' and 'en-us') or just add those two and keep ('en', 'English'), ('fi', 'Finnish') there too? If we replace the shorter formats completely in this PR, the formats.py files probably need to be replaced here also.

To help with #604, I would say we want to have `en-fi` and `fi` (`fi-fi` doesn't add any more information).

> A course is only available in English, and user tries to access it in Finnish, what languages are there in the 404 error page, i.e. is there only English as the course language does not contain information about location or is there English (FI), English (UK) etc?

They should see list of all possible languages, which would match, i.e., `English (US), English (FI), Finnish` for a course with langs `['en', 'fi']` and LANGUAGES `['en-fi', 'en-us', 'fi']`, when they try to access the course with some other language. This isn't perfect, but I think the best we could do.

Better option would be having a different UI language and course language, but that is not the case for current design. However, I think in most of the cases, the system will be configured with a single English locale.

> When using only ?hl=en, do we want to show the error page or direct the user to ?hl=en-fi? I think it would be better to not show the error but the logic might get complicated if we need to guess the longer language code from the shorter one and change it every time.

Technically, we should show an error, but as you said that it would be annoying. Hence, let's make an educated guess by redirecting the user to the first matching locale in the LANGUAGES list (it's an ordered list).

> How about with Finnish? When we probably have only fi-fi, it would make sense to accept both fi and fi-fi as language codes and not show the error page when query parameter is ?hl=fi.

I think it should be sufficient to use just `fi` in the defined languages (it's fully qualified name, as there is no sub languages. Related languages have their own tags, e.g., `fit` for [Tornedalen Finnish](https://en.wikipedia.org/wiki/Me%C3%A4nkieli)).

Thus, we probably should also do lookup for a shorter `hl` entry too. I.e., if we have `?hl=tag1-tag2-tag3-Tag4`, we would first check if there is `tag1-tag2-tag3-Tag4` in LANGUAGES, if not then check `tag1-tag2-tag3`, `tag1-tag2` and finally `tag1`. And as always, select the first from the list of matches and redirect.

As a last remark, I would expect the settings to contain something like this:

```python
LANGUAGES = (
  ('en-FI', _('English')),
  ('fi', _('Finnish')),
  ('sv', _('Swedish')),
)
```

Or in the US:

```python
LANGUAGES = (
  ('en-US', _('English')),
)
```

Or in the Mexico (I guess):

```python
LANGUAGES = (
  ('es-MX', _('Spanish')),
  ('en-US', _('English')),
)
```

p.s. the region code should be styled with uppercase letters, but I'm not sure how Django handles that in this settings value. The `translation.activate` handles it correctly, if I remember right.

These details (i.e. `es-MX` vs. `es`) only help with formats in our case for now.

veeraalt: Your suggestion seems otherwise good, but I'm a bit worried whether things are going to break if we remove `en` from `settings.py` when it's still used as course instance language. I tested that today with the test database and it deleted the `en` from language field in course instances where it was used, I suppose because it wasn't supported in `settings.py`.
We could have `English (US)` or `English (UK)` with language code `en` as the default, and option to change it to `English (FI)` at least from profile page, which then allows to use the Finnish formats.
What do you think of this? 
```
LANGUAGES = [
    ('en-FI', 'English (FI)'),
    ('en', 'English (US)'),
    ('fi', 'Finnish'),
]
```

Mankro: I think we have to have a discussion in a meeting about these language code issues (on Tuesday?). I'm afraid @raphendyr is making this really complicated and I hope it wouldn't need to be so. I can see this is not "as easy as it looks" and that languages and localizations have many twists. However, I would like to consider whether we could still aim for a simple, minimal approach. That is, language choice would focus on the language, not the region, and short 2-character language codes would be used as much as possible. How about continuing with this after the daily?

veeraalt: @Mankro Sounds good for me.

raphendyr: Yes, rest of the software doesn't yet handle language codes longer than 2 characters. I would like to get this one part working, so rest of the components can be fixed while moving forward. In the end, there are only few places which need to be fixed.

> However, I would like to consider whether we could still aim for a simple, minimal approach. That is, language choice would focus on the language, not the region, and short 2-character language codes would be used as much as possible.

I wish it was so simple. We are just mixing locale information and course languages to the same variable, where the course language doesn't often include enough information for the rest of the system.

Even ISO 639-3 codes (3 letter language codes) would already work for many cases, but not all, e.g., `pt-BR` (though, I have no idea why it doesn't have ISO code).

raphendyr: We made a decision to limit the language strings to 2 character codes (i.e. ISO 639-2 codes), which should work for the currently known use cases and has less change to break stuff. The code should in most cases ensure the longer or invalid codes are cut or dropped, so that only valid codes are used.

lib/middleware.py
```diff
@@ -1,16 +1,25 @@
...
 
     def process_request(self, request):
-        if hasattr(request, 'user') and request.user.is_authenticated:
+        query_language = request.GET.get('hl')
+        if query_language:
+            language = query_language
+            lang_codes = [lang[0][:2] for lang in settings.LANGUAGES]
+            if language and (language[:2] not in lang_codes):
+                raise Http404('Page is not available in this language.')   
+        elif hasattr(request, 'user') and request.user.is_authenticated:
```
raphendyr, line 22: I think this requires the is_anonymous check too.

lib/middleware.py
```diff
@@ -1,16 +1,25 @@
...
 class LocaleMiddleware(MiddlewareMixin):
 
     response_redirect_class = HttpResponseRedirect
 
     def process_request(self, request):
-        if hasattr(request, 'user') and request.user.is_authenticated:
+        query_language = request.GET.get('hl')
+        if query_language:
+            language = query_language
+            lang_codes = [lang[0][:2] for lang in settings.LANGUAGES]
```
raphendyr, line 19: A set could be used here, so the `in` check below would be faster.

templates/404.html
```diff
@@ -16,12 +16,24 @@
                         <h1>404 Not found</h1>
                     </div>
                     {% include "_messages.html" %}
+                    {% if languages %}
+                    <div class="alert alert-danger">
+                        {% blocktrans %}
```
raphendyr, line 21: Can you add `trimmed` here, so the translation file wouldn't include `\n` chars.

raphendyr (approved):
It's good from my point of view. I think you could squash and rebase and stuff it now.

raphendyr (approved):

course/viewbase.py
```diff
@@ -117,6 +126,15 @@ def get_course_instance_object(self):
             course__url=self.kwargs[self.course_kw],
         )
 
+    def handle_exception(self, exc):
+        if isinstance(exc, TranslationNotFound):
```
Mankro, line 130: I tried to test this myself. When I used a non-existing language code in the URL of a course page (`?hl=sv`), the 404 page did not show the available languages. The exception in this function was apparently not of type `TranslationNotFound` and the `languages` template context variable was then not defined. The exception was actually `No CourseInstance matches the given query. <class 'django.http.response.Http404'>` (from a debug print). Maybe that comes from the `LocaleMiddleware` instead of `CourseInstanceBaseMixin`. Can you try this yourself and fix it if the bug is real?

Mankro (changes requested):
This looks good, but I found one bug.

course/viewbase.py
Mankro: I just noticed one important point: `sv` is not enabled in the A+ settings.py. If I use `fi` in an English course, then the 404 page shows the link to the English version of the course page. (`fi` and `en` are the default languages in settings.py.)

raphendyr: I think this was a undocumented design decisionI think this was a undocumented design decision.

The `?hl=` is first validated in the middleware, where it has to be something defined in the settings file, hence the `?hl=sv` raises error there. If the variable is valid, then the processing continues to view classes, which include this better solution.

Making the course available to the middleware wouldn't be feasible and ignoring the language error in the middleware might raise other problems.

In short, 404 without language list is application level error, while the other is course level error. Though, that application page _could_ list all configured languages instead of not mentioning anything.

Mankro: I see. Which one is better in the following? If `LocaleMiddleware` raises 404 due to the GET query `hl` parameter (unknown language for the site), then...

* the 404 page shows a link to the same page without any `hl` query parameter. That link would work if the `hl` parameter was the only reason for the 404 error. The user would probably get at least somewhere and understand that the language is not supported. Better to reach some content than to assume that the whole course URL is broken.
* the 404 page shows language links (using `hl` query parameter) to all language versions of the page. All languages would be taken from the site-wide settings.py variable. Not all courses support those languages, so the links would lead to a new 404 page that would finally show working links to the course page.

Mankro: The first alternative without the `hl` seems more reasonable after our daily discussion. Would the link text be something like "open this page in the default language"?

raphendyr: One alternative would be doing an automatic redirect, but add like `?msg=automated redirect due to invalid hl parameter`, which would be rendered as a warning flash. I don't remember if we had such a feature yet or not.

Mankro: Since time is running out, we can only either add a link to the 404 page or keep it as is without any link.

raphendyr: I personally don't see a huge problem, as it should be less likely to include values which are not valid for any course than have a wrong value for that course (or that specific page in a course).

Hence, I would go what there is now and make issue to fix this better later.

veeraalt: I just managed to do the changes regarding the first approach, so I could push those and let me know if it looks good enough or do we want to discard them and go by what we have now.

raphendyr: That works too. Btw. do those pages still return 404?

raphendyr: > Btw. do those pages still return 404?

They do, so that is good :)

Mankro (approved):
Great, the 404 got the small improvement that we discussed. I think this could be squashed and merged. @raphendyr any objections?

lib/middleware.py
```diff
@@ -1,16 +1,28 @@
+from urllib.parse import urlparse
+
 from django.conf import settings
-from django.http import HttpResponseRedirect
+from django.http import Http404, HttpResponseRedirect
 from django.utils import translation
 from django.utils.cache import patch_vary_headers
 from django.utils.deprecation import MiddlewareMixin
+from django.shortcuts import redirect, render_to_response
```
raphendyr, line 8: `redirect` is not used?

lib/middleware.py
```diff
@@ -1,16 +1,28 @@
+from urllib.parse import urlparse
+
 from django.conf import settings
-from django.http import HttpResponseRedirect
+from django.http import Http404, HttpResponseRedirect
```
raphendyr, line 4: `Http404` is not used?

lib/middleware.py
raphendyr, line 8: redirect is not used?

lib/middleware.py
```diff
@@ -1,16 +1,28 @@
+from urllib.parse import urlparse
```
raphendyr, line 1: `urlparse` is not used?

raphendyr (commented):
Found some small things. You can fix and rebase those. Thanks.

PR #642, opened 2020-08-07
Remove mention of media-folder from README.md

saarasat:
Mention of media-folder is removed from `README.md` as it is not being used for any permanent data storage.

# Testing

All existing unit- and selenium-tests pass

This will make README to be up-to-date

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,4 +49,3 @@ Code Organization
 * [templates/](templates) : General site templates
 * [assets/](assets) : Static web server assets, e.g., JavaScript, styles and images
 * [assets_src/](assets_src) : Asset packages (as npm packages), which should populate `assets/` with compiled files.
-* [media/](media) : User uploaded files


raphendyr:
Is this the problematic test? https://github.com/apluslms/a-plus/blob/master/exercise/tests.py#L741

saarasat:
> Is this the problematic test? https://github.com/apluslms/a-plus/blob/master/exercise/tests.py#L741

Yes

raphendyr:
I think the problem is with some write permissions. I think a better solution would be using virtual buffers for the files, instead of writing them to the disk. For that, there is https://docs.python.org/3/library/io.html

```python
file_a = io.BytesIO(b'\x00\x00') # the png data
file_b = io.StringIO("Tekijät ja Hyyppö")

                response = self.client.post(exercise.get_absolute_url(), {
                    "key": "value",
                    "file1": file_a,
                    "file2": file_b,
                })
```

raphendyr:
I replaced the filesystem stuff with virtual buffers in 090ca115df62075a2327160d0d135428ad3db565, which might fix the problem for you. Can you verify that the problem is fixed by it?

saarasat:
> I replaced the filesystem stuff with virtual buffers in [090ca11](https://github.com/apluslms/a-plus/commit/090ca115df62075a2327160d0d135428ad3db565), which might fix the problem for you. Can you verify that the problem is fixed by it?

Yes, the problem was fixed, great! However, there is now the empty, broken link in README.md,  should that be removed? or at least the link removed, even if the mention of the folder remains?

raphendyr:
It would be a good idea to remove `media/` from the README as the data is not stored there in containers or in production.

saarasat:
> It would be a good idea to remove `media/` from the README as the data is not stored there in containers or in production.

Did that and updated the commit message of this very very short PR.

Issue #601, opened 2020-06-29
Standardise button colours


mtthwhggns:
Separating this out of #598 

> ## Consistency
>
>As highlighted in #585 , we have some inconsistent use of button colours. 
>
>We have some cases where we use colours which have a functional or semantic meaning because we want different colours. >One example might be the "danger red" 'Login using Google' button. 
>
>Since we shouldn't rely on colour to make differences between things clear, (WCAG 1.4.1) One idea would be that we have a single style of button, and replace all buttons with this style. Perhaps we have a second style for "destructive" actions (delete).
>
>I'll give some thought to the right way to approach this and add a new comment to this issue soon. 

It seems to make sense to deal with the two issues in isolation. 

I had a discussion with @etanttila and @saarasat about this. I'll write up the results soon.

mtthwhggns:
## Proposed Guidelines

- In every view you can have a maximum of one "default" button. It should be the one we expect most users to click
- All other buttons should be a "secondary" style.
- Differentiate between buttons with words, and, if it's useful, icons. Never with colour. 
- The exception is a button which has "serious destructive effects". There shouldn't be many of these at all. I've not found any student-facing ones yet. These should stand out even more. 

For now, this doesn't consider the "multistate buttons" from [js-jquery-toggle](https://github.com/Aalto-LeTech/js-jquery-toggle) or the colour-coded "color labels" - they need evaluating separately. 

While I say "in every view", these might not always be whole web pages. It will likely make sense in each modal, and each embedded exercise to be considered a different view. 

## Implementation

To start with, we could use the existing Bootstrap classes we have now. In this case, I'd suggest mapping "Default button" to `btn-primary`, "Secondary" to `btn-default` and "Destructive" to `btn-danger`.

![plus cs aalto fi_accounts_accounts_ (1)](https://user-images.githubusercontent.com/6674046/86112187-53327100-bad0-11ea-8e1a-c1ce0b5d790d.png)

In the future, we should consider the colours, maybe considering the default button in a more distinctive colour (compared with, for example, the top navigation bar). We also don't need to stick with Bootstrap here. We could consider making them ourselves if we want to tweak the styles a little. 

Changing these would be a small task, as we can list all uses of `btn-*` classes in each project, set them all to the "Secondary" style, then pick the "default" action per view.

## Example

Here's a page with lots of buttons. I did a quick find & replace on the classes as described above. I also added a "seriously destructive" button to see how that could stand out. 

Before:

[...]

Issue #568, opened 2020-06-10
Fine-tuning the contrasts of colors of A+ for better accessibility


saarasat:
At the moment, some of the colors used in A+ UI do not have strong enough contrast-ratio, for them to pass the AA-level of [web content accessibility guidelines ](https://www.w3.org/WAI/WCAG21/quickref/#contrast-minimum). The contrast-ratio can be assessed at: https://contrast-ratio.com/

In order to fix this issue, some of the colors present in A+ should be fine-tuned (eg. made slightly darker or lighter) to meet the AA-level standards, for instance the MOOC-badge here:

![image](https://user-images.githubusercontent.com/43036333/84268484-646f0a00-ab30-11ea-9efe-f7b0f6422749.png).

saarasat:
Alright, I'll then fix those badges, which do not use the font color selection logic.

Issue #609, opened 2020-07-03
Increase font size, improve readability and make a better use of font sizes in general.


jaguarfi:
Currently A+ use a `font-size: 14px` and it makes it difficult to read the text in small devices. Of course, we can make use of the zoom but increasing the base font might be beneficial. 
https://css-tricks.com/accessible-font-sizing-explained/

Depends on #594

mtthwhggns:
Strongly agree that this would be good.

Since we rely so much on Bootstrap at the moment, I'd suggest this be checked after https://github.com/apluslms/a-plus/issues/594, as it will be much easier to change on top of that.

Mankro:
Sounds good! Let's do this during the summer when the time is right concerning related issues.

PR #645, opened 2020-08-10
Implement custom button styles

mtthwhggns:
# Description

Issue #601 concerns the fact that button styles are used inconsistently and without semantic meaning. It was agreed that to resolve this, we would develop new button guidelines. [These proposed guidelines](https://apluslms.github.io/a-plus-design-system/components/button.html) are now available in the design system. 

There was, however, some unease at the specific styles proposed, so we waited until we could easily change at least the button colours. This PR introduces a new button component, which is designed to fit the new guidelines perfectly, while maintaining some consistency with the Bootstrap buttons, which it will take a while to migrate away from.

Additionally, this PR deals with #568 . It adds better theme support, and uses this to define "functional colours", which are both fed back into Bootstrap using SASS, and used directly by the new components. It also darkens the existing colours enough to not have any immediate contrast failures where any Bootstrap buttons are used. 

Here are the new buttons (left) alongside their equivalent (according to the new guidelines) in the current style (right). 

localhost_8000_accounts_login_

And here's the danger button by default, then with `:hover`:

localhost_8000_accounts_login_ (1)

Over and above what Bootstrap offers, these buttons have a smooth transition to their `:hover` and `:active` states and bolder text on the default action. 

This PR does not replace the buttons everywhere. It replaces them on the login page only. 

# Testing

These changes concern only available styles, which are currently not tested directly. 

If you want to see them in action, I recommend checking the branch out and swapping the classes (sizes `--xs`, `--sm`, `--lg` and `--default` or `--danger`) on the buttons which were added to the login page. 

Documentation regarding how and when to use these buttons belongs in the design system, which can be updated once the styles are available. 


userprofile/templates/userprofile/login.html
```diff
@@ -58,7 +58,7 @@ <h3>
 							</p>
 						{% endif %}
 						<p>
-							<a class="btn btn-danger btn-block" role="button" href="{% url 'social:begin' 'google-oauth2' %}?next={{ next|iriencode }}">{% trans 'Log in using Google' %}</a>
+							<a class="ap-button--secondary ap-button-lg" role="button" href="{% url 'social:begin' 'google-oauth2' %}?next={{ next|iriencode }}">{% trans 'Log in using Google' %}</a>
```
raphendyr, line 61: Typo `ap-button-lg`

userprofile/templates/userprofile/login.html
```diff
@@ -34,7 +34,7 @@ <h3>
 							</p>
 						{% endif %}
 						<p>
-							<a class="btn btn-primary btn-block" role="button" href="{% url 'shibboleth-login' %}?next={{ next|iriencode }}">
+							<a class="ap-button--secondary ap-button-lg" role="button" href="{% url 'shibboleth-login' %}?next={{ next|iriencode }}">
```
raphendyr, line 37: Typo `ap-button-lg`

assets/sass/components/_button.scss
```diff
@@ -0,0 +1,89 @@
...
+	border-width: 1px;
+}
+
+.ap-button {
+	&--default {
+		@include button-basics;
+		@include button-color($aplus-success, saturate($aplus-success, 20%));
+		font-weight: bold;
+	}
+	&--secondary {
```
raphendyr, line 66: `--secondary` doesn't have `@include button-basics;`

raphendyr:
Hmm.. I think it might be easier to replace the bootstrap button styles with these, as replacing all the classes would be a lot of work. However, as we don't have styles for all bootstrap classes, we probably should merge some (e.g. use same styles for primary and success). However, the names for the new classes seem to be better.

Is there reason to include basic button style for both style classes and size classes? Should it be possible to just use class `ap-button--default` or `ap-button--xs` without the other?`

Also, I think we can add 3 characters more to the classes, so they would become `aplus-button--default`.

assets/sass/components/_button.scss
mtthwhggns: Good point. I tried removing them for different combinations to try and figure out what works best. 

It's unnecessary to include them for both the sizes and the colours, and just results in duplicate code, so I'll switch to only including them in the colour, so a size won't work independently of a colour.

mtthwhggns:
> Hmm.. I think it might be easier to replace the bootstrap button styles with these, as replacing all the classes would be a lot of work. However, as we don't have styles for all bootstrap classes, we probably should merge some (e.g. use same styles for primary and success). However, the names for the new classes seem to be better.
> 

Easier is definitely true. 

I opted for this way because that would do a pretty poor job of migrating them with the correct meanings. It would result in some pages having many identical "default" buttons (which the guidelines say are one per page). The same amount of work needs doing to check the meanings are preserved whether it's in the context of switching to use new classes, or checking whether the updated styles work.

Additionally, I think some courses have created custom interactive exercises, grader output, etc. assuming that Bootstrap is present. In the future, it would be great if they would be migrated to use the `.ap-*` styles and follow our guidelines, but forcing a colour change on them when they _may_ be using colour to convey meaning (even though they shouldn't be) seems reckless. 

Given that, do you think I should re-impement this that way?

> Is there reason to include basic button style for both style classes and size classes? Should it be possible to just use class `ap-button--default` or `ap-button--xs` without the other?`

No, I somewhat addressed this above. 

In my opinion, the size shouldn't be used independently, as we don't have a plain, meaningless button in the new guidelines. You must specify a semantic meaning when creating a button (the same reason we don't have an `ap-btn` class defined). 

If felt more comfortable/safer to require a size to be declared, as we may want to adjust the sizes of everything at some point (#609), but I'm also OK with no size specified rendering the medium-sized button if you think that's better. 

My preference would be to not use `basics` in the sizes at all, but they would be required for the colours, and both must be specified whenever a button is added. 

> Also, I think we can add 3 characters more to the classes, so they would become `aplus-button--default`.

I'd be happy with this. I chose these to be familiar to people who are used to using Bootstrap, but I'd much prefer a proper word too.

mtthwhggns:
@raphendyr edited his original comments while I was answering, so one more answer.

Namespacing: I just went with this as many popular CSS frameworks use `.button` or `.btn`. If we want to be able to bring in new options in the future without too much hassle. It's not as specific a name as `.points-indicator` or something like that. It might be useful to namespace all of our custom components, so that those writing graders/exercises/whatever can choose to use them or not, without having things which should have semantic meanings sprinkled into their interfaces unpredictably.

raphendyr:
> Given that, do you think I should re-impement this that way?

I think we should consider that. However, you raised good points (e.g. we need to check all views).

> Namespacing: I just went with this as many popular CSS frameworks use .button or .btn. If we want to be able to bring in new options in the future without too much hassle. It's not as specific a name as .points-indicator or something like that. It might be useful to namespace all of our custom components, so that those writing graders/exercises/whatever can choose to use them or not, without having things which should have semantic meanings sprinkled into their interfaces unpredictably.

Yeah. I personally prefer using namespaces. For assessment tools, I think they should always use versioned prefixed (like quiz1 stuff in Saara's work). The style stuff were not specified at all for the assessment services, so it's been a bit of wild west for now. Though, I think someone should think this a bit so it wouldn't become too confusing and would consider what happens when some versions change (`aplus1` -> `aplus2` or `quiz1` -> `quiz2` etc).

assets/sass/colors/_functional.scss
```diff
@@ -0,0 +1,9 @@
+// These are the colours which add some kind of semantic meaning.
+// Avoid changing these when customising the template if possible.
+// Change these with extreme caution, especially notinf the contrast
```
Mankro, line 3: typo "notinf"

Mankro:
To me, it sounds like it is necessary to add new classes (`aplus-button-*`) instead of just overwriting the old Bootstrap button classes. We want to do this right and thus, the buttons on all pages must be checked anyway so that they follow the new guideline (only one default action).

Are we changing the default button colour from blue to green? Is the top navigation bar also changing colour?

mtthwhggns:
> Are we changing the default button colour from blue to green? Is the top navigation bar also changing colour?

This PR does change the colour of the "default" action button that way. The idea was to make the button (`$aplus-success`) stand out from the majority of the other interface elements (`$aplus-brand-primary`, currently Aalto blue but designed to be customised per-institution). So each view only has one big green button. 

So top navigation stays in the theme's main colour, the default action should be distinct from that.

assets/sass/colors/_functional.scss
mtthwhggns:
```suggestion
// Change these with extreme caution, especially noting the contrast
```

mtthwhggns:
This is rebased and squashed now.

raphendyr:
Should we still rename `ap-button` to `aplus-button`?

mtthwhggns:
> Should we still rename `ap-button` to `aplus-button`?

Oh, yes. I forgot this was outstanding.

It's done and squashed now.

PR #646, opened 2020-08-11
Add points indicator component

mtthwhggns:
# Description

Creates a new component with the associated view. The component isn't used anywhere yet.

The component displays an SVG icon alongside text to show progress on a question, chapter or course. The icon is coloured using CSS, and the data provided to the partial view are used to decide which icon/words to show.

Here's how the indicators look (small size, default and inverse):

localhost_8000_accounts_login_ (6)

# Testing

Since this component isn't used anywhere yet, it should be tested by placing one somewhere with code like `{% include "common/points_indicator/component.html" with result='points' score=7 max_points=12 %}`. 

This should be documented properly in the Design System repository.


assets/sass/main.scss
```diff
@@ -22,4 +22,5 @@ $brand-primary: #005EB8;
 /*!
  * A+ Styles
  */
-@import 'protocols/quiz1';
+ @import 'protocols/quiz1';
```
raphendyr, line 25: Why extra space in the start of the line?

templates/common/points_indicator/component.html
```diff
@@ -0,0 +1,30 @@
...
+	<svg class="icon">
+		{% if result == "not-started" %}
+			{% include "common/points_indicator/icons/not-started.svg" %}
+		{% elif result == "waiting" %}
+			{% include "common/points_indicator/icons/waiting.svg" %}
+		{% elif result == "points" %}
+			{% if score == max_points %}
+				{% include "common/points_indicator/icons/points-full.svg" with score=score max_points=max_points %}
+			{% else %}
+				{% include "common/points_indicator/icons/points.svg" with score=score max_points=max_points %}
```
raphendyr, line 11: We could use term `points` in place of `score` as that is used in other locations of the code.

templates/common/points_indicator/component.html
```diff
@@ -0,0 +1,30 @@
+<div class="points-indicator--{{ result }}">
+	<svg class="icon">
```
raphendyr, line 2: Should this be `points-indicator__icon`?

mtthwhggns: I will swap it that way to avoid any risks, yes.

The idea here was that all styles are nested inside the container, so shouldn't leak out, but I guess things leaking "in" might be as much of a concern.

assets/sass/main.scss
mtthwhggns: Because this PR is. a draft and there are still things to tweak, code style included.

templates/common/points_indicator/component.html
mtthwhggns: Will do this, thanks

templates/common/points_indicator/component.html
raphendyr: > I guess things leaking "in" might be as much of a concern.

I have noticed that to be challenging. If everything would be perfect, it would be nice :slightly_smiling_face:

Mankro:
The design is discussed in the design-system: https://github.com/apluslms/a-plus-design-system/issues/3

This PR is unfinished since the ideas and needs behind it started to evolve a lot. The current state of this PR does not match all expectations, but it contains the implementation for the points indicator component that should already work in basic use cases.

PR #651, opened 2020-08-14
Add different language alert message

### LINKED ISSUE #518: In A+ it should be possible to set student's default language in a specific course

veeraalt:
~~This is based on the #641, so that should be merged first.~~

# Description

When user is directed with language query parameter `?hl=<language>` at the end of the url to course content in different language than set in their preferences (in `UserProfile` or `Enrollment` object), show an alert message (in their saved language) and a button to change the language back to their preferred one. 
* For students (aka users who are enrolled on the course), there is also a button to set the current language as their preferred language for this course.
alert-with-language-icon2

* For course staff (users who are not enrolled), there is a text indicating that they can change the preferred language from their profile.
alert-with-language-icon-teacher

If there is no language preference saved or no language query parameter at the end of the url or the preferred language is same than the query language, alert is not shown.

Fixes #518 

# Testing

* Existing unit tests pass
* Testing has been done manually by checking that these use cases above work like expected

* No, I don't know if it's relevant.


raphendyr:
So, it might be a good idea to render the worning in the preferred language, which then means that there will be area in the screen in a different language, than what the rest of the page is currently. Maybe we should render the language change symbol on the left of the panel, to indicate that this box is about language stuff (considering the user's preferred language is not understandable by them for some reason).

course/templates/course/course_base.html
```diff
@@ -75,6 +75,36 @@ <h4>{% trans "Calendar" %}</h4>
                 {% endif %}
             </div>
         {% endif %}
+        {% if query_language %}
+            {% language user_language %}
+            <div lang={{user_language}}>
+                <div class="alert alert-default">
+                    {% trans "You are browsing this course in different language than set in your preferences." %}
```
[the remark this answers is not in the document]
veeraalt: This word choice is complicated as "preferences" here might refer to two different places. After this PR https://github.com/apluslms/a-plus/pull/632, it's possible to set multiple different language preferences for one user, one for the profile (affects the profile page and A+ front page etc) and one for each course separately (set from the language change symbol in course menu). 

So the word choice could be different depending on where the language is changed. When changed from profile, "set in your profile" is good, but what is a good way to indicate to user that the language is the one they have set for this specific course? 
If we would use one message, it could have some other word than "preferences" but I wouldn't use "profile" as it doesn't always come from profile.

veeraalt:
@mtthwhggns Thanks for the comments! 

> One extra question: If you use the "set X as course language" button - is there a way to undo this change?

Yes, it can be undone by changing the language from the language change symbol located in course menu. "Set X as course language" actually uses the same function as that.

veeraalt:
> Maybe we should render the language change symbol on the left of the panel, to indicate that this box is about language stuff (considering the user's preferred language is not understandable by them for some reason).

Good idea to somehow indicate to user that the message is about language stuff, but I don't know about the symbol. Would it be confusing if the symbol in course menu opens the drop-down menu where user can change the language and then the same symbol is used in the alert message but it doesn't do anything there?

course/templates/course/course_base.html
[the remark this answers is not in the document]
mtthwhggns: I understand. I think this is the right way to do it for now, I'd confused the places.

It's a little awkward though, as I'm afraid that if we have almost any word in a sentence like that (but especially "preferences"), a user will be looking for a "preferences" (or whatever) link to check the setting.

raphendyr:
> @raphendyr Good idea to somehow indicate to user that the message is about language stuff, but I don't know about the symbol. Would it be confusing if the symbol in course menu opens the drop-down menu where user can change the language and then the same symbol is used in the alert message but it doesn't do anything there?

Good points. Do you @saarasat or @etanttila have any opinions on this topic?

course/templates/course/course_base.html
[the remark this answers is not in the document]
veeraalt: That is true that it might be confusing for the user, I'll have to think a bit for better word choices.

course/templates/course/course_base.html
```diff
@@ -79,26 +79,38 @@ <h4>{% trans "Calendar" %}</h4>
...
             <div lang={{user_language}}>
                 <div class="alert alert-default">
-                    {% trans "You are browsing this course in different language than set in your preferences." %}
+                    {% trans "You are browsing this course in a different language than is set in your preferences." %}
                     {% if user_course_data %}
                         <div class="row">
                             <div class="col-lg-4">
                                 <form action="{{ url_without_language }}">
-                                    <button class="btn btn-default btn-block" type="submit">{% blocktrans with lang=user_language|language_name_translated %}browse the course in {{ lang }}{% endblocktrans %}</button>
+                                    <button type="submit" class="aplus-button--secondary aplus-button--md btn-block">
```
raphendyr, line 87: Is the `btn-block` required? If so, then the feature must be copied over to `aplus-button-*` system.

course/templates/course/course_base.html
```diff
@@ -79,26 +79,38 @@ <h4>{% trans "Calendar" %}</h4>
...
+                                        {% blocktrans with lang=user_language|language_name_translated %}
```
raphendyr, line 88: Can you add `trimmed` here. It will make the translation strings nicer. (`{% blocktrans trimmed with lang=user_language|language_name_translated %}`)

course/templates/course/course_base.html
veeraalt: Without `btn-block` it doesn't look very good:
buttons-grey

mtthwhggns: Did you try without the columns? Just two buttons next to one another. 

I didn't implement the `*-block` because it wasn't being used much and generally left-aligned buttons in forms seem to follow the route the eye takes along the left-aligned field labels. 

But, of course we can implement one if we need it.

veeraalt: Without columns: 
buttons-grey2
Do you know an easy way to get the buttons next to each other? Now they are just one above the other.

mtthwhggns: Yes! I'll send you an example in the morning.

raphendyr: Would button groups be a suitable idea for these two buttons? (see https://github.com/apluslms/a-plus-design-system/issues/1#issuecomment-675594108 )

mtthwhggns: In this particular case, I think it makes sense to have separate buttons. They're distinct actions with different effects.

Additionally, the way this PR is implemented now would be a problem, as each button is wrapped inside a `<form>`.

raphendyr: @veeraalt how does it look now?

course/templates/course/course_base.html
```diff
@@ -81,33 +81,29 @@ <h4>{% trans "Calendar" %}</h4>
...
-                                    {% csrf_token %}
-                                    <input name="language" type="hidden" value="{{ query_language }}" />
-                                    <button type="submit" class="aplus-button--secondary aplus-button--md btn-block">
-                                        {% blocktrans with lang=query_language|language_name_translated %}
-                                        Set {{ lang }} as course language
-                                        {% endblocktrans %}
-                                    </button>
-                                </form>
-                            </div>
+                        <div class="button-container">
```
raphendyr, line 84: How does this look now? Additionally, `aplus-button-container` would be more defining name (or possibly `aplus-button__container` @mtthwhggns ?).

mtthwhggns: From my point of view, this is part of the alert, not of the button. To me it wouldn't make much sense to namespace it as if it's used inside the buttons. Of course, the alert isn't our component. 

Since it's only used in one place, and we might choose to have a custom component to do the job of the alert in the future (which would probably have an "actions" block, which might take buttons or whatever else), I think having a class separate from the component like this should be OK.

raphendyr: Check. Let's go with this and let's assume that none else will use `button-container` for anything else for now.

veeraalt: I updated the look of the buttons to the PR description.

> Check. Let's go with this and let's assume that none else will use button-container for anything else for now.

I think it's ok even if someone would use that somewhere else because now it's nested under the `alert alert-default`.

raphendyr: The problem isn't for thiese style definitions to apply somewhere else, but if someone adds definitions for this class without nesting, then those other definitions will apply here too. But we can assume that won't happen.

etanttila:
> > @raphendyr Good idea to somehow indicate to user that the message is about language stuff, but I don't know about the symbol. Would it be confusing if the symbol in course menu opens the drop-down menu where user can change the language and then the same symbol is used in the alert message but it doesn't do anything there?
> 
> Good points. Do you @saarasat or @etanttila have any opinions on this topic?



> > @raphendyr Good idea to somehow indicate to user that the message is about language stuff, but I don't know about the symbol. Would it be confusing if the symbol in course menu opens the drop-down menu where user can change the language and then the same symbol is used in the alert message but it doesn't do anything there?
> 
> Good points. Do you @saarasat or @etanttila have any opinions on this topic?

Personally I don't think it would be very confusing. I think it might even help the user realize that these things are associated with each other. It helps that in a menu people assume things to be links or buttons while in a pop-up or alert people assume things to be texts or images unless they look like links or buttons. However, I think it's a good question to ask for the sake of consistancy. I think it would be important that it wouldn't look like a button or be in a location where a button would be in the alert (such as the top right corner), but would be fine in the top left. But this is just my opinion.

raphendyr:
I was thinking something like this

image

I added into the alert div:

```html
<img src="/static/images/uls.svg" alt="" style="float: left; height: 70px; margin-right: 15px;">
```

and `min-height: 100px;` to the alert div itself.

etanttila:
@raphendyr that is one big icon! I would go with a smaller one. Perhaps one still larger than the one in the menu, but something smaller than your demo. Finding the best size probably requires some playing around. (And would it be better to use relative units than pixels?)

mtthwhggns:
> Would it be confusing if the symbol in course menu opens the drop-down menu where user can change the language and then the same symbol is used in the alert message but it doesn't do anything there?

The icon in the course menu behaves like a button - maybe it should look like a button?

This way it would behave predictably, and we can reuse the icon elsewhere to refer to it without people expecting the icon to always behave like a button?

Näyttökuva 2020-8-24 kello 11 54 37

etanttila:
> > Would it be confusing if the symbol in course menu opens the drop-down menu where user can change the language and then the same symbol is used in the alert message but it doesn't do anything there?
> 
> The icon in the course menu behaves like a button - maybe it should look like a button?
> 
> This way it would behave predictably, and we can reuse the icon elsewhere to refer to it without people expecting the icon to always behave like a button?
> 
> Näyttökuva 2020-8-24 kello 11 54 37

I agree that it should look like a button. I thought it got darker surroundings on hover to make it seem more like a button (just one without borders), but it doesn't even have that!! Oh wow, it's an input and not a button. Maybe it should also be a button since it behaves like one?

raphendyr:
> And would it be better to use relative units than pixels?

Yes, I used pixel so I was able to calculate a correct min-height, as the padding for the alert div has pixels.

> The icon in the course menu behaves like a button - maybe it should look like a button?

Yes, but that isn't a concern for this PR. But yes, it should behave like a button or a dropdown.

mtthwhggns:
> Yes, but that isn't a concern for this PR. But yes, it should behave like a button or a dropdown.

I already did this in #661

etanttila:
> > The icon in the course menu behaves like a button - maybe it should look like a button?
> 
> Yes, but that isn't a concern for this PR. But yes, it should behave like a button or a dropdown.

But it is something that should be considered here. Another thing being broken that should be fixed in the near future shouldn't be the main reason why something else that could be smart shouldn't be done. Having the language icon in the menu look like a button would reduce the risk of confusion, which is something that is good to take into consideration when thinking if using the same icon here in the alert is confusing.

veeraalt:
I added the language icon, looks like this now:

alert-with-language-icon

course/templates/course/course_base.html
```diff
@@ -79,6 +79,7 @@ <h4>{% trans "Calendar" %}</h4>
             {% language user_language %}
             <div lang={{user_language}}>
                 <div class="alert alert-default">
+                    <img src="/static/images/uls.svg" alt="" style="float: left; height: 2em; margin-right: 10px;">
```
raphendyr, line 82: The style should be implemented with a class and relevant settings in the css file.

userprofile/views.py
```diff
@@ -60,7 +60,7 @@ def get_context_data(self, **kwargs):
 def set_user_language(request):
     """"Overrides set_language function from  django.views.i18n."""
     LANGUAGE_PARAMETER = 'language'
-
+
```
raphendyr, line 63: :thinking:

course/viewbase.py
```diff
@@ -81,20 +85,23 @@ def get_resource_objects(self):
...
+                                self.query_language = query_language
                     else:
                         raise TranslationNotFound
                 else:
-                    languages = []
-                    if self.user_course_data and self.user_course_data.language:
-                        languages.append(self.user_course_data.language)
-                    if is_real_user and user.userprofile.language:
-                        languages.append(user.userprofile.language)
-                    languages.append(get_language())
```
raphendyr, line 96: This line is removed, so there is no cases when it would have meaning?

veeraalt: Not sure, removing was probably accident

course/templates/course/course_base.html
```diff
@@ -75,6 +75,45 @@ <h4>{% trans "Calendar" %}</h4>
...
+            {% language user_language %}
+            <div lang={{user_language}}>
+                <div class="alert alert-default">
+                    <img class="language-icon" src="/static/images/uls.svg" alt="">
+                    {% trans "You are browsing this course in a different language than is set in your preferences." %}
+                    {% if user_course_data %}
+                        <div class="button-container">
+                            <form action="{{ url_without_language }}">
+                                <button type="submit" class="aplus-button--secondary aplus-button--md">
+                                    {% blocktrans trimmed with lang=user_language|language_name_translated %}
```
Mankro, line 88: Should all of these forms include the `{% csrf_token %}`? Why is it included in only one?

Mankro: Oh right, some forms use HTTP GET while some POST.

veeraalt: That is the only one of those with `post` method, which needs the token

locale/fi/LC_MESSAGES/django.po
```diff
@@ -2520,16 +2541,17 @@ msgid "Download"
 msgstr "Lataa"
 
 #: exercise/templates/exercise/staff/_submissions_table.html:37
-#, python-format
 msgid ""
 "\n"
 "    %(limit)s latest submissions.\n"
-"    <a class=\"aplus-button--secondary aplus-button--sm\" href=\"%(url)s\">Show all</a>\n"
+"    <a class=\"aplus-button--secondary aplus-button--xs\" role=\"button\" "
```
Mankro, line 2547: Now I see that the Finnish string does not have the `role="button"` like the English one has. You probably didn't change this string yourself, but `manage.py makemessages` decided that the line was too long and it divided it to multiple lines.

raphendyr: It seems like the file was not updated after it was added here.

raphendyr: Can you rebase on top of current master and rerun makemessages?

veeraalt: I can, but that is the last thing I can do for now 😁 So hope it is otherwise good

veeraalt: So they both should have the "role=button" or neither should have?

veeraalt: Oops actually I can't do that because I just wiped my work computer

Mankro: Jaakko fixed that string in a separate commit in master, so we will need to rebase this PR onto master and remove the conflicting changes in django.po. Sounds like it will be my task next week!

Mankro (approved):
Good. I fixed the git conflict in django.po myself by rebasing and force-pushed. I also removed some trailing whitespace from the _main.scss file and added the missing `#, python-format` modifiers in django.po.

PR #652, opened 2020-08-17
Accessibility statement

mtthwhggns:
# Description

It's a legal requirement that we show a statement of our compliance with the EU accessibility directive.

This adds the relevant templates.

No translations are done yet. 

# Testing

### DIFF
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-07-21 12:51+0300\n"
+"POT-Creation-Date: 2020-08-19 09:22+0300\n"
 "PO-Revision-Date: 2019-08-14 12:16+0200\n"
 "Last-Translator: Saara Satokangas <saara.satokangas@aalto.fi>\n"
 "Language-Team: Finnish <>\n"
@@ -111,15 +111,15 @@ msgstr "Käyttäjät"
 msgid "Input an URL identifier for this course."
 msgstr "Syötä URL-osoitteessa käytettävä tunnus kurssille."
 
-#: course/models.py:68 course/models.py:732 exercise/exercise_models.py:127
+#: course/models.py:68 course/models.py:733 exercise/exercise_models.py:127
 msgid "Taken words include: {}"
 msgstr "Varatut sanat: {}"
 
-#: course/models.py:239
+#: course/models.py:240
 msgid "internal"
 msgstr "sisäinen"
 
-#: course/models.py:241
+#: course/models.py:242
 msgid ""
 "The user profile contains a student number and has logged in via local "
 "organisation authentication"
@@ -127,11 +127,11 @@ msgstr ""
 "Käyttäjän profiililla on opiskelijanumero ja hän on kirjautunut sisään "
 "organisaation kirjautumisen kautta"
 
-#: course/models.py:247
+#: course/models.py:248
 msgid "external"
 msgstr "ulkoinen"
 
-#: course/models.py:249
+#: course/models.py:250
 msgid ""
 "The user profile doesn't have a student number, thus the user has logged in "
 "from a different organization or via social authentication"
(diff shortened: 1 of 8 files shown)


userprofile/views.py
```diff
@@ -128,5 +128,44 @@ def get_common_objects(self):
         self.privacy_text = privacy_text
         self.note("privacy_text")
 
+class AccessibilityStatementView(UserProfileView):
```
raphendyr, line 131: Can you add extra newlines before and after the class (i.e. 2 lines between different classes.

userprofile/views.py
```diff
@@ -128,5 +128,44 @@ def get_common_objects(self):
...
+    access_mode=ACCESS.ANONYMOUS
```
raphendyr, line 132: Even though this is a copy-paste, can you add spaces around `=`.

userprofile/views.py
```diff
@@ -128,5 +128,44 @@ def get_common_objects(self):
...
+
+    def get_common_objects(self):
+        super().get_common_objects()
+        lang = "_" + get_language().lower()
+        key = make_template_fragment_key('accessibility_statement', [lang])
+        accessibility_statement = cache.get(key)
+        if not accessibility_statement:
+            local_template_name = "institution_accessibility_text{}.html"
+            local_template = try_get_template(local_template_name.format(lang))
+            if not local_template and len(lang) > 3:
```
raphendyr, line 143: This logic should be refactored to a function for all of the 3 uses (1 in privacy notice and 2 times here), but I think it can be left for later. Can you add a comment `# TODO: refactor to a helper function`?

mtthwhggns:
Update on the status of this: It currently contains a statement where I've swapped out the name of the University for "we" and "us" - so that others can use this as a template if their institution isn't providing one.

The Aalto-specific template has moved elsewhere.

From my point of view, I think I've done everything suggested on this PR.

raphendyr (approved):

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3697,6 +3697,14 @@ msgstr ""
...
 msgstr "Unohda automaattinen uudelleenohjaus."
 
-#: userprofile/views.py:78
+#: userprofile/views.py:126
 msgid "No privacy notice. Please notify administration!"
 msgstr "Ei tietosuojailmoitusta. Ota yhteys ylläpitäjään!"
+
+#: userprofile/views.py:153
+msgid "No local accessibility statement. Please notify the site's owner!"
+msgstr ""
```
Mankro, line 3706: "No local accessibility statement. Please notify the site's owner!"
"Ei paikallista saavutettavuusselostetta. Ilmoitathan tästä sivuston ylläpitäjälle!"

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3697,6 +3697,14 @@ msgstr ""
...
+
+#: userprofile/views.py:165
+msgid "No system-wide accessibility statement found."
```
Mankro, line 3709: "No system-wide accessibility statement found."
"Järjestelmätason saavutettavuusselostetta ei löytynyt."

Mankro (changes requested):
It's good. Would you add the two small translations in django.po that I commented on?

mtthwhggns:
> It's good. Would you add the two small translations in django.po that I commented on?
> Edit: you could rebase and squash then.

Oh, yes! I forgot about these.

I've added them now, squashed and rebased.

userprofile/views.py
```diff
@@ -128,5 +128,47 @@ def get_common_objects(self):
...
+                system_template = try_get_template(system_template_name.format(lang[:3]))
+            if not system_template:
+                logger.warning("No localized system accessibility content for language %s", lang)
+                system_template = try_get_template(system_template_name.format(''))
+            if not system_template:
+                logger.error("No system accessibility content at all!")
+
+            system_accessibility_statement = system_template.render() if system_template else _("No system-wide accessibility statement found.")
+
+            accessibility_statement = f'{local_accessibility_statement}{system_accessibility_statement}'
```
Mankro, line 167: Is this the new Python string formatting syntax that is only available from Python 3.6 onwards? Our production server is still on Python 3.5 and we must therefore remain compatible with Python 3.5.

mtthwhggns: Ok. 

I assumed the docket container was using the same version as production. 

I switched to this because concatenation of the two strings with `+` fails if the file is missing. 

@Mankro What's the right way to do this in our version of Python?

mtthwhggns: I'm no Python expert, but this seems to work: 

```
str(local_accessibility_statement) + str(system_accessibility_statement
```

Mankro: `_` refers to the function `ugettext_lazy`, but we don't need lazy translation in this case. The template `render` method does not return lazy strings, either. I think you can call `gettext` instead of `_` here and then the string concatenate `+` operator will work directly. You need to import `gettext` to use it. Note that Django has the `format_lazy` function for doing string formatting on lazy strings.

mtthwhggns: I've done this now.

Mankro:
@mtthwhggns did you notice yet that the git commit message got broken?

mtthwhggns:
> @mtthwhggns did you notice yet that the git commit message got broken?

I didn't, thanks. Fixed now.

PR #654, opened 2020-08-17
Switch to specific focus state styles

### LINKED ISSUE #653: Frontpage cards don't match the page focus state
The new cards used on the front page and course archive don't appear yellow anywhere when in focus.

This breaks the consistent focus state which was previously implemented.

The issue seems to be that the wrapping `<a>` is displayed `inline`. If it's overridden as `display: block` it gets the background, but doesn't look great because of the background/border conflict. This component might benefit from something bespoke.

mtthwhggns:
# Description

When we started the focus states, I noted that with SASS it would be easier to target a focus state at certain elements with a `@mixin`. This implements two mixins:

- `focus-with-outline-only` the default, applied to all interactive elements automatically, which doesn't change any of the elements styles, but adds a new `outline`. It is pretty safe, and shouldn't be able to break anything.
- `focus-with-background-and-color` is much easier to see on small elements and bright backgrounds, as it changes the whole block colours. This is good for use in coloured menus and links inside text, which might be harder to see with only an outline.

The default is now `focus-with-outline-only` with `focus-with-background-and-color` applied to navbar links, the footer and `a`s which appear inside text. 

As a bonus, this removes the dependence on `!important`. 

Fixes #653

# Testing

This needs testing manually, as the only changes are style-related.

This should be documented in the design system, as it relates to the use of A+ components. 


assets/sass/legacy/_main.scss
```diff
@@ -58,21 +60,13 @@ table.src tr td.src {
...
 
-.js-focus-visible input.focus-visible, 
-.js-focus-visible select.focus-visible, 
-.js-focus-visible div.focus-visible {
-    outline: 3px solid #ffdd00 !important;
+.js-focus-visible a,
+.js-focus-visible input,
+.js-focus-visible select,
+.js-focus-visible div {
+	@include focus-with-outline-only;
```
raphendyr, line 69: Can this fallback be moved out of the legacy to new scss files? For example in it's own file to make clear that it's a fallback.

assets/sass/mixins/_focus.scss
```diff
@@ -0,0 +1,14 @@
+@mixin focus-with-outline-only {
+	&.focus-visible{
+		outline: 3px solid $aplus-focus;
+	}
+}
+
+@mixin focus-with-background-and-color {
+	&.focus-visible{
+		background-color: $aplus-focus;
+    color: $aplus-foreground;
```
raphendyr, line 10: mix of different indentation chars?

assets/sass/components/_button.scss
```diff
@@ -1,9 +1,13 @@
+@import '../mixins/focus';
+
 $extra-small-height: 16px;
 $small-height: 18px;
 $medium-height: 20px;
 $large-height: 24px;
 
 @mixin button-basics {
+	@include focus-with-background-and-color;
```
saarasat, line 9: I was going through this an looks good in general! I liked the way the two different versions work together. And Sass probably indeed works nicely here :) 

Just a question about the buttons. Now they seem to get a bit mixed styling. For instance on the front page of a course, the filter buttons  look like this:

image

but then again the enroll-button looks like this:
image

What was the meant outlook for buttons?

assets/sass/components/_frontpage-cards.scss
```diff
@@ -17,6 +19,10 @@
 				font-size: 18px;
 				margin-top: 20px;
 				margin-bottom: 10px;
+
+				a {
+					@include focus-with-background-and-color;
+				}
```
saarasat, line 25: And this one I am not a huge fan of, but I guess now that we made only the upper-part of the card clickable, and the footer is not clickable, there is not much choice here:

Screenshot from 2020-08-17 16-27-22

assets/sass/components/_frontpage-cards.scss
```diff
@@ -17,6 +19,10 @@
 				font-size: 18px;
 				margin-top: 20px;
 				margin-bottom: 10px;
+
+				a {
+					@include focus-with-background-and-color;
```
saarasat, line 24: And if these are merged, some guidelines for the usage would be welcome :) so, when to use which one.

mtthwhggns: Yes, agreed.

I didn't do anything here, as I thought that's a job for the design system.

assets/sass/components/_button.scss
mtthwhggns: Hmm, yeah, I didn't touch the Bootstrap buttons because removing them from all student-facing things is on my to-do list for this week and it's not so easy to hook into the Bootstrap styles here.

Maybe this can be merged after that?

assets/sass/mixins/_focus.scss
mtthwhggns: I have no idea why this keeps happening, but thanks.

assets/sass/legacy/_main.scss
mtthwhggns: Good point, it should definitely come out of the legacy file. 

I'd be inclined to leave the extra ones I've hooked into where they are (like `.navbar`, where the class didn't get any re-writing other than adding the mixin).

assets/sass/components/_frontpage-cards.scss
mtthwhggns: I felt quite the same, but it's important that it outlines the clickable area, so I didn't see much alternative - other ideas are welcome!

assets/sass/components/_button.scss
Mankro: Finalizing this pull request after the buttons sounds like a good plan.

assets/sass/legacy/_main.scss
Mankro: Is the fallback needed anymore?

mtthwhggns: Yes, but maybe fallback isn't the best description.

Maybe something like "generic focus style" would be a better description.

I was trying to say that it applied to components which don't have a specific focus style defined for the component - at present, this one is used for all form inputs, as we've not got our own components for those.

raphendyr: > I'd be inclined to leave the extra ones I've hooked into where they are (like .navbar, where the class didn't get any re-writing other than adding the mixin).

I think that is fine. Not best, but fine.

assets/sass/mixins/_focus.scss
raphendyr: Yeah. The `.editorconfig` seems to be used by your editor as it removes whitespace in end of lines.

assets/sass/components/_frontpage-cards.scss
raphendyr: I think it works well, when there are clickable buttons in the bottom, which will be focused before moving to the next card.

assets/sass/components/_frontpage-cards.scss
raphendyr: For that, I think you can write documentation in the SCSS file. Or in the style guides repo (assuming these are going there some day).

assets/sass/mixins/_focus.scss
Mankro: The weird indentation is still here.

mtthwhggns: Fixed it manually. No idea why the editor is currently so hit-and-miss.

Mankro (approved):
This looks good! I tested it myself and didn't see errors. You can rebase and squash now. Do you have time to update focus state styles in the aplus-design-system?

PR #655, opened 2020-08-17
Show submission points in course points API endpoint

### LINKED ISSUE #647: Course points API endpoint should contain points for all submissions
This comes from the IntelliJ project: https://github.com/Aalto-LeTech/intellij-plugin/issues/302

The `/api/v2/courses/<course-id>/points/me/` endpoint should be able to provide points for all submissions (for one student in all exercises of one course). We need to still consider if all points are always included in the output or only when some parameter is given in the request GET query parameters. All points should already be available in the points cache: https://github.com/apluslms/a-plus/blob/9e595a0a902d19bcadeeaff8b3160873b0265f43/exercise/cache/points.py#L98

Let's not modify the existing submissions URL list in order to preserve backwards-compatibility. A new key shall be added to the output.

Example snippet for the output (the existing submissions list only contains the URLs of the submissions):

```
"submissions_and_points": [
  {
    "url": "https://plus.cs.aalto.fi/api/v2/submissions/123/",
    "points": 10
  },
  {
    "url": "https://plus.cs.aalto.fi/api/v2/submissions/456/",
    "points": 5
  }
]
```

Jaakko says that it could be best to add the grade field to the existing `SubmissionBriefSerializer` class. It could be more uniform with the rest of the API.

HenrikSundell:
# Description

Add key `submissions_and_points` list to `/api/v2/courses/<course-id>/points/me/`
```
 "submissions_and_points": [
                        {
                            "id": 567,
                            "url": "http://0.0.0.0:8000/api/v2/submissions/567/",
                            "submission_time": "2020-07-24T08:55:11.895107Z",
                            "grade": 10
                        },
                        {
                            "id": 566,
                            "url": "http://0.0.0.0:8000/api/v2/submissions/566/",
                            "submission_time": "2020-07-24T08:54:24.499804Z",
                            "grade": 10
                        }
]
```

Fixes #647

# Testing

### DIFF
--- a/exercise/api/custom_serializers.py
+++ b/exercise/api/custom_serializers.py
@@ -49,10 +49,21 @@ class ExercisePointsSerializer(serializers.Serializer):
                 'submission_id': submission_id
             }, request=request)
 
+        def submission_obj(submission_cached):
+            id_ = submission_cached['id']
+            return {
+                'id': id_,
+                'url': submission_url(id_),
+                'submission_time': submission_cached['date'],
+                'grade': submission_cached['points'],
+            }
+
+        submissions = [submission_obj(s) for s in entry['submissions']]
         exercise_data = {
             'url': exercise_url(entry['id']),
             'best_submission': submission_url(entry['best_submission']),
-            'submissions': [submission_url(s['id']) for s in entry['submissions']],
+            'submissions': [s['url'] for s in submissions],
+            'submissions_with_points': submissions,
         }
         for key in [
             # exercise
--- a/exercise/api/serializers.py
+++ b/exercise/api/serializers.py
@@ -40,6 +40,7 @@ class SubmissionBriefSerializer(AplusModelSerializer):
         model = Submission
         fields = (
             'submission_time',
+            'grade',
         )
         extra_kwargs = {
             'url': {


exercise/api/custom_serializers.py
```diff
@@ -49,10 +49,20 @@ def submission_url(submission_id):
                 'submission_id': submission_id
             }, request=request)
 
+        def submission_obj(submmission_cached):
```
Mankro, line 52: typo `submmission_cached`

exercise/api/custom_serializers.py
```diff
@@ -49,10 +49,20 @@ def submission_url(submission_id):
...
+            id_ = submmission_cached['id']
+            return {
+                'id': id_,
+                'url': submission_url(id_),
+                'submission_time': submmission_cached['date'],
+                'grade': submmission_cached['points'], # NOTE: model has this in field `grade`, TODO rename
```
Mankro, line 58: Many exercise related API endpoints use the word `points` for the current best points of the student, but the submission detail endpoint uses `grade`. Then it makes sense to use `grade` here as well since this object refers directly to the submission detail URL. The comment in the code could be removed, or was there something else to consider here?

exercise/api/custom_serializers.py
```diff
@@ -49,10 +49,20 @@ def submission_url(submission_id):
...
+            }
+
         exercise_data = {
             'url': exercise_url(entry['id']),
             'best_submission': submission_url(entry['best_submission']),
             'submissions': [submission_url(s['id']) for s in entry['submissions']],
+            'submissions_and_points': [submission_obj(s) for s in entry['submissions']],
```
Mankro, line 65: We need to consider the word used in the output. `submissions_and_points` is not the best one necessarily. Since the word `grade` is used in the object, should this be `submissions_with_grades`?

exercise/api/serializers.py
```diff
@@ -40,6 +40,7 @@ class Meta(AplusModelSerializer.Meta):
         model = Submission
         fields = (
             'submission_time',
+            'grade',
```
Mankro, line 43: What does this change do and was it required for changing the course points endpoint?

HenrikSundell: Probably not necessary anymore, I will remove it

exercise/api/custom_serializers.py
HenrikSundell: @raphendyr

exercise/api/custom_serializers.py
HenrikSundell: Yes `submissions_with_grades` could be good if we use the key grade. makes sense to use the same term here.

exercise/api/custom_serializers.py
raphendyr: Pointed out that why it's so and to mention that someone should probably do something about it some day. It's totally possible to add also variable `points` (in all locations) and then remove `grade` when moving to `v3` API.

exercise/api/custom_serializers.py
raphendyr: Depends if we want to prefer word `points` or `grade` for this purpose. Details only has `grade`, because that is the name in the model. I should have used `points`, but I didn't understand enough at the time.

exercise/api/serializers.py
raphendyr: I think this would be fine too. Then the "brief" would include the same information in all locations.

exercise/api/custom_serializers.py
raphendyr: Also, these might not be the best names for the function and the variable. Feel free to rethink them once or twice.

exercise/api/custom_serializers.py
Mankro: I think the comment is misleading. If `grade` is changed here to `points`, the change should be done in a few other places too, not just here. Even if `points` is the future choice, it would be confusing if `points` is used here right now and `grade` in the submission detail endpoint.

exercise/api/custom_serializers.py
Mankro: Maybe `submissions_with_points` so that it is ready for the future if `grade` is changed to `points` some day.

exercise/api/custom_serializers.py
raphendyr: The comment should be written better, for sure. And yes, the key should be now grade, but it would be fine all these locations with a comment to do something about it. Or one could just write an issue about it and leave out the comment (maybe).

exercise/api/custom_serializers.py
raphendyr: :+1: `with` is better than `and`.

exercise/api/serializers.py
Mankro: Including useful info is OK. I wanted to check if this was really needed for the actual issue, but let's keep it now if there are no unexpected consequences from this.

Mankro:
@HenrikSundell we must merge this on Wednesday 19.8. Please finish the small details on Wednesday after the webinar.

exercise/api/custom_serializers.py
HenrikSundell: I removed the comment, for now, I didn't come up with a good comment for this, might be easier to write an issue. I can add a comment here if someone has a good suggestion

exercise/api/custom_serializers.py
```diff
@@ -49,10 +49,20 @@ def submission_url(submission_id):
...
+                'url': submission_url(id_),
+                'submission_time': submission_cached['date'],
+                'grade': submission_cached['points'],
+            }
+
         exercise_data = {
             'url': exercise_url(entry['id']),
             'best_submission': submission_url(entry['best_submission']),
             'submissions': [submission_url(s['id']) for s in entry['submissions']],
+            'submissions_with_points': [submission_obj(s) for s in entry['submissions']],
```
raphendyr, line 65: Can you make this so, that we use the same urls for the other list too and don't do reverse map multiple times?

i.e.

```python
submissions = [submission_obj(s) for s in entry['submissions']]

       exercise_data = {
            'submissions': [s['url']) for s in submissions],
            'submissions_with_points': submissions,
```

Mankro (approved):
Good, final changes have been done. Please do git rebase and squash and let's merge this PR.

PR #656, opened 2020-08-18
Implement aplus-material-v1, aplus-assess-v1, and aplus-hook-v1 inferfaces by adding User-Agent and correct X-Aplus-Event headers

raphendyr:
# Description

Every self respecting web client should provide a good User-Agent string to the server. A+ was sending only `python-requests/2.21.0`. Now, it sends `a-plus/1.7.0-63-gab401233 (+http://plus.example.com) python-requests/2.21.0`, where `1.7.0*` is the current A+ version  defined in `aplus.version.__version__` or resolved by `git describe` (first for real deployments and second for git clone installations) and `http://plus.example.com` is the content of `settings.BASE_URL`.

Additionally, send header `X-Aplus-Event` with a interface version and event type info. Following events are known:

* `aplus.assess.v1/assess-submission` - when sending submission for grading
* `aplus.assess.v1/retrieve-exercise` - when retrieving exercises
* `aplus.assess.v1/retrieve-model` - when retrieving the model file (or files)
* `aplus.assess.v1/retrieve-template` - when retrieving the template file (or files)
* `aplus.hook.v1/post-assessment` - send as part of the post hook
* `aplus.material.v1/retrieve-chapter` - when retrieving chapters

To support this. the `load_exercise_page` in `exercise/protocol/aplus.py` is split to two functions, one for exercises and one for chapters. Additionally, new content selectors are added, so content and exercise services may update their outputs. Even though chapters are internally exercises and should be kept as such from the points point of view, that doesn't mean the interface for them should be identical. I think this separation will make the interface less error prone and will also make the code more clear on what is what.

Finally, start using requests session in `lib/remote_page.py`, which allows reusing http connections. This requires a web server with a http/1.1 support. For example, MOOC-Grader has too old Django for that, but MOOC-Jutut is updated to 2.2 already, which supports this.

# Testing

Hook requires testing with a service, which can react to the hook some way. The following container can be used to print requests.

```yaml
services:
  hookservice:
    image: mendhak/http-https-echo
    depends_on:
      - grader
      - plus
```

MOOC-Grader can be used to test  requests send to it. The following middleware can print all the required info:

```python
class LogRequestsMiddleware(MiddlewareMixin):
    def process_request(self, request):
        logger.warning(" -- MIDDLEWARE DEBUG -- ")
        logger.info("Request %s by %s", request.method, request.META.get('HTTP_USER_AGENT', 'unknown'))
        logger.info("Aplus event: %s", request.META.get('HTTP_X_APLUS_EVENT', 'unknown'))
```

To verify connection pooling, you need an assessment service with http/1.1 support, e.g., MOOC-Jutut (master). Additionally, you can log urllib3 info with the following in A+ settings:

```python
LOGGING['loggers'].update({
    'urllib3': {
        'level': 'DEBUG',
        'propagate': True,
    },
})
```

When the pool is started, the following entry should be in the log:

```
plus_1         | [2020-08-18 14:52:11,867: DEBUG/connectionpool] Starting new HTTP connection (1): 172.19.0.5:8082
```

If the server doesn't support keep-alive or it has expired, then you may see

```
plus_1         | [2020-08-18 15:02:52,799: DEBUG/connectionpool] Resetting dropped connection: grader
```

If the pool entry is reused, then there is no log entry.

The A+ protocol docs will include details of the headers added here as part of the protocol doc.

- [x] I (developer) ~~have created unit tests and~~ the tests pass
- ~~I (developer) have created functional tests (Selenium tests) if applicable~~

### DIFF
--- a/aplus/__init__.py
+++ b/aplus/__init__.py
@@ -1,0 +1,20 @@
+
+# Version
+try:
+    from .version import __version__
+except ImportError:
+    try:
+        __version__ = (
+            __import__('subprocess')
+                .check_output(('git', 'describe'))
+                .decode('utf-8')
+                .strip()
+                .lstrip('v')
+        )
+    except Exception as error:
+        __import__('warnings').warn(
+            'Unable to determine aplus version. %s: %s' % (
+                error.__class__.__name__,
+                error,
+            ))
+        __version__ = 'unknown'
--- a/course/models.py
+++ b/course/models.py
@@ -19,6 +19,7 @@ from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 from django_colortag.models import ColorTag
 
+from aplus import __version__ as aplus_version
 from apps.models import BaseTab, BasePlugin
 from lib.email_messages import email_course_error
 from lib.fields import PercentField
@@ -628,6 +629,10 @@ class CourseHook(models.Model):
         ("post-grading", "Post grading"),
     )
 
+    HOOK_EVENT_MAP = {
+        'post-grading': 'post-assessment',
+    }
+
     hook_url = models.URLField()
     hook_type = models.CharField(max_length=12, choices=HOOK_CHOICES, default="post-grading")
     course_instance = models.ForeignKey(CourseInstance, on_delete=models.CASCADE,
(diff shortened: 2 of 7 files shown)


Mankro:
@raphendyr did you yourself think that this PR could be merged and deployed in the December 2020 A+ upgrade? Or should we first upgrade MOOC-Grader's Django version from 1.9?

raphendyr:
> @raphendyr did you yourself think that this PR could be merged and deployed in the December 2020 A+ upgrade? Or should we first upgrade MOOC-Grader's Django version from 1.9?

If I remember right, this PR is independent of changes in graders. I.e. there shouldn't be any need to update any grading services for these changes to work.

For connections to be reused, the server needs to support http/1.1, i.e., that requires django update. With older server (http/1.0), A+ will just create a new connection per request. Thus, between A+ and MOOC-Grader, there should be less _new_ SSL connections once the MOOC-Grader is upgraded too.

PR #658, opened 2020-08-19
Remove list-styling from error summary

saarasat:
This removes the list-styling and bullets from the
error summary component.

Related to changes in: https://github.com/apluslms/a-plus-design-system/pull/9
Also the only implementation of this component so far, has been matched to this style in:
https://github.com/apluslms/mooc-grader/pull/82

# Testing

Outlook has been tested with multiple and only one erroneous field:

image

image

Not relevant


mtthwhggns (approved):
This looks very simple and the screenshots look good.

assets/css/main.css
```diff
@@ -1,9 +1,8276 @@
-﻿/*!
```
Mankro, line 1: Is this intentional that `main.css` seems to contain the whole Bootstrap CSS? This is not the built minimized output of SASS.

Mankro (changes requested):
Looks good but I get the feeling that the CSS has been compiled in a wrong way.

assets/css/main.css
saarasat: That is weird, I thought I ran `drone exec`. Apparently committed at the wrong phase then. In any case, thanks for pointing out, this has now been fixed!

raphendyr: `main.css` always contains the bootstrap stuff, but that should be in a single line normally. It sometimes could happen that one runs drone exec, but has the sass watch container running at the same time, and then they touch a source file, which triggers a rebuild, which overwrites the `main.css` again. One thing is to run `git diff --cached` to verify that you are committing what you think you are. However, `main.css` doesn't format there so well.

saarasat: Good to know!

raphendyr (approved):

Mankro (approved):
I rebased and updated the compiled CSS output in order to resolve the git conflict (and force-pushed).

PR #688, opened 2020-09-17
Improve accessibility of the  profile page.

### LINKED ISSUE #590: Fix the order and arrangement of elements in the profile page
Currently, the profile page looks like this: 

Screenshot of profile page

The current issues are:

- Avatar is just described as "avatar"
- `input` (API key) not associated with any label
- `select` (language) not associated with any label
- The explanation about the avatar comes much later in the reading and tab orders than the element itself (due to the columns).

Other possible considerations:

- Are the avatars used anywhere?
  - If not many places, do we need them?
  - If many places, can we justify forcing every student to get a WordPress.com account to set one?
- "you must update them on your identity provider." is quite jargon-heavy and could be replaced with something much clearer, maybe identity-provider-specific with a fallback.

Related: https://apluslms.github.io/accessibility-audit/patterns/fixable-a/semantic-markup-and-sequences/

jaguarfi:
# Description

The profile.html template has some accessibility issues that are listed
primarily in issue #590. Therefore, the UI is updated in this PR, and the following changes are addressed:

1. The avatar is removed
3. Labels will be added to the API key and Language input fields.
4. The instructions about changing the avatar will be removed. It is not relevant to provide such information.
5. The navigation order will be fixed.
    a. Language selector
    b. Save Language button.
    c.  API Access Token field.
    d. API access Token copy - button.
    e. If there are LTI services with the automatic login option saved. The navigation will jump to the Forget buttons.

Based on the mockup proposed by @mtthwhggns I will propose a slightly modified version of it. 

Fixes #590

Desktop version

image

Mobile version

image

# Testing

The accessibility issues were tested with the WAVE plugin and the screen reader ChromeVox.

There is no need for updating any documentation.


raphendyr:
I think the "change" button is unclear about the fact that it leads out of the system, i.e, it's link to an external site. It could be enought to just add a glyph (and screen reader text) about this.

jaguarfi:
It's good to hear your suggestions @raphendyr 😃. So maybe we can use a tooltip and the screen reader text with the text "change avatar"

userprofile/templates/userprofile/profile.html
```diff
@@ -12,62 +12,60 @@
...
 <div class="row">
-    <div class="col-md-2">
-        <div class="thumbnail">
-            <img src="{{ profile.avatar_url }}" alt="avatar" />
-        </div>
-        <dl>
-            {% trans "[Missing]" as miss %}
-        	  <dt>{% trans "Student ID" %}</dt>
+    <div class="col-md-offset-2 col-xs-10 col-md-3">
+		<h2>Personal Information:</h2>
```
Mankro, line 35: The issue #590 already had a reason for separating the user data stored in A+ and the data from the identity provider. I prefer that approach.

userprofile/templates/userprofile/profile.html
```diff
@@ -12,62 +12,60 @@
...
+			</a>
+		</div>
+		<h1>
+			{% if user.first_name or user.last_name %}
+			{{ user.first_name }} {{ user.last_name }}
+			{% else %}
+			{% trans "[Unknown name]" %}
+			{% endif %}
+		</h1>
+		<h3 style="margin-top: -10px;" class="text-muted">{% firstof user.username miss %}</h3>
```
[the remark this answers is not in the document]
Mankro: The variable `miss` is not yet defined at this point so I think it will always render the empty string here. The variable could be defined earlier in the code or you can use a literal string parameter in the `firstof` tag.

userprofile/templates/userprofile/profile.html
```diff
@@ -12,62 +12,60 @@
...
-						</div>
-					</div>
-				</form>
-			</dd>
-        </dl>
+			<dd>{% firstof user.username miss %}</dd>
+		</dl>
+		<form class="form-inline profile-form">{% csrf_token %}
+			<div class="form-group">
+			  <label for="apiAccessToken">{% trans "API Access Token" %}</label>
```
Mankro, line 47: Have we used camelCase `apiAccessToken` or hyphenated `api-access-token` style in HTML ids before?

Hyphens are used in the style guide: https://apluslms.github.io/contribute/styleguides/html/

Mankro (changes requested):
Comments below. Your desktop screenshot shows wide horizontal margins in the page. Where did that come from since the Bootstrap3 `container-fluid` is used?

userprofile/templates/userprofile/profile.html
```diff
@@ -12,62 +12,60 @@
...
-        {% if user.first_name or user.last_name %}
-        {{ user.first_name }} {{ user.last_name }}
-        {% else %}
-        {% trans "[Unknown name]" %}
-        {% endif %}
-    </h1>
+<div class="row">
+	<div class="col-md-offset-2 col-md-2 page-header">
+		<div class="gravatar">
+			<a href="http://gravatar.com" target="_blank" rel="noopener noreferrer">
```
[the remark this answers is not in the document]
jaguarfi: I can remove it. However, I would like to argue that I do not know a single website that does not provide an avatar or profile picture. I am sure that there is not much value added by this feature, but it seems to be a standard. Do we really want to remove the icon/picture?

Regarding the wide margins, I added the `col-md-offset-2` property. If you visit websites such as GitHub, Facebook,  StackOverflow, among many others, you can see that when the content is shown in really wide screens the page make use of wide margins and centred content. If I am correct, that will improve the readability for certain users. Of course, if they use the zoom the text will adapt and the margins will be removed.

https://www.w3.org/TR/low-vision-needs/#margins-and-borders
https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-visual-presentation.html

userprofile/templates/userprofile/profile.html
jaguarfi: I read the conversation, but I did not see the benefits of separating the data stored in A+ from the data that is comming from the  identity provider. Why do we want to do that?. Besides, the amount of data that is shown in this webpage is not really huge, and having one header for showing only two values does not seem practical. It  is more practical two have one header and four values below.

userprofile/templates/userprofile/profile.html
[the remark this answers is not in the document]
Mankro: The user profile avatars are used so little in A+ that people hardly notice that they exist at all. However, they depend on external Wordpress.com accounts, which is not nice from the modern privacy perspective. Removing the icons completely is the easiest solution. It is not worthwhile right now to develop a new profile icon feature in A+.

Wide margins: I know many web sites use them, but A+ hasn't. If the profile page is the only page in A+ that uses wide margins, then it is not consistent with the rest of the system. Since I am not used to seeing wide margins in A+, my initial reaction was that it looked unusual and inconsistent. We can talk more about this later.

userprofile/templates/userprofile/profile.html
Mankro: The language is a user preference set in A+ and API tokens are a feature in A+. We could have more API tokens in the future and the user should also be able to delete and regenerate them. The list of automatic LTI redirections is also specific to A+. Then there are user fields that originate from the identity provider and the user can not change them in A+. Using different sections emphasizes the difference for accessibility and usability. We may also add new A+-specific settings in the future.

userprofile/static/userprofile/copy_token.js
```diff
@@ -0,0 +1,21 @@
+function copyToken(text, el) {
+  var elOriginalText = el.attr("data-original-title");
+
+  $("#api-access-token").select();
+  try {
+    const succeed = document.execCommand("copy");
+    const msg = succeed ? "Copied!" : "Not copied!";
```
Mankro, line 7: A+ has a function for translating strings in JavaScript: `_("Copied!")`

The translations for JS code are defined in these JSON files:
https://github.com/apluslms/a-plus/tree/master/assets/js-translations

assets/sass/legacy/_main.scss
```diff
@@ -60,6 +60,24 @@ table.src tr td.src {
   user-select: none;
 }
 
+/* profile */
+.local-storage-fields li.list-group-item{
```
Mankro, line 64: Would this be a good time to define the styles of the profile page in a new SASS CSS file instead of editing the legacy CSS code?

jaguarfi:
- The avatar was removed.
-  The profile page now separates the information that is stored directly in A+ from the information that is gathered from the identity provider.
- Translations were applied, but someone needs to provide a more accurate translation.

assets/sass/pages/_profile-page.scss
```diff
@@ -0,0 +1,17 @@
+/* profile */
```
raphendyr, line 1: Putting these rules to a different scss file, but including them to the same "main.css" doesn't help to scope these.

All of these rules could be wrapped to a class for this page, thus they wouln't effect any other page by accident.

Alternatively,  per page styles could be compiled to a separate css file (i.e., just create `page_profile.scss` next to the `main.scss` and the scripts should generate the second css. Of course, that means that you have to import all variables and mixings to the second scss file too.

Any case, for scoping, something like this.

```scss
body.page-profile {
  dt.provider-information {
	text-align: left;
  }
  li.list-group-item.clearfix {
    padding-top: 0px;
    padding-bottom: 0px;
    padding-right: 0px;
  }
}
```

userprofile/static/userprofile/copy_token.js
```diff
@@ -0,0 +1,21 @@
+function copyToken(text, el) {
```
raphendyr, line 1: There is at least a second similar script in A+. Could we share the implementation to multiple locations? I.e. could there be a "copy-to-clipboard" component?

userprofile/templates/userprofile/profile.html
```diff
@@ -9,89 +9,114 @@
...
+	<div class="col-sm-offset-2 col-sm-8 col-md-4">
+		{% for item in backends.associated %}
+			{% if item.provider == 'google-oauth2' %}
+				<h2>{% trans "Information provided by Google:" %}</h2>
+			{% endif %}
+		{% endfor %}
+
+		{% for item in backends.not_associated %}
+			{% if item == 'google-oauth2' %}
+				<h2>{% trans "Information provided by Aalto:" %}</h2>
```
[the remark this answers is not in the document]
Mankro: Yes, so that it would be easy to change it for other universities.

userprofile/templates/userprofile/profile.html
```diff
@@ -9,89 +9,114 @@
...
+			{% trans "[Missing]" as miss %}
+			<dt class="provider-information">{% trans "Student ID" %}</dt>
+			<dd>{% firstof profile.student_id miss %}</dd>
+			<dt class="provider-information">{% trans "E-mail" %}</dt>
+			<dd>{% firstof user.email miss %}</dd>
+			<dt class="provider-information">{% trans "Username" %}</dt>
+			<dd>{% firstof user.username miss %}</dd>
+		</dl>
+	</div>
+	{# TODO: check the user type and show editing options/help #}
```
Mankro, line 116: Are these TODOs relevant anymore? The enrolled courses are now actually highlighted on the A+ front page.

Mankro (changes requested):
I have many comments, but they are mostly small fixes. I was not able to run this in the local Docker container because the profile page crashes with this error:
```
ProgrammingError at /accounts/accounts/

relation "social_auth_usersocialauth" does not exist
LINE 1: ...", "social_auth_usersocialauth"."extra_data" FROM "social_au...
                                                             ^
```

assets/sass/pages/_profile-page.scss
Mankro: Good points!

userprofile/templates/userprofile/profile.html
```diff
@@ -9,89 +9,114 @@
...
+		<br>
+		<div>
+			<label for="local-storage-services">{% trans "Automatically log in to:" %}</label>
+			<ul id="local-storage-services" class="local-storage-fields"
+				data-forget-text="{% trans "Forget automatic redirection" %}"></ul>
+		</div>
 		<div class="well">
-			<h4>{% trans "Automatic service redirections" %}</h4>
 			<p>
 				{% blocktrans %}
```
raphendyr, line 61: I recommend to add `trimmed` to the blocktrans, so the white space edit's don't matter so much etc.

```django
{% blocktrans trimmed %}
  First sentence.
  Second paragraph.
{% endblocktrans %}
```

userprofile/static/userprofile/copy_token.js
Mankro: https://github.com/apluslms/a-plus/blob/32b86bf17a33a5c001cc4a3ac87c41b50d8e4100/assets/js/aplus.js#L179
This JS has been used in the submitted file modal dialog in order to copy the file contents to the clipboard.

userprofile/static/userprofile/copy_token.js
```diff
@@ -0,0 +1,21 @@
...
+  var elOriginalText = el.attr("data-original-title");
+
+  $("#api-access-token").select();
+  try {
+    const succeed = document.execCommand("copy");
```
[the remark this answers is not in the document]
Mankro: Also note @raphendyr's comments about this JS.

userprofile/static/userprofile/local_storage_external_services.js
```diff
@@ -19,9 +19,10 @@ $(function() {
 			if (key.indexOf("external_service_") != 0 || data === null)
 				continue;
 			console.log(data);
-			const li = $("<li>" + (data.title === undefined ? key : data.title) + "</li>");
-			li.addClass('list-group-item');
-			const btn = $('<button class="aplus-button--secondary aplus-button--sm pull-right"><i class="glyphicon glyphicon-remove"></i></button>');
+			const li = $("<li>" + (data.title === undefined ? key : data.title.split(' ')[0]) + "</li>");
```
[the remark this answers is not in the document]
raphendyr: I think these locations might be relevant:

I think the title is defined here: https://github.com/apluslms/a-plus/blob/master/external_services/templates/external_services/_launch.html#L23

And here is the JS code, which stores it in the local storage (with update logic) https://github.com/apluslms/a-plus/blob/master/external_services/static/js/external_launcher.js#L35

assets/sass/pages/_profile-page.scss
jaguarfi: > Putting these rules to a different scss file, but including them to the same "main.css" doesn't help to scope these.

My intention is not to put these rules in a different file to scope them. What I am doing is organizing the [architecture](https://sass-guidelin.es/#architecture) of the scss files.

>All of these rules could be wrapped to a class for this page, thus they wouln't effect any other page by accident.

Yes, this is a much better approach. I will create some classes

> Alternatively, per page styles could be compiled to a separate css file (i.e., just create page_profile.scss next to the main.scss and the scripts should generate the second css. Of course, that means that you have to import all variables and mixings to the second scss file too.

Having one single **main.css** file may help to improve the user experience, as all the styles are loaded only once, and then cached styles are used instead. I thought that was the reason why the CSS was minified.

userprofile/templates/userprofile/profile.html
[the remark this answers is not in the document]
jaguarfi: Do you mean the properties defined in the **settings.py** file, such as BRAND_NAME, SHIBBOLETH_TITLE_TEXT and so on?

userprofile/static/userprofile/local_storage_external_services.js
```diff
@@ -19,9 +19,10 @@ $(function() {
...
+			li.addClass('list-group-item clearfix');
+      const btn = $('<button aria-label="Forget automatic login for ' + data.title.split(' ')[0] +
```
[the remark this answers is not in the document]
jaguarfi: > One line uses different indentation than others (tabs vs. spaces).
Yes, it is weird. my text editor does not show the indentation in the same way. Anyway, I will fix it.

jaguarfi:
> I have many comments, but they are mostly small fixes. I was not able to run this in the local Docker container because the profile page crashes with this error:
> 
> 
> ```
> ProgrammingError at /accounts/accounts/
> 
> relation "social_auth_usersocialauth" does not exist
> LINE 1: ...", "social_auth_usersocialauth"."extra_data" FROM "social_au...
>                                                              ^
> ```

For some reason the `INSTALLED_APPS += ('shibboleth_login',)` and the `INSTALLED_APPS += ('social_django',)`  are deactivated in the development environment. Therefore, you must uncomment those setting from the **settings.py** file before testing these changes.

userprofile/static/userprofile/copy_token.js
jaguarfi: Actually, A+ uses https://clipboardjs.com/. Therefore, we can make use of this library in different locations.

userprofile/templates/userprofile/profile.html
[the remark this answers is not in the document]
Mankro: yes, could the word "Aalto" in the translated string come from the BRAND_NAME variable? You need to modify the view in order to add new template context variables.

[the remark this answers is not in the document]
Mankro: There is actually already a template tag for rendering the brand name:
https://github.com/apluslms/a-plus/blob/32b86bf17a33a5c001cc4a3ac87c41b50d8e4100/course/templatetags/base.py#L29

Mankro:
> > I have many comments, but they are mostly small fixes. I was not able to run this in the local Docker container because the profile page crashes with this error:
> > ```
> > ProgrammingError at /accounts/accounts/
> > 
> > relation "social_auth_usersocialauth" does not exist
> > LINE 1: ...", "social_auth_usersocialauth"."extra_data" FROM "social_au...
> >                                                              ^
> > ```
> 
> @Mankro For some reason the `INSTALLED_APPS += ('shibboleth_login',)` and the `INSTALLED_APPS += ('social_django',)` are deactivated in the development environment. Therefore, you must uncomment those setting from the **settings.py** file before testing these changes.

Ahh, I forgot about INSTALLED_APPS. It is also possible to override the `aplus-cont-settings.py` file in the run-aplus-front container and define INSTALLED_LOGIN_APPS there. Shibboleth and social login are disabled by default because they can't be used in the local container.

userprofile/static/userprofile/local_storage_external_services.js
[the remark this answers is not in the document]
jaguarfi: I think we do not need to show the whole `data.title`, e.g., Rubyric (https://rubyric.cs.hut.fi). The name of the service must be enough.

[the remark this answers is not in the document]
Mankro: I don't think showing the URL would be harmful. Nevertheless, if you split by spaces, then it doesn't work when the name contains spaces, e.g., `Lab queue`.

userprofile/templates/userprofile/profile.html
```diff
@@ -9,89 +9,114 @@
...
 				These are all services that you have marked for automatic redirection on this device.
-				You can remove automatic redirection here and you will be shown the privacy information next time you are being redirected to the service.
+				You can remove automatic redirection here and you will be shown the privacy information next time you
```
Mankro, line 63: Should this text about automatic redirection be placed before the list of the services? Users now see the list before the instructions explaining the list. Additionally, if there are no services, then these instructions don't connect to anything. Can you add text like "no services marked" if the list is empty?

userprofile/static/userprofile/local_storage_external_services.js
```diff
@@ -19,9 +19,10 @@ $(function() {
...
+      '" class="aplus-button--danger aplus-button--sm pull-right"><span class="glyphicon glyphicon-remove"></span>Forget</button>');
```
Mankro, line 25: The string "Forget" should be translated.

userprofile/static/userprofile/copy_token.js
jaguarfi: Anycase, I will create a **copy-to-clipboard.js** file with a more generic implementation.

assets/sass/pages/_profile-page.scss
```diff
@@ -0,0 +1,17 @@
...
+.local-storage-fields li.list-group-item{
+	line-height: 1.5em;
+}
+dt.provider-information {
+	text-align: left;
+}
+li.list-group-item.clearfix {
+    padding-top: 0px;
```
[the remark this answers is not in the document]
jaguarfi: yes, my editor uses tabs 🤔 .Maybe I forgot to format the file.

userprofile/static/userprofile/local_storage_external_services.js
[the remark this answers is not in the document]
raphendyr: > my text editor does not show the indentation in the same way

Text editors typically render tabs instead of showing them as tabs (e.g. `^I`). `git diff` is usally good at showing this, as people usually don't use 8 characters in editor and I would expect they don't change the tab width in terminal.

To be clear, rest of the file uses tabs to indent while the new line uses space. The file could be added to the .editorconfig as an exception, if you like. If I remember right, there is quite many different identations in different files.

userprofile/static/userprofile/copy_token.js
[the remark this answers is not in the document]
jaguarfi: https://clipboardjs.com/#browser-support 🤔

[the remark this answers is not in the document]
Mankro: Let's keep using clipboard.js since it is already used in A+ and it seems to work at the moment even though it also depends on obsolete APIs.

assets/sass/pages/_profile-page.scss
jaguarfi: I decided to nest the selectors under the `.local-storage-fields` class. This class is only used in the **profile.html** page

userprofile/templates/userprofile/profile.html
raphendyr: I started to use a local githook to stop pushes, when to code contains keyword "FIXME". For TODO, I have used the idea that it means "this could be made better, but it's not so important". So, from my point of view, every "FIXME" probably should have an issue. Some TODOs could have (i.e., things that should be cleaned after some time and such).

assets/sass/pages/_profile-page.scss
raphendyr: > Having one single main.css file may help to improve the user experience, as all the styles are loaded only once, and then cached styles are used instead. I thought that was the reason why the CSS was minified.

In my mind, this is double-edged sword. If everything is in a single file, then every client needs to download a lot of stuff, which isn't ever used. On the other hand, HTTP/1.0 and HTTP/1.1 are not so good with multiple files. Hence, to my best undestanding, the best would be to group files so that both concerns are covered. E.g., for A+, I probably would have all the css used to render student pages in main.css and all the admin extra in an another file. Profile page is 50/50, rarely used, but also doesn't contain many rules.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3673,78 +3673,68 @@ msgstr ""
...
 
-#: userprofile/templates/userprofile/profile.html:88
+#: userprofile/templates/userprofile/profile.html:75
 msgid "Forget automatic redirection"
-msgstr "Unohda automaattinen uudelleenohjaus."
+msgstr "Unohda automaattinen uudelleenohjaus"
+
+#: userprofile/templates/userprofile/profile.html:104
+msgid "Information provided by %(university)s"
+msgstr "Tiedot tarjoaa %(university)s"
```
Mankro, line 3717: `"Tiedot tarjoaa %(university)s"` sounds funny in Finnish.
`Instituutiolta %(university)s saadut tiedot`

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3673,78 +3673,68 @@ msgstr ""
...
+
+#: userprofile/templates/userprofile/profile.html:106
+msgid "Information provided by Google:"
```
Mankro, line 3720: I think the ending colon should be removed from these headings.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3673,78 +3673,68 @@ msgstr ""
...
+msgstr "Googlen toimittamat tiedot:"
+
+#: userprofile/templates/userprofile/profile.html:108
+msgid "The following settings should be changed in your identity provider"
+msgstr "Seuraavat asetukset tulisi muuttaa henkilöllisyystarjoajasi kanssa"
```
Mankro, line 3725: Finnish: 
`Seuraavat asetukset tulisi muuttaa tunnistustietojen tarjoajan järjestelmässä.`

userprofile/templates/userprofile/profile.html
```diff
@@ -2,96 +2,120 @@
...
-            {% endblocktrans %}
-        </div>
-		<div class="well">
-			<h4>{% trans "Automatic service redirections" %}</h4>
-			<p>
-				{% blocktrans %}
-				These are all services that you have marked for automatic redirection on this device.
-				You can remove automatic redirection here and you will be shown the privacy information next time you are being redirected to the service.
+	<div class="col-sm-offset-2 col-sm-8 col-lg-4">
+		<h2>{% trans "A+ Preferences:" %}</h2>
```
Mankro, line 52: I don't think this section heading "A+ Preferences" shoud have a colon in the end.

userprofile/templates/userprofile/profile.html
```diff
@@ -2,96 +2,120 @@
...
-			</dd>
+			<ul id="local-storage-services" class="local-storage-fields"
+			data-forget-text="{% trans 'Forget automatic redirection' %}" role="group"></ul>
 		</div>
-    </div>
-    <div class="col-md-4">
-
-        {# TODO: list enrolled courses #}
-    </div>
+		<form action="{% url 'set-user-language' %}" method="post" class="form form-horizontal">{% csrf_token %}
```
Mankro, line 77: I think the language form is the most important interactive part in the page and it should be the first item under "A+ Preferences".

userprofile/templates/userprofile/profile.html
```diff
@@ -2,96 +2,120 @@
...
 </div>
+<div class="row">
+	<div class="col-sm-offset-2 col-sm-8 col-md-4">
+		{% if not is_google %}
+			{% brand_university_name as university %}
+				<h2>{% blocktrans with university=university trimmed %}Information provided by {{ university }}{% endblocktrans %}</h2>
+		{% elif is_google %}
+			<h2>{% trans "Information provided by Google:" %}</h2>
+		{% endif %}
+		<p class="text-warning">{% trans "The following settings should be changed in your identity provider" %}.</p>
```
Mankro, line 108: The ending period should be part of the translated string since different languages may have different punctuation rules.

userprofile/templates/userprofile/profile.html
```diff
@@ -2,96 +2,120 @@
...
+		<dl class="dl-horizontal">
+			{% trans "[Missing]" as miss %}
+			<dt class="provider-information">{% trans "Student ID" %}</dt>
+			<dd>{% firstof profile.student_id miss %}</dd>
```
Mankro, line 112: Note that the student ID is always empty when Google login is used and it can not be set in Google. Maybe that is clear enough even though the page says that the info "should be changed at Google".

Mankro (changes requested):
There are language issues to correct and other small UX details.

assets/js-translations/copy-to-clipboard.fi.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
jaguarfi: Definitely, it is something that needs to be analysed.

assets/js/copy-to-clipboard.js
```diff
@@ -0,0 +1,45 @@
...
+ *  data-translation
+ * 	rel="preload"
+ *  as="fetch"
+ *  crossorigin="anonymous"
+ *  type="application/json;"
+ *  hreflang="fi"
+ * href="{{ STATIC_URL }}js-translations/copy-to-clipboard.fi.json"
+ * >
+ */
+const copyToClipboard = (btnClassSelector, targetIdSelector) => {
```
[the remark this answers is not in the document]
jaguarfi: Maybe it is good to start using some [ES6 syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). But, basically, this is the same as `const copyToClipboard = function (btnClassSelector, targetIdSelector)` or ` function copyToClipboard (btnClassSelector, targetIdSelector)`

userprofile/templates/userprofile/profile.html
jaguarfi: Or maybe we can add the student id to the header.
image

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3673,78 +3673,68 @@ msgstr ""
...
-"\t\t\t\tOlen merkinnyt nämä kaikki palvelut automaattisesti "
-"uudelleenohjattaviksi tällä laitteella.\n"
-"\t\t\t\tVoit poistaa automaattisen uudelleenohjauksen. Sinulle näytetään "
-"yksityisyydensuojasta tieto ensikerralla kun sinut ollaan ohjaamassa "
-"palveluun.\n"
-"\t\t\t\t"
+"Olen merkinnyt nämä kaikki palvelut automaattisesti uudelleenohjattaviksi "
+"tällä laitteella.Voit poistaa automaattisen uudelleenohjauksen. Sinulle "
+"näytetään yksityisyydensuojasta tieto ensikerralla kun sinut ollaan "
+"ohjaamassa palveluun"
```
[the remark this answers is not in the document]
jaguarfi: I will switch messages. I will put the message **"No services enabled for automatic redirection on this device."** in the template and the **"You have enabled automatic redirection to these services on this device. You may remove the automatic redirection here. The privacy information will be shown to you next time you open the service."** will be changed only if automatic logins are found in the local storage. The reason for switching these messages is that if the connection is slow, the default message will be changed only until the automatic logins are found.

userprofile/templates/userprofile/profile.html
Mankro: Interesting idea. I would now keep the student id under the institution section since the student id really comes from the institution for Aalto students. Do you strongly feel like moving the student id to the top?

jaguarfi: For me, it looks nice, and it could solve the confusion when the students log in with a Google account, but I do not have a strong opinion about this. We can also leave it where it is now.

Mankro (approved):
Great! I'd like to change that one string below and then you can rebase and squash.

jaguarfi:
Done!

PR #691, opened 2020-10-09
Extract the header, footer and menu-notifications from the legacy/_main.scss file to to individual components and fix some accesibility issues related to these comoponents

### LINKED ISSUE #628: Small template-level accessibility issues
Two small known accessibility issues:

- Text contrast in the page footer
- The gradient in the top navigation area makes automated contrast checks fail (and it looks a little out of place)

Also #609 came up here, but already had an issue.

jaguarfi:
# Description

The base.html template is used as the base page for the entire A+ website. Therefore, the accessibility issues that are raised in this template are spread throughout the entire application. 

This PR address some of the accessibility issues reported in the https://apluslms.github.io/accessibility-audit/. Additionally, the `.topbar`, `.site-footer` and `.menu-notification` styles are migrated from the **legacy/_main.scss**  to their own files, in this case, the new files are located in the  **layout/_header.scss**,  **layout/_footer.scss**  and in the  **components/_menu-notification.scss** respectively.

Fixes #628

# Testing

The contrast of the footer elements was tested using the WAVE extension.
The styles of the topbar were tested manually in localhost.

It is not necessary to update any documentation. 


assets/sass/components/_topbar.scss
```diff
@@ -0,0 +1,141 @@
...
+	& .menu-notification .badge {
+		position: absolute;
+		left: 28px;
+		top: 15px;
+		-webkit-animation: bounce 10s ease-in-out;
+		-moz-animation: bounce 10s ease-in-out;
+		-o-animation: bounce 10s ease-in-out;
+		animation: bounce 10s ease-in-out;
+	}
+	@-webkit-keyframes bounce {
```
Mankro, line 122: Does it make any difference that these CSS animations are now nested inside `.topbar` is the SASS code? In the old CSS, they are not nested.

assets/sass/components/_topbar.scss
```diff
@@ -0,0 +1,141 @@
...
+		color: #ffffff;
+		background: transparent;
+	}
+
+	& a.navbar-brand:hover,
+	& .navbar-nav > li > a:hover,
+	& .navbar-nav > .open > a,
+	& .navbar-nav > .open > a:focus,
+	& .navbar-nav > .open > a:hover,
+	.navbar-inverse .navbar-toggle:focus,
```
Mankro, line 26: The old CSS has the rule `.navbar-inverse .navbar-toggle:focus` without nesting under `.topbar`. However, if I understand SASS correctly, now this rule is nested under `.topbar` even if the line does not start with `& `. That is semantically different than the original code, though maybe it does not affect the outcome.

assets/sass/components/_topbar.scss
```diff
@@ -0,0 +1,141 @@
...
+
+		&:hover {
+			background: #eee;
+		}
+	}
+
+	& .menu-notification .glyphicon {
+		margin-right: 18px;
+	}
+	& .menu-notification .badge {
```
Mankro, line 113: The original CSS did not start with `.topbar` in the rule `.menu-notification .badge`. It probably works as well, though. It is important to test these notifications since they don't appear under typical local testing. I think you can create a fake notification manually in the Django admin interface.

assets/sass/main.scss
```diff
@@ -41,6 +41,8 @@ $font-size-base:  $aplus-font-size-base;
 /*!
  * A+ Styles
  */
+
+@import 'components/topbar';
```
Mankro, line 45: Is it necessary to import topbar first or could we maintain alphabetic sorting in these imports?

templates/base.html
```diff
@@ -263,7 +263,7 @@ <h4 class="modal-title"></h4>
                         <li class="not-found"><a>{% trans "No matches" %}</a></li>
                     </ul>
                 </span>
-                <input type="text" class="form-control" placeholder="{% trans 'Search for...' %}" />
+                <input type="text" aria-label="search" class="form-control" placeholder="{% trans 'Search for...' %}" />
```
Mankro, line 266: The `aria-label` value should be translated. You can use the same value as the `placeholder` attribute so that there is no new string in the po file.

Mankro (changes requested):
Good! There are very minor requested changes and questions about the SASS coding style.

assets/sass/components/_topbar.scss
```diff
@@ -0,0 +1,141 @@
+.topbar {
+	background-color: $aplus-brand-primary;
+	border-bottom: none;
+	-webkit-box-shadow: 0 1px 4px rgba(0, 0, 0, 0.25),
+		inset 0 -1px 0 rgba(0, 0, 0, 0.1);
+	-moz-box-shadow: 0 1px 4px rgba(0, 0, 0, 0.25),
+		inset 0 -1px 0 rgba(0, 0, 0, 0.1);
+	box-shadow: 0 1px 4px rgba(0, 0, 0, 0.25), inset 0 -1px 0 rgba(0, 0, 0, 0.1);
+
+	& .navbar-brand,
```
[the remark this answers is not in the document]
jaguarfi: Now that I read the article and some additional webpages, I think we can avoid using the `&` symbol when we are only nesting elements under the parent selector.

assets/sass/components/_topbar.scss
jaguarfi: Yes, you are right. it was a typo.

assets/sass/components/_topbar.scss
jaguarfi: It seems like I added the curly braces in the wrong place. I will review this file again.

assets/sass/components/_topbar.scss
jaguarfi: Same as above

assets/sass/components/_topbar.scss
jaguarfi: Actually, having these classes under the topbar may increase the specificity of the rules. But, I think I will stick to the previous sass properties.

assets/sass/components/_topbar.scss
jaguarfi: same as above

assets/sass/main.scss
jaguarfi: based on the current implementation we can keep using alphabetic sorting. I will move this line below.

jaguarfi:
I applied the required changes, but I also change the structure of the Sass files a little bit. 

I think we need to define and document the structure of these Sass files and also the architecture.

Mankro (changes requested):
This looks nice. One file had extra whitespace and we can discuss the variable name `$layout-base-color`.

assets/sass/colors/_theme.scss
```diff
@@ -10,3 +10,7 @@ $aplus-brand-secondary: #EF353B;
 // strongly not recommended at the moment.
 $aplus-background: #FFF;
 $aplus-foreground: #000;
+
+// Additional colors used in some components and/or pages
+$layout-base-color: #236aa7;
```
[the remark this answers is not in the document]
jaguarfi: I actually created the variable because that specific colour was used in a couple of places, and I did not want to have the same colour hardcoded in each property. I will explore the `choose-contrast-color` function and I will see what can be done with that function.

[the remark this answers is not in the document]
jaguarfi: I will remove the variable. Using the hexadecimal colour value directly in the CSS properties should work fine.

jaguarfi:
Since the requested changes were rather small I decided to rebase the branch

Mankro (approved):

assets/sass/colors/_theme.scss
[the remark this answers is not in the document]
Mankro: OK, let's merge it now. I thought you might rename the variable. The concern with removing the variable is that now a certain shade of blue is hardcoded in some CSS hover effects, but if some other A+ site (namely Tampere) changes the primary color Sass variable, then will the hardcoded shade of blue work very well with other primary colors? If it doesn't, then we must either add new Sass variables or calculate helper colors with a function that uses the primary color variable as input.

Mankro:
I force-pushed because I removed some whitespace from the Sass theme file and sorted the imports in the Sass main file. 

I wrote new comments in the `$layout-base-color` discussion above.

assets/sass/components/_menu-notification.scss
```diff
@@ -0,0 +1,127 @@
...
+.menu-notification .badge {
+	position: absolute;
+	left: 28px;
+	top: 15px;
+	-webkit-animation: bounce 10s ease-in-out;
+	-moz-animation: bounce 10s ease-in-out;
+	-o-animation: bounce 10s ease-in-out;
+	animation: bounce 10s ease-in-out;
+}
+@-webkit-keyframes bounce {
```
raphendyr, line 13: FYI: this might be a good place for a mixing. See https://stackoverflow.com/a/33203668

PR #693, opened 2020-10-09
Solve accessibility issues on the login page.

### LINKED ISSUE #629: Small accessibility issues on login page
- Heading levels don't match
- It would make sense to use `section` for types of login
- The old button contrast issue
- Input ttype should be programmatically determinable - add some `autocomplete` attributes.

jaguarfi:
# Description

This PR fixes some of the accessibility issues raised in the issue #629. Therefore, the different login options were grouped in sections, the headings were organised and the autocomplete attribute was added to the login form. This last change required to implement the Django form manually, instead of using the Django template form.
Fixes #629

# Testing

The changes were tested with the WAVE and ChromeVox extension. The old implementation of the form was compared with the new implementation and behaviours such as error validation, and the error feedback were validated manually.

There is no need to update the README or any other documentation


assets/sass/main.scss
```diff
@@ -46,3 +46,4 @@ $font-size-base:  $aplus-font-size-base;
 @import 'components/focus';
 @import 'components/frontpage-cards';
 @import 'protocols/quiz1';
+@import 'components//login';
```
Mankro, line 49: These statements should be sorted so that all components are imported before the protocols. In addition, double slash `//`.

userprofile/templates/userprofile/login.html
```diff
@@ -6,104 +6,148 @@
...
+									{% endfor %}
+							</div>
+						{% endif %}
+						{% if form.username.errors %}
+							{% for error in form.username.errors %}
+								<div class="form-group has-error">
+									<label class="control-label"
+										for="{{ form.username.id_for_label }}">{{ form.username.label_tag }}</label>
+									<input class="form-control" id="{{ form.username.id_for_label }}" type="text"
+										name="{{ form.username.html_name }}" required autocomplete="current-password" />
```
Mankro, line 98: `autocomplete="current-password"` must be a mistake for the username field.

assets/sass/pages/_login.scss
```diff
@@ -0,0 +1,12 @@
+.login-section {
```
Mankro, line 1: Like Jaakko said in the profile page PR #688, these page-specific rules could be nested under a page-specific selector. Currently, the body element has the attribute `data-view-tag` that can be set in the HTML template with block view_tag. The CSS could use rule like `body[data-view-tag=login]`.

userprofile/templates/userprofile/login.html
```diff
@@ -20,84 +20,130 @@ <h1>{% trans "Login" %}</h1>
 	<div id="login-box-row" class="row">
 
 		{% if shibboleth_login %}
-			<div class="col-sm-6 col-md-3 login-box">
-				<div class="thumbnail">
-					<div class="caption">
+			<section aria-label="{{ shibboleth_title_text|safe }} login" class="col-sm-6 col-md-3 login-box">
```
Mankro, line 23: I don't think the `safe` filter is safe to use here since unescaped citation marks would break the HTML attribute. Is `shibboleth_title_text` translated in the view? I think so, and then appending "login" to the end does not work for every language.

userprofile/templates/userprofile/login.html
```diff
@@ -20,84 +20,130 @@ <h1>{% trans "Login" %}</h1>
...
+					{% blocktrans trimmed %}
+						If you have been provided with credentials specifically
+						for {{ brand_name }}, use this login.
+					{% endblocktrans %}
+				</p>
+				<form method="post" role="form">
+					{% csrf_token %}
+						{% if form.non_field_errors %}
+							<div class="alert alert-danger">
+								<a class="close" aria-label="Close error list" data-dismiss="alert">×</a>
```
Mankro, line 86: `aria-label` should be translated.

userprofile/templates/userprofile/login.html
```diff
@@ -20,84 +20,130 @@ <h1>{% trans "Login" %}</h1>
...
+									{% for error in form.non_field_errors %}
+										{{ error|escape }}
+									{% endfor %}
+							</div>
+						{% endif %}
+						{% if form.username.errors %}
+							{% for error in form.username.errors %}
+								<div class="form-group has-error">
+									<label class="control-label"
+										for="{{ form.username.id_for_label }}">{{ form.username.label_tag }}</label>
```
Mankro, line 96: I think `{{ form.username.label_tag }}` outputs the `<label>` element, but you have wrapped it manually inside a label.
https://docs.djangoproject.com/en/2.2/topics/forms/#rendering-fields-manually

userprofile/templates/userprofile/login.html
```diff
@@ -20,84 +20,130 @@ <h1>{% trans "Login" %}</h1>
...
+						{% if form.non_field_errors %}
+							<div class="alert alert-danger">
+								<a class="close" aria-label="Close error list" data-dismiss="alert">×</a>
+									{% for error in form.non_field_errors %}
+										{{ error|escape }}
+									{% endfor %}
+							</div>
+						{% endif %}
+						{% if form.username.errors %}
+							{% for error in form.username.errors %}
```
Mankro, line 93: Doesn't the error loop render the username input multiple times for each error? Usually, the error loop only renders the errors and not the input. The input could be rendered after the loop.

Mankro (changes requested):
Minor changes.

assets/sass/pages/_login.scss
jaguarfi: In this particular case, I do not see the need for adding a parent selector since the `login-section` class is specific to the login page. This class is not used in any other part of the application.

userprofile/templates/userprofile/login.html
jaguarfi: There is a problem. Since I have to implement the input field manually to add the autocomplete parameter I have to manage the errors manually too. As you can see in the image below using only the `{{ form.username.label_tag }}` tag will not highlight the test when an error is detected.
image

jaguarfi: But, I think I have to take a closer look at this.

Mankro (changes requested):
Nice! Minor changes needed. I thought that the login sections wouldn't need any aria-label attributes, but do they?

aplus/settings.py
```diff
@@ -37,6 +37,7 @@
 SHIBBOLETH_BUTTON_TEXT = 'Aalto login'
 MOOC_TITLE_TEXT = 'Users external to Aalto'
 MOOC_BODY_TEXT = 'Some of our courses are open for everyone. Log in with your user account from one of the following services.'
+MOOC_ACTION_TEXT = 'Users external to Aalto Login'
```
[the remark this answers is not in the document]
jaguarfi: https://www.w3.org/TR/wai-aria-practices/examples/landmarks/region.html. In this article they mention the use of the `aria-label`. We could use the `aria-labelledby` but in this case, the headers of each section do not mention that this is a login section. It only mentions the type of user, i.e. `'Users external to Aalto'`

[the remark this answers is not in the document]
jaguarfi: I can also hardcode the text in the `aria-label`  attribute. @Mankro which solution do you think is better?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3628,15 +3628,19 @@ msgstr "{:d} pisteet"
...
 msgstr "Kirjaudu sisään"
 
-#: userprofile/templates/userprofile/login.html:61
+#: userprofile/templates/userprofile/login.html:63
 msgid "Log in using Google"
 msgstr "Kirjaudu Google-tunnuksella"
 
-#: userprofile/templates/userprofile/login.html:72
+#: userprofile/templates/userprofile/login.html:70
+msgid "Local users login"
```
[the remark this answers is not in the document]
jaguarfi: The headers of each section do not mention that this is a **login** section. It only mentions the type of user, i.e, `'Local users'`

userprofile/templates/userprofile/login.html
```diff
@@ -20,84 +20,130 @@ <h1>{% trans "Login" %}</h1>
...
+						for {{ brand_name }}, use this login.
+					{% endblocktrans %}
+				</p>
+				<form method="post" role="form">
+					{% csrf_token %}
+						{% if form.non_field_errors %}
+							<div class="alert alert-danger">
+								<a class="close" aria-label="{% trans 'Close error list' %}" data-dismiss="alert">×</a>
+									{% for error in form.non_field_errors %}
+										<p>{{ error|escape }}</p>
```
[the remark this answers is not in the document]
jaguarfi: I don not know where the global autoescape is defined.

aplus/settings.py
```diff
@@ -37,6 +37,7 @@
 SHIBBOLETH_BUTTON_TEXT = 'Aalto login'
```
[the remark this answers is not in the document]
jaguarfi: What about ... "Log in with your Aalto University user account by clicking on the button below. FiTech, Open University and Program students as well as faculty members must log in here."

aplus/settings.py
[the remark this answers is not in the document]
Mankro: All right. I would imagine that when the page starts with the main title "Log in", then people understand that the section title "Aalto University users" means the login for those people without repeating the word "login". Therefore, we can use `aria-labelledby="id of the visible section heading"` in each login type section and we don't need to add new strings. I trust that this is clear because people must notice the main title "Log in" first. I can't test this with screen readers or other accessible technology, but I don't think they would just skip over the main title above the login sections. Even the metadata `<title>` html element contains the word "login" (could be changed to the verb "log in" as many other web sites seem to do so in order to emphasize the action).

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3628,15 +3628,19 @@ msgstr "{:d} pisteet"
 msgid "Login"
```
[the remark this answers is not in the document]
Mankro: The html template uses the word "login" also in the block title (`<title>` element in the page `<head>`), but that title could also be changed to the verb "Log in".

locale/fi/LC_MESSAGES/django.po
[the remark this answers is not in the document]
Mankro: I think the login `<section>` can use `aria-labelledby="id of the visible h2 heading"` like was discussed above.

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
Mankro: The global setting is given in `a-plus/settings.py` in the TEMPLATES variable. We don't define anything which means that autoescaping is enabled by default (like it should). It is also possible to control the autoescaping inside one template with the autoescape templatetag. To summarize, autoescaping is enabled here like it should be and you don't need to call the escape filter manually. Manual escaping should not result in double escaping in Django templates (Django protects against that too), but it is unnecessary to add escaping manually when it is automatically applied in all variables. 

If something were really escaped twice in a row, then that would change the meaning to something else. E.g., `<` in the content becomes first `&lt;` and then `&amp;lt;`, which is rendered as `&lt;` in the browser, not `<`

aplus/settings.py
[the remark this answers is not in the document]
Mankro: Yes, almost. Degree programmes use the longer spelling PROGRAMME. Not all staff members are faculty members, like all the technical support services.
`Log in with your Aalto University user account by clicking on the button below. FiTech, Open University and programme students as well as staff members must log in here.`
Finnish:
`Kirjaudu palveluun Aalto-yliopiston käyttäjätunnuksella alla olevasta painikkeesta. FiTechin, avoimen yliopiston ja koulutusohjelmien opiskelijoiden sekä henkilökunnan täytyy kirjautua tästä.`

userprofile/templates/userprofile/login.html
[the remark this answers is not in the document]
Mankro: I'm looking at this again, but the escape filter is still here. It shall be removed since automatic escaping is active.

Mankro (changes requested):
Great! There are still tiny changes and then this is finished.

jaguarfi:
> I'm looking at this again, but the escape filter is still here. It shall be removed since automatic escaping is active.

It is weird. I thought I had removed it. Any case, I applied the changes and I rebased

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3625,18 +3626,18 @@ msgstr "{:d} pisteet"
 
 #: userprofile/templates/userprofile/login.html:6
 #: userprofile/templates/userprofile/login.html:17
-msgid "Login"
-msgstr "Kirjaudu sisään"
+msgid "Log in to A+"
```
Mankro, line 3629: I now realize that this new version of the string hardcodes the BRAND_NAME setting. We will need to change this later since the BRAND_NAME can be customized on each site.

selenium_test/test/locators.py
```diff
@@ -10,6 +10,7 @@ class FirstPageLocators(object):
     BANNER = (By.XPATH, "//*[@class='page-header']/h1")
     APLUS_TEST_COURSE_INSTANCE_BUTTON = (By.XPATH, "//div[contains(@class, 'card')]/a[contains(@href, '/aplus1/basic_instance/')]//*[contains(@class, 'card-title')]")
     HOOK_EXAMPLE_BUTTON = (By.XPATH, "//div[contains(@class, 'card')]/a[contains(@href, '/aplus1/hook_instance/')]//*[contains(@class, 'card-title')]")
+    SHOW_LOGIN_BUTTON = (By.XPATH, "//*[contains(@class, 'show-extra-login-btn')]")
```
Mankro, line 13: The extra login button is not visible if Shibboleth and Google logins have not been enabled in the settings. The login page test might fail if the button does not exist, but that is easy to fix.

Mankro (approved):
Awesome!

PR #695, opened 2020-10-23
Add instructions for using NGINX dynamic modules and small doc improvements

Gehock:
# Description

This pull request adds instructions about using NGINX dynamic modules to the deployment documentation, as well as improves documentation in other parts.

## Testing

These changes are a result of the Aalto A+ exam server deployment and are currently used in production.

## Is it [Done](https://wiki.aalto.fi/display/EDIT/Definition+of+Done)?

### DIFF
--- a/doc/DEPLOYMENT.md
+++ b/doc/DEPLOYMENT.md
@@ -319,7 +319,36 @@ Following instructions expect that the applocation is installed under `/srv/aplu
 This guide bases on NGINX module [nginx-http-shibboleth](https://github.com/nginx-shib/nginx-http-shibboleth).
 This module uses fastcgi and shibboleth scripts to provide similar integration as Apache 2 plugin.
 
- 1. With Ubuntu xenial or before NGINX 1.11
+ 1. Starting from Ubuntu Bionic or NGINX 1.11 we can use dynamic modules
+
+        # as root
+        cd /usr/src
+
+        # edit /etc/apt/sources.list if necessary, make sure to enable the
+        # -updates repository to get the latest version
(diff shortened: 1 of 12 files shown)


doc/DEPLOYMENT.md
```diff
@@ -372,7 +373,34 @@ This module uses fastcgi and shibboleth scripts to provide similar integration a
...
+          --add-dynamic-module=../nginx-http-shibboleth \
+          --add-dynamic-module=../headers-more-nginx-module \
+          FLAGS_HERE
+        make modules
+        chmod 644 \
+           objs/ngx_http_shibboleth_module.so \
+           objs/ngx_http_headers_more_filter_module.so
+        cp objs/ngx_http_shibboleth_module.so \
+           objs/ngx_http_headers_more_filter_module.so \
+           /usr/lib/nginx/modules/
```
raphendyr, line 400: Can this be `/usr/local/lib/nginx/modules/`?

p.s. in Aalto, could CS IT or Aalto package prebuild modules as a debian package and install that with the nginx on relevant servers?

Mankro (approved):
Nice! I don't know NGINX magic myself but I trust you do. :)
Were you planning to still change something since you did something just today? I will wait a while before merging.

doc/DEPLOYMENT.md
Gehock: Module path changed to `/usr/local/lib/nginx/modules`. Technically the module could be prebuilt as a debian package but I'd say that setting up shibboleth is such a rare case that it's not worth the effort

Mankro (approved):
Excellent! Thank you so much!

PR #699, opened 2020-10-30
Change background-color and color of the badges and progress-bar

### LINKED ISSUE #670: Colour contrast of points badges
As I've discussed here a lot, colour contrast is quite crucial for users with low vision, general readability and WCAG compliance.

The biggest problem are the points badges shown when a user has partial points, with white on orange. The white on green also fails to meet the requirements. 

Issue https://github.com/apluslms/a-plus/issues/568 was meant to address this, but was closed after the ability to change the colours was completed without changing all of them. 

We still have the problems. 

I implemented new colours in `assets/sass/colors/_functional.scss` which mirror the Bootstrap colour names. These were initialised with darker versions of their original colours, but weren't used by the badges (the badges are not part of Bootstrap). They caused the progress bars to appear in a different colour than the badges, which isn't ideal.

I tested replacing the badge colours with the darker colours to meet the required contrast ratio for small text, but it becomes much harder to tell them apart (dark green, dark orange, etc all look quite alike). 

Screenshot 2020-08-27 at 13 34 10

This was where the suggestion to replace the badge with an icon, and beside the icon would be text in the normal colour, originated. The shape of the icon would provide a clear way to tell them apart, and colour could be used as a bonus. 

This was implemented in PR https://github.com/apluslms/a-plus/pull/646 however, since then, the discussion in https://github.com/apluslms/a-plus-design-system/issues/3 has stretched the requirements a little, and there was not time to finish it this summer.

If this cannot be solved by January's deployment time, I would recommend returning the Bootstrap `warning` colour to the bright orange, and add it as a known issue to the accessibility statement.

jaguarfi:
# Description

I changed the `background-color` and  `color` of the badges and progress-bars. I used black as text color for the badges to keep a high contrast. 

Fixes #670

# Testing

I check the contrast using the WAVE plugging in some of the pages that used badges. 

Badge colours
image

progress-bar
image

...


Mankro (approved):
Nice! This is already acceptable, but I have some concerns. The red badge is a little hard to read with my normal colour vision. Could this be improved? Additionally, the points badges now use different colours than the points progress bars (in the module panels). Though, at the moment I don't feel like the different shades of colour are really a problem. You notice them, but they are not distracting. Do people understand that the different shades still convey the same meaning (dark and light green both imply full points)? Using the same colours and symbols consistently for the same meaning is probably better for accessibility.

jaguarfi:
>The red badge is a little hard to read with my normal colour vision. Could this be improved? 

I will select a different red maybe something more bright. 

> Additionally, the points badges now use different colours than the points progress bars (in the module panels). Though, at the moment I don't feel like the different shades of colour are really a problem. You notice them, but they are not distracting. Do people understand that the different shades still convey the same meaning (dark and light green both imply full points)? Using the same colours and symbols consistently for the same meaning is probably better for accessibility.

I agree. Maybe it is meaningful to use the exact same colour in the badges and in the progress bar.

assets/sass/colors/_functional.scss
```diff
@@ -8,3 +8,12 @@ $aplus-info: #1d7792; // This is a darkened version of the bootstrap 3 default.
 $aplus-warning: #a6670e; // This is a darkened version of the bootstrap 3 default. Contrast must be >=4.5 with white.
 $aplus-danger: #a50000; // There should be a contrast ratio above 7 between this and white.
 $aplus-focus: #ffdd00; // This should be clearly different fron those defned above and the brand colours.
+
+// This colours may be used to represent the status of the grading.
```
Mankro, line 12: typo This -> These

assets/sass/colors/_functional.scss
```diff
@@ -8,3 +8,12 @@ $aplus-info: #1d7792; // This is a darkened version of the bootstrap 3 default.
...
+// The best color to use in order to have a good contrast level should be #000.
```
Mankro, line 13: What is this comment trying to say? That the contrast must be high against black? There is some English grammar mistake here.

assets/sass/colors/_functional.scss
```diff
@@ -8,3 +8,12 @@ $aplus-info: #1d7792; // This is a darkened version of the bootstrap 3 default.
...
+$aplus-primary-status-bg: #42a5f5;
+$aplus-default-status-bg: #c5c5c5;
+$aplus-success-status-bg: #8bc34a;
+$aplus-warning-status-bg: #ffb74d;
+$aplus-danger-status-bg: #ff5c72;
+$aplus-info-status-bg: #00bcd4;
```
Mankro, line 19: If we add new variables for these colours, then my initial reaction is that the variable names should be semantically correct. "Danger" is correct in the sense that it is dangerous to have failed submissions that do not pass point requirements, but it is more accurate to say "not passed" or "failed". Thus, the variable name could be `$aplus-failed-status` (value red colour).

Similarly,
* `aplus-full-points-status` (green)
* `aplus-passed-status` (yellow)
* `aplus-not-submitted-status` (gray)
* `aplus-unofficial-status` (gray)

Is it necessary to include `bg` (background) in the variable name at all? Do we have other variables for the text colour or is it always black?

What do you think of these? If we use semantic variable names now, then the same variables would maybe make more sense in the new points progress component too (#646).

assets/sass/mixins/_badge.scss
```diff
@@ -0,0 +1,4 @@
+@mixin badge-color($bg-color, $color) {
```
Mankro, line 1: This is cool and it matches the existing button mixins.

assets/sass/components/_badge.scss
```diff
@@ -0,0 +1,18 @@
+@import "../mixins/badge";
+
+.badge {
+	@include badge-color($aplus-default-status-bg, #000);
+
+	&-success {
```
Mankro, line 6: (no change requested here) Concerning the variable names, since A+ uses CSS classes `badge-success` and `badge-danger` etc. everywhere, we still want to keep those class names. We don't want to change them everywhere at this time.

Mankro (changes requested):
I started to ponder about the Sass variable names. Share your views in the comments.

assets/sass/colors/_functional.scss
jaguarfi: yes, I meant the color css property. But, maybe the comment is not needed since we can choose any other colour based on the contrast we want to achieve.

assets/sass/colors/_functional.scss
Mankro: We agreed that at the moment we can keep using the Bootstrap names "success", "danger" etc. They are consistent with the CSS class names that are used everywhere in the HTML and the Bootstrap progress bar component also uses these names in the Sass variables for the progress bar colours.

jaguarfi: As discussed over the phone. We will keep the current names since the names somehow follow the bootstrap naming convention. Regarding the #646, if needed we can rename the variable names and use semantic names for the status of the grades.

jaguarfi:
I decided to use the`choose-contrast-color($bg-color, $light-color, $dark-color)` function to objectively choose the best colour to contrast the background colour of the badges.

Mankro (approved):

PR #708, opened 2020-11-09
Fix issues related to the course pagination element

### LINKED ISSUE #704: Navigation links at the bottom and top of the page do not follow a logical tab navigation
The navigation links at the bottom of the page are not arranged sequentially. Therefore, when using the tab key to navigate the course content, the navigation goes to the right link (Next) then the left link(Previous) and finally the centre link(Course Material)
bottom_navigation

jaguarfi:
# Description

The navigation order of the course content nav, located at the bottom of
the page, did not follow a logic keyboard navigation. This commits
fixes the issue, and now the keyboard navigation goes from left to
right. This new navigation order follows the Web Content Accessibility
Guidelines.

Fixes #704

# Testing

The tab navigation was tested manually. 

There is no documentation that needs to be updated.


Mankro (changes requested):
The top bar is not well aligned. The vertical position varies between the previous/course/next links. Could this be fixed? The bottom bar has the same problem. The tab order works well.
image

If you prefer to merge this PR and start a new issue about the vertical position, let's do that. It looks like the old code has the vertical alignment problem too, but perhaps it got even worse now with your changes.

jaguarfi:
I  changed the alignment of the links inside the pagination element.  I also rebased the commits.

jaguarfi:
I will address the issue #713 in this PR.

assets/sass/components/_pagination.scss
```diff
@@ -0,0 +1,22 @@
+nav.row.siblings {
+    display: flex;
```
Mankro, line 2: Two lines use spaces for indentation while the rest uses tabs as intended.

assets/sass/components/_pagination.scss
```diff
@@ -0,0 +1,22 @@
...
+    flex-direction: row;
+	flex-wrap: wrap;
+	justify-content: space-between;
+	margin: 0 5px;
+
+	a.nav-flex-left{
+		justify-content: flex-start;
```
Mankro, line 9: Firefox shows me a warning that `justify-content` has no effect since the element is not a flex container. Is it better to just remove that rule since everything works fine without it? Firefox suggests adding `display: flex` or `display: grid`, but it doesn't look like we would need that.

Mankro (changes requested):
Great! I found a problem in the bottom pagination bar, though. 

The bottom course pagination bar disappears in narrow displays, but the elements are not hidden by CSS. The bottom privacy bar simply covers them. It would make sense to fix this, right? It is better to show the bottom pagination bar properly even in small displays. We don't want to remove it.

assets/sass/components/_pagination.scss
jaguarfi: Yes, it is true. The `justify-content` property is not necessary.

assets/sass/components/_pagination.scss
```diff
@@ -0,0 +1,25 @@
+nav.row.siblings {
+	display: flex;
+	flex-direction: row;
+	flex-wrap: wrap;
+	justify-content: space-between;
+	margin: 0 5px;
+
+	a.nav-flex-left {
```
Mankro, line 8: Since `a.nav-flex-left` and `a.nav-flex-right` have identical rules now, they can be combined with the comma `a.nav-flex-left, a.nav-flex-right { ... }`

Mankro (approved):
Excellent! You can rebase and clean up git history.

jaguarfi:
@Mankro Done!

Mankro:
I fixed the git conflict in the built main.css file myself and force-pushed.

PR #715, opened 2020-11-12
Hide courses marked as not visible to students from students in UI

### LINKED ISSUE #684: A+ frontpage shows hidden courses to students
The A+ frontpage was heavily changed in the summer 2020. Unfortunately, it now shows hidden courses to students at least if the student has previously enrolled in the course. If the `CourseInstance` field `visible_to_students` is `False`, then the course must not be shown to students, not even the title of the course in the frontpage.

sayravai:
# Description

- In course/cache/menu.py: Check that a course is visible to students before adding it into the menu.
- In course/views.py: Create a list of all course instances (filtering out courses not visible for current user) before checking which courses the user has enrolled in.

Fixes #684 

# Testing

Unit tests and Selenium tests have been run. I did not add any new tests.

Manual testing:
- Create a new course instance and make it open for current date
- Register for the course instance as a student
- Mark the course instance as not visible for students
- Check that the enrolled student does not see the course instance

No.

No.

### DIFF
--- a/course/views.py
+++ b/course/views.py
@@ -57,10 +57,13 @@ class HomeView(UserProfileView):
                 if instance not in my_instances:
                     my_instances.append(instance)
             
-            for instance in user.userprofile.enrolled.all().filter(ending_time__gte=end_threshold):
+            for instance in user.userprofile.enrolled.all().filter(
+                    ending_time__gte=end_threshold,
+                    visible_to_students=True,
+                ):
                 if instance not in my_instances:
                     my_instances.append(instance)
-        
+
         all_instances = CourseInstance.objects.get_visible(user).filter(ending_time__gte=end_threshold)
         all_instances = [c for c in all_instances if c not in my_instances]
         
--- a/course/cache/menu.py
+++ b/course/cache/menu.py
@@ -36,7 +36,8 @@ class CachedTopMenu(CachedAbstract):
 
         enrolled = []
         for instance in profile.enrolled.all():
-            enrolled.append(course_entry(instance))
+            if instance.visible_to_students:
+               enrolled.append(course_entry(instance))
 
         teaching = []
         for course in profile.teaching_courses.all():


course/views.py
```diff
@@ -58,10 +60,9 @@ def get_common_objects(self):
                     my_instances.append(instance)
             
             for instance in user.userprofile.enrolled.all().filter(ending_time__gte=end_threshold):
-                if instance not in my_instances:
+                if instance not in my_instances and instance in all_instances:
```
raphendyr, line 63: Here is a check `instance in <queryset>`. Does it generate a sql query per evaluation?

As an alternative solution, should `user.userprofile.enrolled.all()` limit the output like the `get_visible(user)` does?

sayravai: Seems to me, that only one query is created from that line, even with several enrolled courses.

raphendyr: Ok good. Then I guess the django sql cache works as it's supposed to.

Mankro:
@sayravai Shall we fix the author email in the git commit? Currently, it's a fake email that was set by your computer by default. Github.com connects commits to your account based on the email in the commit and in your Github profile. Github can then give you credit for your contribution when the work is linked to the Github user. It is also possible to use a fake Github email if you don't want to reveal real email addresses publicly. The commit metadata is obviously public. I have used my real email in commits without trouble, but it is easy to set up the fake Github email as well. It can be found in the Github account settings.

`git config --global user.email 'first.last@email.org'`

Changing the author data of existing commits changes the git history since the commit checksum will change. I can help you with that, or I can just change the email myself for you.

course/views.py
```diff
@@ -58,10 +60,9 @@ def get_common_objects(self):
                     my_instances.append(instance)
             
             for instance in user.userprofile.enrolled.all().filter(ending_time__gte=end_threshold):
```
Mankro, line 62: I think it would make sense to check the visibility in the same filter call:
`for instance in user.userprofile.enrolled.all().filter(ending_time__gte=end_threshold, visible_to_students=True):`
Then it is not necessary to run `instance in all_instances`. If the user is a teacher, then the above lines have already added the instance to `my_instances`.

What do you think?

Mankro (approved):
Great first contribution! I had a question about the way visible courses are filtered and about the commit author email.

course/views.py
sayravai: Yes, you are right. I also tried to think of a way to save one DB query here, but doesn't seem feasible as the third one is done via userprofile, not course instances. The queries made here using your suggestion are (in pseudocode):
SELECT COURSES WHERE user is assistant OR user is teacher OR course is visible for students AND course is ongoing
SELECT COURSES WHERE user is teacher AND course is ongoing
SELECT COURSES WHERE user is enrolled AND course is visible for students AND course is ongoing

PR #716, opened 2020-11-13
Add icon and a warning for screen readers to links that open in new tabs

### LINKED ISSUE #659: Make clear if a link opens in a new tab
[Guideline 3.2 of WCAG](https://www.w3.org/TR/2016/NOTE-UNDERSTANDING-WCAG20-20161007/consistent-behavior.html) says that links should only open in new tabs where it's necessary and [technique H83](https://www.w3.org/TR/WCAG20-TECHS/H83.html) says the user should be warned of this.

I just spotted that in course/templates/course/course_base.html we are opening links from the `advert` in `target="_blank"` with no warning, but whoever does this could quickly grep te codebase for `_blank`to see if there are any more.

If things need to open in a new tab, adding words like

> (opens in a new tab)

after the link text should suffice.

jaguarfi:
# Description

Add an icon and a hidden warning for the screen readers that says "open in a new tab"
Fixes #659
image

# Testing

I tested the changes manually.

There is no related documentation.


jaguarfi:
I wonder if the icon is pretty enough, maybe we can find some alternatives. I am also concern about the menu items that open in new tabs. 
image
Should remove the icon for those menu items? or remove the icons for all the links and keep only the screen reader description?

assets/js/aplus.js
```diff
@@ -86,6 +86,37 @@ $(function() {
...
+    }
+
+    function addScreenReaderMessage(link, message) {
+      if (!link.querySelector('.screen-reader-only')) {
+        link.insertAdjacentHTML('beforeend', `<span class="screen-reader-only"> (${message})</span>`);
+      }
+    }
+
+    document.querySelectorAll('a[target="_blank"]').forEach(link => {
+      console.log("open in a new tab");
```
Mankro, line 115: `console.log` should be removed.

assets/js/aplus.js
```diff
@@ -86,6 +86,37 @@ $(function() {
...
+    }
+
+    function addExternalLinkIcon(link) {
+      if (!link.querySelector('.icon')) {
+        link.insertAdjacentHTML('beforeend', `<span class="icon external-link"></span>`);
+      }
+    }
+
+    function addScreenReaderMessage(link, message) {
+      if (!link.querySelector('.screen-reader-only')) {
```
Mankro, line 109: Why did you make a new class `screen-reader-only` instead of using the `sr-only` in Bootstrap?

assets/js/aplus.js
```diff
@@ -86,6 +86,37 @@ $(function() {
...
+        link.insertAdjacentHTML('beforeend', `<span class="screen-reader-only"> (${message})</span>`);
+      }
+    }
+
+    document.querySelectorAll('a[target="_blank"]').forEach(link => {
+      console.log("open in a new tab");
+      addLinkType(link, 'noopener');
+      addExternalLinkIcon(link)
+      addScreenReaderMessage(link, 'opens in a new tab');
```
Mankro, line 118: The message `opens in a new tab` must be translated (in Finnish it is `avautuu uudessa välilehdessä`).

assets/sass/helpers/_new_tab_links.scss
```diff
@@ -0,0 +1,27 @@
...
+.icon::after {
+	display: inline-block;
+	font-style: normal;
+	font-variant: normal;
+	text-rendering: auto;
+	-webkit-font-smoothing: antialiased;
+}
+
+.external-link::after {
+	font-family: "Font Awesome 5 Free";
```
Mankro, line 13: Did you download a font library (Font Awesome) for this to work? I don't see any library in the pull request.

assets/sass/helpers/_new_tab_links.scss
```diff
@@ -0,0 +1,27 @@
...
+	font-weight: 900;
+	content: "\f35d";
```
Mankro, line 15: Is it necessary to use the `::after` pseudo-element? You already add new `<span>` elements in JS, so could the icon content be included directly in them? If that is infeasible, then of course the after pseudo-element is the right solution.

assets/sass/helpers/_new_tab_links.scss
```diff
@@ -0,0 +1,27 @@
+// The following classes are used in assets/js/aplus.js. Therefore, you must
```
Mankro, line 1: I think this file belongs to the components directory. Helpers are supposed to be for example Sass functions that can be used in other Sass code, but this file really contains styles for certain elements. In other words, it is a component.

Mankro (changes requested):
I think that the current icon looks good enough, also in the course menu.

I added a Rubyric link to the course menu. Somehow in Firefox, the link actually did not open in a new tab unless I specifically clicked on the icon. Clicking on the text Rubyric opened the target in the same window. Could you investigate?

assets/js/aplus.js
jaguarfi: I forgot about that class. I will use `sr-only` instead

assets/js/aplus.js
jaguarfi: When I was searching for my `console.log()` I found some others in the code. How many of those are intentional?
image

assets/sass/helpers/_new_tab_links.scss
jaguarfi: There is a link in the **base.html** file https://github.com/apluslms/a-plus/blob/6a3d1e5342f234fc85f915e66a31b099093cc609/templates/base.html#L21

assets/js/aplus.js
Mankro: Let's ignore the old console.logs now, but your new code does not need it.

assets/sass/helpers/_new_tab_links.scss
jaguarfi: The `<span>` that I added is `sr-only` and it only contains the text that is going to be read by the screen reader. I decided to use the pseudo-element  `::after`, because that symbol is a decorative feature. Of course, we can also add a second `<span>` or `<i>` element to add the icon using JS. Which one do you prefer?

jaguarfi:
> I added a Rubyric link to the course menu. Somehow in Firefox, the link actually did not open in a new tab unless I specifically clicked on the icon. Clicking on the text Rubyric opened the target in the same window. Could you investigate?

Regarding this issue, it seems that having plain text inside the `<a>` tag is causing the problem. As you see in the image below, the "Rubyric" text is just plain text and does not have a proper HTML tag.

image

This issue is solved by wrapping the text inside an HTML element such as a `<span>`

image


Actually, I have seen a lot of text that is added, as plain text, to the HTML code using the Django template language without any HTML tag. Maybe we can check the code and add tags whenever is possible.

assets/sass/helpers/_new_tab_links.scss
Mankro: `::after` is OK. Let's keep it.

Mankro:
> > I added a Rubyric link to the course menu. Somehow in Firefox, the link actually did not open in a new tab unless I specifically clicked on the icon. Clicking on the text Rubyric opened the target in the same window. Could you investigate?
> 
> Regarding this issue, it seems that having plain text inside the `<a>` tag is causing the problem. As you see in the image below, the "Rubyric" text is just plain text and does not have a proper HTML tag.
> 
> image
> 
> This issue is solved by wrapping the text inside an HTML element such as a `<span>`
> 
> image
> 
> Actually, I have seen a lot of text that is added, as plain text, to the HTML code using the Django template language without any HTML tag. Maybe we can check the code and add tags whenever is possible.

It is normal to have text directly in an `<a>` element. I would like to avoid adding extra `<span>` elements everywhere. I will take a look at this now myself too.

Mankro:
After looking into this, I think that the new tab issue comes from other JavaScript code not related to this pull request. The language link fixes last summer included JS that modifies all links in order to keep the current GET query language parameters in the links. That JS might redirect the window location without taking `target="_blank"` links into account. However, it can be fixed.

https://github.com/apluslms/a-plus/blob/8b45edde82f9f0bef6046ea87559c0227a38110c/assets/js/aplus.js#L648

Mankro (approved):
I make small fixes myself. It looks like some of the icon code is unnecessary since the icon appears even without the fa html classes.

Mankro:
@jaguarfi Another thing to note here is that your JS adds the icon to links without checking if the link points to an external site or the same A+ site. Though, the LTI links in the course menu always point to the same A+ site, but it makes sense to show the icon there even though there is a separate warning page about leaving the site.

Mankro:
@jaguarfi I fixed the bug about target blank links. The language URL PR from the summer included JS that always opened links in the current window. The new fix is here:
https://github.com/apluslms/a-plus/commit/bca5fe14626e50c2fa9ef67ea18358e9654713b3

jaguarfi:
Regarding this https://github.com/apluslms/a-plus/pull/716#discussion_r531553097. I realised that at some point I pushed some changes that I was testing locally. Therefore, the link was added using only JS, and the font awsome classes are doing nothing at this point. I will create a new PR for this. I am sorry about this.

PR #718, opened 2020-11-16
Remove the role="presentation" fom the <li> tags inside navigation menus 

### LINKED ISSUE #714: It’s unclear why every item has role="presentation" in the Course Navigation

jaguarfi:
# Description

Fixes #714

# Testing

Tested manually with [ChromeVox](https://chrome.google.com/webstore/detail/chromevox-classic-extensi/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en)

The system design document has a mention of this issue. Therefore, the system design document needs to be updated.

### DIFF
--- a/course/templates/course/_course_menu.html
+++ b/course/templates/course/_course_menu.html
@@ -14,11 +14,10 @@
 </li>
 
 {% if instance.language|first == "|" %}
-</li>
-<li role="presentation" class="header visible-xs">
+<li class="header visible-xs">
 	<h4 id="course-menu-heading">{% trans "Language" %}</h4>
 </li>
-<li role="presentation" class="visible-xs dropdown">
+<li class="visible-xs dropdown">
 	<a href="#" class="dropdown-toggle" id="lang-dropdown-mobile" data-toggle="dropdown" role="button"
 		aria-haspopup="true" aria-expanded="false">
 		{% trans "Change language" %} <span class="caret"></span>
@@ -36,7 +35,7 @@
 		{% endfor %}
 	</ul>
 </li>
-<li role="presentation" class="header">
+<li class="header">
 	<div class="row">
 		<div class="col-sm-10">
 			<h4 id="course-menu-heading">{% trans "Course" %}</h4>
@@ -62,24 +61,24 @@
 		</div>
 	</div>
 {% else %}
-<li role="presentation" class="header">
+<li class="header">
 	<h4 id="course-menu-heading">{% trans "Course" %}</h4>
 </li>
 {% endif %}
 
-<li role="presentation" class="menu-home">
+<li class="menu-home">
 	<a href="{{ instance|url }}">
 		<span class="glyphicon glyphicon-home" aria-hidden="true"></span>
 		{{ course.code }}
 	</a>
 </li>
-<li role="presentation" class="menu-toc">
+<li class="menu-toc">
 	<a href="{{ instance|url:'toc' }}">
 		<span class="glyphicon glyphicon-book" aria-hidden="true"></span>
 		{% trans "Course materials" %}
 	</a>
 </li>
-<li role="presentation" class="menu-results">
+<li class="menu-results">
 	<a href="{{ instance|url:'results' }}">
 		<span class="glyphicon glyphicon-stats" aria-hidden="true"></span>
 		{% trans "Exercise results" %}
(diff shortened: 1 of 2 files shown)


Mankro (changes requested):
Like I wrote in the issue, I interpreted the aria role menu and menuitem so that they really should NOT be used in the A+ course menu. I got the impression that the role presentation could be removed from the li elements. The nav element already conveys the semantics of the navigation landmark. We have multiple navigation landmarks and thus they require aria labels (or aria-labelledby links to other titles). 

Can you look into this again? It would good to verify still that the role presentation is unnecessary within the nav element and that the role menu is not intended for normal web menus.

course/templates/course/course_base.html
```diff
@@ -26,7 +26,7 @@
 {% block content %}
 <div data-taggings="{{ get_taggings|join:' ' }}"  class="row">
     <div class="col-sm-2 hidden-xs">
-        <nav class="course-menu" id="main-course-menu" aria-labelledby="course-menu-heading">
+        <nav class="course-menu" id="main-course-menu" aria-label="Course Navigation"></nav>
```
Mankro, line 29: I think this is wrong. `aria-labelledby="course-menu-heading"` was already good because the course-menu-heading is the title with the word "course" above the course menu (in the desktop screen size). But the page has multiple elements with the same id, so some ids should be changed.

I would use aria-labelledby. If aria-label were used, then note that the current text is not translated and it repeats the word navigation. Since it is the label on a nav element, it should NOT contain the word "navigation". Screen readers already announce the type of elements, thus the label "course navigation" is probably announced as "course navigation navigation". The word "course" is enough to describe the purpose of the navigation landmark and the nav element signals clearly that this is a navigation landmark.

course/templates/course/_course_menu.html
```diff
@@ -14,7 +14,7 @@
 </li>
 
 {% if instance.language|first == "|" %}
-<li role="presentation" class="header">
+<li  class="header">
```
Mankro, line 17: You left extra whitespace in the `<li>` elements in this file. Do you mind cleaning it up?

Mankro (changes requested):
Removing role "presentation" is OK. What has happened to the id `course-menu-heading` since the page can have multiple elements with the same id? Did you intentionally use it on some elements that have the title "Language" or "change language"? I think the course menu heading should be simply the word "course". I suppose both the desktop and mobile menu can use that heading. The same id should not be repeated in the document, so could the mobile menu use an id like "mobile-course-menu-heading"?

course/templates/course/course_base.html
jaguarfi: Yes, you are right. Using `aria-labelledby` is more convenient in this case

course/templates/course/course_base.html
```diff
@@ -26,7 +26,7 @@
 {% block content %}
 <div data-taggings="{{ get_taggings|join:' ' }}"  class="row">
     <div class="col-sm-2 hidden-xs">
-        <nav class="course-menu" id="main-course-menu" aria-labelledby="course-menu-heading">
+        <nav class="course-menu" id="main-course-menu" aria-labelledby="course-menu-heading"></nav>
```
Mankro, line 29: You weren't supposed to close the nav element on this line.

Mankro (changes requested):
OK. You can fix that one mistake in the nav element and rebase and resolve the git conflicts.

course/templates/course/course_base.html
jaguarfi: The IDE doing autocorrections 🤦🏽‍♂️

course/templates/course/_course_menu.html
```diff
@@ -14,11 +14,10 @@
 </li>
 
 {% if instance.language|first == "|" %}
-</li>
```
jaguarfi, line 17: I accidentally introduced this closing tag in the previous PR

course/templates/course/_course_menu.html
```diff
@@ -14,11 +14,10 @@
...
-<li role="presentation" class="header visible-xs">
+<li class="header visible-xs">
 	<h4 id="course-menu-heading">{% trans "Language" %}</h4>
```
Mankro, line 18: I wonder why the subtitle "Language" has the id `course-menu-heading`. This is probably a mistake. This id is already used in the heading above the menu. @jaguarfi could you fix that? The language heading should use a different id.

Mankro (approved):
Good!

PR #735, opened 2020-12-01
Add a hidden link to open exercises in a new tab / full-page

### LINKED ISSUE #669: Provide a modal-free workaround for assignment submissions
We know that modals are breaking the keyboard and screen reader accessibility and as of yet, I've not got close to finding a good solution (we have too many, called in too many places and inconsistently to find a single straightforward solution).

There has been talk of dropping modals altogether, but this is a long way off. 

I just had the idea that we could temporarily work around this in exercises embedded inside chapters by:

- Having a (maybe hidden like the skip links) link right before the exercise, which says something like "open exercise 5 in a new tab".
- This opens the full-page view showing the exercise only (which I believe we have already, but don't link to), where all of the links point to pages, rather than opening modals.
- This page also includes links to submissions and results.

Can anyone think of a problem with this solution?

jaguarfi:
# Description

Aplus does not support keyboard navigation for modal windows, and all
the current exercises use modal windows for different tasks, e.g., show
model response, inspect submission among other tasks. Therefore, the
exercises must be opened in full-page for users who use keyboard
navigation. We added hidden links that are shown only when the user
navigates the course with the keyboard.

Fixes #669
# Testing

I tested the changes manually.

There is no documentation related to this issue.

### DIFF
--- a/exercise/templates/exercise/exercise_plain.html
+++ b/exercise/templates/exercise/exercise_plain.html
@@ -37,6 +37,9 @@
     </head>
     <body>
         <div id="exercise-all">
+            <a href="{{ exercise|url:'exercise' }}" target="_blank" rel="noopener" class="skip-link">
+                {% trans "Open this exercise in a new tab (recommended for keyboard and assistive technology users)" %}
+            </a>
             <div class="overlay-parent">
 
                 {% if issues and not submission_allowed %}
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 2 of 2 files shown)


Mankro (changes requested):
I was surprised that you used JavaScript for adding the link. I would like to check first if this could be done easily in the backend. It is better to add common navigation in the backend without relying on frontend JS. The browser may have disabled JS completely.

If you add the hidden link close to the navigation bar of the embedded chapter exercise (link below), does it work straight away or do you need small JS changes in the chapter.js code?
https://github.com/apluslms/a-plus/blob/0e72da9e1e8007f23cf710d3bc09a6b9d44b2a87/exercise/templates/exercise/exercise_plain.html#L39

Note that chapter.js uses a lot of selectors that depend on the HTML, e.g., the `id="exercise-all"`.

Mankro (approved):
Excellent! Simple and effective.

PR #736, opened 2020-12-02
Modify <a> event handler and remove target='_blank'

### LINKED ISSUE #732: Submitted file links are decorated with external link icons in the inspect submission page
File link in the inspect submission page is appended with the external link icon. That should not happen. Opening the link seems to always open a new tab instead of the modal dialog. It should open in a modal dialog.

In addition, double-check submitted file links in the students' feedback pages (modal dialogs in the chapter pages).

jaguarfi:
# Description

* Remove the attribute `target='_blank'` from the `<a>` tag utilised to
display/open the submitted files. Since the `<a>` tag is used to
display/open the file submitted by the students in a modal window and
not on a new tab, the `target='_blank' is unnecessary.
* In the commit 2b5c311 an event handler was added to listen to every
click event in the UI. However, this event handler only added new logic
to the `<a>` tag click event. Unfortunately, the new logic caused the
modal windows to malfunction. It is also important to notice that such
an event handler, which capture every click, may have an impact on
performance. Therefore, we replaced the listener, and now it only
responds to the clicks of the `<a>` tags, which in theory are in charge
of the course navigation.

Fixes #732

# Testing

I tested changes manually, The new code seems to work, but a more exhaustive course navigation should be done in order to verify that the query parameter (?hl=en) is working as intended.

There is no documentation related to these changes

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -667,11 +667,21 @@ $(function() {
  */
 (function (window) {
   $(document).on("click", "a[href]", function (event) {
+    // Remove the language query when the user navigate to the home page or profile page.
+    let keepLangQuery = true;
+    if (
+      event.target.href.replace(/\/$/, "") == window.location.origin ||
+      event.target.href.includes("accounts/accounts")
+    ) {
+      keepLangQuery = false;
+    }
+    // Keep the language query.
     if (
       event.target.className != "dropdown-toggle" &&
       event.target.protocol === window.location.protocol &&
       event.target.host === window.location.host && // hostname:port
-      (event.target.search === "" || event.target.search === "?") // no search
+      (event.target.search === "" || event.target.search === "?") && // no search
+      keepLangQuery
     ) {
       event.target.search = window.location.search; // copy query params
     }
--- a/course/viewbase.py
+++ b/course/viewbase.py
@@ -8,7 +8,7 @@ from django.utils.translation import get_language, get_language_info
 
 from authorization.permissions import ACCESS
 from exercise.cache.content import CachedContent
-from lib.helpers import remove_query_param_from_url, update_url_params
+from lib.helpers import remove_query_param_from_url, update_url_params, switch_lang_query_param_from_url
 from lib.viewbase import BaseTemplateView
 from userprofile.viewbase import UserProfileMixin
 from .cache.students import CachedStudent
@@ -68,13 +68,14 @@ class CourseInstanceBaseMixin(object):
             self.is_teacher = self.course.is_teacher(user)
             self.is_course_staff = self.is_teacher or self.is_assistant
             self.get_taggings = lambda: CachedStudent(instance, user.id).data['tag_slugs']
+            self.url_secondary_query_language = None
             self.url_without_language = remove_query_param_from_url(self.request.get_full_path(), 'hl')
             self.query_language = None
             self.user_language = None
 
             self.note(
                 "course", "instance", "content", "user_course_data", "is_student", "is_assistant",
-                "is_teacher", "is_course_staff", "get_taggings", "url_without_language",
+                "is_teacher", "is_course_staff", "get_taggings", "url_without_language", "url_secondary_query_language",
                 "query_language", "user_language"
             )
 
@@ -98,6 +99,13 @@ class CourseInstanceBaseMixin(object):
                         language = query_language
                         if languages:
                             self.user_language = languages[0]
+                            # Since the prompt to continue browsing the course in the enrollment
+                            # language is presented when the user visit the course using a query
+                            # parameter with a different language, we set the
+                            # self.url_secondary_query_language to the user_language. This allows to
+                            # change the query parameter in the url and fetch the document in the
+                            # correct language.
+                            self.url_secondary_query_language = switch_lang_query_param_from_url(self.request.get_full_path(), 'hl', languages[0])
                             if self.user_language[:2] != query_language[:2]:
                                 self.query_language = query_language
                     else:
--- a/course/templates/course/course_base.html
+++ b/course/templates/course/course_base.html
@@ -82,15 +82,14 @@
                     {% trans "You are browsing this course in a different language than is set in your preferences." %}
                     {% if user_course_data %}
                         <div class="button-container">
-                            <form action="{{ url_without_language }}">
-                                <button type="submit" class="aplus-button--secondary aplus-button--md">
-                                    {% blocktrans trimmed with lang=user_language|language_name_translated %}
-                                    Browse the course in {{ lang }}
-                                    {% endblocktrans %}
-                                </button>
-                            </form>
+                            <a href="{{ url_secondary_query_language }}" class="aplus-button--secondary aplus-button--md">
+                                {% blocktrans trimmed with lang=user_language|language_name_translated %}
+                                Browse the course in {{ lang }}
+                                {% endblocktrans %}
+                            </a>
                             <form action="{{ instance|url:'set-enrollment-language' }}" method="post">
                                 {% csrf_token %}
+                                <input name="next" type="hidden" value="{{ url_secondary_query_language }}" />
                                 <input name="language" type="hidden" value="{{ query_language }}" />
                                 <button type="submit" class="aplus-button--secondary aplus-button--md">
                                     {% blocktrans trimmed with lang=query_language|language_name_translated %}
--- a/course/views.py
+++ b/course/views.py
@@ -245,7 +245,8 @@ class GroupsView(CourseInstanceMixin, BaseFormView):
         return kwargs
 
     def get_success_url(self):
-        return self.instance.get_url('groups')
+        url = self.instance.get_url('groups') + "?hl=" +self.request.GET.get('hl')
+        return url
 
     def form_valid(self, form):
         form.save()
--- a/course/views.py
+++ b/course/views.py
@@ -20,7 +20,7 @@ from django.utils.translation import gettext_lazy as _
 from authorization.permissions import ACCESS
 from exercise.cache.hierarchy import NoSuchContent
 from exercise.models import LearningObject
-from lib.helpers import settings_text, remove_query_param_from_url
+from lib.helpers import settings_text, remove_query_param_from_url, add_lang_query_param_from_url
 from lib.viewbase import BaseTemplateView, BaseRedirectMixin, BaseFormView, BaseView, BaseRedirectView
 from userprofile.viewbase import UserProfileView
 from .forms import GroupsForm, GroupSelectForm
@@ -285,8 +285,9 @@ class LanguageView(CourseInstanceMixin, BaseView):
     
     def post(self, request, *args, **kwargs):
         LANGUAGE_PARAMETER = 'language'
-        
-        next = remove_query_param_from_url(request.POST.get('next', request.GET.get('next')), 'hl')
+        # If the user change the language, the URL will use the hl query parameter
+        # to redirect the resources to the required language
+        next = add_lang_query_param_from_url(request.POST.get('next', request.GET.get('next')), 'hl')
         if ((next or not request.is_ajax()) and
                 not is_safe_url(url=next,
                                 allowed_hosts={request.get_host()}, 
--- a/lib/helpers.py
+++ b/lib/helpers.py
@@ -95,6 +95,32 @@ def remove_query_param_from_url(url, param):
     query.pop(param, None)
     return urlunsplit(url._replace(query=urlencode(query, True)))
 
+def switch_lang_query_param_from_url(url, param, user_language):
+    """
+    Take an url with or without language query parameters. Return url with  an updated query parameter.
+    """
+    url = urlsplit(url)
+    query = parse_qs(url.query, keep_blank_values=True)
+
+    if query.get("hl") is not None:
+        query["hl"] = [user_language]
+
+    return urlunsplit(url._replace(query=urlencode(query, True)))
+
+def add_lang_query_param_from_url(url, param, lang="en"):
+    """
+    Take an url with or without language query parameters. Return url with an updated query parameter.
+    """
+    url = urlsplit(url)
+    query = parse_qs(url.query, keep_blank_values=True)
+
+    if query.get("hl") is not None:
+        if(query.get("hl")[0] ==  "en"):
+            query["hl"] = ["fi"]
+        else:
+            query["hl"] = ["en"]
+
+    return urlunsplit(url._replace(query=urlencode(query, True)))
 
 FILENAME_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ._-0123456789"
 


assets/js/aplus.js
```diff
@@ -666,27 +670,18 @@ $(function() {
...
+  if (
+      event.target.protocol === window.location.protocol
+      && event.target.host === window.location.host // hostname:port
+      && window.location.search.includes('?hl=en') // hl search
+  ) {
+      event.target.search = window.location.search; // copy query params
+  }
+}
+
+$(document).on('click', 'a[href]', addQueryParameter);
```
Mankro, line 683: I would first assume that adding the event handler like this does not work for any link that is added after the initial page load by JS. For example, the exercise descriptions in chapters are loaded with AJAX. They are not part of the initial page load. The original event handler in the whole window or body element works even if the DOM is modified later on.

assets/js/aplus.js
```diff
@@ -666,27 +670,18 @@ $(function() {
...
-            } else {
-                window.location.href = event.target.href; // redirect the current window
-            }
-        }
-    }, false);
+  const addQueryParameter = (event) => {
+  if (
+      event.target.protocol === window.location.protocol
+      && event.target.host === window.location.host // hostname:port
+      && window.location.search.includes('?hl=en') // hl search
```
Mankro, line 677: How could the `includes('?hl=en')` part work reliably? The `hl` query parameter is not always the first parameter, thus the question mark can not be used in the condition. The language could be any language, not always English (en).

assets/js/aplus.js
```diff
@@ -666,27 +670,18 @@ $(function() {
...
+  ) {
+      event.target.search = window.location.search; // copy query params
```
Mankro, line 679: If you use two spaces for indentation like the style guide says, then `event.target.search` should have six spaces before it (function body, arrow function body, if body).

assets/js/aplus.js
```diff
@@ -666,27 +670,18 @@ $(function() {
...
-                    event.target.href,
-                    '_blank',
-                    'location,menubar,resizable,scrollbars,status,noopener');
-            } else {
-                window.location.href = event.target.href; // redirect the current window
-            }
-        }
-    }, false);
+  const addQueryParameter = (event) => {
+  if (
```
Mankro, line 674: The arrow function body should be indented. Add two spaces before the `if`.

assets/js/aplus.js
```diff
@@ -666,27 +670,18 @@ $(function() {
...
+      event.target.protocol === window.location.protocol
+      && event.target.host === window.location.host // hostname:port
+      && window.location.search.includes('?hl=en') // hl search
+  ) {
+      event.target.search = window.location.search; // copy query params
+  }
+}
+
+$(document).on('click', 'a[href]', addQueryParameter);
+
```
Mankro, line 684: Does the new `<a>` event handler let the normal click handler work too? The `_blank` target if branches were removed, but I suppose the standard event handler would take care of it (open the link in a new window or the same window).

exercise/templates/exercise/_file_link.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n %}
 {% if file.exists %}
-	<a href="{{ file.get_absolute_url }}"{% if not file.is_passed %} class="file-modal"{% endif %} target="_blank">{{ file.filename }}</a>
+	<a href="{{ file.get_absolute_url }}"{% if not file.is_passed %} class="file-modal"{% endif %}>{{ file.filename }}</a>
```
Mankro, line 3: Hmm, the `target _blank` attribute probably didn't cause issues in the past because the file modal jQuery plugin overrode the click event handler and target `_blank` was ignored. The language hl click event handler changed the event handlers and the order of them. It probably overrode the file modal handler completely, which is clearly a problem in the language handler.

Mankro (changes requested):
The language hl click event handler from the summer 2020 has issues that must be fixed and this PR verifies that. However, I think this PR has some issues too. See the comments below.

assets/js/aplus.js
Mankro: Actually, the event handler is still attached to the document and it should work even after new `<a>` elements are added to the DOM. I somehow misread this. Is there any difference between the **window** and the **document** in event handlers?

assets/js/aplus.js
jaguarfi: This event handler does not overwrite any other event handler. This is basically one event handler on top of the existing ones.

exercise/templates/exercise/_file_link.html
jaguarfi: Indeed, The attribute `target=_blank` did not cause any issues in the past but it may be good to make it clear in the code that this link will not open a new tab #716

assets/js/aplus.js
jaguarfi: It is true, perhaps we can use regex.

jaguarfi:
This PR is more like a quick fix for #745. I think that we must take the following approach to manage multilingual URLs https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#language-prefix-in-url-patterns. Currently, adding the hl parameter to the URL seems a bit awkward and may cause some unwanted behaviours. Instead, adding the prefix in the sever side will be cleaner, and the navigation between content will be also more clear for the user.

exercise/templates/exercise/_file_link.html
Mankro: Some browsers could have JavaScript disabled in which case we can not depend on JS to fix the behaviour. In those cases, it would make sense to let the browser open the file in a new tab. Progressive Enhancement also suggests that web pages should remain functional at least for the basic needs even without JS. Accessible technology in particular might disable JS.

In this case, the target attribute has a very small impact. Opening files in new tabs makes some sense with disabled JS but it is not strictly required. What do you think?

https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement

Mankro:
> This PR is more like a quick fix for #745. I think that we must take the following approach to manage multilingual URLs https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#language-prefix-in-url-patterns. Currently, adding the hl parameter to the URL seems a bit awkward and may cause some unwanted behaviours. Instead, adding the prefix in the sever side will be cleaner, and the navigation between content will be also more clear for the user.

Jaakko @raphendyr specifically did not want to use the language URLs implemented in Django. I am not sure of his reasons, but we spent a lot of effort in the summer in implementing the language URLs with the query parameters. It is too late now to change the whole architecture at this point. At least it would waste all the previous efforts. It would be nice to know why Jaakko was so certain that the existing Django implementation (https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#language-prefix-in-url-patterns) would not work in A+.

assets/js/aplus.js
jaguarfi: It shouldn't be any difference between **window** and **document**. **document** is just a child of the window object

jaguarfi:
Ok, so we need to start analysing the cases where the URL change. So far, I have seen problems with the submit buttons inside the forms.

exercise/templates/exercise/_file_link.html
jaguarfi: I think we should focus on the browsers we support, such as Firefox, and Chrome. Actually, I am not sure if we mention somewhere the browsers we support. But, if we target A+ to Firefox and Chrome we should have no problem with Javascript. We can also add the following script to the **base.html** template
```html
<noscript>
 For full functionality of this site it is necessary to enable JavaScript.
 Here are the <a href="https://www.enable-javascript.com/">
 instructions how to enable JavaScript in your web browser</a>.
</noscript>
```

assets/js/aplus.js
```diff
@@ -665,28 +669,22 @@ $(function() {
...
-            } else {
-                window.location.href = event.target.href; // redirect the current window
-            }
-        }
-    }, false);
+(function (window) {
+  // The following regex takes the language values from the LANGUAGES variable in the aplus/settings.py, i.e., en and fi.
+  // Therefore, the values must be adjusted accordingly.
+  const regexp = /\?+.*hl=(en|fi)+/gi;
+  const addQueryParameter = (event) => {
```
raphendyr, line 676: Pointing out to remember that IE 11 does not support arrow functions. From style point of view, I don't see reason to use one here either. Default version is even a bit shorter.

```javascript
const addQueryParameter = (event) => {
function addQueryParameter(event) {
```

assets/js/aplus.js
```diff
@@ -665,28 +669,22 @@ $(function() {
...
+    if (
+      event.target.protocol === window.location.protocol &&
+      event.target.host === window.location.host && // hostname:port
+      window.location.search.match(regexp) // hl search
+    ) {
+      event.target.search = window.location.search; // copy query params
```
raphendyr, line 682: It might be a better idea to copy only the `hl` parameter here, instead of replacing all query parameters. [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) might help with that (note, no IE 11 support, but there is a polyfill of course).

assets/js/aplus.js
```diff
@@ -665,28 +669,22 @@ $(function() {
...
  * Note! Translation requires that at least 'hl' query parameter is copied
  */
-(function(window) {
-    window.addEventListener('click', function(event) {
-        if (
-            event.target.nodeName === 'A'
-            && event.target.href
-            && event.target.protocol === window.location.protocol
-            && event.target.host === window.location.host // hostname:port
-            && (event.target.search === '' || event.target.search === '?') // no search
```
raphendyr, line 675: Was this line supposed to be removed? It tested if the `href` had a search string already.

assets/js/aplus.js
```diff
@@ -665,28 +669,22 @@ $(function() {
...
+      event.target.protocol === window.location.protocol &&
+      event.target.host === window.location.host && // hostname:port
+      window.location.search.match(regexp) // hl search
+    ) {
+      event.target.search = window.location.search; // copy query params
+    }
+  };
+
+  // Add event to all the <a> tags with a href attribute.
+  $(document).on("click", "a[href]", addQueryParameter);
```
raphendyr, line 687: To my understanding `jquery.on(event, filter)` is identical to old implementation in performance point of view. Both should capture all click events and do check if the filter match. In old case, we checked if the target object is `a` tag and in new one there is a jquery filter. In a new implementation, our code is not executed for every click, but jquery is. It's possible that I'm wrong and web browsers have an API to do the filtering before JS code.

assets/js/aplus.js
```diff
@@ -665,28 +669,22 @@ $(function() {
...
-                    'location,menubar,resizable,scrollbars,status,noopener');
-            } else {
-                window.location.href = event.target.href; // redirect the current window
-            }
-        }
-    }, false);
+(function (window) {
+  // The following regex takes the language values from the LANGUAGES variable in the aplus/settings.py, i.e., en and fi.
+  // Therefore, the values must be adjusted accordingly.
+  const regexp = /\?+.*hl=(en|fi)+/gi;
```
raphendyr, line 675: I think something like `/(?|&)hl=[a-zA-z-](&|$)/gi` is needed here (i.e., starting with `?` or `&` and ending with `&` or end of stting.

raphendyr:
> Jaakko @raphendyr specifically did not want to use the language URLs implemented in Django. -- It would be nice to know why Jaakko was so certain that the existing [Django implementation](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#language-prefix-in-url-patterns) would not work in A+.

I think the main factor was to follow the design where the language is not part of the resource. Of course, as it has been found out, the language _should_ be part of the resource identifier (e.g. URL). Originally, it seemed easier to calculate points no mather the language the exercise was done with (e.g. do 2 weeks in English, then switch to Finnish).

Another, I think I didn't want to change how request routing works as there is already a lot of complications in the URL patterns. Remember that all urls start with a variable, the course key, thus adding another variable to the same spot could result complication. Of course, we could first check if the course name is a language code and if not, then use it as a course key. However, it likely would require a custom implementation of the middleware.

Finally, during the development, it became clear that we needed a custom version of the language selection subsytem in both locations, in the middleware and the view class. I don't remember the exacts, but at least the default language is selected differently than in the Django code as we don't use the language cookie anymore.

> I think that we must take the following approach to manage multilingual URLs https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#language-prefix-in-url-patterns. Currently, adding the hl parameter to the URL seems a bit awkward and may cause some unwanted behaviours. 

Of course, that is an option, but I don't personally see good reasons for it. I have personally always been againts defining the language this way, mainly because it's ugly, but also because language typically is not part of the resource. In lthe case of learning platforms, this most likely is not the case (i.e. language is a important part to identify the resource). Also, in many cases there are cleaner ways to define language. One of my favorites was to use language specific URLs (e.g. `/homepage` and `/etusivu`). One cool example is espoo.fi vs. esbo.fi . However, they also append the locale string to the path (note `sv-FI` for Swedish in Finland. Though, `en-US` for English and not `en-FI`).

Adding the hl in click action _seemed_ the easiest way to do this, without breaking most of the existing code. Better option would have been to edit templates and add the parameter on the server side when needed. This would have been a better option as it would have worked also when javascript is disabled.

> Instead, adding the prefix in the sever side will be cleaner, and the navigation between content will be also more clear for the user.

Same can be done for the query parameters too. However, there are so many locations, that fixing them all is a lot of work. It could be possible to rewrite the Django's reverse template tag to do this. MOOC-Jutut has [a code](https://github.com/Aalto-LeTech/mooc-jutut/search?q=updated_query) in the direction of the first case.

In relation to this, I think the language thing is one of the major areas for design in the next generation of a platform or such. Specially, how points (progress), urls and languages are linked to each other.

Current problems didn't become apparent to me, but maybe move comments in the code help a bit. Feel free to ask me more.

jsorva:
Hi! I've just become aware of this thread. I didn't have the time to read all of the above now (sorry, attention is a limited resource and all that), but I'd just like to *strongly* agree with the point that language should be part of the URL and different-language versions of the same material must have different URLs. So any efforts towards that goal have a very large thumbs-up from me, even if there are some bumps along the road.

Mankro:
> Hi! I've just become aware of this thread. I didn't have the time to read all of the above now (sorry, attention is a limited resource and all that), but I'd just like to _strongly_ agree with the point that language should be part of the URL and different-language versions of the same material must have different URLs. So any efforts towards that goal have a very big thumbs-up from me, even if there are some bumps along the road.

@jsorva yes, this has been implemented and deployed to production. The language can be added to the URL in the query parameters.
Example:
https://plus.cs.aalto.fi/accounts/privacy-notice/?hl=en
https://plus.cs.aalto.fi/accounts/privacy-notice/?hl=fi

You can do that in your course materials. But the JavaScript has bugs that must be urgently fixed.

jsorva:
> @jsorva yes, this has been implemented and deployed to production. The language can be added to the URL in the query parameters.

Great! Looking forward to using it.

assets/js/aplus.js
jaguarfi: I think that we should not support IE at all. Even Microsoft is going to stop supporting that browser. https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666

Again,  I think we should state somewhere the browsers we support.

assets/js/aplus.js
jaguarfi: My idea was to use the `jquery.on()` function to create a delegate event handler. 

>Delegated event handlers have the advantage that they can process events from descendant elements that are added to the document at a later time.

https://api.jquery.com/on/#direct-and-delegated-events

jaguarfi:
@raphendyr message explains well why we took the ´hl´ approach. Anycase, I feel the responsibility of sharing some of the Google recommendations. https://developers.google.com/search/docs/advanced/crawling/managing-multi-regional-sites#locale-specific-urls.

**Now**, regarding this PR, I will take a different approach. The problem here is not related to the ´<a>´ tags and to the use of those tags in the course navigation. We actually care about the document language and the ´location.search´ parameters. Therefore, I will implement a script that takes these two things into account.

Mankro:
Does it help to reorder the JS click event handlers? The event handlers are executed in a certain order and it depends on the order the handlers are attached. The order could probably be reset. There are many jQuery plugins, thus we can't really know in which order the plugins are executed in each page, but one handler could always force a certain ordering for the handlers. Should the hl param handler be always the first?

raphendyr:
> @raphendyr message explains well why we took the ´hl´ approach. Anycase, I feel the responsibility of sharing some of the Google recommendations. https://developers.google.com/search/docs/advanced/crawling/managing-multi-regional-sites#locale-specific-urls.

Yes, those are good points of geotargeting. In A+ case, the language is expected to be stored per user on a course (based on original design, don't copy to a new one), hence the language query is extra and less often seen feature, which is not supposed to be obvious for the user (in contrast to geotargeting).

assets/js/aplus.js
```diff
@@ -665,28 +669,36 @@ $(function() {
...
-(function(window) {
-    window.addEventListener('click', function(event) {
-        if (
-            event.target.nodeName === 'A'
-            && event.target.href
-            && event.target.protocol === window.location.protocol
-            && event.target.host === window.location.host // hostname:port
-            && (event.target.search === '' || event.target.search === '?') // no search
-        ) {
-            event.preventDefault(); // stop default redirect
```
raphendyr, line 677: I'm pretty sure this was the line that broke every other JS stuff. Assuming this handler was called fist, then the event stopped here and rest of the handlers were not called. It might be enough not to do the redirect below (`window.location.href = event.target.href`), but keep the editing of the link (`event.target.search = ...`), thus the link content should be edited.

If that doesn't work, because it's possible that the event.target.href is already read into variable, and thus editing it now doesn't change default click action), then I think we can somehow edit the information within the event. In other words, the event handlers could be like this:

1. this function: add `hl=` parameter to the link if some condiftions apply
2. any other event handlers, which may end the event (result: some other action is taken, but the link href will contain `hl=`
3. default click handler in the browser: this should now read the updated event.target.href and redirect with the `hl=`

jaguarfi: It seems like we want to keep the same logic and code :). But, I already mention some of the problems above. I will use the jquery `on()` method because I am sure that delegation is something we want to have. Besides, the piece of code that assesses the `target="_blank" is also causing some unwanted behaviours in the modal windows.

raphendyr:
On the language in the url topic. I was thinking that pattern `/<course>/<instance>/<lang>/<module>/<exercise>` might be most suitable. I f I remember right, part `/<lang>/<module>/<exercise>` is passed to our custom router, thus allowing us to handle it easily. Furthermore, courses with a single language could ignore the `<lang>` element.

jaguarfi:
Could @Mankro comment on this PR? and I will also appreciate if @raphendyr can also provide some feedback about the functionality and use cases. I am trying to cover several uses cases here and I do not know if I am covering all of them.

Mankro:
Very interesting comments, thank you! It will take some days before I am able to focus on this better. I was thinking that we could deploy a simple hotfix that simply removes the new JS click event handler for the hl parameter. Then, everything works and the links in the page will simply not have the hl parameter, which may change the language when the user opens links. However, that is a minor issue. This hotfix would be a **temporary** fast, simple solution while we are working on more sophisticated solutions. @jaguarfi what do you think and can you make a new pull request about this idea (only remove the problematic hl click event handler)?

assets/js/aplus.js
```diff
@@ -666,27 +670,30 @@ $(function() {
...
-                    'location,menubar,resizable,scrollbars,status,noopener');
-            } else {
-                window.location.href = event.target.href; // redirect the current window
-            }
         }
-    }, false);
+    });
+
+    if (!window.location.search) {
+      if(document.documentElement.lang == "en-gb" || document.documentElement.lang == "en")
```
Mankro, line 688: Is it necessary to check the documentElement's lang attribute here? The original idea was to copy to the hl query parameter only if the current location (URL) contains the hl parameter. It can be copied from the location, thus it doesn't matter what values the document lang attribute has.

The issue with the document lang attribute can be fixed (https://github.com/apluslms/a-plus/issues/748), but I don't think it is necessary in this PR.

course/viewbase.py
```diff
@@ -98,6 +99,11 @@ def get_resource_objects(self):
                         language = query_language
                         if languages:
                             self.user_language = languages[0]
+                            # Since the option to continue browsing the course in
+                            # enrollment language is presented in the case the user navigates
+                            # the course with a secondary language, it makes sense to set the
+                            # self.url_secondary_query_language to the user_language
+                            self.url_secondary_query_language = switch_lang_query_param_from_url(self.request.get_full_path(), 'hl', languages[0])
```
Mankro, line 106: I don't understand yet why you are adding these secondary language parameters. I would prefer a simple solution instead of adding more hacks. We can talk in more detail soon.

jaguarfi: It is important to handle the language query parameters in the POST requests. Therefore, we should always redirect the user to the corresponding URL and add the query parameter. Otherwise, the query parameter is deleted, and the URL to which the user is redirected will not include the `hl` parameter.  Currently, we are handling the navigation through the course, assuming that everything is a GET request (basically, that was the reason to add the JS event handler), but we should also think about the management of the POST request. In this PR, I added a third commit that shows an example of how a POST request generated when the user tries to create a group can break the management of the `hl` parameter in the URL.

jaguarfi:
@Mankro feel free to review this PR. I will be able to slowly apply the required changes.

Mankro:
I am closing this PR because I think it approaches the problem from a wrong angle and I don't think this will work correctly. There are also programming style issues, such as hardcoding fi and en language codes in the code. That wouldn't work when we add support for new languages.

I have recently made small fixes in the hl click event handler: the commit https://github.com/apluslms/a-plus/commit/73a3e667ccab1ed958314e381149005b9e92f410 and a few commits before that.

PR #738, opened 2020-12-02
Enable editing teachers from course details

### LINKED ISSUE #700: Teacher should be able to change the teachers of the course
Currently, only A+ admins may change the teachers of a course (model `Course` defines the teachers for all `CourseInstances` of the course). However, this has been an unnecessary hindrance that causes extra manual work and it is annoying to contact admins for such a trivial task. A teacher should be able to add and remove other teachers in the course himself/herself. It should be even possible to remove oneself as the teacher, after which one can no longer access the edit course page.

The edit course form already displays the teachers of the course. It can be changed so that the teachers are edited there. The assistants are already edited in the same form.

sayravai:
# Description

The teachers of a course should be editable via the course instance details page by the teachers so, that they don't need admin assistance whenever teachers need to be edited. However, teachers are defined for a course, not a course instance, so they cannot be simply saved for the course instance object.

Fixes #700

This pr has only been tested by hand so far, as it doesn't work as it should. Saving new teachers works ok, but for some reason the select multiple list item does not list the existing teachers as **selected** on page load, so the jQuery plugin does not convert them into buttons like it does for the assistants. Also due to this, the existing teachers do not get submitted via the form unless manually added via the search field. I assume this is because some user id's don't get matched at Django side, but haven't been able to pinpoint the actual problem.

### DIFF
--- a/edit_course/course_forms.py
+++ b/edit_course/course_forms.py
@@ -74,6 +74,10 @@ class CourseInstanceForm(forms.ModelForm):
         initial_queryset=UserProfile.objects.none(),
         required=False) # Not required because a course does not have to have any assistants.
 
+    teachers = UsersSearchSelectField(queryset=UserProfile.objects.all(),
+        initial_queryset=UserProfile.objects.none(),
+        required=False) # Not required because a course does not have to have any teachers.
+
     class Meta:
         model = CourseInstance
         fields = [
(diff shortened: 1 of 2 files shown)


Mankro:
@sayravai I found the problem. The initial_queryset is working and the hidden select element has the current teachers in the HTML form output. However, those option elements do not have the selected attributes. The JavaScript is activated by the selected attributes and you can see that the assistants field has those attributes in the hidden select element. 

https://github.com/apluslms/a-plus/blob/24bd64eae94923a95729a8793bdf6a8babe49b61/assets/js/ajax_search_select.js#L75
https://github.com/apluslms/a-plus/blob/24bd64eae94923a95729a8793bdf6a8babe49b61/lib/fields.py#L92

The teachers field is a ManyToManyField to UserProfile like the assistants field.
https://github.com/apluslms/a-plus/blob/24bd64eae94923a95729a8793bdf6a8babe49b61/course/models.py#L56

The teachers field is not part of the CourseInstance model. The CourseInstanceForm is a ModelForm for the CourseInstance model. This may be the reason why the form widget does not render the selected attributes for teachers. It looks difficult to modify the widget's HTML output since that is part of Django, not a-plus code. It is much easier to create a new ModelForm for the Course model and modify teachers with that form. (Assuming that I am correct about the problem.) If we create a new form for modifying teachers, we might need to add a new page/tab to the A+ edit course section. I think it is more important to implement this feature than to worry about separating the teachers from the other course settings. It may look a little dirty with multiple tabs under the edit course section, but it will work just fine.

Mankro:
The new PR #803 replaced this one.

PR #764, opened 2021-01-14
Hotfix january modal window and event handler for the <a> tags

### LINKED ISSUE #745: The JavaScript for copying the hl language parameter copies some unwanted parameters too
A+ has new JavaScript that copies GET query parameters from the current page's URL into links. The links are modified in the click event handler. This is intended for the `hl=en` parameter so that the same language is kept in the URLs and the pages do not change the language when you open links.

https://github.com/apluslms/a-plus/blob/57031587030092ac4c482fde53ff9d9577da0f0f/assets/js/aplus.js#L669-L689

However, there are cases when this behaviour causes issues. There are pages that depend on GET query parameters in the URL, but the parameters should not be copied to all link targets.

* When you are unauthenticated and you open a page that requires authentication, you are redirected to the login page. The URL includes the `next` query parameter so that you could be redirected back to the target page after logging in. However, the next parameter should not be copied to any link targets. If you open the front page from the top navbar from the login page, the `next` parameter stays in the URL for any link you open.

jaguarfi:
# Description

This is a hotfix to solve some of the issues that have been caused for the JavaScript code added in the commit 2b5c311. However, I recommend working in the Pull Request #736 in order to allow the users to navigate the course content while using the `?hl=` query parameter.

Fixe #732
Fixes #745

# Testing

I tested the changes manually. I tested with some chapters of the O1 course and also the aplus manual.

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -665,28 +665,16 @@ $(function() {
  * to the next
  * Note! Translation requires that at least 'hl' query parameter is copied
  */
-(function(window) {
-    window.addEventListener('click', function(event) {
-        if (
-            event.target.nodeName === 'A'
-            && event.target.href
-            && event.target.protocol === window.location.protocol
-            && event.target.host === window.location.host // hostname:port
-            && (event.target.search === '' || event.target.search === '?') // no search
-        ) {
-            event.preventDefault(); // stop default redirect
-            event.target.search = window.location.search; // copy query params
-            if (event.target.getAttribute('target') === '_blank') {
-                // Open a new window or tab.
-                window.open(
-                    event.target.href,
-                    '_blank',
-                    'location,menubar,resizable,scrollbars,status,noopener');
-            } else {
-                window.location.href = event.target.href; // redirect the current window
-            }
-        }
-    }, false);
-})(window);
+(function (window) {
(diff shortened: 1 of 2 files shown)


assets/js/aplus.js
```diff
@@ -665,28 +665,16 @@ $(function() {
...
-})(window);
+(function (window) {
+  $(document).on("click", "a[href]", function (event) {
+    if (
+      event.target.className != "dropdown-toggle" &&
+      event.target.protocol === window.location.protocol &&
+      event.target.host === window.location.host && // hostname:port
+      (event.target.search === "" || event.target.search === "?") // no search
+    ) {
+      event.target.search = window.location.search; // copy query params
```
Mankro, line 676: I think this copy of the search query can be changed so that it only copies the hl parameter. If `window.location.search` has other parameters, they should not be copied at all.

Since this copies all parameters, the issue https://github.com/apluslms/a-plus/issues/745 has not been resolved. You actually close that issue in the commit message, but it is not correct. The problem clearly still exists in the login page as was described in the issue https://github.com/apluslms/a-plus/issues/745.

Mankro (approved):
Thanks, I will merge this. However, the issue https://github.com/apluslms/a-plus/issues/745 has not been resolved even though you close it in the commit message.

Issue #602, opened 2020-06-29
Add roles to enrollments


Mankro:
Currently, the Course model has a many-to-many field to UserProfile for setting the teachers of the course and the CourseInstance model has a many-to-many field to UserProfile for setting the assistants. 

Separating students from the course staff has been problematic. For example, the course points page mixes assistants with the students: https://github.com/apluslms/a-plus/issues/540

There are two approaches to this problem.

1. We don't change the database schema and try to filter teachers and assistants out from the course points page and probably the participants page too.
2. We change the Enrollment model so that it includes a role of the user: student, assistant or teacher. It would make sense to enroll all users in the course, but separate their roles. Currently, teachers should not enroll so that they are not included in the course points page. Enrollment is also needed for storing the pseudonymized name of the user that may be used with, e.g., LTI connections. The old teacher and assistants fields in the Course and CourseInstance fields would be removed.
This approach enables adding new roles in the future, for example, a role for teaching support personnel that are not part of the course staff, but they need to access the course data without having full admin access to the system.

I wrote this issue in favor of alternative 2, but before we commit to it, we should plan ahead. How much would old code have to be changed for alternative 2? At least some user API endpoints are affected, as well as the course points page and the course participants page. All references to assistants, teachers, and enrollments should be checked.

Mankro:
I am first assigning this to @veeraalt . If there is no time for this in July, then @HenrikSundell could step in.

raphendyr:
Someone should consider if we could include "removed" as a role or should it be a separate boolean. Removing teacher/assistant could drop role first to student. Toggling between student and removed should be easy for the person it self (unless we want to support banning).

Mankro:
@raphendyr you mean that instead of deleting enrollments, the enrollment is just flagged as "deleted"? That seems to be the right way to go when we don't want to immediately delete all of the user data in the course. Moodle also has a "deleted" flag for this purpose instead of immediately deleting the data. This also means that when we check for the existence of the enrollment, it is **crucial to check the deleted flag**.

Based on my gut feeling, I think that using "removed" or "deleted" as the role is strange. It also loses the information about the role before deletion. Adding a separate boolean deleted field is simple, but it is worth exploring the role idea.

When we support the removal of enrollments like this, it is essential to enable teachers to remove/unenrol students from the course (so yes, we want to support banning).

raphendyr:
[...]

PR #766, opened 2021-01-19
Change enrollment checks to use QuerySet get method.

### LINKED ISSUE #600: Refactor enrollment checks to use the uniqueness constraint
After the PR https://github.com/apluslms/a-plus/pull/599, enrollments are guaranteed to be unique in a course instance. Therefore, it is not necessary to assume that there could be multiple enrollments in this way: `enrollment = Enrollment.objects.filter(course_instance=course_instance, user_profile=user.userprofile).first()`,
https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/external_services/lti.py#L96
Lines like that could be refactored if there is a cleaner way. In lti.py, the QuerySet method `get` could be used with exception handling if the enrollment does not exist.

PasiSa:
Relies on the uniqueness of enrollments based on (course_instance, user_profile).
Requires exception handling for nonexisting entries.

Fixes #600.

# Description

Getting familiar with PR process by taking an old easy issue for fixing.
Note two inline question comments in code (i.e., this is not final PR):
1) forms.py is not affected by this issue (different query attributes), right?
2) In lti.py, should we still raise PermissionDenied exception in negative case or is the default raised by get method ok (ObjectDoesNotExist)? I couldn't yet figure out how to test lti.py

I think these were the only places in code affected by this issue (did a recursive search for enrollment using grep). Am I correct?

Fixes #600

# Testing

Tested models.py in both succesful and nonexisting case.
Couldn't yet figure out how to test lti.py.
No unit tests modified.

None relevant, I think.

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -539,7 +539,10 @@ class CourseInstance(UrlMixin, models.Model):
         UserTagging.objects.create(tag=tag, user=user.userprofile, course_instance=self)
 
     def get_enrollment_for(self, user):
-        return Enrollment.objects.filter(course_instance=self, user_profile=user.userprofile).first()
+        try:
+            return Enrollment.objects.get(course_instance=self, user_profile=user.userprofile)
+        except Enrollment.DoesNotExist:
+            return None
 
     def get_user_tags(self, user):
         return self.taggings.filter(user=user.uesrprofile).select_related('tag')
--- a/external_services/lti.py
+++ b/external_services/lti.py
@@ -2,8 +2,8 @@ from hashlib import md5
 from urllib.parse import urlsplit, urljoin
 
 from django.conf import settings
-from django.core.exceptions import PermissionDenied
-from django.utils.translation import get_language
+from django.http import Http404
+from django.utils.translation import get_language, ugettext_lazy as _
 from rest_framework.reverse import reverse
 from rest_framework.settings import api_settings
 from oauthlib.common import urldecode
@@ -93,9 +93,10 @@ class LTIRequest(object):
     def user_info(self, course_instance, user):
         if self.service.is_anonymous:
             # Anonymize user information
-            enrollment = Enrollment.objects.filter(course_instance=course_instance, user_profile=user.userprofile).first()
-            if not enrollment:
-                raise PermissionDenied()
+            try:
+                enrollment = Enrollment.objects.get(course_instance=course_instance, user_profile=user.userprofile)
+            except Enrollment.DoesNotExist:
+                raise Http404(_("Course enrollment required for accessing the LTI service."))
             # Creates anon name and id for pre-pseudonymisation Enrollments
             if not (enrollment.anon_name or enrollment.anon_id):
                 # the model's post_save functions take care of the creation
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -2908,6 +2908,10 @@ msgstr ""
 "Palauttaaksesi tehtäviä sinun pitää rekisteröityä ja ilmoittautua kurssin "
 "etusivulla."
 
+#: external_services/lti.py
+msgid "Course enrollment required for accessing the LTI service."
+msgstr "Kurssille ilmoittautuminen vaaditaan LTI-palveluun pääsemiseksi."
+
 #: external_services/models.py
 msgid "Url can not contain scheme or domain part."
 msgstr "Url ei voi sisältää skeema- tai verkkotunnusosaa."


course/models.py
```diff
@@ -539,7 +539,10 @@ def tag_user(self, user, tag):
         UserTagging.objects.create(tag=tag, user=user.userprofile, course_instance=self)
 
     def get_enrollment_for(self, user):
-        return Enrollment.objects.filter(course_instance=self, user_profile=user.userprofile).first()
+        try:
+            return Enrollment.objects.get(course_instance=self, user_profile=user.userprofile)
+        except ObjectDoesNotExist:
```
raphendyr, line 544: Did you consider `Enrollment.DoesNotExist`?

course/forms.py
```diff
@@ -32,6 +32,7 @@ def clean(self):
         self.member_profiles = [self.profile]
         for key in self.fields.keys():
             if key in self.cleaned_data and self.cleaned_data[key]:
+                # PS: I guess this is not covered by issue #600
```
Mankro, line 35: I suppose this comment can be removed from the final version.

external_services/lti.py
```diff
@@ -93,9 +93,8 @@ def __init__(self, service, user, instance, request, title, context_id=None, lin
     def user_info(self, course_instance, user):
         if self.service.is_anonymous:
             # Anonymize user information
-            enrollment = Enrollment.objects.filter(course_instance=course_instance, user_profile=user.userprofile).first()
-            if not enrollment:
-                raise PermissionDenied()
+            # PS: is the default get exception ok, or should we raise PermissionDenied?
+            enrollment = Enrollment.objects.get(course_instance=course_instance, user_profile=user.userprofile)
```
Mankro, line 97: I would first guess that the exception directly affects the HTTP response status code. PermissionDenied leads to 403 Forbidden and other errors to 500 Internal Server Error. If this is the case, then we would want to keep 403 here.

course/models.py
Mankro: Yes, it is absolutely the best practice to use the DoesNotExist attribute from the target model class. That should be clearly highlighted in the official Django docs on making basic database queries.
https://docs.djangoproject.com/en/2.2/topics/db/queries/#retrieving-a-single-object-with-get

Mankro (changes requested):
Thus far it is looking good. There are some comments.

external_services/lti.py
raphendyr: The `DoesNotExist` should yield 404, so it's choice between 403 and 404. I don't remember reasoning why 403 is used here. I tried to check git history, but there were no details. PR could include more details.

On a side note, I'm not sure how common `PS` is as a tag. I have used `FIXME` for things that shouldn't get merged/pushed, `TODO` for things that _should_ have an issue (or they have). `NOTE` to highlight things that are not apparent from the source and to make the comment more important than just comment without any tag. In this respect, I would write `FIXME` or `TODO` here to indicate that this part requries rework before merging. And this is just my personal opinion.

Mankro: Oh yes of course, if an object does not exist, the error must be HTTP 404 Not Found. Using 404 in this case could make sense since the problem really is that the enrollment is not found. This is typical for teachers and LTI services are the only reason why teachers need to enroll at the moment. The course points page does not separate teachers from students if everyone is enrolled, but that is another issue. Therefore, we could use the 404 response here unless we find a reason why PermissionDenied 403 has been used in the past.

Adding roles to enrollments so that teachers could always be enrolled:
https://github.com/apluslms/a-plus/issues/602

course/forms.py
PasiSa: yes

course/models.py
PasiSa: Thanks, will fix.

external_services/lti.py
PasiSa: So if I leave it as is, raising DoesNotExist would cause HTTP 404 by the framework? Or should I do something else to cause 404 response?

Mankro: I think we shall switch to the 404 error in this case. It would be best if the user can still get a clear error message instead of a very generic "404 not found". The message should say that the enrollment is missing and is required to access the LTI service. A+ already has custom error messages in at least 403 Forbidden pages, so I guess it is easy to do it in the 404 page too. Maybe it was done with the `messages` module, but I don't remember exactly.

This is one example of a customized 404 error message, but it is defined in a Permission class and this LTI code does not use permissions.
https://github.com/apluslms/a-plus/blob/24bd64eae94923a95729a8793bdf6a8babe49b61/course/permissions.py#L57

Django has the shortcut `get_object_or_404` that could be used assuming that you can modify the error message with it. At any rate, there is also the exception `Http404` that can probably be raised with a message string. Note that this message to the user shall be translated to Finnish too. (There is the function `_("hello message")` for translated strings.) Our A+ orientation wiki and the a-plus/doc/README.md also explain how to add translations (to the django.po file).

https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#get-object-or-404
https://docs.djangoproject.com/en/2.2/topics/i18n/translation/

PasiSa:
Comments are now addressed, I believe. Now also successful and failed lti.py cases are tested.

Mankro (approved):
Great job!

Mankro:
I will now fix the conflict in the PO file myself and then I can merge this. FYI, I will rebase the branch onto master, fix the git conflict and force-push to your branch so that it updates this pull request.

PR #772, opened 2021-01-26
Add instances page and redirect to the last course

jaguarfi:
# Description

**What?**

I added a view that lists all the course instances of any particular course. In addition, I  implement a redirect feature, thus teachers can share a standard URL that will always redirect to the last instance of their course.

**Why?**

The need started from the O1 course, but we think it may benefit multiple courses.

**How?**

I also added a new URL https://SERVER/course_url/instances. This URL query all the course instances and show them organized by year.

I add a new URL pattern that takes the https://SERVER/course_url/ URL and redirect it to the last instance, based on the starting date.

Fixes #578

# Testing

**Remember to add or update unit tests for new features and changes.**

* How to [test your changes in A-plus](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations)
* How to [test accessibility](https://wiki.aalto.fi/display/EDIT/How+to+check+the+accessibility+of+pull+requests)

**What type of test did you run?**

- [X] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Other test. *(Add a description below)*
- [X] Manual testing.

* I tested that the URL that displays all course instances was listing all the instances of the course.
* I tested that the user was redirected to the last instance of the course when the course instance is not specified in the URL. 

---
1. The course EXISTS. 
    
    CONDITIONS
    a. The course URL is localhost:8000/def
    b. The course URL for showing the course instances is localhost:8000/def/instances

    i. At least one instance of the course exists, and it is visible to students. 
        
        RESULTS:
        a. The user is redirected to the existing instance
        b. The instances page is displayed.
    ii. At least one instance  of the course exists, but the course instance is hidden from students

        RESULT:
        a. 403 Forbidden Error
        b. The instances page is displayed with the following message "The course instances of this course 
        are not visible to students."
    iii. There is more than one course instance, and all instances are visible to students.

        RESULT:
        a. The student is redirected to the last instance of the course
        b. All the course instances are listed
    iv. There is more than one course instance, but the last course instance is hidden.

        RESULT:
        a. 403 Forbidden Error
        b. Only the visible course intances are displayed
    v. There is more than one course instance, but some of them are hidden, not the last one.

        RESULT:
        a. The student is redirected to the last instance of the course
        b. Only the visible course intances are displayed
    vi. No course instances have been created

        RESULT:
        a. 404 Page not found Error with the message "There are no course instances for this course."
[...]

PR #784, opened 2021-02-10
Add file with reserved words and validate course instance URL

### LINKED ISSUE #781: Course instance cloning does not validate reserved URL keys
The course instance cloning feature under A+ edit course allows the reserved key/url "instances". Could it use the `CourseInstance.clean()` method for validation? The course instance becomes broken if it uses the reserved value for the key/url.

This issue originates from #772.

jaguarfi:
# Description

**What?**

The Django admin has a list of reserved words that cannot be used in the
course instance URL. This validation needs to be added to the UI that
allows users to clone course instances. 
This PR actually continues PR #772 

**Why?**

No course instance should be named "instances"

**How?**

A simple ´in´ validation was added

Fixes #781

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I validated that new course instances cannot be cloned with the URL "instances"

- [X] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -21,6 +21,7 @@ from django.utils.translation import gettext_lazy as _
 from django_colortag.models import ColorTag
 
 from apps.models import BaseTab, BasePlugin
+from lib.reserved_words import ReservedWordsCourseInstances
 from lib.email_messages import email_course_error
 from lib.fields import PercentField
 from lib.helpers import (
@@ -451,7 +452,7 @@ class CourseInstance(UrlMixin, models.Model):
     def clean(self):
         super().clean()
         errors = {}
-        RESERVED = ("instances",)
+        RESERVED = tuple(item.value for item in ReservedWordsCourseInstances)
         if self.instance_name in RESERVED:
             errors['instance_name'] = format_lazy(_("You cannot use word '{}' as an instance name."), self.instance_name)
         if self.url in RESERVED:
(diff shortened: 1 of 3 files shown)


course/models.py
```diff
@@ -451,7 +452,7 @@ def __str__(self):
     def clean(self):
         super().clean()
         errors = {}
-        RESERVED = ("instances",)
+        RESERVED = tuple(item.value for item in ReservedWordsCourseInstances)
```
Mankro, line 455: I think the reserved keywords should be defined in the `class CourseInstance`. You have now moved the definition to the `lib/reserved_words.py`, but I feel that the reserved words are part of the CourseInstance and they should be defined in the class. It is true that the reason that the word `instances` is reserved is actually part of the URL config, but those URLs are closely related to courses.

When we compare this to the constants `ENROLLMENT_AUDIENCE` etc. at the start of `class CourseInstance`, we can see that those constants are also used in the field definitions in the same class. Thus, it would not make sense to define those constants outside the class. The reserved keywords are only used in the URL config outside the class and the `clean` (validation) method in this class. Nevertheless, since the reserved keywords are used in this class and the keyword is really related to this class, I would define the keywords in the class as a constant. Note that you can refer to `CourseInstance.ENROLLMENT_AUDIENCE.INTERNAL_USERS` anywhere after importing `CourseInstance`.

Note that A+ has a custom `Enum` class that is different from the class in the Python standard library. The A+ Enum includes description strings for the values too.

Since there is only one keyword now, the `AutoName` trick is really not needed and it can be removed.

Mankro (changes requested):
The validation in the course instance cloning view is OK, but I disagree with the Enum definition.

Mankro:
Closing since the issues have not been fixed in this pull request,

PR #809, opened 2021-04-20
Implement new exercise tree endpoint. Fixes #603

### LINKED ISSUE #603: Implement exercise/chapter tree as an endpoint to the API
# Course exercise API endpoint does not sort the exercises

The API endpoint for listing course exercises (http://localhost:8000/api/v2/courses/1/exercises/) does not seem to sort the output. The exercises should be output in the same order as they are listed in the course hierarchy (see the course pages "course materials" or "exercise results"). Note that the API endpoint outputs course modules (exercise rounds) and lists the module's exercises under it. 

Sorting the course modules is easy based on their `order` fields, but sorting the exercises is hard because learning objects can be nested (typically, there are many chapters in a module and each chapter has child exercises). The A+ web pages use the cache for ouputting the course contents and the sorting is done in the cache class (https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/exercise/cache/hierarchy.py, https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/exercise/cache/content.py). Could the API use the same cache for outputting a sorted list of exercises or should the API implement the sorting in another way?

The API endpoint: https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/course/api/views.py#L63
That line refers to CourseModules, but the exercises are retrieved in the serializer: https://github.com/apluslms/a-plus/blob/d34f94b7c9a8e4d22cd28d08968b36787bb344c4/course/api/full_serializers.py#L44

skulonen:
# Description

**What?**

Add new exercise tree endpoint (`/api/v2/courses/{course_id}/tree`) to A+ API. The new endpoint presents the learning objects of a course in a tree-like hierarchy, unlike the regular `exercises` endpoint, which presents them as a flat list.

**Why?**
The flat exercise list does not sort the exercises in the hierarchical order. This new endpoint is also necessary for the IntelliJ plugin.

**How?**

Created a new viewset (`CourseExerciseTreeViewSet`) which accesses the cached content of the course instance, and two new serializers for outputting the data (`TreeCourseModuleSerializer`, `TreeExerciseSerializer`). The viewset doesn't derive from `ReadOnlyModelViewSet` and the serializers don't inherit from `AplusModelSerializer` because they don't operate on model objects. Instead, they operate on `dict`s that the course instance's cached content contains.

The reason for using the cached content is that it is already in a hierarchical structure.

Fixes #603

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the exercise tree was returned correctly in cases where there are multiple modules, where the hierarchy is deeply nested, where a module has no exercises, and other cases.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/api/full_serializers.py
+++ b/course/api/full_serializers.py
@@ -2,6 +2,7 @@ from rest_framework import serializers, exceptions
 
 from lib.api.fields import NestedHyperlinkedIdentityField
 from lib.api.serializers import AplusModelSerializer, NestedHyperlinkedIdentityFieldWithQuery
+from exercise.api.full_serializers import TreeExerciseSerializer
 from exercise.api.serializers import ExerciseBriefSerializer
 from userprofile.api.serializers import UserBriefSerializer, UserListField
 from ..models import (
@@ -21,6 +22,7 @@ __all__ = [
     'CourseStudentGroupSerializer',
     'CourseUsertagSerializer',
     'CourseUsertaggingsSerializer',
+    'TreeCourseModuleSerializer',
 ]
 
 
@@ -52,6 +54,7 @@ class CourseSerializer(CourseBriefSerializer):
     ...
     """
     exercises = NestedHyperlinkedIdentityField(view_name='api:course-exercises-list', format='html')
+    tree = NestedHyperlinkedIdentityField(view_name='api:course-tree-list', format='html')
     points = NestedHyperlinkedIdentityField(view_name='api:course-points-list', format='html')
     students = NestedHyperlinkedIdentityField(view_name='api:course-students-list', format='html')
     taggings = NestedHyperlinkedIdentityField(view_name='api:course-taggings-list', format='html')
@@ -83,6 +86,7 @@ class CourseSerializer(CourseBriefSerializer):
             'visible_to_students',
             # links
             'exercises',
+            'tree',
             'points',
             'students',
             'taggings',
@@ -120,7 +124,7 @@ class CourseUsertagSerializer(CourseUsertagBriefSerializer):
         )
 
     def validate(self, data):
-        # Name is enough, slug will be automatically generated 
+        # Name is enough, slug will be automatically generated
         data = super().validate(data)
         if not data['name']:
             raise serializers.ValidationError("At minimum, field 'name' is required")
@@ -173,7 +177,7 @@ class CourseUsertaggingsSerializer(AplusModelSerializer):
         tag_dict = validated_data['tag']
 
         first_in_required = [ f for f in self._required if f in user_dict ][0]
-        try:    
+        try:
             user = {
                 'id': lambda: UserProfile.objects.get(user__id=user_dict['id']),
                 'student_id': lambda: UserProfile.get_by_student_id(user_dict['student_id']),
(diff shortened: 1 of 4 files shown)


skulonen:
I asked Olli Kinnunen for feedback regarding the structure of the response, and after discussing it with his team, Olli confirmed that it is satisfactory.

course/api/full_serializers.py
```diff
@@ -219,3 +223,16 @@ class Meta(CourseStudentGroupBriefSerializer.Meta):
             }
         }
 
+
+class TreeCourseModuleSerializer(serializers.Serializer):
+    id = serializers.IntegerField()
```
Mankro, line 228: I wonder if we should include other additional `CourseModule` fields in this serializer.

course/api/views.py
```diff
@@ -1,4 +1,4 @@
-from rest_framework import filters, generics, permissions, viewsets, status, mixins
+from rest_framework import filters, generics, permissions, viewsets, status, mixins, views
```
Mankro, line 1: You added import for `views`, but is it actually used anywhere?

course/api/views.py
```diff
@@ -68,6 +68,20 @@ def get_object(self):
         return self.get_member_object('module', 'Exercise module')
 
 
+class CourseExerciseTreeViewSet(CourseResourceMixin,
+                                viewsets.ViewSet):
+    serializer_class = TreeCourseModuleSerializer
+
+    def list(self, request, *args, **kwargs):
+        serializer = self.serializer_class(
+            self.content.data['modules'],
```
Mankro, line 77: Just writing down that `self.content` comes from the super class `CourseResouceMixin - CourseInstanceBaseMixin`. It is the course cache `CachedContent`.

* https://github.com/apluslms/a-plus/blob/46386ac304ebfb8dc00bfd2bffdb7d0d01d6ae8a/course/viewbase.py#L61
* https://github.com/apluslms/a-plus/blob/46386ac304ebfb8dc00bfd2bffdb7d0d01d6ae8a/exercise/cache/content.py#L10

Mankro (approved):
This looks excellent in a quick overview!

* Does the tree include both exercises and chapters? Looks like yes, just double-checking.
* You could add doc strings to the new classes. It is hard to see from the code that the data comes from the cache. The new serializer classes are very dependent on the structure of the cache object.

course/api/views.py
skulonen: It is not, I'll remove it.

course/api/full_serializers.py
skulonen: The IntelliJ team was happy with the fields currently available in the response. But if you think some specific fields should be added, I can do that.

skulonen:
@Mankro I've added the doctrings, removed the unnecessary import and squashed the commits.

>Does the tree include both exercises and chapters? Looks like yes, just double-checking.

Yes. It has to, because exercise can be children of chapters, and things would become really messy if we tried to exclude the chapters but include their child exercises.

Mankro (approved):

PR #805, opened 2021-04-13
Improve the usability of the admin insterface. Fixes #707

skulonen:
# Description

**What?**

Improve the usability of the A+ admin interface by increasing page load performance, adding search functionality and exposing previously hidden fields.

**Why?**

A+ admins waste a lot of time using the admin interface because of its slowness and lacking features.

**How?**

- For every model that references the `UserProfile` model, the referencing field was added to `raw_id_fields` in the corresponding `ModelAdmin` class. This turns the field's input control into a simple textbox with a search functionality, instead of a pre-populated dropdown menu. With large numbers of users, populating the dropdown menu was very slow.
- Added some model fields to `search_fields` for every corresponding `ModelAdmin`. This adds a search box onto the list page of the admin interface, and makes all of those fields searchable.
- Updated the `UserProfile` to include the user's email in the `__str__` method. This causes the user's email to be displayed along with name, username and student ID in the admin interface.
- Removed some list filter options from submissions list sidebar. It listed all users and exercises, which caused the page to load slowly.
- Added `submission_time` to submissions' admin page, and `timestamp` to enrollments' admin page. These were originally hidden because they're read-only fields, but they can be made visible using `ModelAdmin.readonly_fields`.
- Changed all `ModelAdmin` classes to use a consistent style that conforms to the style guide (https://apluslms.github.io/contribute/styleguides/python/).

Fixes #707

# Testing

**Remember to add or update unit tests for new features and changes.**

* How to [test your changes in A-plus](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations)
* How to [test accessibility](https://wiki.aalto.fi/display/EDIT/How+to+check+the+accessibility+of+pull+requests)

**What type of test did you run?**

- [ ] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [x] Django unit tests.
- [ ] Selenium tests.
- [ ] Other test. *(Add a description below)*
- [x] Manual testing.

Created 100,000 test users in the database, along with smaller amounts of other test data. Tested that all admin pages of models that reference `UserProfile` now load quickly. Tested that the search box appears on every list page and is functional. Tested that the previously hidden information was now visible on the admin pages.

**Did you test the changes in**

- [ ] Chrome
- [x] Firefox
- [ ] This pull request cannot be tested in the browser.

**Think of what is affected by these changes and could become broken**

# Translation

[...]

PR #780, opened 2021-02-08
Improve performance Notification Django Admin view

jaguarfi:
# Description

**What?**

Improved the performance of the Notification Django Admin view

**Why?**

Currently, the UI may crash or may return a timeout error when an admin user tries to access the details of any notification.

**How?**

`search_fields` and `autocomplete_fields` are added to the admin.py pages

# Testing

**Remember to add or update unit tests for new features and changes.**

* How to [test your changes in A-plus](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations)
* How to [test accessibility](https://wiki.aalto.fi/display/EDIT/How+to+check+the+accessibility+of+pull+requests)

**What type of test did you run?**

- [X] [Django debug toolbar](https://django-debug-toolbar.readthedocs.io/en/latest/)
- [ ] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Other test. *(Add a description below)*
- [ ] Manual testing.

**SQL Query previous to this PR**
notification

**SQL Query after this PR**
improved-notification

**Did you test the changes in**

- [X] Chrome
- [X] Firefox
- [ ] This pull request cannot be tested in the browser.

**Think of what is affected by these changes and could become broken**

# Translation

- [ ] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

# Programming style

- [X] Did you follow our [style guides](https://apluslms.github.io/contribute/styleguides/)?

# Have you updated the README or other relevant documentation?

- [ ] documents inside the doc directory.
- [ ] README.md.
- [ ] Aplus Manual.
- [ ] Other documentation (mention below which documentation).

# Is it Done?

- [ ] Reviewer has finished the code review
- [ ] After the review, the developer has made changes accordingly
- [ ] Customer/Teacher has accepted the implementation of the feature

*Clean up your git commit history before submitting the pull request!*

PR #813, opened 2021-04-25
Add Django Debug Toolbar

jaguarfi:
# Description

**What?**

Add Django Debug toolbar for debugging and testing purposes.

**Why?**

The Django Debug Toolbar  is a tool recommended in the Django project to
measure some performance information.

https://docs.djangoproject.com/en/3.1/topics/performance/#performance-benchmarking

This toolbar allows the developer to objectively measure some performance metrics. I saw that in the PR https://github.com/apluslms/a-plus/pull/805  the `search_fields` were added to improve the performance of the admin website. However,  in the test section of I found this comment.
>Created 100,000 test users in the database, along with smaller amounts of other test data. Tested that all admin pages of models that reference UserProfile **NOW LOAD QUICKLY**. Tested that the search box appears on every list page and is functional. Tested that the previously hidden information was now visible on the admin pages.

There is nothing bad with measuring the time manually and report that the page **"now loads quickly"**. But, it may be better to use an approach similar to the one used in the PR https://github.com/apluslms/a-plus/pull/780, where the time is measure in Real-time and the improvements can be reported objectively.

**How?**

The Django Debug Toolbar is added to the **requirements.txt** file, the configuration options are added to the **settting.py**, and the debug toolbar is added to the **urls.py** file.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Other test. *(Add a description below)*
- [ ] Manual testing.

I tested that the Debug toolbars are loaded when A+ is run outside the docker containers. I also tested that the Debug Toolbar is loaded within the Docker Container.
If you want to visualise the debug toolbar, you have to change the `DEBUG` configuration to `True`. This configuration is inside the **settings.py** file.

image

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


aplus/urls.py
```diff
@@ -53,4 +53,9 @@
 if settings.DEBUG:
     import django.views.static
     urlpatterns.insert(0, url(r'^media/(?P<path>.*)$', django.views.static.serve,
-        { 'document_root': settings.MEDIA_ROOT }))
+                              {'document_root': settings.MEDIA_ROOT}))
+
+    import debug_toolbar
+    urlpatterns.insert(0,
+                       url(r'^__debug__/', include(debug_toolbar.urls)),
+                       )
```
Mankro, line 61: I think the following indentation looks nicer. We can also use the newer `path` function instead of the deprecated `url`. If regular expressions are really required, then the new function is `re_path`.
```suggestion
    urlpatterns.insert(
        0,
        path('__debug__/', include(debug_toolbar.urls)),
    )
```

aplus/settings.py
```diff
@@ -126,6 +126,9 @@
 
     'js_jquery_toggle',
     'django_colortag',
+
+    # Debug Performance
+    'debug_toolbar',
```
Mankro, line 131: I don't think we want to enable this app in production, thus it must not be always included in `INSTALLED_APPS`. The end of this file already has if statements for modifying these settings. I suppose the app is only needed when `DEBUG = True`.

aplus/settings.py
```diff
@@ -189,6 +196,7 @@
     'social_django.middleware.SocialAuthExceptionMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'debug_toolbar.middleware.DebugToolbarMiddleware',
```
Mankro, line 199: Same for `MIDDLEWARE` as for `INSTALLED_APPS`.

Mankro (changes requested):
Thanks! Small changes requested below.

jaguarfi:
Ready for a second review.

PR #820, opened 2021-05-05
Add separate button for Aalto login.

### LINKED ISSUE #815: Add separate button for Aalto login (aside Haka login)
In a recent change (pull request #804 ) Aalto (or other local organization) login button was replaced by general Haka login, that directs user to organization selector, to allow login also using other Haka organization accounts than Aalto. This was an intermediate step (due to difficulties in shibboleth configuration), and a separate button for local organization login should now be added back, as majority of students would be using it, and usually some additional guidance may need to be added, e.g. for open university students to use local organization account instead of other organization.

PasiSa:
Add back the Aalto login button that was replaced by Haka login
button in recent commit. Now both local Aalto login button
and Haka login button for external users is shown.
Closes #815.

# Description

**What?**

Making this through pull request to check with others whether the URL management style is good, or whether it would be better done some other way. This works at minus.cs.aalto.fi. (Rest of PR template skipped)

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/aplus/urls.py
+++ b/aplus/urls.py
@@ -31,7 +31,7 @@ all_sitemaps = {
 #  Pay attention to the order the URL patterns will be matched!
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
-    url(r'^shibboleth/', include(shibboleth_login.urls)),
+    url(r'^', include(shibboleth_login.urls)),
     url('', include(social_django.urls, namespace='social')),
     url(r'^api/v(?P<version>(2))/', include(api.urls_v2)), # why version in url? doc/api_versioning.md
     url(r'^accounts/', include(userprofile.urls)),
--- a/userprofile/templates/userprofile/login.html
+++ b/userprofile/templates/userprofile/login.html
@@ -22,6 +22,29 @@
 		{% if shibboleth_login %}
 			<section aria-labelledby="login-shibboleth-users" class="col-sm-6 col-md-3 login-box">
 				<div class="login-section">
+					{% if shibboleth_title_text %}
+						<h2 id="login-shibboleth-users">
+							{{ shibboleth_title_text|safe }}
+						</h2>
+					{% endif %}
+					<div class="login-section-body">
+						{% if shibboleth_body_text %}
+							<p>
+								{{ shibboleth_body_text|safe }}
+							</p>
+						{% endif %}
+						<p>
+							<a class="aplus-button--secondary aplus-button--lg" role="button"
+								href="{% url 'shibboleth-login' %}?next={{ next|iriencode }}">
+								{{ shibboleth_button_text|safe }}
+							</a>
+						</p>
+					</div>
+				</div>
+			</section>
+		
+			<section aria-labelledby="login-haka-users" class="col-sm-6 col-md-3 login-box">
+				<div class="login-section">
 					{% if haka_title_text %}
 						<h2 id="login-haka-users">
 							{{ haka_title_text|safe }}
@@ -35,7 +58,7 @@
 						{% endif %}
 						<p>
 							<a class="aplus-button--secondary aplus-button--lg" role="button"
-								href="{% url 'shibboleth-login' %}?next={{ next|iriencode }}">
+								href="{% url 'haka-login' %}?target={% url 'shibboleth-login' %}">
 								{{ haka_button_text|safe }}
 							</a>
 						</p>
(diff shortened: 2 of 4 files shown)


aplus/urls.py
```diff
@@ -31,7 +31,7 @@
 #  Pay attention to the order the URL patterns will be matched!
 urlpatterns = [
     url(r'^admin/', admin.site.urls),
-    url(r'^shibboleth/', include(shibboleth_login.urls)),
+    url(r'^', include(shibboleth_login.urls)),
```
Mankro, line 34: Is `^` required here when it is otherwise empty? 

By the way, `url` is deprecated and an alias to the newer `path_re` function. If regex is not needed, the more simple `path` function is preferred. As you can see, we have never refactored old urls to the new path functions.

userprofile/templates/userprofile/login.html
```diff
@@ -20,6 +20,29 @@ <h1>{% blocktrans %}Log in to {{ brand_name }}{% endblocktrans %}</h1>
 	<div id="login-box-row" class="row">
 
 		{% if shibboleth_login %}
+			<section aria-labelledby="login-shibboleth-users" class="col-sm-6 col-md-3 login-box">
```
Mankro, line 23: `aria-labelledby` values and the target element ids do not match. Check both Shibboleth and Haka sections in this template. Each section needs a unique aria label. In this case, the label is a normal visible `<h2>` title and `aria-labelledby` refers to its id attribute.

So, this one should stay as `login-shibboleth-users`, but the Haka section below needs to be corrected, right?

userprofile/templates/userprofile/login.html
```diff
@@ -20,6 +20,29 @@ <h1>{% blocktrans %}Log in to {{ brand_name }}{% endblocktrans %}</h1>
...
+				<div class="login-section">
+					{% if shibboleth_title_text %}
+						<h2 id="login-haka-users">
```
Mankro, line 26: This `<h2>` has the wrong id. This element is referred to by the section's `aria-labelledby` attribute. The ARIA attributes are important for accessibility. This A+ page was rewritten for accessibility last year so let's keep it accessible.

Change `login-haka-users` to `login-shibboleth-users`.

Mankro (changes requested):
Overall, this is good. At least the HTML template must be changed because some id attributes are wrong.

userprofile/templates/userprofile/login.html
```diff
@@ -35,7 +58,7 @@ <h2 id="login-haka-users">
 						{% endif %}
 						<p>
 							<a class="aplus-button--secondary aplus-button--lg" role="button"
-								href="{% url 'shibboleth-login' %}?next={{ next|iriencode }}">
+								href="{% url 'haka-login' %}?target={% url 'shibboleth-login' %}">
```
Mankro, line 61: Is the query parameter (after `?`) safe without any encoding? Special characters might require encoding to be safe and standards-compliant. Could you check if it is required?

aplus/urls.py
PasiSa: Yep, but I thought to keep it similar to several other entries on the same file. I also noticed about the deprecated API, but maybe we should to the refactoring as separate commit? (I can take that next on my todo list after this, if we want to do it for June release)

userprofile/templates/userprofile/login.html
PasiSa: good catch, will do

userprofile/templates/userprofile/login.html
PasiSa: It comes from our own source code,in urls.py and equals to "shibboleth/login/", so I thought the encoding is not needed.

aplus/urls.py
Mankro: Refactoring is not urgent. The current code is fine for this PR.

userprofile/templates/userprofile/login.html
Mankro: Yeah. I wasn't sure if Django includes the domain (https://plus.cs.aalto.fi) in the output of the `url` template tag. I suppose this is fine.

Mankro (approved):
HTML template id and aria-labelledby attributes have been fixed.

PR #838, opened 2021-06-29
Add a language pack for English and replace translatable strings in source code with keys

etanttila:
# Description

**What?**

* Add a language pack for English and replace translatable strings in source code with string keys
* Update documentation relating to updating the translations
* Fix formatting of lazy translation strings to use a formatting method, which is suitable for lazy strings

**Why?**

* In order to make editing user interface strings more easy (and less likely to break translations)
* Lazy formatting: fix a bug identified by Markku

**How?**

* Created a language file for English and copied all the strings to msgstrs with a script
* Created a script which replaces selected keys (which I had written in the PO-file) in both PO-files as well as the source code
* Lazy formatting: used `git grep` to search where `format` is used, and replaced those instances.

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Manually checked that the strings seem to appear normally in both English and Finnish.
Did not test for accessibility as change should affect only how the texts are displayed in the source code.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [x] README.md.
- [ ] Aplus Manual.
- [x] Other documentation (mention below which documentation).

I also worked on updating the documentation in the [styleguide for translations](https://apluslms.github.io/contribute/styleguides/gettext/), for which the PR can be found [here](https://github.com/apluslms/apluslms.github.io/pull/7).

# Is it Done?


course/models.py
```diff
@@ -368,34 +368,34 @@ class CourseInstance(UrlMixin, models.Model):
...
+        ('TOC', 1, _('TABLE_OF_CONTENTS')),
+        ('LAST', 2, _('LAST_VISITED_LINK')),
+        ('EXPERIMENT', 10, _('EXPERIMENTAL_SETUP')),
     ])
     CONTENT_NUMBERING = Enum([
-        ('NONE', 0, _("No numbering")),
-        ('ARABIC', 1, _("Arabic")),
-        ('ROMAN', 2, _("Roman")),
-        ('HIDDEN', 3, _("Hidden arabic")),
+        ('NONE', 0, _('NUMBERING_NONE')),
```
skulonen, line 390: These enum values are prefixed with NUMBERING, but the enums right before this one are not prefixed. Is this a consistency issue?

course/models.py
```diff
@@ -700,25 +696,25 @@ class CourseModule(UrlMixin, models.Model):
     opening times and deadlines for exercises.
     """
     STATUS = Enum([
-        ('READY', 'ready', _("Ready")),
-        ('UNLISTED', 'unlisted', _("Unlisted in table of contents")),
-        ('HIDDEN', 'hidden', _("Hidden")),
-        ('MAINTENANCE', 'maintenance', _("Maintenance")),
+        ('READY', 'ready', _('STATUS_READY')),
+        ('UNLISTED', 'unlisted', _('STATUS_UNLISTED')),
+        ('HIDDEN', 'hidden', _('HIDDEN_capitalized')),
```
skulonen, line 703: Should this be STATUS_HIDDEN, to be consistent with the other enum values?

course/templates/course/module.html
```diff
@@ -24,13 +24,11 @@
         <br />
         <span class="dates">
             <em>
-                {% blocktrans with deadline=current.late_time|date:"DATETIME_FORMAT" %}
-                Late submissions are allowed until {{ deadline }}.
-                {% endblocktrans %}
+                {% blocktranslate trimmed with deadline=current.late_time|date:"DATETIME_FORMAT" %}
+                    LATE_SUBMISSIONS_ALLOWED_UNTIL -- {{ deadline }}{% endblocktranslate %}
```
skulonen, line 28: These {% endblocktranslate %} tags should probably be on their own rows, since the {% blocktranslate %} tags are on their own rows.

course/templates/course/staff/_tag_remove_modal.html
```diff
@@ -18,32 +18,28 @@
...
             </div>
             <div class="modal-body">
-                {% blocktrans %}
-                Remove tagging
-                <span id="tag-remove-modal-tagging"></span>
-                from <span id="tag-remove-modal-amount"></span> users:
-                <ul id="tag-remove-modal-users"></ul>
-                {% endblocktrans %}
+                {% blocktranslate trimmed %}
+                REMOVE_TAGGING_FROM_USERS_MODAL{% endblocktranslate %}
```
skulonen, line 26: This {% endblocktranslate %} tag should probably be on its own row, since the {% blocktranslate %} tag is on its own row.

exercise/submission_models.py
```diff
@@ -120,12 +120,12 @@ class Submission(UrlMixin, models.Model):
     A submission to some course exercise from one or more submitters.
     """
     STATUS = Enum([
-        ('INITIALIZED', 'initialized', _("Initialized")),
-        ('WAITING', 'waiting', _("In grading")),
-        ('READY', 'ready', _("Ready")), # graded normally
-        ('ERROR', 'error', _("Error")),
-        ('REJECTED', 'rejected', _("Rejected")), # missing fields etc
-        ('UNOFFICIAL', 'unofficial', _("No effect on grading")),
+        ('INITIALIZED', 'initialized', _('INITIALIZED')),
```
skulonen, line 123: Should these enum values use a prefix?

locale/en/LC_MESSAGES/django.po
```diff
@@ -3178,948 +2859,776 @@ msgstr ""
...
 msgstr "Home"
 
 #: templates/base.html userprofile/templates/userprofile/privacy.html
-msgid "Privacy Notice"
+msgid "PRIVACY_NOTICE"
 msgstr "Privacy Notice"
 
 #: templates/base.html userprofile/templates/userprofile/accessibility.html
-msgid "Accessibility Statement"
+msgid "ACCESSABILITY_STATEMENT"
```
skulonen, line 3232: Misspelled "accessibility"

locale/en/LC_MESSAGES/django.po
```diff
@@ -3178,948 +2859,776 @@ msgstr ""
...
 msgstr "No privacy notice. Please notify administration!"
 
 #: userprofile/views.py
-msgid "No local accessibility statement. Please notify the site's owner!"
+msgid "NO_LOCAL_ACCESSIBILITY_STATEMENT"
 msgstr "No local accessibility statement. Please notify the site's owner!"
 
 #: userprofile/views.py
-msgid "No system-wide accessibility statement found."
+msgid "NO_SYSTEM-WIDE_ACCESSABILITY_STATEMENT"
```
skulonen, line 3629: Misspelled "accessibility"

course/models.py
etanttila: In the Enum above, it was difficult to decide whether to prefix them or not (especially as several of the phrases above are used in other contexts such as "User results" and "Table of contents". I thought that with the numberings, the words on their own without contexts are strange, so the prefixes seemed important there.
What do you think concerning about the INDEX_TYPE Enums, does it need prefixes (or some of the phrases, which appear only there), and if so, what would be a good prefix?

course/models.py
etanttila: I considered that, but then we need several keys that match to the word "Hidden". However, that might be good if one wants to change the phrase or word displayed instead of hidden in the status context but not everywhere else. Which do you think would be better?

course/templates/course/module.html
etanttila: Good catch, thanks, I'll fix these!

course/templates/course/staff/_tag_remove_modal.html
etanttila: Good catch, thanks, I'll fix this!

locale/en/LC_MESSAGES/django.po
etanttila: Oops, great catch, I'll fix these!

locale/en/LC_MESSAGES/django.po
etanttila: Oops, great catch, I'll fix these!

exercise/submission_models.py
etanttila: I was of two minds on that.  "Rejected" is used elsewhere as well, and I thought that some of the others might be used somewhere else (at least in the future), but at the moment they aren't. Looking back at this, at least "initialized" and "error" would be good to have the "STATUS" prefix up front.
"Rejected" can or can not (depending on whether we want to have separate keys for both "rejected" uses, (or if we would want the points badge to use "STATUS_REJECTED" as well, which is also an option). What do you think?

"in grading" and "No effect on grading" are also difficult. They could be left without the prefix (as they are now, which would accent that it's not just the same word as the enum value), or the keys could just be "STATUS_WAITING" and "STATUS_UNOFFICIAL" while the msgstr would remain "In grading" and "No effect on grading". Personally I'd probably prefer the latter. What do you think?

skulonen:
The issues I mentioned in the review have been fixed and the commit history has been cleaned up. I'll merge this.

PR #840, opened 2021-07-05
Add a teachers list page

lainets:
# Description

**What?**

A page for listing active teachers.

**Why?**

An up to date list of teachers is important for communication.

**How?**

A new page at /accounts/teachers that is only accessible to super users.

# Testing

- [X] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [X] Django unit tests.
- [X] Selenium tests.
- [X] Manual testing.

Tested that the buttons and translations work.

- [X] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

Added a new ADMINISTRATION.md file for instructions related to administration.

# Is it Done?


lainets:
Is there a better place than the new file for mentioning where to find the teacher list page?

locale/en/LC_MESSAGES/django.po
```diff
@@ -1221,6 +1222,12 @@ msgstr ""
 "Request for a course configuration failed with error '{error!s}'. "
 "Configuration of course aborted."
 
+#: edit_course/operations/configure.py
+msgid ""
+"Request for a course configuration failed with error '{error!s}'. "
+"Configuration of course aborted."
+msgstr ""
```
etanttila, line 1229: Is this message supposed to be here at all?

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1224,6 +1225,12 @@ msgstr ""
 "Kurssikonfigurointipyyntö epäonnistui virheellä '{error!s}'. Konfigurointi "
 "keskeytettiin."
 
+#: edit_course/operations/configure.py
+msgid ""
+"Request for a course configuration failed with error '{error!s}'. "
+"Configuration of course aborted."
+msgstr ""
```
etanttila, line 1232: Is this message supposed to be here at all?

userprofile/views.py
```diff
@@ -221,3 +224,71 @@ def get_context_data(self, **kwargs):
...
+    access_mode = ACCESS.SUPERUSER
+    template_name = "userprofile/teachers.html"
+
+    def entry(self, teacher: UserProfile, instance: CourseInstance):
+        return [
+            teacher.user.first_name + " " + teacher.user.last_name,
+			teacher.user.email,
+			str(instance),
+			str(instance.starting_time),
+			str(instance.ending_time),
```
etanttila, line 238: What are these extra indents? (Does the indentation of this array use both spaces and tabs?)

skulonen:
I tested this, seems to work well! Here are some comments.

- "Sisällä assistentit"? Should probably be "sisällytä"?
- Style guide: "surround top-level function and class definitions with two blank lines". This rule is not respected in this PR.
- The code uses double quotes for strings that are not user-facing.
- Unused imports in teacher_list_test.py: WebDriverWait, EC
- Type hinting: some parameters are typed properly, but some are missing (e.g. TeacherListView.entries, TeacherListPage.\_\_init\_\_). In addition, return types are missing and some lists are untyped (query in TeacherListPage.\_\_init\_\_, data in TeacherListPage.get_table_data, entries in TeacherListView.entries).

Notes about usability:
- The page doesn't look great on mid-sized screens (wider than smartphone but narrower than 1080p). Try docking the window to the left or right side of your screen and you'll see that the form at the top becomes unnecessarily narrow and the buttons at the bottom can't contain their labels.
- Speaking of the buttons at the bottom, should they actually be above the table? If the list can get long (hundreds of users), I'd rather not have to scroll to the bottom to use the buttons.
- I thought it was a bit annoying that if you set the start of the date range first, the end automatically gets the same value instead of staying unspecified. If it stayed unspecified, you could easily search for e.g. all users who have taught a course within the last year. It's not a big deal though, since you can manually clear the end date.

lainets:
> * Style guide: "surround top-level function and class definitions with two blank lines". This rule is not respected in this PR.

I fixed this but left locators.py as is on purpose as double spacing doesn't really make sense there.

> * The code uses double quotes for strings that are not user-facing.

The style guide doesn't say you can't use them for other things. In fact, it only specfies the usages "for keys and types" and "for strings shown to the user", anything else seems up to the coder. This also isn't how it is used in the codebase as is. (I also happen to think this whole rule is stupid, there is no explanation for the rule). I did change the dict keys to single quotes though.

> * Type hinting: some parameters are typed properly, but some are missing (e.g. TeacherListView.entries, TeacherListPage.__init__). In addition, return types are missing and some lists are untyped (query in TeacherListPage.__init__, data in TeacherListPage.get_table_data, entries in TeacherListView.entries).

I think this PR was done before the type hinting rules and I was just adding them willy-nilly 😄. Should be fixed now

> * The page doesn't look great on mid-sized screens (wider than smartphone but narrower than 1080p). Try docking the window to the left or right side of your screen and you'll see that the form at the top becomes unnecessarily narrow and the buttons at the bottom can't contain their labels.

Fixed. Also, I cannot for the life of me figure out how to center the form with bootstrap when "max-width" is specified, so I just moved it to the left side.

> * Speaking of the buttons at the bottom, should they actually be above the table? If the list can get long (hundreds of users), I'd rather not have to scroll to the bottom to use the buttons.

I mean it shouldn't be very long but I guess it's better on top.

> * I thought it was a bit annoying that if you set the start of the date range first, the end automatically gets the same value instead of staying unspecified. If it stayed unspecified, you could easily search for e.g. all users who have taught a course within the last year. It's not a big deal though, since you can manually clear the end date.

I just set the fields to fill to today by default. This fixes the problem and is also more intuitive.

etanttila:
>The style guide doesn't say you can't use them for other things. In fact, it only specfies the usages "for keys and types" and "for strings shown to the user", anything else seems up to the coder.

Just as a comment, I believe the goal is to use double quotes for only user-facing strings in Python (unless necessary due to existence of single quotes in the string itself) and vice versa, or at least that was how I understood Jaakko meant it when I was reading and commenting the style guide. It might not have been directly said, but I'm assuming the point of the describing the usages is that those are the two categories: (1) strings shown to users, (2) others, including keys and types, etc.
I believe the idea of the rule is to be quickly be able to identify if a string is one that is displayed to the users or just a back-end thing. Of course, with almost everything having translations, most user-facing strings can be identified by those.

skulonen:
Alright, I checked this again. Everything seems to be fixed. Just a small nitpick: the declaration of `TeacherListPage.__init__` goes past the line length limit defined in the style guide (119) now that types have been added to it. It should be separated to multiple lines. Notice how function declarations are indented in the style guide:
```python
def foobar(
        argument1,
        argument2,
        keyword1=None,
        keyword2=None, # remember the comma
        ): # here this seems to be part of the arguments, which is good
```

list_data and list_data_with_assistants in the same file also have this problem, but I don't really take issue with them because they're just data and their contents don't matter when reading the code.

Also, do we need to discuss the single vs double quote thing further? Seems like there's some mild disagreement on how to interpret the style guide.

etanttila:
Concering single vs. double quote thing, it might be a good idea to at some point meet about it and discuss it further and come up with clear guidelines and edit the guide to make it clearer and add justifications, but whether it's necessary at the moment I'd say is up to you guys.
I entirely forgot to review this in detail. If I remember correctly when I saw this originally, I had a very short amount of time so I thought I'd skim it and mention what sticked out to me quickly and later come back to it and review it in more detail, but I entirely forgot about it, and am not sure if it's worth it for me to review it anymore as Sampo seems to have done an excellent job, but if you guys wish, I can still take a look at this and review it. So basically I don't have much of an opinion on this PR, I just quickly commented yesterday as well.

lainets:
> Alright, I checked this again. Everything seems to be fixed. Just a small nitpick: the declaration of `TeacherListPage.__init__` goes past the line length limit defined in the style guide (119) now that types have been added to it. It should be separated to multiple lines. Notice how function declarations are indented in the style guide:

Fixed.

> Also, do we need to discuss the single vs double quote thing further? Seems like there's some mild disagreement on how to interpret the style guide.
>
> Concering single vs. double quote thing, it might be a good idea to at some point meet about it and discuss it further and come up with clear guidelines and edit the guide to make it clearer and add justifications, but whether it's necessary at the moment I'd say is up to you guys.

I was going to bring up whether this rule should even exist in the daily.

> I entirely forgot to review this in detail. If I remember correctly when I saw this originally, I had a very short amount of time so I thought I'd skim it and mention what sticked out to me quickly and later come back to it and review it in more detail, but I entirely forgot about it, and am not sure if it's worth it for me to review it anymore as Sampo seems to have done an excellent job, but if you guys wish, I can still take a look at this and review it. So basically I don't have much of an opinion on this PR, I just quickly commented yesterday as well.

I don't think there is a need for you to review it.

skulonen:
I would be fine with removing the rule. Now that almost all user-facing strings have translations, and the translations are referenced using keys, the number of _actually_ user-facing strings in the code is very low. This makes the rule almost pointless.

With that being said, the style guide should probably still mention which type of quotes we should prefer in our code, or mention that anything goes if that's what we decide.

skulonen:
This is good now. After rebasing and cleaning the commit history we can merge this.

lainets:
Made a small change: I limited the width of the export button container to max 600px.

skulonen:
Looks fine to me. Is this ready to be merged or do you still need to do something with the commit history?

lainets:
I think it's good now. Did something about the commits catch your eye?

skulonen:
No, I just asked in case you weren't happy with them yet. I'll go ahead and merge this.

PR #850, opened 2021-08-12
Type hinting for mixins

skulonen:
# Description

**What?**

This PR adds type hinting to all mixin classes.

**Why?**

The team decided to start using type hints, and adding them to mixins facilitates adding them to other places, since many classes depend on mixins.

**How?**

The following things have been done:
- Create base classes that declare methods that the mixins use through `self` or `super()`. The base classes are replaced with an empty class at runtime, so that the method declarations don't appear in the MROs of the mixin classes. As such, they exist solely for type checking purposes. They're defined in `authorization/types.py`, `course/types.py` and `exercise/types.py` (depending on which classes they use) or `lib/types.py` (when they only use Python builtin/Django/DRF classes), and are named "Supports*".
- Add type hints to method parameters and return values.
- Remove unused imports.
- `cast` was needed in two situations: when getting a LearningObject from an arbitrary dict (`edit_course/exercise_forms.py`) and when using getattr (`lib/api/mixins.py`).

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

Tested manually that the main site and API work.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lib/types.py
+++ b/lib/types.py
@@ -37,7 +37,7 @@ class SupportsGetFields:
 @empty_at_runtime
 class SupportsGetObjectOrNone:
     TModel = TypeVar('TModel', bound=Model)
-    def get_object_or_none(self, kwarg: str, model: Type[TModel]) -> TModel: ...
+    def get_object_or_none(self, kwarg: str, model: Type[TModel], field: str = 'pk', **extra: Any) -> TModel: ...
 
 
 @empty_at_runtime


lainets:
I thought the point was to directly inherit from the appropriate classes unless there were multiple options like AuthorizationMixin/APIView... The other problem is that if every method has its own Supports* class, you get zero information out of it, might as well do `# type: ignore`. I had thought we would make a single class that represents AuthorizationMixin and APIView, this would give the programmer the contextual information that this class should have one of these as a super class.

From my opinion type hints have two purposes: (1) check the correctness of the code and (2) give the programmer information about types. The reason why I think Supports* should only be used when needed is that it effectively says "ignore type errors for these methods" i.e. we lose the first purpose. The second purpose is also lost if in addition every method has it's own class (or you could add a comment saying what usually implements the method in question, but I think it would make more sense to have the Supports* classes represent actual classes, though you will most likely need the comment in this case too)

lainets:
Note that the Supports* works fine when used as Protocols on function parameters because then the linter can check that the methods exist. This just doesn't work when typing super classes.

skulonen:
Okay, seems like there was a misunderstanding between us at some point and I got the impression that you preferred these Supports* classes in general, and I just went along with it because the consistency appealed to me. But you're right that we lose all static guarantees when we do this. Well, no big deal, I'll get back to this in a while.

Mankro (commented):
Should this PR be closed without merging? We don't want to continue with a dozen SupportsMethod classes for static typing.

Mankro:
> Should this PR be closed without merging? We don't want to continue with a dozen SupportsMethod classes for static typing.

Closing now.

PR #853, opened 2021-08-16
Improvement of texts and translations in the user interface

etanttila:
# Description

**What?**

Improve texts and translations in the user interface. Also add verbose names or labels and translations to models and fields, which are displayed in forms and on the admin site.

**Why?**

Because many texts were quite awkward and grammatically incorrect. Some terms were also used inconsistently.

**How?**

Reading through strings and fixing them. I also made a table of frequently used terms and their translations, which helped in providing consistency and selecting good translations. Several people also went through the table and commented their thoughts.  This table and those comments were extremely valuable in improving texts.

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

I went through most of the pages and checked that the translations seem good and appear correctly.

- [x] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

- [x] Customer/Teacher has accepted the implementation of the feature


etanttila:
Note: one of the selenium tests that are run on GitHub fails for some reason https://github.com/etanttila/a-plus/runs/3340814277
I'm not sure what causes it to fail.

etanttila:
I had to update one of the Selenium test locators since the text had changed, so this fixup commit fixes the problem with the Selenium tests.

locale/en/LC_MESSAGES/django.po
```diff
@@ -290,53 +629,110 @@ msgstr "Maintenance"
...
+msgstr "learning object category"
+
+#: course/models.py
+msgid "MODEL_NAME_LEARNING_OBJECT_CATEGORY_PLURAL"
+msgstr "learning object categories"
 
 #: course/permissions.py
 msgid "COURSE_VISIBILITY_PERMISSION_DENIED_MSG"
-msgstr "Permission denied by course visibility"
+msgstr "Permission denied by course visibility."
```
annirytkonen, line 735: I'm not quite sure what the "Permission denied by course visibility" means. How can course visibility deny anything? What is the context for this?

locale/en/LC_MESSAGES/django.po
```diff
@@ -46,23 +171,23 @@ msgstr "Any authenticated student"
...
 
 #: authorization/permissions.py
 msgid "ACCESS_COURSE_ASSISTANT"
-msgstr "Assistant of the course"
+msgstr "An assistant of the course"
 
 #: authorization/permissions.py
 msgid "ACCESS_GRADING"
-msgstr "Grading. Assistant if course has that option or teacher"
+msgstr "Grading. An assistant if course has that option or a teacher"
```
annirytkonen, line 182: I'm not quite sure what this string means. I would use commas, e.g. "An assistant, if course has that option, or a teacher"

locale/en/LC_MESSAGES/django.po
```diff
@@ -290,53 +629,110 @@ msgstr "Maintenance"
...
+msgstr "confirm the level"
+
 #: course/models.py
 msgid "LEARNING_OBJECT_CATEGORY_LEVEL_CONFIRMATION_EXERCISE_HELPTEXT"
 msgstr ""
-"Once exercise is graded non zero it confirms all the points on the hierarchy "
-"level. Implemented as a mandatory feedback feature."
+"Once an assignment is graded non-zero, it confirms all the points on the "
+"hierarchy level (e.g. within the chapter or the module). Implemented as a "
+"mandatory feedback feature."
```
annirytkonen, line 712: Is this string visible to end-users? The info "Implemented as a mandatory feedback feature." is quite technical in that context

locale/en/LC_MESSAGES/django.po
```diff
@@ -894,11 +1299,11 @@ msgstr "There are no course instances for this course."
 
 #: course/views.py
 msgid "ENROLLMENT_ERROR_CANNOT_ENROLL_OR_ALREADY_ENROLLED"
-msgstr "You cannot enroll, or have already enrolled, in this course."
+msgstr "You cannot enrol, or have already enrolled, in this course."
```
annirytkonen, line 1302: heh :) This string should be edited to include more concrete status update, but that's most likely requires some development

locale/en/LC_MESSAGES/django.po
etanttila: This appears, for example, if a user tries to access a course that isn't visible to them (e.g. an external student tries to open a course for only internal students, or a student has copy/pasted or guessed a link to course whose material is not visible to students yet.) The error messages can be something like this:
image
image
image


I agree that the wording isn't great, but didn't have better ideas, either.

locale/en/LC_MESSAGES/django.po
etanttila: I'm not sure in which contexts it's used (basically it's a description of the access level, so if the course allows TA's to grade, if the access level is "GRADING", both teachers and TA's can access the page or whatever, if TA's aren't allowed grading, it's just teachers). Good point about the commas. Do you think knowing the context now that commas or parentheses would be better, for example
"An assistant, if the course has allowed that option, or a teacher" or "An assistant, if the course has allowed assistant grading, or a teacher"
"An assistant (if the course has allowed that option) or a teacher" or "An assistant (if the course has allowed assistant grading) or a teacher"

locale/en/LC_MESSAGES/django.po
etanttila: It is, it's visible to teachers when defining category settings (as a help text for the "Confirm the level" (the name isn't very easy to understand):
image
I think the phrase is quite techincal and awkward (and difficult to understand), do you have any better ideas how to get the point across? (I think even the "on the hierarchy level" is difficult to understand, which is why hopefully the added specification helps.)

jsorva: I’d label the feature according to what it’s meant for, i.e., making things mandatory.  Maybe something like:
_[ ] **Mandatory component**
Students must have non-zero points in this category before their other points in the same chapter/module become official. Suggested usage: check this box for feedback activities that students must fill to score points for a course component._

locale/en/LC_MESSAGES/django.po
jsorva: This particular this string aside, is the plan to adopt/use British English in A+? I’d vote against that, since almost all the English learning materials served on A+ are (presumably) in American English, and this would just create an unattractive contrast between the platform and (most of) the contents.

locale/en/LC_MESSAGES/django.po
etanttila: Sounds great, thanks!

locale/en/LC_MESSAGES/django.po
jsorva: Suggestion:
_“Member of course staff with grading rights: a teacher or (if allowed by course settings) an assistant.”_ 

BTW, the existing text seems to suggest that TA grading is set at a course level when in fact it can be adjusted at a much finer grain, even for individual assignments.

locale/en/LC_MESSAGES/django.po
jsorva: Incidentally, there’s an unrelated minor UX issue in that screenshot: the pane is titled just "Category" and there is no indication of which category is being edited.

jsorva: Another off-topic comment now that I’m at this: I’ve just realized that even after using A+ for years, I don’t know what each of the options in the Status menu does. Maybe don’t need to know, and I guess there’s documentation for that somewhere, but this view could also be more helpful.

etanttila: >Incidentally, there’s an unrelated minor UX issue in that screenshot: the pane is titled just "Category" and there is no indication of which category is being edited.

Good point! I suppose the "Name" field is how to distinguish between the categories, but since it's not first in the list, it doesn't seem like an identifier.

locale/en/LC_MESSAGES/django.po
jsorva: Yes, this message definitely needs changing. It’s a very visible part of A+ that people often encounter (I surmise). 

As far as I can tell, the point of the message is that a page is not accessible, at least not now. That is, the message’s point is not what technical entity might be "denying" or "forbidding" something, so that can be left out entirely. 

Maybe something along these lines? 

If the page is unavailable to all: _“The page at this address is not available at this time. The course staff may choose to make it public later. Please also check that you are attempting to access the correct page.”_

If the page is unavailable to external students: _“Sorry! The page at this address is only available to students with an Aalto University user account. If you have an Aalto account, please make sure you use it to log in to A+.”_

Also, I’m not sure **Forbidden** in big letters is how we want to prefix this message to would-be learners.

(BTW, Ella’s message above referred to Aalto specifically, so I take it that that’s OK in this context. Or should this message be more generic?)

locale/en/LC_MESSAGES/django.po
etanttila: >(EDIT: The existing text says "non-zero", as does my suggestion, but is that actually correct? Does the "Points to pass" setting affect this?)

Excellent point! I'm not entirely sure. Does @Mankro know?

locale/en/LC_MESSAGES/django.po
etanttila: >(BTW, Ella’s message above referred to Aalto specifically, so I take it that that’s OK in this context. Or should this message be more generic?)

The message uses the institution name, and I really hope that all of the institutions have names that the genetive form is formed by just adding an "n" to the end (e.g. yliopisto -> yliopiston; koulu -> koulun; lukio -> lukion), since I didn't find a way to use genetive forms properly.

locale/en/LC_MESSAGES/django.po
etanttila: At least I understood that that was the plan, since the materials and names at Aalto in general are in British English (e.g. "Learning Centre"), and the language code had been defined as en-gb. However, if most of the learning materials are in face in American English, that would definitely be a good reason to stick with American English.

jsorva: I don’t know this for a fact, but my gut feeling is that AmE is way more common in learning materials. And yeah, aligning with learning materials on the same site seems more important than aligning with generic Aalto policies that have little impact on the grassroots level of teaching and learning.

locale/en/LC_MESSAGES/django.po
Mankro: that assumption (genetive adding n) won't always work. Since languages are different, I think the only way is to have separate strings for different forms of the word. You also shouldn't manipulate any translated string by adding something to it in the code (like adding "n" to a translated string "koulu"). That won't work in all languages since they present things so differently.

locale/en/LC_MESSAGES/django.po
Mankro: I haven't tried that. Since the old text says "non-zero", let's keep that.

locale/en/LC_MESSAGES/django.po
etanttila: Yep, it doesn't work in all languages and would require different forms of the word. I originally didn't intend to use the Finnish brand name in the Finnish message at all, but localization and lazy formatting didn't allow not including the string in the message. Likewise having different forms of the word wouldn't work since in English here one would just say "students within {name}" while in Finnish one would need the genetive form, and providing both forms doesn't work since one would need to be left out anyway. It's a shame that localization files and formatting can't work in a way where you could provide a word with several different cases and you can just select the proper one in the translation string. I would love to just be able to write, say "Tämä kurssi on vain {institution.genetive} opiskelijoille (eli sisäisille käyttäjille)." But yep probably here it's necessary to remove the name of the institution entirely from the error message from both languages, unless it's reworded in a much longer fashion where it would be ok to say "oppilaitoksen {institution} opiskelijoille" or something like that.

selenium_test/test/locators.py
```diff
@@ -77,7 +77,7 @@ class SubmissionPageLocators(object):
     INSPECTION_LINKS = (By.XPATH, "//table//a[contains(@href, '/inspect/')]")
 
 class StudentFeedbackPageLocators(object):
-    ASSISTANT_FEEDBACK_LABEL = (By.XPATH, "//h4[text()='Assistant feedback']")
```
PasiSa, line 80: Do you want this to be a separate commit, or rebased to one of the earlier?

etanttila: The idea is to be rebased to one of the earlier, that's why it's a fixup commit :)

locale/en/LC_MESSAGES/django.po
```diff
@@ -381,8 +381,8 @@ msgstr "external"
 #: course/models.py
 msgid "EXTERNAL_USER_DESCRIPTION"
 msgstr ""
-"The user profile doesn't have a student number, thus the user has logged in "
-"from a different organization or via social authentication"
```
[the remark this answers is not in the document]
etanttila: I believe in the past the very general (and strange) term of "social authentication" has been used in case we intend to add other forms of social authentication, such as Facebook. Of course, since we don't support others, it could be "trimmed" down to just say Google.
Good point about the student IDs!  Are the users that have logged in via Haka and have a different organization that the local one defined as external?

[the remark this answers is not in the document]
etanttila: Would it be better to use ["social login"](https://en.wikipedia.org/wiki/Social_login) rather than "social authentication", as it would still be general if we later on decided to add on, say Microsoft login?

PasiSa:
My problem was more about the word "social". Google is not really a social network service these days, is it? Perhaps "external authentication" would be better, if we do not want to say "Google", although in my opinion it would be better to directly tell which provider we are using.

- Pasi

PasiSa:
Currently the logic is such that if user's organization is different than the local organization configured in settings.py, it is determined as external.

etanttila:
> My problem was more about the word "social". Google is not really a social network service these days, is it? Perhaps "external authentication" would be better, if we do not want to say "Google", although in my opinion it would be better to directly tell which provider we are using. - Pasi

Hmm, "external authentication" could work as well, and I suppose "through a different organization" would belong in that as well (in addition to just these social logins). And you do have a point in Google not necessarily being a social network service, although in a way I suppose it is social (email, Google classroom, sharing files in Google Drive, etc. is quite social), and Wikipedia does list them on their "Social login" page. But I think "external authentication" or "external login" could be good, as it would include Haka as wel.

locale/en/LC_MESSAGES/django.po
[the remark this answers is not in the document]
Mankro: like you said, external users are now defined from the new organisation field of the UserProfile. Only one organisation counts as internal (set in settings.py). All Haka users have some organisation. Google users have no organisation.

"Social login" is fine for me. At the moment Google is the only method, but indeed there could be more. We aren't planning to add any new social login providers, though.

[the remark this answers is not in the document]
etanttila: How about combining "from a different organization" and "social authentication/login" into "external authentication" or "external login" as suggested below?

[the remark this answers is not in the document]
PasiSa: good idea

locale/en/LC_MESSAGES/django.po
etanttila: The general "forbidden" -message etc. is something that probably should be addressed in a later issue and commit. The error messages could also be improved a lot at that time. Currently with the time limits I'll change the messages to something a bit "nicer" and hopefully easier to understand ("Unfortunately you are not permitted to see this ..."), but more detailed error messages based on the context would be a good improvement that probably take too much time to fix for this RC.
I'll also change the genetive form of the institution to "oppilaitoksen {institution} opiskelijoille", etc.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -748,15 +749,15 @@ msgstr "Tälle kurssille ei ole vapaata pääsyä."
 #, python-brace-format
 msgid "COURSE_ENROLLMENT_AUDIENCE_ERROR_ONLY_INTERNAL -- {institution}"
 msgstr ""
-"Tämä kurssi on vain {institution}n opiskelijoille (eli sisäisille "
-"käyttäjille)."
+"Tämä kurssi on vain oppilaitoksken {institution} opiskelijoille (eli "
```
PasiSa, line 752: typo! oppilaitoksken

locale/fi/LC_MESSAGES/django.po
```diff
@@ -382,8 +382,8 @@ msgstr "ulkopuolinen"
 #: course/models.py
 msgid "EXTERNAL_USER_DESCRIPTION"
 msgstr ""
-"Käyttäjän tiedoissa ei ole opiskelijanumeroa, joten hän on kirjautunut "
-"sisään toisesta organisaatiosta tai sosiaalisen palvelun kautta"
+"Käyttäjän organisaationa ei ole paikallista organisaatiota, joten hän on "
+"kirjautunut sisään jonkin toisen palvelun kautta."
```
PasiSa, line 386: olisiko "tunnistuspalvelun" parempi?

locale/en/LC_MESSAGES/django.po
etanttila: I created an issue about the access denied messages and the 403 page: #858 
If @jsorva you want to write more there, feel free to :)

PR #855, opened 2021-08-19
Add feedback link to footer

etanttila:
# Description

**What?**

Add a link leading to a feedback form to the footer.

**Why?**

So users can easily give feedback on the UI, etc.

**How?**

By editing the base template and adding a translation.

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -229,6 +229,7 @@
 						<li><a href="{% url 'privacy_notice' %}">{% translate "PRIVACY_NOTICE" %}</a></li>
 						<li><a href="{% url 'accessibility_statement' %}">{% translate "ACCESSIBILITY_STATEMENT" %}</a></li>
 						<li><a href="{% url 'support_channels' %}">{% translate "SUPPORT" %}</a></li>
+						<li><a href="https://link.webropol.com/s/aplus-feedback" target="_blank">{% translate "FEEDBACK" %}</a></li>
 					</ul>
 				</div>
 			</div>
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -3309,6 +3309,10 @@ msgid "SUPPORT"
 msgstr "Support"
 
 #: templates/base.html
+msgid "FEEDBACK"
+msgstr "Feedback"
+
+#: templates/base.html
 msgid "LOADING_FAILED"
 msgstr "Loading failed!"
 
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -3314,6 +3314,10 @@ msgid "SUPPORT"
 msgstr "Tuki"
 
 #: templates/base.html
+msgid "FEEDBACK"
+msgstr "Palaute"
+
+#: templates/base.html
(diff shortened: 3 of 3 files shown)


templates/base.html
```diff
@@ -229,6 +229,7 @@ <h4>{% translate "SITE" %}</h4>
 						<li><a href="{% url 'privacy_notice' %}">{% translate "PRIVACY_NOTICE" %}</a></li>
 						<li><a href="{% url 'accessibility_statement' %}">{% translate "ACCESSIBILITY_STATEMENT" %}</a></li>
 						<li><a href="{% url 'support_channels' %}">{% translate "SUPPORT" %}</a></li>
+						<li><a href="https://link.webropol.com/s/aplus-feedback">{% translate "FEEDBACK" %}</a></li>
```
Mankro, line 232: Does this link open in the same tab/window? It should open in a new tab. That requires the attribute `target="_blank"`.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target

Mankro (changes requested):

templates/base.html
etanttila: Excellent point, thanks, I'll fix that!

etanttila: By the way, all of the other links in the footer open in the same tab as well. Is this intended, and do we want this one to behave in a different way?

etanttila: Also, accessibilty-wise, if a link opens in a new tab, apparently it should say it (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#external_links_and_linking_to_non-html_resources). At least Firefox automatically adds an icon and screen-reader text for the icon if it's defined to open in a new tab (and it looks kind of bad, at least in my opinion):
image

So is it better to have it open in a new tab and be less minimalistic, or have it open in the same tab (at least I automatically click "open in new tab" if opening a link while I don't want the page I'm on to close.

Mankro: This Webropol link should open in a new tab because it takes you out from A+. The privacy and other footer links are still part of the A+ site. 

We already have JS code that adds the external icon to links that open in new tabs. No need to change that.

Mankro: The external link icon is necessary so that the user knows that the link will lead outside the current site. It may look bad, but it is necessary.

etanttila: Ah, all right, so that external icon I was wondering about is from the JS code. And makes sense that the new tab or not is dependent on whether it takes you out of A+ or not. I just didn't realize that that was the "linjaus". (And I suppose that that icon that looks kind of bad would be one more reason to at some point consider having the feedback form embedded into A+)

Mankro: The external icon could be made look better, but not in this PR :)

Mankro (approved):
Great! Please clean up the git history.

templates/base.html
etanttila: Yep, I think it works well in text, etc. but looks kind of strange in menus and the footer.

PR #878, opened 2021-09-28
Adding and modifying courses from API

PasiSa:
# Description

**What?**

Add POST and PUT methods to courses API so that courses can be added and modified using the API. Requires admin rights. See additional information from commit messages.

**Why?**

This was requested by Tampere University, but it could be useful also to us in some form later, when we proceed with integrations with Aalto systems.

**How?**

See the related four commits for details. Also a few basic tests were added (although there would be room for much better test coverage, as always.

# Testing

- [X] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested the API with different correct and incorrect inputs using VScode REST tool.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/api/full_serializers.py
+++ b/course/api/full_serializers.py
@@ -1,3 +1,4 @@
+from typing import OrderedDict
 from rest_framework import serializers, exceptions
 
 from lib.api.fields import NestedHyperlinkedIdentityField
@@ -5,8 +6,10 @@ from lib.api.serializers import AplusModelSerializer, NestedHyperlinkedIdentityF
 from exercise.api.full_serializers import TreeExerciseSerializer
 from exercise.api.serializers import ExerciseBriefSerializer
 from userprofile.api.serializers import UserBriefSerializer, UserListField
+from django.contrib.auth.models import User
 from ..models import (
     CourseInstance,
+    Course,
     CourseModule,
     UserTag,
     UserTagging,
@@ -19,6 +22,7 @@ from .serializers import *
 __all__ = [
     'CourseModuleSerializer',
     'CourseSerializer',
+    'CourseWriteSerializer',
     'CourseStudentGroupSerializer',
     'CourseUsertagSerializer',
     'CourseUsertaggingsSerializer',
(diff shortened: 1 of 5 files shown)


course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+
+        data = request.data
+        form = CourseInstanceForm(data)
+
+        if not form.is_valid():
+            return Response({'status': 'error', 'errors': form.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        formdata = form.cleaned_data
+        try:
+            course: CourseInstance = CourseInstance.objects.get(id=self.kwargs['course_id'])
```
Mankro, line 186: (Just an anecdote) The `update` method might also provide the course instance id in the `pk` parameter (based on the DRF docs). But this works too.

course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+            formdata['id'] = course.id
+            course.instance_name = formdata.get('instance_name')
+            course.url = formdata.get('url')
+            course.language = formdata.get('language')
+            course.starting_time = formdata.get('starting_time')
+            course.ending_time = formdata.get('ending_time')
+            course.visible_to_students = formdata.get('visible_to_students')
+            course.configure_url = formdata.get('configure_url')
+            self.set_teachers(course, data)
```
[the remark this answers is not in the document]
Mankro: The same applies to many other fields. What if the API user wants to preserve the existing values, but modify only one or two fields? In addition, how can you override, e.g., existing ending_time with None if the parameter is excluded in the request data? Your form validation even explicitly states that many fields are not required.

(Input validation should be done with a serializer, not Django forms.)

DRF also provides the `partial_update` method, which is interestingly mapped to HTTP PATCH.
https://www.django-rest-framework.org/api-guide/generic-views/#updatemodelmixin
https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions

Should all fields be required in a PUT request? That could be accompanied with the HTTP PATCH method (`partial_update` class method). Alternatively, it is possible to implement optional fields in the PUT request too. Then, you would not override database fields if they are excluded in the request data.

course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+
+            course.save()
+
+        except CourseInstance.DoesNotExist:
+            return Response('Course instance does not exist.',
```
Mankro, line 200: Does DRF actually call this `update` method if the course id does not exist? It might check existence beforehand based on the ViewSet queryset, but I am not sure what DRF does.

course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+
+        data = request.data
+        form = CourseInstanceForm(data)
+
+        if not form.is_valid():
+            return Response({'status': 'error', 'errors': form.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        else:
+            code = form.cleaned_data.get('code')
+            course: Course = Course.objects.filter(code=code).first()
```
Mankro, line 130: Is it safer to fail fast (without modifying anything) if the course code is not unique? In the worst case, people modify the wrong Course when they don't realize they have duplicate course codes in the database.

course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+                    )
+                except:
+                    return Response('Course creation failed.',
+                                    status=status.HTTP_400_BAD_REQUEST)
+
+            formdata = form.cleaned_data
+            instance_name = formdata.get('instance_name')
+            if (not formdata.get('url')):
+                # if url is not given, compose it from instance name, removing special chars
+                formdata['url'] = ''.join(i for i in instance_name if i.isalnum())
```
Mankro, line 146: Based on docs (https://docs.python.org/3/library/stdtypes.html#str.isalnum), `isalnum` allows much more than the ASCII printing characters. In the URL (path), should we allow Scandinavian or Chinese letters? Do they actually work? I guess they should work correctly, though they would usually be hex encoded. Is that a problem? Other opinions?

At least, this removes whitespace and special characters (that's good).

course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+        if not form.is_valid():
+            return Response({'status': 'error', 'errors': form.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        else:
+            code = form.cleaned_data.get('code')
+            course: Course = Course.objects.filter(code=code).first()
+            if not course:
+                try:
+                    course = Course.objects.create(
+                        name = form.cleaned_data.get('name'),
```
Mankro, line 134: The docs could be more explicit about `name` and `course_url`. They are only used if a new Course is created for the course code, otherwise they are ignored. Existing Course records are never modified in this method, which I suppose is the way we should do this now. The API docs could simply be more explicit.

course/api/views.py
```diff
@@ -64,6 +95,112 @@ def get_queryset(self):
...
+            instance_name = formdata.get('instance_name')
+            if (not formdata.get('url')):
+                # if url is not given, compose it from instance name, removing special chars
+                formdata['url'] = ''.join(i for i in instance_name if i.isalnum())
+            
+            try:
+                new_instance: CourseInstance = CourseInstance.objects.create(
+                    instance_name = instance_name,
+                    url = formdata.get('url'),
+                    language = formdata.get('language'),
```
Mankro, line 152: What happens if the request data excludes some parameters? E.g., language may be blank in the model validation, but does it allow assigning the `None` value here? Assigning the empty string "" when the value is really empty would be more correct concerning the field types.

Mankro (changes requested):
There are issues in the way this implementation uses the DRF (Django REST framework). The greatest concern is using Django Forms instead of DRF Serializers for input validation in the API. I also want to clarify the behaviour of optional fields in the POST and PUT requests.

course/api/views.py
```diff
@@ -32,13 +35,16 @@
...
+    CourseInstanceForm,
+)
 from .serializers import *
 from .full_serializers import *
 
 
 class CourseViewSet(ListSerializerMixin,
                     CourseResourceMixin,
-                    viewsets.ReadOnlyModelViewSet):
+                    viewsets.ModelViewSet):
```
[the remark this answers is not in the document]
Mankro: Even worse, is any logged-in user allowed to delete since the course list has had very lenient permission classes? (FYI, the default is set in settings.py, REST_FRAMEWORK['DEFAULT_PERMISSION_CLASSES'])

course/api/views.py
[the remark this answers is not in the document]
PasiSa: I think PUT is defined to overwrite the resource in its entirety, and PATCH should be used for partial updates ([RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789)). But maybe we leave PATCH for future work if someone really needs it.

course/api/tests.py
```diff
@@ -0,0 +1,150 @@
...
+    setUp = CourseTest.setUp
+
+    def test_get_courselist(self):
+        """
+        Test if list of courses are given correctly via REST.
+        This does not need any authentication.
+        """
+        client = APIClient()
+        client.force_authenticate(user=self.superuser)
+        response = client.get('/api/v2/courses/')
```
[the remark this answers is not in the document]
PasiSa: Good question. The argument for hardcoded URL is, that if URL changes, there may be applications somewhere that will break, so maybe the tests should break too.

PasiSa:
Made a new commit based on using a new serializer for writing, that is intended to replace the earlier implementation based on form. Most of the logic is moved from views.py full_serializers.py. Don't merge yet, will clean up the commit history related to earlier implementation. I also decided to drop the logic that tries to compose url automatically if one is not given, and require the url to be included in request always.

course/api/views.py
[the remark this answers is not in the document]
PasiSa: Permission classes should now be used appropriately, I hope.

course/api/views.py
PasiSa: Decided to give up on building URL automatically, and require that request always includes it. I don't anyone really requested this feature, but it was more an own idea.

course/api/views.py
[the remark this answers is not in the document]
PasiSa: After latest modification, the serializer does not accept request if teachers field is not present. Empty list is allowed.

course/api/full_serializers.py
```diff
@@ -105,6 +109,101 @@ class Meta(CourseBriefSerializer.Meta):
...
+            userprofile: UserProfile = None
+            try:
+                user = User.objects.get(username=i)
+            except User.DoesNotExist:
+                # If user does not exist, create a new user.
+                # If external authentication (e.g. Shibboleth) is used, other
+                # attributes will be updated when user logs in for the first time.
+                user = User.objects.create_user(i)
+
+            userprofile = UserProfile.objects.get(user=user)
```
Mankro, line 149: You don't need to make an explicit query to model UserProfile since Django automatically reverses the OneToOne relationship: `userprofile = user.userprofile`

course/api/full_serializers.py
```diff
@@ -105,6 +109,101 @@ class Meta(CourseBriefSerializer.Meta):
...
+        data = super().validate(data)
+        return data
+
+    def create(self, validated_data: OrderedDict) -> CourseInstance:
+        course = Course.objects.filter(code=self.code).first()
+        if not course:
+            try:
+                course = Course.objects.create(
+                    code = self.code,
+                    name = self.name,
```
Mankro, line 169: Style: no whitespace around `=` in function keyword parameters.

Mankro (approved):
OK! Two minor style issues could be still fixed.

PR #880, opened 2021-10-12
Delayed feedback

### LINKED ISSUE #830: Support for delayed feedback
By default task grading results are shown to student immediately. There should be a configuration option to delay the feedback so that student can see the outcome only after specific time, also in cases when grading happens manually. If feedback is delayed, it should be indicated to the student clearly in the UI, along with which task is considered in grading (typically the last submission).

This feature is needed particularly in exams, but should be available also otherwise.

skulonen:
# Description

**What?**

This PR introduces new course configuration features:
- The time at which exercise results and feedback are revealed to students can now be controlled with a reveal rule. When the results are not revealed, the student cannot see the points they received anywhere in the UI or in the API. The reveal rule allows multiple configuration options for defining the reveal time:
  - IMMEDIATE: results are revealed immediately after the student has submitted and the submission has been graded (this is the default, which means A+ will behave as it previously has unless the reveal rule is explicitly configured).
  - MANUAL: results are never revealed to students unless the teacher manually reveals them.
  - TIME: results are revealed at a specific date and time.
  - DEADLINE: results are revealed when the deadline for the exercise has passed. If the student has any deadline deviations, those also delay the revealing of the results.
  - DEADLINE_ALL: same as the previous one, except every student's deadline deviations are considered.
  - COMPLETION: results are not revealed until the student has used all of their submissions or received full points.
- In addition to exercise results, the reveal time of exercise model solutions can also be configured with a reveal rule. For model solution reveal rules, the default reveal time is DEADLINE.
- The grading mode of an exercise can be switched between BEST and LAST. BEST behaves like A+ has behaved previously: the student's best submission determines their final score for the exercise. In LAST mode, the student's last submission determines the final score. Most often, the designer of the course won't have to explicitly set this option, because the default value depends on the exercise result reveal mode. In IMMEDIATE (default) mode, the default grading mode is BEST. In other modes, the default grading mode is LAST.

**Why?**

This was one of the feature requests from Aalto teachers in 2021.

**How?**

The `submission_feedback_reveal_mode`, `model_solutions_reveal_rule` and `grading_mode` fields have been added to BaseExercise. The first two fields reference a RevealRule model, which contains the configuration described above. The model's methods `is_revealed` and `get_reveal_time` are used by providing a "reveal state" object, that provides methods for calculating whether the rule is revealed or not (deadline, current points, etc). The BaseRevealState class acts as an interface, and the ExerciseRevealState class is the implementation used in this PR. In the future, other reveal states may be implemented for controlling the visibility for other parts of the course.

Most of the work went into ensuring that the results are never revealed to students anywhere in the UI and in the API, when the exercise's results should be hidden. This is why there are now effectively 2 versions of each student's CachedPoints cache: one where points are revealed (initialization parameter `is_staff` is true; used when course staff is viewing a student's results) and one where the points are hidden (normal student UI). In addition, there are changes to templates, including a new `format_points` template tag that displays a question mark instead of the actual points, if the points should be hidden. Additionally, questionnaire exercises can now populate the student's previous answers to the form without also displaying the hints and results received from mooc-grader.

Since there are multiple places where  points are calculated directly with a queryset (CourseAggregateDataViewSet, CourseResultsDataViewSet and ListSubmittersView), and the way points are calculated is now much more complicated, the point calculation has been extracted into its own queryset method, `annotate_submitter_points`.

See also: a-plus-rst-tools PR: apluslms/a-plus-rst-tools#129

Fixes #830

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

All new features have been tested manually. I've checked all views that students have access to, and ensured that hidden points are not leaked in them. Points caching has also been tested extensively, since it has changed significantly (two separate caches as described previously, also the cache can now be invalidated after some time has passed, to support the TIME, DEADLINE and DEADLINE_ALL reveal modes). New unit tests have been written.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/static/exercise/chapter.js
+++ b/exercise/static/exercise/chapter.js
@@ -3,121 +3,121 @@
  *
  */
 ;(function($, window, document, undefined) {
-	"use strict";
+  "use strict";
 
-	var pluginName = "aplusChapter";
-	var defaults = {
-		chapter_url_attr: "data-aplus-chapter",
-		exercise_url_attr: "data-aplus-exercise",
-		active_element_attr: "data-aplus-active-element",
-		loading_selector: "#loading-indicator",
-		message_attr: {
-			load: "data-msg-load",
-			submit: "data-msg-submit",
-			error: "data-msg-error"
-		},
-		modal_selector: "#page-modal",
-	};
+  var pluginName = "aplusChapter";
+  var defaults = {
+    chapter_url_attr: "data-aplus-chapter",
+    exercise_url_attr: "data-aplus-exercise",
+    active_element_attr: "data-aplus-active-element",
+    loading_selector: "#loading-indicator",
+    message_attr: {
+      load: "data-msg-load",
+      submit: "data-msg-submit",
+      error: "data-msg-error"
+    },
+    modal_selector: "#page-modal",
+  };
 
-	function AplusChapter(element, options) {
-		this.dom_element = element;
-		this.element = $(element);
-		this.settings = $.extend({}, defaults, options);
-		this.ajaxForms = false;
-		this.url = null;
-		this.modalElement = null;
-		this.loader = null;
-		this.messages = null;
-		this.quizSuccess = null;
-		this.aeOutputs = {}; // Add active element outputs to chapter so they can be found by id later.
-		this.init();
-	}
+  function AplusChapter(element, options) {
+    this.dom_element = element;
+    this.element = $(element);
+    this.settings = $.extend({}, defaults, options);
+    this.ajaxForms = false;
+    this.url = null;
+    this.modalElement = null;
+    this.loader = null;
+    this.messages = null;
+    this.quizSuccess = null;
+    this.aeOutputs = {}; // Add active element outputs to chapter so they can be found by id later.
+    this.init();
+  }
 
-	$.extend(AplusChapter.prototype, {
+  $.extend(AplusChapter.prototype, {
 
-		/**
-		 * Constructs contained exercise elements.
-		 */
-		init: function() {
-			this.ajaxForms = window.FormData ? true : false;
-			this.url = this.element.attr(this.settings.chapter_url_attr);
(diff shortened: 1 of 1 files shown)


Mankro:
> submission_feedback_reveal_rule, default_submission_feedback_reveal_rule and a third similarly named property on top of these. But I think adding docstrings to the properties would alleviate the problem.

Docstrings sound good!

>>    The submission dropdown menus and the submission feedback dialogs never indicate if only the last submission is really counted or if some submission is manually marked as final by the teacher. It leaves the impression that the best points would be used.

> Do you think there should be an indicator in these places if a submission has been marked as final by the teacher? I think it could be useful, but I'm not sure if it's really related to this PR since it's about the "mark as final" functionality.

No need to do it now in this PR. We can return to it later especially if the teachers request it after seeing it in action.

>>    Though, when the separate "best points" badge is shown, it shows the final points.

> Are you referring to the points badge next to the submission dropdown menu in the top-left corner of the exercise? Yes, that one shows the final points for the exercise.

Yes. And also the points shown in the Exercise results page (which lists all modules). They show the correct final points.

skulonen:
The commit history has been cleaned. There is now basically one big commit and 4 smaller ones. There were relatively few clearly separate new features/fixes that could've been their own commits, unlike in the assessment UI update.

What about the remaining issues? Should I make Github issues for them? Or just start working on the most important ones right now? Here's the list of unresolved issues that have been mentioned on this issue:
- Chapter.js hides messages
- Performance bottleneck in r_reveal_feedback
- Indicate overridden points in submissions dropdown
- Moving max deviation logic into deviation classes
- Moving get_feedback_visibility logic into RevealRule class
- Moving BeautifulSoup code into ExercisePage class
- Refactoring two separate caches into one cache + template tag
- Refactoring points calculation code

PR #911, opened 2021-11-19
CachedPoints duplication and related fixes

### LINKED ISSUE #887: Refactoring the duplication of CachedPoints
(From #880, delayed feedback)

As a result of the delayed feedback update, the `CachedPoints` cache is now duplicated for each student: there is one instance that respects reveal rules and hides points where necessary, and one instance that does not hide points (the latter is used when course staff is viewing the points of a student). The reasoning for this is explained below. In this issue, research if it would be possible to remove this duplication, and instead store two sets of points in the cache (real points, and points where unrevealed exercises are not included) and use a template tag to automatically pick the correct points based on whether the user is staff. Would this complicate the code that uses the cache, or increase the likelihood of accidentally leaking unrevealed points? If the refactoring seems to be worth it, implement it.

---

**Background discussion from #880:**

> Why is it necessary to store two versions of `CachedPoints` for the same student (`is_staff` True/False)? There are other approaches too:
> 
> - store separate fields in the same cache, `real_points` and `display_points` or something like that. The user of the cache needs to choose which one to use. (`real_points` could be called just `points` as before.)
> - just store the real points in the `points` field and require the user of the cache to hide it in some cases.
> 
> You already store `feedback_revealed` and `feedback_reveal_time` fields in the cache, and they would also be useful in the above approaches.
> 
> Why did you choose the approach with the `is_staff` parameter?

> Originally, my changes to `CachedPoints` were much more rudimentary. I basically only stored the `feedback_revealed` and `feedback_reveal_time` values for exercises, and nothing else. This was a problematic approach because there were multiple places in the UI that displayed point sums (e.g. the module headings, the `points_progress` elements). Before delayed feedback, these elements could simply get the `points` value of the container (module/category/whatever) and display that. But now we couldn't do that anymore, because some exercises within the container might use delayed feedback. Originally I had made a new template tag that traversed the cache and calculated the displayed points, but that added unnecessary repetitive computations and made it more complicated to use the cache. So, I decided to move the point calculation logic inside the cache.
> 
> But now that I think about it, I could've moved the calculation code inside the cache without duplicating the cache. I think my idea was to make it easier to use the cache: templates that use it don't need to worry whether to use the `real_points` or `display_points` value, instead you could just use `points` and that would be correct as long as you had initialized the `CachedPoints` instance correctly. In a way, the idea was to reduce the chance for developers to accidentally leak unrevealed points by using the wrong cache field. Although I guess I could've just created a new template tag to automatically pick the correct points to be displayed in the template. I'd be open to refactoring this like that.

skulonen:
# Description

**What?**

`CachedPoints` was duplicated in the delayed feedback update because that was the easiest way to store revealed and unrevealed exercise results simultaneously. This PR fixes the duplication.

Additionally, the PR contains several related bug fixes:
- `InspectSubmitterView` and `NextUnassessedSubmitterView` used the student version of the cache, which means they would redirect the user to the submitter's last submission if delayed feedback was enabled, even though staff members are supposed to be able to know which submission is the best one.
- The "my submissions" list would evaluate the exercise's reveal rule for each submission separately. This was unnecessary because the result would be the same for each submission. This has been optimized.
- The cache key `best_submission` would reveal the best submission through the student cache before feedback was supposed to be revealed. It is now reverted to the last submission in the student cache when feedback is not revealed.
- Some comments were outdated after the "delayed feedback" and "force exercise points" updates.
- The cache keys `feedback_revealed` and `feedback_reveal_time` were missing from submissions.
- The helper functions `has_more_points` and `is_newer` now accept a submission as the second argument, instead of the current cache entry. Because of this, `submission_date` (exercise key) no longer has to be stored in cache.

**Why?**

The duplicated cache took some extra memory. In addition, it would have to be generated twice since there were two versions of it, so this is also a performance improvement.

**How?**

`CachedPoints` now generates the staff and student cache immediately after one another, using the same query results for both passes. Because of this, the generation time is reduced. After both versions have been generated, they are merged by recursively comparing the values in the caches. When there is a conflicting value (e.g. points are revealed in staff cache but not in student cache), a tuple with both values is stored. This completely eliminates the duplication if a course doesn't use delayed feedback at all, and greatly reduces it when it does since only the differing keys have duplicated values. In the `CachedPoints` initializer, only one set of these duplicate keys is left in the dict, while the other set is removed. This has a small runtime cost (~4-5 ms), but makes it easier to use the cache in views and templates (the caller doesn't have to check which value should be used every time).

The "my submissions" list was optimized by allowing an `is_revealed` argument to be passed to the `points_badge` template tag. If it is provided, the reveal rule will not be evaluated by the template tag.

Fixes #887

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

Tested manually that the cache still returns the expected values. In addition, the amount of memory taken by the cache was measured by accessing memcached via telnet.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -19,16 +19,18 @@ from .content import CachedContent
 from .hierarchy import ContentMixin
 
 
-def has_more_points(submission: Submission, current_entry: Dict[str, Any]) -> bool:
-    if submission.grade == current_entry['points']:
-        return is_newer(submission, current_entry)
-    return submission.grade >= current_entry['points']
+def has_more_points(submission: Submission, current_best_submission: Optional[Submission]) -> bool:
+    if current_best_submission is None:
+        return True
+    if submission.grade == current_best_submission.grade:
+        return is_newer(submission, current_best_submission)
+    return submission.grade >= current_best_submission.grade
 
 
-def is_newer(submission: Submission, current_entry: Dict[str, Any]) -> bool:
+def is_newer(submission: Submission, current_best_submission: Optional[Submission]) -> bool:
     return (
-        current_entry['submission_date'] is None
-        or submission.submission_time >= current_entry['submission_date']
+        current_best_submission is None
+        or submission.submission_time >= current_best_submission.submission_time
     )
 
 
@@ -152,7 +154,6 @@ class CachedPoints(ContentMixin, CachedAbstract):
                         'submission_count': 0,
                         'submissions': [],
                         'best_submission': None,
-                        'submission_date': None,
                         'points': 0,
                         'formatted_points': '0',
                         'passed': entry['points_to_pass'] == 0,
@@ -195,6 +196,9 @@ class CachedPoints(ContentMixin, CachedAbstract):
         })
 
         if is_authenticated:
+            # Keep track of the final and last submission of each exercise.
+            final_submissions = {}
+            last_submissions = {}
             # Augment submission data.
             for submission in submissions:
                 try:
@@ -203,6 +207,8 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     self.dirty = True
                     continue
                 entry = tree[-1]
+                current_final_submission = final_submissions.get(entry['id'])
+                current_last_submission = last_submissions.get(entry['id'])
                 ready = submission.status == Submission.STATUS.READY
                 unofficial = submission.status == Submission.STATUS.UNOFFICIAL
                 if ready or submission.status in (Submission.STATUS.WAITING, Submission.STATUS.INITIALIZED):
@@ -222,28 +228,29 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     'unofficial': unofficial,
                     'date': submission.submission_time,
                     'url': submission.get_url('submission-plain'),
+                    'feedback_revealed': True,
+                    'feedback_reveal_time': None,
                 })
-                # TODO: implement way to select algorithm for the best
                 if submission.exercise.grading_mode == BaseExercise.GRADING_MODE.BEST:
                     is_better_than = has_more_points
                 elif submission.exercise.grading_mode == BaseExercise.GRADING_MODE.LAST:
                     is_better_than = is_newer
                 else:
                     is_better_than = has_more_points
-                # Update best submission if one of these is true
+                # Update best submission if exercise points are not forced, and
+                # one of these is true:
                 # 1) current submission in ready (thus is not unofficial) AND
                 #    a) current best is an unofficial OR
-                #    b) current submission has better grade
+                #    b) current submission is better depending on grading mode
                 # 2) All of:
                 #    - current submission is unofficial AND
                 #    - current best is unofficial
-                #    - current submission has better grade
+                #    - current submission is better depending on grading mode
                 if submission.force_exercise_points:
-                    # This submission is chosen as the best submission and no
+                    # This submission is chosen as the final submission and no
                     # further submissions are considered.
                     entry.update({
                         'best_submission': submission.id,
-                        'submission_date': submission.submission_time,
                         'points': submission.grade,
                         'formatted_points': format_points(submission.grade, True, False),
                         'passed': (ready and submission.grade >= entry['points_to_pass']),
@@ -251,26 +258,41 @@ class CachedPoints(ContentMixin, CachedAbstract):
                         'unofficial': False,
                         'forced_points': True,
                     })
+                    final_submissions[entry['id']] = submission
                 if not entry.get('forced_points', False):
                     if (
                         ready and (
                             entry['unofficial'] or
-                            is_better_than(submission, entry)
+                            is_better_than(submission, current_final_submission)
                         )
                     ) or (
                         unofficial and
                         not entry['graded'] and # NOTE: == entry['unofficial'], but before any submissions entry['unofficial'] is False
-                        is_better_than(submission, entry)
+                        is_better_than(submission, current_final_submission)
                     ):
                         entry.update({
                             'best_submission': submission.id,
-                            'submission_date': submission.submission_time,
                             'points': submission.grade,
                             'formatted_points': format_points(submission.grade, True, False),
                             'passed': (ready and submission.grade >= entry['points_to_pass']),
                             'graded': ready, # != unofficial
                             'unofficial': unofficial,
                         })
+                        final_submissions[entry['id']] = submission
+                # Update last submission based on the same logic as best
+                # submission, except only compare submission times regardless
+                # of grading mode.
+                if (
+                    ready and (
+                        entry['unofficial'] or
+                        is_newer(submission, current_last_submission)
+                    ) or (
+                        unofficial and
+                        not entry['graded'] and
+                        is_newer(submission, current_final_submission)
+                    )
+                ):
+                    last_submissions[entry['id']] = submission
                 if submission.notifications.count() > 0:
                     entry['notified'] = True
                     if submission.notifications.filter(seen=False).count() > 0:
@@ -321,6 +343,10 @@ class CachedPoints(ContentMixin, CachedAbstract):
                         'feedback_revealed': is_revealed,
                         'feedback_reveal_time': reveal_time,
                     })
+                    if not is_revealed:
+                        # Revert the best submission to the last one.
+                        last_submission = last_submissions.get(entry['id'])
+                        entry['best_submission'] = last_submission.id if last_submission else None
 
                     for submission in entry['submissions']:
                         submission.update({
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -4,9 +4,6 @@ from typing import Any, Dict, Iterable, List, Optional, Type, Tuple, Union
 
 from django.contrib.auth.models import User
 from django.db.models.base import Model
-from django.db.models.aggregates import Max
-from django.db.models.expressions import Exists, OuterRef
-from django.db.models.query_utils import Q
 from django.db.models.signals import post_save, post_delete, m2m_changed
 from django.utils import timezone
 
@@ -36,13 +33,17 @@ def is_newer(submission: Submission, current_entry: Dict[str, Any]) -> bool:
 
 
 class CachedPoints(ContentMixin, CachedAbstract):
-    KEY_PREFIX = 'points'
+    """
+    Extends `CachedContent` to include data about a user's submissions and
+    points in the course's exercises.
 
-    @classmethod
-    def invalidate(cls, course_instance: CourseInstance, user: User) -> None:
-        # Invalidate both the staff cache and the non-staff cache
-        super().invalidate(course_instance, user, modifiers=[str(False)])
-        super().invalidate(course_instance, user, modifiers=[str(True)])
+    Note that the `data` returned by this is dependent on the `is_staff`
+    parameter. When `is_staff` is `False`, reveal rules are respected and
+    exercise results are hidden when the reveal rule does not evaluate to true.
+    When `is_staff` is `True`, reveal rules are ignored and the results are
+    always revealed.
+    """
+    KEY_PREFIX = 'points'
 
     def __init__(
             self,
@@ -54,8 +55,8 @@ class CachedPoints(ContentMixin, CachedAbstract):
         self.content = content
         self.instance = course_instance
         self.user = user
-        self.is_staff = is_staff
-        super().__init__(course_instance, user, modifiers=[str(is_staff)])
+        super().__init__(course_instance, user)
+        self._extract_tuples(self.data, 0 if is_staff else 1)
 
     def _needs_generation(self, data: Dict[str, Any]) -> bool:
         return (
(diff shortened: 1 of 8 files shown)


skulonen:
This PR supersedes #907.

skulonen:
The current implementation stores duplicated values as tuples, and duplicated values are identified based on their type `tuple` in the `_extract_tuples` method. This works because there are no other tuples stored in the cache. But should we still use some other way to identify duplicated values, so we could theoretically store other tuples in the cache? Duplicated keys like in the original PR? A custom class that derives from `tuple`?

Mankro (changes requested):
Fantastic work! I wonder if we could optimize the memory consumption. I'm not sure if the current way of turning large QuerySets into lists is the most efficient solution.

In addition, the new dictionaries `final_submissions` and `last_submissions` consume some memory. You added them because the submissions are not grouped by the exercise and thus, the submissions loop may jump between exercises back and forth. What if the submissions database query grouped the submissions by the exercise so that you could loop over all submissions in one exercise at once? You wouldn't need to remember all best submissions in the whole course. What do you think? (The last submission would also be obvious if the submissions were sorted, though sorting has a significant cost too.)

Mankro:
> The current implementation stores duplicated values as tuples, and duplicated values are identified based on their type `tuple` in the `_extract_tuples` method. This works because there are no other tuples stored in the cache. But should we still use some other way to identify duplicated values, so we could theoretically store other tuples in the cache? Duplicated keys like in the original PR? A custom class that derives from `tuple`?

I think using tuples is fine. There is currently no obvious need to store tuples as "normal data" in the cache, hence using tuples for a special purpose is perfectly fine. Any additional extra identification field would just consume more memory.

exercise/cache/points.py
```diff
@@ -167,62 +228,78 @@ def r_augment(children: List[Dict[str, Any]]) -> None:
                     'unofficial': unofficial,
                     'date': submission.submission_time,
                     'url': submission.get_url('submission-plain'),
+                    'feedback_revealed': True,
+                    'feedback_reveal_time': None,
```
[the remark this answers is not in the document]
skulonen: Good point and yes, they are the same between all submissions. The main (and I think only) reason I decided to store these for every submission back when I implemented delayed feedback, was `points_badge`. `points_badge` is designed to work with three alternative input types (see `_points_data`): a `UserExerciseSummary` object, a `Submission` object or a CachedPoints dict. The last option does not differentiate between different types of cache entries - it may be an exercise, a submission, a module or a category. If we got rid of these values in the submission dict, the submission dropdowns on the course front page would not know if feedback is revealed or not. We would need extra logic that passes these values to `points_badge` along with the submission dict itself.

Note that these aren't the only values that are redundantly stored for each submission. `max_points` is also the same for all submissions. And this:
```python
'submission_count': 1, # to fool points badge
```

exercise/cache/points.py
```diff
@@ -73,6 +76,68 @@ def _generate_data(
             user: User,
             data: Optional[Dict[str, Any]] = None,
             ) -> Dict[str, Any]:
+        # Perform all database queries before generating the cache.
+        exercises = list(
```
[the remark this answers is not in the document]
skulonen: The memory consumption is a good point. I didn't consider that, and instead only optimized for minimal execution time and minimal number of database queries. The generation code seemed to work ever so slightly faster when using `list()`, so I opted to use it. In addition, I logged all points when the generation code executes a query, and it seemed that when not using `list()`, some queries were executed during the second pass (when `is_staff=False`) because of reveal rules. I thought it might be more efficient to do all database queries upfront.

>I was under the impression that QuerySets cache the results inside the QuerySet object and you could iterate over the same QuerySet twice while loading the data only once from the database. I am not sure, though.

>Then again, is the QuerySet's cache actually in memory, thus either the memory is consumed by this list variable or a QuerySet variable?

Yes, the results are cached in the `QuerySet` object, so I would imagine either the list or the queryset would consume the memory. I'm not sure if it's the same amount of memory, though, since I didn't measure. It might be that the queryset's cache is somehow more optimized?

exercise/cache/points.py
```diff
@@ -73,6 +76,68 @@ def _generate_data(
...
+            .select_related('submission_feedback_reveal_rule')
+            .only('id', 'submission_feedback_reveal_rule')
+        )
+        if user.is_authenticated:
+            submissions = list(
+                user.userprofile.submissions.exclude_errors()
+                .filter(exercise__course_module__course_instance=instance)
+                .select_related()
+                .prefetch_related('exercise__parent', 'notifications')
+                .only('id', 'exercise', 'submission_time', 'status', 'grade', 'force_exercise_points')
```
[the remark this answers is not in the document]
skulonen: Reverse relationship names can't be used in `only()`. I tested and the code queries notifications only once when generating the cache.

exercise/cache/points.py
```diff
@@ -415,6 +491,44 @@ def submission_ids(
...
+            for index, inner_value1 in enumerate(value1):
+                inner_value2 = value2[index]
+                self._pack_tuples(inner_value1, inner_value2, value1, index)
+        else:
+            if value1 != value2:
+                parent_container[parent_key] = (value1, value2)
+
+    def _extract_tuples(self, value, tuple_index, parent_container=None, parent_key=None):
+        """
+        Finds tuples within a data structure, end replaces them with the value
```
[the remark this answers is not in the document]
skulonen: Fixed.

skulonen:
>In addition, the new dictionaries final_submissions and last_submissions consume some memory.

Probably not very much. The submission objects are already stored in the `submissions` list. Since dicts store things by reference, these dicts will only map some integers to some memory addresses.

>You added them because the submissions are not grouped by the exercise and thus, the submissions loop may jump between exercises back and forth.

I added `final_submissions` so that I could pass the current final submission to the `is_better_than` functions, since they now take a submission as the second argument. Previously, the exercise cache entry was passed to those functions, which is why `submission_date` needed to be stored in the exercise cache entry. I added `last_submissions` so that I could set `best_submission` to point to the last submission in case the reveal rule evaluates to false.

>What if the submissions database query grouped the submissions by the exercise so that you could loop over all submissions in one exercise at once? You wouldn't need to remember all best submissions in the whole course. What do you think?

Sure. In that case we wouldn't need to store the final submissions as a dict, and instead it would only be one pointer to one submission object. `last_submissions` would still be needed because reveal rules are evaluated after submissions have been looped. Unless we change the logic so that a single exercise's reveal rule is handled right after after all of its submissions have been looped.

>though sorting has a significant cost too

The submissions are already sorted because of the `ordering` field in the Submission Meta class:

```sql
ORDER BY "exercise_submission"."id" DESC
```

exercise/cache/points.py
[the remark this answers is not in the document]
Mankro: That makes sense. Though, you have already added the `is_revealed` parameter to the `points_badge` template tag, but the parameter isn't always used (like in the course's Your points page). Let's not change this right now. We can leave this as is. It would be possible to avoid repeating these fields for every submission, but the changes would take too much time.

skulonen:
I have now implemented the previously discussed change to the submissions loop. Commit: https://github.com/apluslms/a-plus/pull/911/commits/579fa02cf1115aa97b041458fb5989b2173ae28b

The main changes are:
- The submissions list is now ordered by exercise and then descending submission time.
- No more `final_submissions` and `last_submissions` dicts. Now we have `final_submission` and `last_submission` variables instead.
- Some variables are only updated when a new exercise is encountered (`exercise`, `entry`, `is_better_than`).
- Reveal rules are now checked when the last submission of the exercise has been handled (see `check_reveal_rule` function). Previously, they were handled in their own loop. This has the side effect that reveal rules are no longer checked for exercises that have no submissions, so their score will be shown as "0" instead of "?". This can be seen as a good or a bad thing.
- Deviations are now looped before submissions. This is because we have to know the deviations before reveal rules can be evaluated.

To be honest, I'm not sure if this is better or worse than the original version. In a way, I think the original version seemed more structured because each part was handled in its own, neat block (submissions, then deviations, then reveal rules). Bet now they're a bit intertwined. On the other hand, this seems to be a bit more efficient because a) we don't need to store the two dicts, b) `entry` does not have to resolved for each exercise. The difference is very small, though.

Feel free to tell me what you think. We can keep or revert this commit.

exercise/cache/points.py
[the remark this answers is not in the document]
Mankro: According to your test, the QuerySet object consumes even a little more memory than the simple list. Let's keep the list as is.

exercise/cache/points.py
```diff
@@ -22,28 +29,38 @@
 from .hierarchy import ContentMixin
 
 
-def has_more_points(submission: Submission, current_entry: Dict[str, Any]) -> bool:
-    if submission.grade == current_entry['points']:
-        return is_newer(submission, current_entry)
-    return submission.grade >= current_entry['points']
+# Type for the submission compare functions.
+IsBetterThan = Callable[[Submission, Optional[Submission]], bool]
```
Mankro, line 33: Sometimes I have seen `if TYPE_CHECKING:` before listing static types. I'm just curious if this line would be a typical place to use the `if` or not.

(I am still not very familiar with static typing.)

exercise/cache/points.py
```diff
@@ -133,21 +210,103 @@ def r_augment(children: List[Dict[str, Any]]) -> None:
...
+                    last_submission = None
+                    # These variables stay constant throughout the exercise.
+                    exercise = submission.exercise
+                    try:
+                        tree = self._by_idx(modules, exercise_index[submission.exercise.id])
+                    except KeyError:
+                        self.dirty = True
+                        continue
+                    entry = tree[-1]
+                    if submission.exercise.grading_mode == BaseExercise.GRADING_MODE.BEST:
```
Mankro, line 304: `submission.exercise` could be just `exercise` here.

exercise/cache/points.py
```diff
@@ -133,21 +210,103 @@ def r_augment(children: List[Dict[str, Any]]) -> None:
...
+                entry.update({
+                    'best_submission': entry['best_submission'] if is_revealed else last_submission.id,
+                    'points': entry['points'] if is_revealed else 0,
+                    'formatted_points': format_points(entry['points'], is_revealed, False),
+                    'passed': entry['passed'] if is_revealed else False,
+                    'feedback_revealed': is_revealed,
+                    'feedback_reveal_time': reveal_time,
+                })
+
+                for submission in entry['submissions']:
```
Mankro, line 264: Looking at this loop, I started to think that this data could have been inserted correctly originally (the line 314 `entry['submissions'].append({`). The reveal rule can be defined with the exercise and the exercise is known in the beginning of the submissions loop on line 286. The reveal rule could be used to hide the points when submissions are inserted to the `entry['submissions']` list.

Mankro (approved):
I like the new approach. :+1:  Indeed, this probably affects performance less than what I maybe initially thought, but it's done now.

You said you didn't have time to test this much, so could you still spend a moment testing this? I read the code and it feels correct to me.

I commented on something below that could a very easy, small improvement. If you agree, you could still change that.

You commented on exercises with no submissions, using delayed feedback, and the student sees 0 points instead of ? points:
it makes perfect sense that with no submissions, you get zero points. It's good to show that in the UI. So no problem with that!

exercise/cache/points.py
skulonen: Could be done either way. The reason `if TYPE_CHECKING` is usually used is when we want to do things that have no runtime effect. For example:
- For type hinting purposes, importing classes that only exist in the `django-types` type library but not in Django. (`RelatedManager`)
- For type hinting purposes, importing classes that would cause a cyclical import error at runtime.
- Type hinting model fields that Django adds at runtime (id, reverse relationships).

There isn't really a reason to use it here, but it could still be done because it kind of makes sense - `IsBetterThan` has no runtime purpose, it only exists so I could add a type to the `is_better_than` variable, which was initialized to `None`.

exercise/cache/points.py
skulonen: No, the reveal rule must be evaluated after all submissions have been iterated. This is because the `COMPLETION` reveal trigger depends on the submission count and the final points of the exercise. If that trigger option didn't exist, we could do what you said and evaluate the reveal rule right when a new exercise is encountered.

Mankro: Right you are! Then you may fix the commit history and merge we shall!

exercise/cache/points.py
skulonen: Fixed

exercise/cache/points.py
skulonen: I actually decided to remove this `IsBetterThan` type and other initial types from the submission loop variables because they can be inferred by the IDE, and it was technically wrong to initialize them to `None` while annotating them with another type: https://github.com/apluslms/a-plus/pull/911/commits/723b6f8d18be7dea07f4599cb96a7a68ccfde27a.

Mankro: Sounds good!

skulonen:
I realized that `_generate_data_internal` no longer needs to receive the `exercises` list as a parameter, so I removed that.

I have now cleaned the commit history.

skulonen:
By the way, I think CachedContent could benefit from some optimization too. Generating it takes about 200 ms in the A+ manual course, and according to Django Debug Toolbar, it repeats a lot of queries unnecessarily. On the other hand, CachedContent is invalidated much more rarely than CachedPoints.

Mankro (approved):
Excellent work! Now I merge.

Mankro:
> By the way, I think CachedContent could benefit from some optimization too. Generating it takes about 200 ms in the A+ manual course, and according to Django Debug Toolbar, it repeats a lot of queries unnecessarily. On the other hand, CachedContent is invalidated much more rarely than CachedPoints.

New issue #912

PR #914, opened 2021-11-27
Initial Student Information System integration support

PasiSa:
# Description

**What?**

Support for Student Information System integration (such as SISU in Aalto). When cloning a course, a teacher has option to link the course with a corresponding SIS realization, if SIS is configured and course can be found in SIS (lookup is done based on course code). In this case teachers and start/end time is set based on information received from SIS, and SIS identifier is stored with course instance. When SIS identifier is set, students can be enrolled from SIS participants list. Currently this is only done manually.

The actual SIS interface is organization-specific plugin module (Aalto version is [here](https://github.com/apluslms/aalto-sisu)). The PR also includes sis_test.py plugin module that works as example, and can be used for testing. Also includes a couple of new unit tests using this.

**Why?**

In Aalto all participating students must be enrolled in SISU. This helps keeping A+ enrollments and Sisu enrollment in sync.

**How?**

See above

# Testing

- [X] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested using real SISU and the test plugin, both enrolling and unenrolling students.

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


PasiSa:
Next steps (in separate PRs):

- Enroll new users on first login on all active courses where student has enrolled in SISU
- Periodic (e.g. daily) enrollment checks using Celery

course/staff_views.py
```diff
@@ -145,7 +148,43 @@ class EnrollStudentsView(CourseInstanceMixin, BaseFormView):
     form_class = EnrollStudentsForm
     template_name = "course/staff/enroll_students.html"
 
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs["instance"] = self.instance
+        return kwargs
+
+    def enroll_from_sis(self) -> None:
```
PasiSa, line 156: On a second thought, this should really be a CourseInstance method, and not here

Mankro:
Before looking at the code, I remembered this: have you done anything to enrollment questionnaires? How will the student answer to the background questionnaire if the enrollment is added automatically before filling in the questionnaire?

PasiSa:
> Before looking at the code, I remembered this: have you done anything to enrollment questionnaires? How will the student answer to the background questionnaire if the enrollment is added automatically before filling in the questionnaire?

Nothing is done with enrollment questionnaires yet. I think it needs more thinking and discussion how to do it in Sisu world. It may lead to confusing situations if student is officially enrolled on course in Sisu, but not enrolled in A+ because of missing questionnaire. There could be some sort of "confirmed/unconfirmed" enrollment status, or then the questionnaire could be a blocking condition for opening follow-up assignments, even if technical student status is enrolled. For now teacher still has the option of not connecting a course instance with Sisu if this causes problems, so I propose we solve this as a separate issue a bit later.

edit_course/course_forms.py
```diff
@@ -206,9 +207,27 @@ class CloneInstanceForm(forms.Form):
...
+        sis: StudentInfoSystem = get_sis_configuration()
+        if not sis:
+            # Student Info System not configured
+            return
+        
+        instances = sis.get_instances(self.instance.course.code)
+
+        # If there are no SIS instances by this course code, don't show menu
+        if len(instances) > 0:
+            options = [('none', 'None')] + instances
```
Mankro, line 220: The `'None'` text visible in the UI could be just `'---------'` so that there is no need to translate it. Or you can translate "None".
```suggestion
            options = [('none', '---------')] + instances
```

edit_course/course_forms.py
```diff
@@ -206,9 +207,27 @@ class CloneInstanceForm(forms.Form):
...
+            self.fields['sis'] = forms.ChoiceField(
+                    widget=forms.Select,
```
Mankro, line 222: Select is already the default widget, but it can be specified explicitly too.

edit_course/operations/clone.py
```diff
@@ -81,6 +112,9 @@ def clone(
     if clone_assistants:
         instance.set_assistants(assistants)
 
+    if siskey != 'none':
```
Mankro, line 115:
```suggestion
    if siskey and siskey != 'none':
```
We could also allow an empty `siskey` that would mean that no SIS setup is made like with `'none'`.

edit_course/views.py
```diff
@@ -320,6 +320,8 @@ def get_form_kwargs(self):
 
     def form_valid(self, form):
         from .operations.clone import clone
+        if not 'sis' in form.cleaned_data:
+            form.cleaned_data['sis'] = 'none'
```
Mankro, line 324: If the `clone()` function accepts empty `siskey` values too, then these lines are not needed.
```suggestion
```

course/models.py
```diff
@@ -798,10 +810,40 @@ def enroll_student(self, user):
                     user_profile=user.userprofile,
                     role=Enrollment.ENROLLMENT_ROLE.STUDENT,
                     status=Enrollment.ENROLLMENT_STATUS.ACTIVE,
+                    from_sis=from_sis,
                 )
                 return True
         return False
 
+    def enroll_from_sis(self) -> None:
+        from .sis import get_sis_configuration, StudentInfoSystem
```
Mankro, line 819: Docstring for the function would be nice. The function could have a return value, e.g., the number of students enrolled.

course/staff_views.py
```diff
@@ -145,7 +146,15 @@ class EnrollStudentsView(CourseInstanceMixin, BaseFormView):
...
     template_name = "course/staff/enroll_students.html"
 
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs["instance"] = self.instance
+        return kwargs
+
     def form_valid(self, form):
+        if (self.instance.sis_id and form.cleaned_data["sis"]):
+            self.instance.enroll_from_sis()
```
Mankro, line 156: The teacher should receive some feedback from the SIS enrollment. E.g., "10 students from SIS have been successfully enrolled in the course." or a warning "No students were found in SIS.".
The message can be shown with `messages.info()` and `messages.warning()`

Mankro (changes requested):
This is a great start! I am relieved that this Sisu integration can be implemented in such a simple and minimalistic way. 

I made plenty of very small comments. There are also issues in the error handling that must be fixed.

* There are no instructions on how to install the custom SIS module in production. You could update doc/DEPLOYMENT.md. Where is the module copied to?
* What happens if the SIS API requests are very slow or time out? You should catch exceptions so that A+ does not crash. The exceptions could be maybe caught in the local Sis module since its functions could just return empty lists and log warnings in error cases. Don't forget loggin; it helps when people have to debug the servers.
* `requests.get()`: you should use the timeout parameter so that it doesn't hang for minutes.
   https://docs.python-requests.org/en/latest/user/quickstart/#timeouts

Question:
Can teachers change the `CourseInstance.sis_id` on an existing CourseInstance in any way? Or are we as admins doing it for every course? Maybe you already said so some time ago.

I also reviewed your Aalto SIS module at https://github.com/apluslms/aalto-sisu/

* setup.py does not list the requests package as a dependency, but it is needed.
* You should handle exceptions from `requests.get()`: 
  https://docs.python-requests.org/en/latest/user/quickstart/#errors-and-exceptions
*  `response.json()` raises an exception if the data is not valid JSON. Don't crash the A+ server!
*  `if response.status_code != 200: raise Exception`
  requests also has the method `response.raise_for_status()`.
  https://docs.python-requests.org/en/latest/api/#requests.Response.raise_for_status
* Is `f"{settings.AALTO_SISU_ENHANCED_API_KEY}"` better than `str(settings.AALTO_SISU_ENHANCED_API_KEY)`?

course/models.py
```diff
@@ -790,6 +801,7 @@ def enroll_student(self, user):
                     return False
```
Mankro, line 801: If the student already has an active enrollment, the sis flag is not enabled here during enroll_from_sis. Maybe we should check the sis flag here and enable or disable it for existing enrollments too. Why not flag all sis participants in Aplus too?

course/sis.py
```diff
@@ -0,0 +1,72 @@
...
+    '''
+    Loads a proper SIS plugin based on settings
+
+    Returns
+    -------
+    StudentInfoSystem - inherited object that implements university-specific
+    functionality and APIs for SIS interactions.
+    '''
+    if not hasattr(settings, 'SIS_PLUGIN_MODULE') or not hasattr(settings, 'SIS_PLUGIN_CLASS'):
+        return None
```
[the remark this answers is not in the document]
PasiSa: > Need to set up the logger in the top of this module.

Added this warning, but I thought it is ok not to specify SIS settings, if an installation is not going to use SIS (relates to the settings.py comment above)

aplus/settings.py
```diff
@@ -182,6 +182,11 @@
 SOCIAL_AUTH_URL_NAMESPACE = 'social'
 SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL = True
 
+##########################################################################¨
+# Student Information System plugin settings
+#SIS_PLUGIN_MODULE = 'course.sis_test'
```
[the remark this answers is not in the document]
PasiSa: I left this unchanged for now. get_sis_configuration() should be the only place using this, and on the other hand, I think it is safer to check hasattr() there in any case. I don't think we should use course.sis_test by default, because some installations do not want to have SIS set up, which is the old behavior and therefore good default.

course/models.py
PasiSa: The only current reason for from_sis flag is to support the following scenario: if student has been enrolled based on SIS, and if she later unenroll's in SIS, then A+ enrollment should also be removed, to keep these two in sync. If a student had enrolled directly in A+ (e.g. for being a MOOC student), then she should not be unenrolled based on changes in SIS data.

Ideally local students should not be able to enroll directly when SIS is in use. A follow-up PR adds option to disallow direct enrolling by local students, but when a course has both MOOC and local students, then we need to know the enrollment method to support automatic unenrolling. So in summary: I think we should leave this as it is.

PasiSa:
> This is a great start! I am relieved that this Sisu integration can be implemented in such a simple and minimalistic way.
> 
> I made plenty of very small comments. There are also issues in the error handling that must be fixed.

I made corrections in separate commits and force pushed. If the changes look ok, I can clean up the git history, stashing the three last commits with the appropriate original commits.

Did not yet fix aalto-sis issues, but will do that next.

> Question: Can teachers change the `CourseInstance.sis_id` on an existing CourseInstance in any way? Or are we as admins doing it for every course? Maybe you already said so some time ago.

I don't think teachers should directly touch (or even see) the sis_ids as they are only visible in the API. In the beginning admins may need to manually set some of these, but hopefully not in future, if everything works as expected.

course/models.py
Mankro: I meant that if a student has enrolled locally in A+, but his/her enrollment exists in Sisu too, then the enrollment could be flagged to the Sisu mode in A+. Then, the student would be able to uneroll in Sisu (which is propagated to A+) and if we include only Sisu students in the Sisu grades CSV export, then we need to know all students that exist in Sisu.

If local students are not able to enroll locally in A+ (only in Sisu), then the above scenario would never occur. Unless teachers enroll late students manually in A+ since Sisu has been already closed...

course/sis.py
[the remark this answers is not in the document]
Mankro: If the logger is actually going to make unnecessary warnings when Sisu is not used on purpose, then that wasn't such a great idea from me. If the logger used the debug level, it wouldn't be active all the time in a normal production setup.

Mankro (changes requested):
Thanks! I found a few minor issues that are quick to fix. Then this is ready for rebasing and merging.

`logger.exception()` is useful inside (try-)except blocks.
https://docs.python.org/3/library/logging.html#logging.Logger.exception

course/models.py
PasiSa: ok, accepted that it is better to do as you suggest

doc/DEPLOYMENT.md
```diff
@@ -485,3 +485,27 @@ Final steps
...
+    SIS_PLUGIN_CLASS = 'SisTest'
+
+Variable `SIS_PLUGIN_MODULE` specifies the Python module path name of the plugin,
+and `SIS_PLUGIN_CLASS` is the name of the class implementing the generic
+SIS interface. The proper values for these, along with the plugin installation
+instructions, should be described in the plugin's documentation. See the
+[Aalto SISU plugin](https://github.com/apluslms/aalto-sisu) and its README file
+ for an example. In addition to these parameters, a plugin may have its own
+ plugin-specific additional parameters, as should be described in the plugin
+ installation instructions.
```
Mankro, line 511: These lines start with whitespace for no reason.

doc/DEPLOYMENT.md
```diff
@@ -485,3 +485,27 @@ Final steps
...
 
+Integration with Student Information System
+-------------------------------------------
+
+A+ can be integrated with the organisation's Student Information System (SIS,
+a system that manages information on courses and its participants).
+For example, Aalto University, along with a few other Finnish universities
+use "SISU" for this purpose. The integration with the organisation-specific
+SIS system is done using a SIS plugin. For example, Aalto's SISU plugin
+can be found in [GitHub](https://github.com/apluslms/aalto-sisu).
```
Mankro, line 496: We should add a link to the plugin interface class here. Mentioning plugins and the interface without linking the definition of the interface is confusing.

Mankro (approved):
Thanks for your efforts! The comments below are not that important. Let's merge now, so please squash the commits and rebase onto the upstream/master. For some reason, you have rebased today, but not onto the upstream/master but one commit behind it.

```
# squash/fixup
git rebase -i 2e11bf58
# rebase onto the current upstream/master
git fetch --all
git rebase upstream/master
# and force-push
```

This review got rather pedantic, but it will soon be over!

-----------------------------------

Are you still working on the updating the Aalto Sisu plugin?

Mankro:
For some reason, your branch still wasn't rebased onto the upstream/master. I fixed that and force-pushed before merging.

PR #928, opened 2021-12-07
Save deviation granters and grant times

skulonen:
# Description

**What?**

This PR adds the "granter" and "grant time" columns into the deadline deviation list.

**Why?**

This feature was requested by Aalto teachers.

**How?**

The fields `granter` and `grant_time` have been added to the deviation models, and they are populated when deviations are added by users. Their values are displayed in the deadline deviation list, or "Unknown" if the fields are empty (i.e. the deviation was created before this update).

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

Tested manually that the fields are populated when deviations are added using the web UI, and that the values are displayed correctly in the deviation list.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


exercise/staff_views.py
```diff
@@ -264,9 +264,9 @@ def post(self, request, *args, **kwargs):
             submitter=self.submission.submitters.first(),
             defaults={'extra_submissions': 0}
         )
-        MaxSubmissionsRuleDeviation.objects\
-            .filter(id=deviation.id)\
-            .update(extra_submissions=(F('extra_submissions') + 1))
+        deviation.extra_submissions += 1
+        deviation.granter = request.user.userprofile
+        deviation.save()
```
skulonen, line 269: `QuerySet.update` does not update the `auto_now` fields of models.

deviations/models.py
```diff
@@ -86,6 +86,19 @@ class SubmissionRuleDeviation(UrlMixin, models.Model):
         verbose_name=_('LABEL_SUBMITTER'),
         on_delete=models.CASCADE,
     )
+    granter = models.ForeignKey(UserProfile,
+        verbose_name=_('LABEL_GRANTER'),
+        on_delete=models.CASCADE,
+        related_name='+',
```
skulonen, line 92: `submitter` already used the automatically generated related names (`deadlineruledeviation_set` and `maxsubmissionsruledeviation_set`). I could've used `related_name="granted_%(class)_set"` here to avoid the name conflict, but I'm not sure if these related names would even be used ever, since these fields mostly exist for edit tracking purposes.

Mankro: Disabling the reverse relation is good!

deviations/models.py
```diff
@@ -86,6 +86,19 @@ class SubmissionRuleDeviation(UrlMixin, models.Model):
         verbose_name=_('LABEL_SUBMITTER'),
         on_delete=models.CASCADE,
     )
+    granter = models.ForeignKey(UserProfile,
+        verbose_name=_('LABEL_GRANTER'),
+        on_delete=models.CASCADE,
```
Mankro, line 91: `on_delete` should actually be SET_NULL, right? If the teacher user account is deleted, the student's deviation should still exist. This field even already has `null=True` and `blank=True`.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.SET_NULL

If you agree, you can change it here and also in the migration. Don't generate two separate migrations!

deviations/tests.py
```diff
@@ -12,6 +12,10 @@
 
 class DeviationsTest(TestCase):
     def setUp(self):
+        self.teacher = User(username="staff", is_staff=True)
```
Mankro, line 15: Doesn't really matter, but I doubt `is_staff=True` is needed here. It allows access to the Django admin site. It is not related to A+ or "course staff".

https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.is_staff

exercise/staff_views.py
Mankro: The advantage of `update` with the `F` field reference is that it is more robust in case multiple requests update the value simultaneously because the old value is used directly in the database. Using `+=` and `save()` fetches the old value into Python, but the value may be changed in the database by another request before the +1 value is written in this request from the Python code. Then, some updates are lost. But we can ignore this since it is not likely going to cause any issues.

Mankro (changes requested):
Great job! I just think that the `granter` model field should use `on_delete=SET_NULL` instead of CASCADE so that deviations are not deleted when the teacher's user account is deleted. (They should be deleted when the student's user account is deleted.)

You may take care of the git conflict in django.po.

deviations/tests.py
skulonen: Ok, I just copied the rows related to `self.teacher` from `exercise/tests.py`. I suppose the `is_staff` parameter is not needed there either since that file doesn't test the API.

deviations/models.py
skulonen: I agree, fixed.

deviations/migrations/0004_auto_20211207_1459.py
```diff
@@ -0,0 +1,35 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='deadlineruledeviation',
+            name='grant_time',
+            field=models.DateTimeField(auto_now=True, null=True, verbose_name='LABEL_GRANT_TIME'),
+        ),
+        migrations.AddField(
+            model_name='deadlineruledeviation',
+            name='granter',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='+', to='userprofile.userprofile', verbose_name='LABEL_GRANTER'),
```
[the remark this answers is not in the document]
skulonen: Fixed

deviations/migrations/0004_auto_20211207_1459.py
```diff
@@ -0,0 +1,35 @@
...
+    dependencies = [
+        ('userprofile', '0006_auto_20210812_1536'),
+        ('deviations', '0003_auto_20210812_1536'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='deadlineruledeviation',
+            name='grant_time',
+            field=models.DateTimeField(auto_now=True, null=True, verbose_name='LABEL_GRANT_TIME'),
```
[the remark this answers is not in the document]
skulonen: Granter and grant time will be None for pre-existing deviations (I tested). I think this makes sense because we don't know when the old deviations were created.

skulonen:
I fixed the merge conflicts, squashed the SET_NULL fix and force pushed.

Mankro (approved):
Excellent!

PR #929, opened 2021-12-09
Use datetime-local in datetime form fields

skulonen:
# Description

**What?**

This PR adds a datetime picker to all datetime fields in forms.

**Why?**

This was mentioned when discussing the deviation UI with Aalto teachers.

**How?**

A new widget, `DateTimeLocalInput`, has been added. This new widget is otherwise the same as the built-in `DateTimeInput`, except it uses the `datetime-local` HTML input type, which uses a datetime picker provided by the browser. `datetime-local` expects the initial value of the field to be in a specific format, which the widget ensures. `datetime-local` is supported in all major browsers except for IE. If the user's browser doesn't support this widget type, it uses `text` instead (which the A+ forms used before this update anyway).

# Testing

- [x] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

Tested manually that the initial value is populated to the field correctly (in the course configuration form). Tested that the entered value is saved correctly.

- [x] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -1427,7 +1427,7 @@ msgstr "New deadline"
 
 #: deviations/forms.py
 msgid "DEVIATION_NEW_DEADLINE_DATE_HELPTEXT"
-msgstr "New submission deadline in the future in format YYYY-MM-DD HH:MM."
+msgstr "New submission deadline."
 
 #: deviations/forms.py deviations/models.py
 msgid "LABEL_WITHOUT_LATE_PENALTY"
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -1431,7 +1431,7 @@ msgstr "Uusi määräaika"
 
 #: deviations/forms.py
 msgid "DEVIATION_NEW_DEADLINE_DATE_HELPTEXT"
-msgstr "Uusi määräaika formaatissa YYYY-MM-DD HH:MM."
+msgstr "Uusi määräaika."
 
 #: deviations/forms.py deviations/models.py
 msgid "LABEL_WITHOUT_LATE_PENALTY"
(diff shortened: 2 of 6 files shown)


locale/en/LC_MESSAGES/django.po
```diff
@@ -1382,7 +1382,7 @@ msgstr "New deadline"
 
 #: deviations/forms.py
 msgid "DEVIATION_NEW_DEADLINE_DATE_HELPTEXT"
-msgstr "New submission deadline in the future in format YYYY-MM-DD HH:MM."
+msgstr "New submission deadline in the future."
```
Mankro, line 1385:
```suggestion
msgstr "New submission deadline."
```
I don't think it is necessary to emphasize "future" here. If the user tries to enter an invalid date in the past, then the form validation should show a descriptive error.

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1386,7 +1386,7 @@ msgstr "Uusi määräaika"
 
 #: deviations/forms.py
 msgid "DEVIATION_NEW_DEADLINE_DATE_HELPTEXT"
-msgstr "Uusi määräaika formaatissa YYYY-MM-DD HH:MM."
+msgstr "Uusi määräaika tulevaisuudessa."
```
Mankro, line 1389:
```suggestion
msgstr "Uusi määräaika."
```

Mankro (changes requested):
Awesome! I just thought that the label text could be more compact.

skulonen:
I've shortened the labels as you wished.

> If the user tries to enter an invalid date in the past, then the form validation should show a descriptive error.

I agree, but I don't think this is directly related to this PR. There are other issues related to that form's validation, too: it doesn't check that you've entered at least one exercise or module, and when you haven't entered a date or a minute number, it sends you back to the deviation list and shows the error message at the top of the page. These issues could be addressed in PR #401 instead.

skulonen:
I rebased this and force-pushed.

Mankro (approved):

PR #933, opened 2021-12-10
Fix PR #913 issues

lainets:
# Description

**What?**

Fixes the outstanding issues with PR #913. 

**Why?**

That PR needed to rolled out quickly, so the rest of the issues are fixed in this PR.

# Is it Done?


doc/AUTH.md
```diff
@@ -1,14 +1,14 @@
...
 
 Every request A+ makes to graders and other services contain a JWT
-authentication/authorization token. Any request to A+ that requires
-authentication must also contain a JWT. This does not include submission
-grading, as then it includes a token in the GET parameters. Though, this token
-can be sent as part of the JWT's tokens field.
+authentication/authorization token. Any request to A+ by graders that requires
+authentication must also contain a JWT or a exercise/submission token, that A+
+has supplied to the grader, in the GET parameters. Though, this token can also
+be sent as part of the JWT's tokens field.
```
Mankro, line 7:
```suggestion
authentication must also contain one of the following tokens:

1. JWT in the HTTP Authorization header (new preferred method).
1. Exercise/submission token in the GET query parameters.
   A+ supplies this token to the exercise service as part of the `submission_url` parameter in the grader protocol.
   Using this token is the old method and not recommended any longer.
   This token can also be sent as part of the `tokens` field of the JWT.
```
A list makes the differences very clear!

Mankro (changes requested):
Good! I made a suggestion for the AUTH.md text. Do you agree with it?

Mankro (approved):

PR #937, opened 2021-12-14
Add permissions to remote page accesses

lainets:
# Description

**What?**

Send the instance id permission with requests to co

**Why?**

Newest Git Manager requires permissions on all endpoints.

**How?**

Populate permissions on requests.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Seems to work.

- [X] Chrome
- [ ] Firefox

If it works on one browser, it works on all.

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lib/remote_page.py
+++ b/lib/remote_page.py
@@ -1,11 +1,14 @@
+from bs4 import BeautifulSoup
 import logging
 import posixpath
 import re
 import requests
+from requests.models import Response
 import time
-from bs4 import BeautifulSoup
+from typing import Optional
 from urllib.parse import urlparse, urljoin
 
+from aplus_auth.payload import Permission, Permissions
 from django.conf import settings
 from django.utils.http import parse_http_date_safe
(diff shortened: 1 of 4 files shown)


lib/remote_page.py
```diff
@@ -1,11 +1,14 @@
+from bs4 import BeautifulSoup
```
Mankro, line 1: The imports were previously grouped correctly, but not anymore. Python built-in modules should be separated from third-party modules.

https://www.python.org/dev/peps/pep-0008/#imports

Mankro (approved):

lib/remote_page.py
lainets: I only changed the order of the imports and not the grouping. Didn't even think about whether they should be grouped differently as well. Fixed in PR #943

PR #939, opened 2021-12-15
Table ordering and grouping

skulonen:
# Description

**What?**

This PR adds the following optional features to the AplusTableFilter jQuery plugin: column ordering and row grouping. Column ordering is now enabled on all tables that had filtering enabled. Grouping by user and module is enabled in the deviation list.

**Why?**

Column ordering and row grouping for the deviation list were requested by Aalto teachers.

**How?**

***Ordering***

Column ordering was originally implemented in `participants.js`, which was unfortunate because it was specific to the participant list page. In this PR, I moved the ordering logic from `participants.js` to AplusTableFilter. It can be enabled by adding the `ordered-table` class to any table (much like filtering is enabled by adding the `filtered-table` class). Ordering can be disabled for individual columns by adding the `data-order-disable="true"` attribute to the `th` element.

***Grouping***

Row grouping uses the concept of parent rows and child rows. Parent rows and child rows are linked by using `data-group-parent` and `data-group-child` attributes, where the parent row and its child rows have a common attribute value (a group id). The Javascript code does not decide these group ids, they must be predetermined in Python code and added to templates. The deviation list view is an example of this.

Parent rows can be expanded to show a group of child rows beneath them. When grouping is enabled (by adding a `grouped-table` class to a table), a expand/collapse button is added next to all parent rows, and a button that expands/collapses all groups is added to the top-left corner of the table. Grouping works together with ordering: the top-level rows are ordered first, and then the child rows are ordered group-by-group and moved below their parent rows.

---

Here is a screenshot of all 3 features (filtering, ordering, grouping) in action:
image

# Testing

- [x] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

Tested manually that all features work together in the deviation list table. Tested that rows were grouped together when they should be, and that they were not grouped in these cases:
- One exercise in the module has a different deviation
- One exercise in the module doesn't have a deviation

The selenium tests did not pass, but it seems that they were already broken in an earlier update.

- [x] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


assets/js/filter.js
```diff
@@ -209,7 +298,7 @@ $(function() {
             } else {
                 input.off('hide.bs.popover');
                 input.popover("destroy")
-                    .removeAttr("class title data-original-title");
+                    .removeAttr("title data-original-title");
```
skulonen, line 301: Not sure why `class` was originally in the list of attributes to be removed. It caused the input box to lose its styling after the popover was closed, so I removed it from the list.

deviations/templates/deviations/list_dl.html
```diff
@@ -22,40 +22,90 @@
...
+							<span class="glyphicon glyphicon-remove" aria-hidden="true"></span>
+							{% translate "REMOVE" %}
+						</button>
+					</form>
+				</td>
+			</tr>
+			{% endfor %}
+			{% empty %}
+			<tr>
+				<td colspan="5">{% translate "NO_DEADLINE_DEVIATIONS" %}</td>
```
skulonen, line 105: This should be `colspan="7"` now. I'll fix this soon.

deviations/views.py
```diff
@@ -16,18 +20,77 @@ class ListDeadlinesView(CourseInstanceBaseView):
...
 
-    def get_common_objects(self):
+    def get_common_objects(self) -> None:
         super().get_common_objects()
-        self.deviations = (
+
+        self.deviation_groups = self.get_deviation_groups()
+        self.note("deviation_groups")
+
+    def get_deviation_groups(self) -> Iterable[Tuple[List[DeadlineRuleDeviation], bool, Optional[str]]]:
```
skulonen, line 29: I spent a lot of time trying different methods for grouping the deviations by user and module (template tags, query trickery, etc), and the current method that uses `itertools.groupby` is definitely both the fastest (it barely has any performance effect with 2000 deviations) and cleanest way in my opinion.

skulonen:
Question: should we rename the AplusTableFilter plugin, since it now has multiple different table-related features and not just filtering?

assets/js/filter.js
```diff
@@ -1,11 +1,32 @@
 $(function() {
```
skulonen, line 1: This file's indentation is incorrect. I'll fix it after the review.

skulonen:
Found a possible bug:

image
image

In the second image, I have filtered the table by granter. This has an unintuitive interaction with grouping: the parent row is hidden, because the granter field is empty for that row. However, the child rows are still visible because they match the filter. Conversely, if the group was not expanded before typing the filter, neither the parent row nor the child rows are visible after filtering, even though the children match the filter.

I think a better behaviour would be: when typing a filter, if a child row matches the filter, also display its parent row. Additionally, if the group is collapsed at the moment of filtering, expand it.

Mankro:
> Question: should we rename the AplusTableFilter plugin, since it now has multiple different table-related features and not just filtering?

Maybe yes. Is it easy to do?

course/templates/course/staff/participants.html
```diff
@@ -51,21 +51,21 @@
...
+                <th>{% trans "STUDENT_ID" %}</th>
+                <th>{% trans "LAST_NAME" %}</th>
+                <th>{% trans "FIRST_NAME" %}</th>
                 <th>{% trans "EMAIL" %}</th>
-								<th>{% trans "STATUS" %}</th>
-                <th>{% trans "TAGS" %}</th>
-								<th></th>
+				<th data-filter-type="none" data-order-disable="true">{% trans "STATUS" %}</th>
+                <th data-filter-type="none" data-order-disable="true">{% trans "TAGS" %}</th>
+				<th data-filter-type="none" data-order-disable="true"></th>
```
Mankro, line 68: Mixed tabs and spaces? Looks like the file used spaces in the past, though the style guide prefers tabs in HTML now.

deviations/templates/deviations/list_dl.html
```diff
@@ -22,40 +22,90 @@
...
+						<span class="glyphicon glyphicon-ok"></span>
+						{% translate "YES" %}
+					{% else %}
+						<span class="glyphicon glyphicon-remove"></span>
+						{% translate "NO" %}
+					{% endif %}
+				</td>
+				<td></td>
+				<td></td>
+				<td></td>
```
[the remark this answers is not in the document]
Mankro: Sounds good!

deviations/templates/deviations/list_dl.html
```diff
@@ -22,40 +22,90 @@
...
+			{% for deviations, can_group, group_id in deviation_groups %}
+			{% if can_group %}
+			<tr data-group-parent="{{ group_id }}">
+				<td>{{ deviations.0.submitter }}</td>
+				<td>
+					{{ deviations.0.exercise.course_module }}
+					<strong>
+						({% blocktranslate trimmed with count=deviations|length %}
+						NUM_OF_EXERCISES -- {{ count }}
+						{% endblocktranslate %})
```
Mankro, line 54: Django translations support pluralization based on `count`: 1 exercise, 2 exercises

https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#pluralization

Mankro (commented):
Wondeful! On a very quick glance, this already looks good.

Does the table group deviations from multiple students into the same group if the deviations are otherwise identical?

skulonen:
> Does the table group deviations from multiple students into the same group if the deviations are otherwise identical?

No, they are always grouped by user and module, although what you asked here is probably doable. If it did group deviations from multiple students, how should such a group appear in the table? Multiple students listed in the "submitter" cell? What if the same deviation has been given to 200 students? Maybe the cell could say "Stacy Student and 199 others", and the user can expand the group to see the individual students?

deviations/templates/deviations/list_dl.html
skulonen: True, but a group must always have more than one deviation (and therefore more than one exercise).

Mankro:
> > Does the table group deviations from multiple students into the same group if the deviations are otherwise identical?
> 
> No, they are always grouped by user and module, although what you asked here is probably doable. If it did group deviations from multiple students, how should such a group appear in the table? Multiple students listed in the "submitter" cell? What if the same deviation has been given to 200 students? Maybe the cell could say "Stacy Student and 199 others", and the user can expand the group to see the individual students?

I was asking since I wasn't sure what it does. Don't change anything due to that question. Sounds like the current functionality is good.

skulonen:
>>Question: should we rename the AplusTableFilter plugin, since it now has multiple different table-related features and not just filtering?

>Maybe yes. Is it easy to do?

Yeah, looks like the name only appears in filter.js. If we do this, filter.js should probably also be renamed, which would also be easy since the file is only referenced in base.html.

Mankro:
> > > Question: should we rename the AplusTableFilter plugin, since it now has multiple different table-related features and not just filtering?
> 
> > Maybe yes. Is it easy to do?
> 
> Yeah, looks like the name only appears in filter.js. If we do this, filter.js should probably also be renamed, which would also be easy since the file is only referenced in base.html.

Ok, go ahead! Better names will make it more understandable to future developers.

course/templates/course/staff/participants.html
skulonen: The indentation in that file was all kinds of messed up. I fixed it now.

deviations/templates/deviations/list_dl.html
skulonen: Fixed

skulonen:
I pushed some new commits. The weird interaction with filtering and grouping has been fixed: the parent row is now always visible if its group contains at least one child that is not hidden by a filter.

I also renamed the plugin from AplusTableFilter to AplusTable, and the js file from filter.js to table.js. Unfortunately, it seems like Github can't show the history of the file past the rename. I don't know if the rename is worth it if we would lose easy access to the file's history.

assets/js/filter.js
skulonen: Fixed

skulonen:
I fixed the indentation of table.js, rebased and cleaned the commit history. This is ready to be merged.

skulonen:
There was still some weird indentation in the script at the bottom of participants.html. I fixed that and force-pushed.

Mankro (approved):
Excellent work again! :+1:

PR #942, opened 2022-01-03
Fix 404 when clicking "start manual assessment"

skulonen:
# Description

**What?**

The "start manual assessment" button, as well as the "inspect" button in the submission list when grouped by submitter, would cause a 404 error if none of the chosen user's submissions were marked as `best_submission` in the points cache. This PR fixes the bug.

**Why?**

Shouldn't return 404 when there actually are submissions.

**How?**

The two views in question used the method `CachedPoints.submission_ids` to get the best submission of the chosen user. This method returned an empty list when called with `best=True` and none of the submissions were considered "best" by the points cache's logic. A new optional parameter, `fallback_to_last` has been added to the method. It ensures that *some* submission is returned even when there is no "best" submission.

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested manually by setting a user's all submissions to the WAITING state, and pressing the "start manual assessment" and "inspect" buttons in the grouped submission list. A 404 error was no longer given, and instead the user's last submission was opened.

The unit tests were already broken by a previous commit: https://github.com/apluslms/a-plus/runs/4690502793?check_suite_focus=true

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -483,6 +483,7 @@ class CachedPoints(ContentMixin, CachedAbstract):
             exercise_id: Optional[int] = None,
             filter_for_assistant: bool = False,
             best: bool = True,
+            fallback_to_last: bool = False,
             raise_404=True,
             ) -> List[int]:
         exercises = self.search_exercises(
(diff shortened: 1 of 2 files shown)


Mankro (commented):
This is a bit suspicious, so I have a question first. Is it really necessary that CachedPoints returns `best_submission` `None` when some submissions really exist? The generate method checks the submission status READY before updating the best submission, but what if it accepted also the statuses INITIALIZED or WAITING for the best submission? Why couldn't the current best submission be a submission with zero points and in the WAITING status? The student would have zero points at that time anyway, so the best submission has zero points. The best submission will be updated when the submission is really graded.

Mankro (approved):
We agreed that for now, the `best_submission` in `CachedPoints` is not changed. If all submissions are in the WAITING state, then best_submission is None. The inspect page uses the new `fallback_to_last` parameter to find the latest submission even if best submission is None when some submissions exist.

PR #943, opened 2022-01-03
Fix get-token 500 error and PR #937 issue 

lainets:
# Description

**What?**

No more 500 error on loading get-token page, and fixes the import grouping for remote_page.py.

**Why?**

500 isn't correct behaviour.

**How?**

Return 400 Bad Request instead, as it makes most sense.

# Is it Done?


Mankro (approved):
:+1:

PR #944, opened 2022-01-04
Allow INITIALIZED and WAITING submissions to be the best submission in CachedPoints

skulonen:
# Description

**What?**

This PR changes CachedPoints to allow INITIALIZED and WAITING submissions to be the best submission, instead of just READY.

**Why?**

The same change was already done to UserExerciseSummary in https://github.com/apluslms/a-plus/commit/87dae322fd9777f707db2d0279ccfd1258bf0587, which changed how points are displayed on chapter pages. After this PR, the same change is visible on the course results page (front page).

image

In the above image, the exercise's grading mode is LAST, which causes the 2nd submission to be the "best" submission. Before this PR, the exercise's points would be shown as 10/10, because the 2nd submission's state is WAITING.

**How?**

In all points in the CachedPoints code where submissions were restricted to READY only, all states except UNOFFICIAL are now allowed (ERROR and REJECTED were already filtered out before the main submission loop).

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested manually that the front page's points now correspond to the chapter pages.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -335,7 +335,7 @@ class CachedPoints(ContentMixin, CachedAbstract):
                         'best_submission': submission.id,
                         'points': submission.grade,
                         'formatted_points': format_points(submission.grade, True, False),
-                        'passed': (ready and submission.grade >= entry['points_to_pass']),
+                        'passed': (not unofficial and submission.grade >= entry['points_to_pass']),
                         'graded': True,
                         'unofficial': False,
                         'forced_points': True,
@@ -343,7 +343,7 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     final_submission = submission
                 if not entry.get('forced_points', False):
                     if (
-                        ready and (
+                        not unofficial and (
                             entry['unofficial'] or
                             is_better_than(submission, final_submission)
                         )
@@ -356,14 +356,14 @@ class CachedPoints(ContentMixin, CachedAbstract):
                             'best_submission': submission.id,
                             'points': submission.grade,
                             'formatted_points': format_points(submission.grade, True, False),
-                            'passed': (ready and submission.grade >= entry['points_to_pass']),
-                            'graded': ready, # != unofficial
+                            'passed': (not unofficial and submission.grade >= entry['points_to_pass']),
+                            'graded': not unofficial,
                             'unofficial': unofficial,
                         })
                         final_submission = submission
                 # Update last_submission to be the last submission, or the last
                 # official submission if there are any official submissions.
-                # Note that the submissions are ordered by descendng time.
+                # Note that the submissions are ordered by descending time.
                 if last_submission is None or (
                     last_submission.status == Submission.STATUS.UNOFFICIAL
                     and not unofficial
--- a/exercise/tests.py
+++ b/exercise/tests.py
@@ -587,15 +587,22 @@ class ExerciseTest(TestCase):
         self.learning_object_category.accept_unofficial_submits = True
         self.learning_object_category.save()
 
+        self.submission_when_late_allowed.set_error()
+        self.late_late_submission_when_late_allowed.set_error()
+        self.submission_when_late_allowed.save()
+        self.late_late_submission_when_late_allowed.save()
+
         self.late_submission_when_late_allowed.set_points(10, 10)
         self.late_submission_when_late_allowed.set_ready()
         self.late_submission_when_late_allowed.save()
         self.assertEqual(self.late_submission_when_late_allowed.grade, 100)
         self.assertEqual(self.late_submission_when_late_allowed.status, Submission.STATUS.UNOFFICIAL)
         summary = UserExerciseSummary(self.base_exercise_with_late_submission_allowed, self.user)
-        self.assertEqual(summary.get_submission_count(), 3)
+        self.assertEqual(summary.get_submission_count(), 1)
         self.assertEqual(summary.get_points(), 0) # unofficial points are not shown here
         self.assertFalse(summary.is_graded())
+        # 3 submissions:
+        # initialized, unofficial, initialized
         self.assertTrue(summary.is_unofficial())
 
         self.submission_when_late_allowed.set_points(5, 10)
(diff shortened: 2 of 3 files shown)


skulonen:
This PR would change how some fields in the public REST API are defined, which may or may not be ok. Here's a list of the fields throughout the codebase that would change their meanings:

- `PointsCache` exercise:
	- `passed`:
		- Before this PR: best submission's `status == READY and grade >= points_to_pass`
		- After this PR: best submission's `status != UNOFFICIAL and grade >= points_to_pass`
	- `graded`:
		- Before this PR: best submission's `status == READY`
		- After this PR: best submission's `status != UNOFFICIAL`
- `UserExerciseSummary`:
	- `is_passed`:
		- Before Markku's previous commit: best submission's `status == READY and grade >= points_to_pass`
		- After Markku's previous commit: best submission's `status != UNOFFICIAL and grade >= points_to_pass`
	- `is_graded`:
		- Before Markku's previous commit: best submission's `status == READY`
		- After Markku's previous commit: best submission's `status != UNOFFICIAL`
- `ExercisePointsSerializer` (public API):
	- `passed`:
		- Directly from `PointsCache` `passed`
		- Before this PR: best submission's `status == READY and grade >= points_to_pass`
		- After this PR: best submission's `status != UNOFFICIAL and grade >= points_to_pass`
	- `official`:
		- Influenced by `PointsCache` `graded`
		- Before this PR: best submission's `status == READY and not unconfirmed`
		- After this PR: best submission's `status != UNOFFICIAL and not unconfirmed`

exercise/tests.py
```diff
@@ -587,15 +587,22 @@ def test_unofficial_submission(self):
...
         self.assertEqual(self.late_submission_when_late_allowed.grade, 100)
         self.assertEqual(self.late_submission_when_late_allowed.status, Submission.STATUS.UNOFFICIAL)
         summary = UserExerciseSummary(self.base_exercise_with_late_submission_allowed, self.user)
-        self.assertEqual(summary.get_submission_count(), 3)
+        self.assertEqual(summary.get_submission_count(), 1)
         self.assertEqual(summary.get_points(), 0) # unofficial points are not shown here
         self.assertFalse(summary.is_graded())
+        # 3 submissions:
+        # initialized, unofficial, initialized
         self.assertTrue(summary.is_unofficial())
```
skulonen, line 606: In this test, there were 3 submissions with the states INITIALIZED, UNOFFICIAL, INITIALIZED. `assertTrue` on line 599/606 would fail because the initialized submissions can now "de-unofficialize" the summary. To be in line with the test's original intentions, I decided to set the states of the INITIALIZED submissions to ERROR, so the assertion would succeed. Perhaps not the cleanest way to update this test? It was broken by the earlier commit.

skulonen: The comments on lines 604-605 are temporary (this is a draft PR). But perhaps these tests could use some extra comments? Without any comments it's pretty annoying to figure out what submissions there are and what is being tested at which point.

exercise/tests_cache.py
```diff
@@ -141,10 +141,13 @@ def test_invalidation(self):
 
     def test_accumulation(self):
         self.submission2.set_points(2,2)
+        self.submission2.set_error()
         self.submission2.save()
         c = CachedContent(self.instance)
         p = CachedPoints(self.instance, self.student, c)
         entry,tree,_,_ = p.find(self.exercise)
+        # 2 submissions:
+        # ready 50, initialized 100
```
skulonen, line 150: In this test, there were two submissions: a READY one with 50 points and an INITIALIZED one with 100 points. `assertEqual` on line 150/153 would fail because the second submission is now considered the best one. This is why I decided to set its state to ERROR so it wouldn't be included in the calculation.

exercise/cache/points.py
```diff
@@ -335,15 +335,15 @@ def check_reveal_rule() -> None:
                         'best_submission': submission.id,
                         'points': submission.grade,
                         'formatted_points': format_points(submission.grade, True, False),
-                        'passed': (ready and submission.grade >= entry['points_to_pass']),
+                        'passed': (not unofficial and submission.grade >= entry['points_to_pass']),
```
Mankro, line 338: Is it possible that `passed` = `True` when submission.status in (REJECTED, ERROR) and `points_to_pass` = 0? Letting a rejected submission be "passed" sounds suspicious?

exercise/cache/points.py
```diff
@@ -356,14 +356,14 @@ def check_reveal_rule() -> None:
                             'best_submission': submission.id,
                             'points': submission.grade,
                             'formatted_points': format_points(submission.grade, True, False),
-                            'passed': (ready and submission.grade >= entry['points_to_pass']),
-                            'graded': ready, # != unofficial
+                            'passed': (not unofficial and submission.grade >= entry['points_to_pass']),
+                            'graded': not unofficial,
```
Mankro, line 360: If all submissions are in the INITIALIZED status, then it is strange to call the exercise "graded" since the submissions really have not been graded yet.

Mankro (commented):
I have some questions below. This feels complicated since the old API for these values has been inconsistent or illogical. We will need to change the data visible through the REST API a bit (passed or graded fields etc.), though, in order to fix the clear bugs.

exercise/cache/points.py
skulonen: No, it's not possible. REJECTED and ERROR are already filtered out before the main submission loop: https://github.com/apluslms/a-plus/blob/c9208d7d917ab86ef5b907e4d17390199e4e1235/exercise/cache/points.py#L91-L92

exercise/cache/points.py
skulonen: So do you think the definition of "graded" should remain unchanged (best submission's status == READY)? In that case, I'll make the same change to UserExerciseSummary as well.

What about "passed"? Do you think it's strange to call an exercise "passed" if all of its submissions are INITIALIZED?

exercise/tests.py
skulonen: Yes, and I could write a couple more comments to further describe what the test is supposed to test.

Mankro:
Closing this PR because it was replaced by #948:
https://github.com/apluslms/a-plus/pull/948#issuecomment-1008847617

PR #946, opened 2022-01-07
Fix couple small JWT access issues

lainets:
# Description

**What?**

Fix couple small JWT access issues.

**Why?**

MOOC Jutut couldn't access an exercise when the module was not open.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Programming style

# Is it Done?

### DIFF
--- a/course/permissions.py
+++ b/course/permissions.py
@@ -226,7 +226,7 @@ class EnrollInfoVisiblePermission(ObjectVisibleBasePermission):
         return True
 
 
-class CourseModulePermission(MessageMixin, Permission):
+class CourseModulePermissionBase(MessageMixin, Permission):
     message = _('MODULE_PERMISSION_MSG_NOT_VISIBLE')
 
     def has_permission(self, request, view):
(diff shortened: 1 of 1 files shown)


course/permissions.py
```diff
@@ -226,7 +226,7 @@ def is_object_visible(self, request, view, course_instance):
         return True
 
 
-class CourseModulePermission(MessageMixin, Permission):
+class CourseModulePermissionBase(MessageMixin, Permission):
```
Mankro, line 229: Does this mean that students are able to read hidden modules by using the API and JWT tokens? Normally, students can not see hidden modules nor any content in them.

Mankro (commented):
One question below.

If there is no issue in access control, then the code is good for merging.

course/permissions.py
lainets: Right now, only teachers or graders/git manager may get a JWT access to a course, so students can't read them. The JWT read access is more akin to administrative read (e.g. the build logs in git manager) than to content read (e.g. course chapter files). I do not see a particular reason to expand the use of the tokens to students as well. Though, it might make sense to use the 'write' permission for all administrator-like purposes.

Mankro: OK, since no secret information is exposed to the students, this is good. :+1:

Mankro (approved):

PR #947, opened 2022-01-09
Initial mass regrade implementation (for comments)

PasiSa:
# Description

**What?**

Initial version of mass regrade implementation. At least the UI needs yet improvements, but I made the PR now, if we happen to have urgent need for this feature after the recent server failures. I'm doubtful about few things here:

* Is it really a good idea to store the Celery task ID in the database? I feel DB should hold more persistent attributes. For example, if the server is restarted in the middle of Celery run, the DB gets into inconsistent state. I did some maneuvers to recover from such scenario, though. Is there an alternative way for remembering the task ID during its existence? (memcache or something)
* The grade method has been using HttpRequest to build URI for grader. However, Celery task does not have one. I worked around this by storing the URI prefix when Celery task is initiated and use that in later grade requests. The current implementation feels clunky however. Is there a better approach for this?

I'd like to get some opinions about the above (and other) aspects. Therefore this is just a draft PR now. The UI translations are not yet done, because the UI is not considered final. Some nicer progress view should be implemented.

There is also a "Simulated" option in the Regrade menu to make it easier to test larger tasks locally. It should be removed before final merge.

**Why?**

Much requested feature.

**How?**

See above.

Fixes #<issue number> (I don't think there was issue for this specific feature)

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Generated few submissions locally and tested.

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


exercise/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+    # FIXME: "Simulated" type is used for testing with arbitrarily higher loads.
+    # It will be removed from the final PR, along with the "iterations" variable.
+    iterations = 1
+    if regrade_type == 'simulated':
+        iterations = 15
+        total = total * iterations
+
+    count = 0
+    while iterations > 0:
+        for i in qs:
```
Mankro, line 32: `i` is a submission so shouldn't the variable be named `submission`? `i` is often associated with an integer variable in loops.

-> use `submission`

exercise/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+from .exercise_models import BaseExercise
+
+logger = logging.getLogger('aplus.exercise')
+
+@app.task
+def regrade_exercises(exerciseid: int, uri_prefix: str, regrade_type: str) -> None:
+    exercise = BaseExercise.objects.get(pk=exerciseid)
+
+    qs = exercise.submissions\
+        .defer("feedback", "submission_data", "grading_data")
```
Mankro, line 16:
```suggestion
    qs = (exercise.submissions
        .defer("feedback", "assistant_feedback", "submission_data", "grading_data")
    )
```

https://apluslms.github.io/contribute/styleguides/python/#indentation-line-length-and-blank-lines

exercise/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if regrade_type == 'incomplete':
+        qs = qs.filter(Q(status='INITIALIZED') | Q(status='WAITING') | Q(status='ERROR'))
```
Mankro, line 19: Don't duplicate constant values here by hardcoding. `status = Submission.STATUS.INITIALIZED` etc.

exercise/templates/exercise/staff/regrade_submissions.html
```diff
@@ -0,0 +1,10 @@
+{% load i18n %}
+{% load static %}
+
+<div class="col-md-12">
+    {% if result %}
+        Mass regrade in progress: state: {{ result.state }} -- info: {{ result.info }}
```
Mankro, line 6: Strings to translate.

exercise/submission_models.py
```diff
@@ -367,12 +367,19 @@ def get_post_parameters(
...
-        self.exercise.as_leaf_class().modify_post_parameters(
-            self._data, self._files, user, students, request, url)
+
+        # When running mass regrade, we do not have applicable request, and in such case
+        # it is set to None. Therefore, in mass regrade cases we are not modifying post
+        # parameters. This means that submissions with file attachments or LTI interaction
+        # are not properly covered. TODO: Check what is appropriate way of handling this.
+        if request:
+            self.exercise.as_leaf_class().modify_post_parameters(
+                self._data, self._files, user, students, request, url)
```
Mankro, line 381: Is it impossible to regrade submissions that include uploaded files? Another reason why the `exercise.grade()` method should not depend on HTTP requests.

exercise/staff_views.py
```diff
@@ -64,7 +67,25 @@ def get_common_objects(self):
...
+        # For the time being only admins can start mass regrade.
+        # This condition may be loosened later on
+        if self.request.user.is_superuser:
+            self.allow_regrade = True
+        else:
+            self.allow_regrade = False
+
+        if self.exercise.regrade_task:
+            result = AsyncResult(self.exercise.regrade_task)
+            if result.state == 'FAILURE':
```
Mankro, line 81: Does Celery have any constants for these constant-looking values `"FAILURE"` or `"PROGRESS"`?

exercise/staff_views.py
```diff
@@ -271,6 +292,28 @@ def post(self, request, *args, **kwargs):
...
+    template_name = "exercise/staff/regrade_submissions.html"
+
+    def get_common_objects(self) -> None:
+        self.regrade_type = self.request.GET.get('regrade', None)
+
+        if self.exercise.regrade_task:
+            self.result = AsyncResult(self.exercise.regrade_task)
+            self.note("result")
+        else:
+            job = regrade_exercises.delay(
```
Mankro, line 308: Are you creating a new background task in an HTTP GET request? Very suspicious. In principle, the database is never modified in GET requests except for some kind of logging. There are good reasons for that. GET is a so-called "safe" method and people assume it is safe to use (idempotent). 

At any rate, `get_common_objects` is not the right method for creating new Celery tasks.

-----------------------------------------------

I remember an old blog post that describes how Google Search bot crawler regularly deleted one site's whole database because items were deleted with GET requests that have query param `action=delete`. Google just followed links with GET requests like they are supposed to do.

exercise/exercise_models.py
```diff
@@ -533,6 +533,11 @@ class BaseExercise(LearningObject):
         default=GRADING_MODE.BEST,
         help_text=_('BASE_EXERCISE_GRADING_MODE_HELPTEXT'),
     )
+    regrade_task = models.CharField(
```
Mankro, line 536: One option would be to define the task identifier in a new model so that it doesn't pollute the existing exercises. If the field is not really an attribute of (~all) exercises, then why should it be part of the same model?

If there were a new model class, it would refer to the exercise with a ForeignKey.

exercise/exercise_models.py
```diff
@@ -962,9 +967,12 @@ def get_load_url(
             )
         return super().get_load_url(language, request, students, url_name, ordinal)
 
-    def grade(self, request, submission, no_penalties=False, url_name="exercise"):
+    def grade(self, request, submission, uri_prefix=None, no_penalties=False, url_name="exercise"):
         """
         Loads the exercise feedback page.
+        Note: request may be None if grading is initated by async Celery task,
+        i.e., for mass regrade. In such cases uri_prefix is used to compose
+        absolute URI for the grader.
```
Mankro, line 975: Now is the time to refactor this grade method so that it can be used without a `request` parameter. I mean the whole `request` parameter is removed. We also have an old issue in building the A+ server URL in the wrong way. We should not trust the request to know the correct name for the server. It is a problem when multiple domains are used (aalto.fi, hut.fi, IP address used directly) and it's always problematic in localhost (Docker internal domain aplus versus the localhost address that is usable in the web browser). In addtition, trusting the request is a security risk since the requests can be forged (source IP address, Host HTTP request header).

The only URL/domain for the server that we can trust is the one we define manually in the server settings (local_settings.py). 

The request variable has been used in many places to build the A+ server's URL, and all of those should be fixed eventually.

Mankro (changes requested):
There are multiple severe issues in this code. As you noticed, using the request variable in the parameters of model methods is just a bad idea that, for example, makes it impossible to run normal processes without an HTTP request (like in Celery background tasks). Rafactoring the `grade` method to remove the whole request variable requires more work.

You also seem to make database changes in GET requests, which breaches the concept of safe, idempotent HTTP methods (GET, HEAD, OPTIONS).

Read this:
https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP

Mankro:
> Is it really a good idea to store the Celery task ID in the database? I feel DB should hold more persistent attributes. For example, if the server is restarted in the middle of Celery run, the DB gets into inconsistent state. I did some maneuvers to recover from such scenario, though. Is there an alternative way for remembering the task ID during its existence? (memcache or something)

The database is the only persistent storage. Perhaps caches could be viable in this case. Django has its own cache API. We use Memcached in production, but Django supports other backends too. However, we can't trust the cache to store these values for hours. We can't really trust them at all, so if it is important that the value is not lost too early, then using caches is not really an option.

https://docs.djangoproject.com/en/3.2/topics/cache/

File system is also permanent storage, but is it really better to store these Celery task identifiers in files when we have a database? In the end, database is also a bunch of files.

exercise/staff_views.py
PasiSa: It seems no, at least all examples I could find use strings.

PasiSa:
Made entirely new pull request #960 that should address these comments. Therefore closing this.

PR #949, opened 2022-01-10
Permission fixes for GraderUser

lainets:
# Description

**What?**

Fixes GraderUser permissions related to Jutut access.

**Why?**

Jutut either couldn't access exercises or it crashed A+.

**How?**

Check that the user is not a GraderUser in normal permissions, and add exercise read permission to exercise tokens.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Programming style

# Is it Done?

### DIFF
--- a/course/permissions.py
+++ b/course/permissions.py
@@ -17,7 +17,7 @@ from authorization.permissions import (
 from exercise.cache.points import CachedPoints
 from exercise.models import BaseExercise
 from exercise.submission_models import Submission
-from userprofile.models import UserProfile
+from userprofile.models import GraderUser, UserProfile
 from .models import (
     Course,
     CourseModule,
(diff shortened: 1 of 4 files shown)


course/permissions.py
```diff
@@ -130,6 +130,9 @@ def is_object_visible(self, request, view, course):
         """
         # NOTE: course is actually course instance
 
+        if isinstance(request.user, GraderUser):
```
Mankro, line 133: What has changed compared to A+ v1.11? I assume that GraderUser was never before using this permission, so why is it now here?

Mankro (approved):
Good, thanks! I can merge this now. I had one question which I would still like to understand better,

course/permissions.py
lainets: It was using it, the GraderUser was just always assumed to be a teacher, so it got skipped on the next if statement.

Mankro: So what happened if GraderUser is now not counted as a teacher?

lainets: Being a teacher implies write permissions to a course, so we do not assume that the GraderUser is always a teacher. They are only counted as a teacher if they have write permissions to the course (listed in the JWT)

Mankro: Thanks for the clarification!

PR #950, opened 2022-01-11
Clarifying the Final label in Inspect submissions view

PasiSa:
The Final label is used in Inspect submissions' submission list
also for submissions that are not marked as final. This modification
uses a different label when Final is not appropriate. It also
clarifies whether the best or last submission is chosen for grading.

# Description

**What?**

See above. This is issue EDIT-768 in Jira.

**Why?**

As described in Jira

**How?**

[ANSWER HERE]

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested different combinations with few submissions by same user.

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/templates/exercise/staff/_submissions_table_compact.html
+++ b/exercise/templates/exercise/staff/_submissions_table_compact.html
@@ -56,11 +56,19 @@
 				</span>
 
 				{% if s == summary.best_submission %}
-				<span
-					class="label label-primary"
-					data-toggle="tooltip"
-					title="{% translate 'SUBMISSION_FINAL_TOOLTIP' %}"
-				>{% translate 'FINAL' %}</span>
+					{% if s.force_exercise_points %}
+					<span
+						class="label label-primary"
+						data-toggle="tooltip"
+						title="{% translate 'SUBMISSION_FINAL_TOOLTIP' %}"
+					>{% translate 'FINAL' %}</span>
+					{% else %}
+					<span
+						class="glyphicon glyphicon-star"
+						data-toggle="tooltip"
+						title="{% translate 'SUBMISSION_CHOSEN_TOOLTIP' %}"
+					></span>
+					{% endif %}
 				{% endif %}
 				{% if s.grader %}
 				<span
--- a/exercise/templates/exercise/staff/inspect_submission.html
+++ b/exercise/templates/exercise/staff/inspect_submission.html
@@ -81,6 +81,9 @@
 						</form>
 					{% endif %}
 				{% endif %}
+				{% if not submission.force_exercise_points %}
+					{{ grading_mode_text }}
+				{% endif %}
 				{% include "exercise/staff/_submissions_table_compact.html" %}
 			</div>
 		</div>
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -3090,6 +3090,10 @@ msgid "MODEL_NAME_REVEAL_RULE_PLURAL"
 msgstr "Reveal rules"
 
 #: exercise/staff_views.py
+msgid "GRADING_MODE_TITLE -- {}"
+msgstr "Grading mode: {}"
+
+#: exercise/staff_views.py
 msgid "ASSESS_SUBMISSION_REVIEW_SAVED_SUCCESS"
 msgstr "The review was saved successfully and the submitters were notified."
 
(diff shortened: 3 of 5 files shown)


exercise/templates/exercise/staff/inspect_submission.html
```diff
@@ -81,6 +81,13 @@
 						</form>
 					{% endif %}
 				{% endif %}
+				{% if not submission.force_exercise_points %}
+					{% if exercise.grading_mode == 1 %}
+						{% translate "GRADING_MODE_BEST_EXPLAIN" %}
+					{% elif exercise.grading_mode == 2 %}
+						{% translate "GRADING_MODE_LAST_EXPLAIN" %}
+					{% endif %}
```
skulonen, line 89: I don't think it's a good idea to hardcode enum values into the template, but I guess it's not possible to access the enum directly from the template. Could you generate the explanation string in the view code instead? Or maybe we could make a generic template tag that compares enum values by their string representation instead? Something like: `exercise.grading_mode|enum_equal:"BEST"`

locale/en/LC_MESSAGES/django.po
```diff
@@ -3867,6 +3871,14 @@ msgstr "Assess next submitter manually"
...
 msgstr "Add one extra submission for the student in this exercise."
 
+#: exercise/templates/exercise/staff/inspect_submission.html
+msgid "GRADING_MODE_BEST_EXPLAIN"
+msgstr "Choosing highest grade"
+
+#: exercise/templates/exercise/staff/inspect_submission.html
+msgid "GRADING_MODE_LAST_EXPLAIN"
+msgstr "Choosing latest submission"
+
```
skulonen, line 3881: Is it clear what "choosing" means? To me, "choosing" makes it sound like the user is expected to choose something.

exercise/templates/exercise/staff/_submissions_table_compact.html
```diff
@@ -56,11 +56,19 @@
...
+						class="label label-primary"
+						data-toggle="tooltip"
+						title="{% translate 'SUBMISSION_FINAL_TOOLTIP' %}"
+					>{% translate 'FINAL' %}</span>
+					{% else %}
+					<span
+						class="glyphicon glyphicon-star"
+						data-toggle="tooltip"
+						title="{% translate 'SUBMISSION_CHOSEN_TOOLTIP' %}"
+					></span>
```
skulonen, line 70: This isn't a huge issue, but I think the gray star next to a colorful label looks a bit awkward. Could it perhaps be part of the points label or have its own colored label next to it?

skulonen:
The logic seems valid to me. I wrote some nitpicks about things that I didn't like, but overall this looks good to me.

exercise/templates/exercise/staff/inspect_submission.html
PasiSa: True. Now choosing the text on the view side and just passing the string

locale/en/LC_MESSAGES/django.po
PasiSa: On a second thought I decided to use the existing translations "Grading mode" and "Best/Last Submission", which should be clear enough for the course staff. Anyway these texts were my own idea, and not a teacher request, so we could just leave it out as well.

exercise/templates/exercise/staff/_submissions_table_compact.html
PasiSa: Now including the star inside a blue label. Any better?

locale/en/LC_MESSAGES/django.po
skulonen: Nice! I like this a lot.

exercise/templates/exercise/staff/inspect_submission.html
skulonen: Looks good!

exercise/templates/exercise/staff/_submissions_table_compact.html
skulonen: To be honest I think this looks way worse :joy:
The star is off center and the label's height doesn't match the points label next to it. Plus it looks like the Somalian flag. I think we should go with your earlier version, it looked good enough.

Mankro (changes requested):
This is good otherwise, but the way translated strings are concatenated does not support all languages. It is quick to change that, thus it could still be done before merging.

PasiSa:
> This is good otherwise, but the way translated strings are concatenated does not support all languages. It is quick to change that, thus it could still be done before merging.

Done requested changes with one deviation: I used "Grading mode: {}" instead of proposed "{}: {}" which would have felt a bit awkward to me, despite the existing "Grading mode" model label. But then again, who knows, perhaps in some exotic language the model label text would not be usable as such in this situation anyway.

Mankro (approved):
Great, thanks!

PR #953, opened 2022-01-14
Fix Jutut access

lainets:
# Description

**What?**

Add exercise read permissions to submission tokens.

**Why?**

Otherwise Jutut requests are denied.

**How?**

Add permissions in grader.py on authentication.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lib/api/authentication/grader.py
+++ b/lib/api/authentication/grader.py
@@ -46,9 +46,18 @@ class GraderAuthentication(ServiceAuthentication[GraderUser], BaseAuthentication
     def add_token_permissions(self, token: str, permissions: ObjectPermissions, payload: Payload):
         if token[0] == "s":
             submission = self.authenticate_submission_token(token[1:])
+            payload.permissions.submissions.add(Permission.READ, id=submission.id)
+            permissions.submissions.add(Permission.READ, submission)
             payload.permissions.submissions.add(Permission.WRITE, id=submission.id)
             permissions.submissions.add(Permission.WRITE, submission)
             exercise = submission.exercise
+
+            # MOOC-jutut needs this but it doesn't seem correct to give exercise read
+            # permissions here. It might not _actually_ even need this but the way
+            # the permissions system works, accessing the submission also requires
+            # access to the exercise.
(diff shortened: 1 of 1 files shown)


lib/api/authentication/grader.py
```diff
@@ -46,9 +46,18 @@ def authenticate(self, request: Request) -> Optional[Tuple[GraderUser, Payload]]
...
+            permissions.submissions.add(Permission.READ, submission)
             payload.permissions.submissions.add(Permission.WRITE, id=submission.id)
             permissions.submissions.add(Permission.WRITE, submission)
             exercise = submission.exercise
+
+            # MOOC-jutut needs this but it doesn't seem correct to give exercise read
+            # permissions here. It might not _actually_ even need this but the way
+            # the permissions system works, accessing the submission also requires
+            # access to the exercise.
+            payload.permissions.exercises.add(Permission.READ, id=exercise.id)
```
Mankro, line 59: Now that both if branches set exercise READ permission, these two lines could moved from the if-elif branches to the end with the course permissions.

Mankro (approved):
Thanks for the quick solution! I don't think it is wrong to grant exercise READ permissions to students who already have permission to create submissions. 

In normal life, how would a student write solutions to exercises if they were not allowed to see the exercise description? :)

lainets:
> Thanks for the quick solution! I don't think it is wrong to grant exercise READ permissions to students who already have permission to create submissions.

AFAIK the tokens are not meant for students to use, and, as I have said before, the READ permission is an administrative read permission i.e. "let me read everything about this exercise", not "allow me to read the exercise description" (though, I'm not sure if there is a difference between these at the moment). The submission token in question is supposed to be used just to modify an existing submission. Though, I do acknowledge that maybe read permissions could be given to the exercise but this should be clearly documented so everyone knows what the tokens are allowed to do.

This doesn't invalidate the fact that you shouldn't need permissions to access the whole course though.

PR #958, opened 2022-01-20
Deviation usability improvements

skulonen:
# Description

**What?**

This PR brings the old PR #401 up to date with the current A+ version, and implements several new features on top of the old PR.

New features from the old PR:
- When adding new deviations, if some of the deviations already existed (same submitter and exercise), the user is taken to the deviation override page where they can pick and choose which deviations will be overridden with the new values, and which ones will keep their original values.
  - The deviation override table has all the same functionality as the deviation list table (filtering, ordering, grouping).
- Multiple deviations can be removed using the deviation removal form.

New features implemented in this PR:
- The deviation user interfaces are now accessible from the course sidebar instead of the Edit Course page.
- The deadline deviation list now includes new deadlines instead of just the minutes.
- The submitter column now links to the submitter's results page.
- Deviations can now be added (and removed) by providing one or more user tags, in which case the deviations are granted to (or removed from) all users with one of the tags.
- Deviations can now be removed by selecting them using checkboxes on the deviation list page and then clicking "Remove selected deviations".
- Minutes for deadline deviations can now be entered by entering days, hours and  minutes separately.

**Why?**

These features were requested by Aalto teachers.

**How?**

When A+ detects that some of the new deviations already existed, the input form data is stored in the session cache. The deviation override page reads the session cache to populate the table with the overlapping deviations. Template tags are used for calculating the new deadlines/minutes on the deadline deviation override page. Otherwise, the override page uses the same logic for rendering and grouping the table as the deviation list page.

The user tag searching works exactly like when a module is selected on the "add deviations" page. All users with one of the tags are selected, and any users in the "submitter" field are selected in addition to those with the tags.

A new "synchronized checkboxes" feature was implemented in table.js. It makes the checkbox on the parent row of a group in a table react to its child rows: it's checked if all children are checked, unchecked if all children are unchecked, and "indeterminate" if some but not all children are checked. In addition, all children can be checked or unchecked by clicking the parent checkbox. This feature is used in the "override" checkboxes on the deviation override page, and the checkboxes that are used to select deviations to be deleted on the deviation list page.

A new widget `DurationInput` and a new field `DurationField` were implemented to facilitate entering different units of time for deadline deviations. The widget displays a set of input boxes, one for each unit (days, hours, minutes by default), and the field combines their values into a single integer (minutes) when the form is submitted.

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [x] Manual testing.

All new features have been tested manually. Many new unit tests were written for the new functionality.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


skulonen:
The commit history is a disaster at the moment, but I think I'll clean it after the final comments from the teachers, in case we still need to make some significant changes.

assets/js-translations/teacher.fi.json
```diff
@@ -141,5 +141,8 @@
   "Vieritä erikseen",
 
   "Download":
-  "Lataa"
+  "Lataa",
+
+  "Are you sure you want to remove the selected deviation? |||| Are you sure you want to remove the %{smart_count} selected deviations?":
+  "Haluatko varmasti poistaa valitun poikkeaman? |||| Haluatko varmasti poistaa valitut %{smart_count} poikkeamaa?"
```
skulonen, line 147: Looks like this syntax wasn't used in A+ yet, but it works because translate.js is based in polyglot.js.

deviations/forms.py
```diff
@@ -1,37 +1,77 @@
...
         label=_('LABEL_EXERCISE'),
-        help_text=_('DEVIATION_EXERCISE_HELPTEXT'),
         widget=SearchSelect,
     )
+    submitter_tag = forms.ModelMultipleChoiceField(
+        queryset=UserTag.objects.none(),
+        required=False,
+        label=_('LABEL_SUBMITTER_TAG'),
+        widget=SearchSelect(display_fields=['slug']),
+    )
```
skulonen, line 35: The user tag field doesn't contain the hardcoded user tags ("Aalto" and "MOOC") because they aren't real tags that are stored in the database and thus not included in the queryset. Including them might be tricky.

deviations/templates/deviations/_remove_modal.html
```diff
@@ -0,0 +1,64 @@
...
+					value="{% translate 'CANCEL' %}"
+					data-dismiss="modal"
+				/>
+				<input
+					id="remove-selected-submit-button"
+					class="aplus-button--danger aplus-button--md"
+					type="submit"
+					value="{% translate 'REMOVE' %}"
+					formaction="{{ remove_url }}"
+				/>
```
skulonen, line 39: The reason I added `formaction` to the remove modal's submit button instead of adding `action` to the form, is that we might want to use the selection checkboxes for something other than deletion in the future. In that case we would likely add another modal with a submit button with a different `formaction`.

deviations/templates/deviations/add_dl.html
```diff
@@ -1,28 +1,29 @@
-{% extends "edit_course/edit_course_base.html" %}
+{% extends "course/course_base.html" %}
```
skulonen, line 1: I made a lot of small changes to the templates because the deviation pages are no longer in the Edit Course menu but in the sidebar. Different base template, different view tags, edittitle -> title, editbreadcrumblist -> breadcrumblist, coursecontent -> columns.

deviations/templates/deviations/_remove_modal.html
skulonen: Also note that the selected deviations are no longer removed with AJAX even though that was originally implemented in #401. I changed that because:
- The user can now delete multiple deviations at a time using the checkboxes, so the page reload is not a big deal anymore.
- The deviations are grouped in the backend. Reloading allows us to regroup the deviations when some of them are deleted.

deviations/viewbase.py
```diff
@@ -0,0 +1,295 @@
...
+            exercise__in=exercises,
+            submitter__in=submitters,
+        )
+
+        if existing_deviations:
+            # Some deviations already existed. Use OverrideDeviationsView to
+            # confirm which ones the user wants to override. Store the form
+            # values in the current session, so they can be used afterwards.
+            self.success_url = self.deviation_model.get_override_url(self.instance)
+            self.request.session[self.session_key] = self.serialize_session_data(form.cleaned_data)
```
skulonen, line 57: There were several options for what to store in the session cache:
- Input form data (this is what I did)
- "Expanded" input form data, i.e. module IDs are expanded to exercise IDs and user tag IDs are expanded to user IDs
- The new deviation objects as JSON
- List of overlapping deviation IDs and new values from the form data (this is what the old PR did)

I ended up going with the first option because it keeps the session cache relatively small and the code mostly straightforward. I didn't like what the old PR did, because it assumed that the database table would stay unchanged between submitting the "add" and "override" forms. My solution is more "stateless".

exercise/exercise_models.py
```diff
@@ -651,7 +652,7 @@ def delta_in_minutes_from_closing_to_date(self, future_date):
         # -> convert future_date to same format as module_close
         string_date = str(future_date)[:16]
         converted = timezone.make_aware(
-                datetime.datetime.strptime(string_date, '%Y-%m-%d %H:%M'),
+                parse_datetime(string_date),
```
skulonen, line 655: I changed this to `parse_datetime` so we wouldn't need to have this hardcoded format string here, plus the format string was not consistent with the way dates are sent from HTML forms (they use the T-format).

skulonen:
TODO: deviation tests failing after the latest push?

deviations/templates/deviations/_remove_modal.html
```diff
@@ -0,0 +1,64 @@
...
+
+			const selected = form.find('input[name="id"]:checked');
+			$("#remove-selected-no-selection-warning").toggleClass("hidden", selected.length > 0);
+			$("#remove-selected-confirmation")
+			.toggleClass("hidden", selected.length === 0)
+			.text(
+				_("Are you sure you want to remove the selected deviation? |||| Are you sure you want to remove the %{smart_count} selected deviations?",
+				{ smart_count: selected.length })
+			);
+			$("#remove-selected-submit-button").attr("disabled", selected.length === 0);
```
Mankro, line 61: This looks suspicious. The `disabled` attribute is a boolean attribute and it does not use any value. If you set a value, the element is still disabled even if the value is "false" or similar.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

Should this be treated like a property?
https://api.jquery.com/prop/

Mankro (approved):
Excellent and truly remarkable work!

What about the commit history? Does this new PR include any significant code from the old PR #401? If not, it is better to remove the old PR's commits completely from the history. If there is something useful, then in order to give credit to the original author, there needs to be at least one commit with the original author. Does it make sense to squash all #401 commits into one? Or completely remove part of the #401 commits?

deviations/models.py
```diff
@@ -150,6 +177,30 @@ def get_new_deadline(self, normal_deadline: Optional[datetime] = None) -> dateti
...
+        else:
+            minutes = int(minutes)
+        self.extra_minutes = minutes
+        self.without_late_penalty = bool(form_data.get('without_late_penalty'))
+
+    def is_equal(self, other: 'DeadlineRuleDeviation') -> bool:
+        return (
+            self.extra_minutes == other.extra_minutes
+            and self.without_late_penalty == other.without_late_penalty
+        )
```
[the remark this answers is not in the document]
skulonen: Yes, I thought about this myself when coding this. It's true that methods named like "is_equal" or "equals" are often used for overloading the `==` operator. Maybe a better name would be `is_groupable` or `is_equivalent`?

deviations/templates/deviations/_remove_modal.html
skulonen: It seems that `.attr("disabled", true)` adds an attribute like `disabled="disabled"` and `.attr("disabled", false)` removes the attribute. `prop` works the same way except the attribute it adds looks like `disabled=""`. So it works either way, however, the jQuery documentation states that

> To retrieve and change DOM properties such as the checked, selected, or disabled state of form elements, use the .prop() method.

so I guess I'll change it to use `prop`.

skulonen:
> What about the commit history? Does this new PR include any significant code from the old PR #401?

Yes, the new views and templates, as well as the logic of transfering data between the "add" and "override" views is based on the old PR, although it has been heavily modified by me.

> If not, it is better to remove the old PR's commits completely from the history. If there is something useful, then in order to give credit to the original author, there needs to be at least one commit with the original author. Does it make sense to squash all #401 commits into one? Or completely remove part of the #401 commits?

I intend to leave at least one commit from the old PR. However, the old PR's commit history is also pretty messy and could use some squashing. If I squash some of the old PR's commits, do they become "my" commits?

deviations/models.py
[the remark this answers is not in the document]
Mankro: `is_groupable` is good since it refers to the purpose of the method, grouping under a module.

Mankro:
> > What about the commit history? Does this new PR include any significant code from the old PR #401?
> 
> Yes, the new views and templates, as well as the logic of transfering data between the "add" and "override" views is based on the old PR, although it has been heavily modified by me.
> 
> > If not, it is better to remove the old PR's commits completely from the history. If there is something useful, then in order to give credit to the original author, there needs to be at least one commit with the original author. Does it make sense to squash all #401 commits into one? Or completely remove part of the #401 commits?
> 
> I intend to leave at least one commit from the old PR. However, the old PR's commit history is also pretty messy and could use some squashing. If I squash some of the old PR's commits, do they become "my" commits?

Squashing preserves the author of the base commit onto which the other commits are squashed. In addition, it is easy to change the author if needed. In an interactive rebase with the "edit" action, you would run `git commit --amend --author="Firstname Surname <name@aalto.fi>".

skulonen:
@Mankro I've cleaned the commit history and rebased this branch onto master. I think this is ready to be merged.

Mankro (approved):

PR #966, opened 2022-02-03
Fix rendering of binary files in submitted file views

### LINKED ISSUE #965: A+ must not try to render binary files in plain text in the submitted file views
Student's submitted files are usually opened in a modal dialog for a preview of the contents. In the teacher's inspect submission page, the preview is within the page without a modal dialog. The preview works for plain text files. A+ does not recognize all binary file types, thus A+ tries to render previews for some binary files too. That results in bad user experience because the preview shows only garbled text and non-printable characters.

A+ must not try to render binary files in plain text. Binary files may only be downloaded without a preview in the web page.

A+ already recognizes some binary file types, but their MIME types are hardcoded, which is not optimal:
https://github.com/apluslms/a-plus/blob/8811feab20862474ac69951ec04315a98e79c659/exercise/submission_models.py#L606

Microsoft Office files (.doc, .docx, .xls, .xlsx) are particularly missing. They are needed urgently.

If there were a better way to recognize binary files without listing them manually, that would be the best. There are libraries that try to recognize files types from the header of the file or try to check if the file contains only printable ASCII/Unicode characters/bytes.

Students see submitted files in submission feedback pages and modal dialogs. Course staff see submitted files in the inspect submission page in addition to the student participant overview and the submission feedback pages.

Related to #728.

Internal ticket:
https://rt.cs.aalto.fi/Ticket/Display.html?id=20621

ihalaij1:
# Description

**What?**

A+ no longer tries to render some binary files in plain text in the submitted file views.

**Why?**

A teacher noticed that this problem was causing performance issues, when large binary files were being loaded for preview.

**How?**

Binary files are now detected using a library called ``binaryornot`` instead of using ``mimetypes``. The library inspects the file contents and based on that guesses whether the file is binary or not.

Fixes #965

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Checked that binary files are no longer rendered with a preview in any of the submitted file views.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/submission_models.py
+++ b/exercise/submission_models.py
@@ -4,6 +4,7 @@ import os
 import json
 from typing import IO, Dict, Iterable, List, Tuple, TYPE_CHECKING
 
+from binaryornot.check import is_binary
 from django.conf import settings
 from django.core.files.storage import default_storage
 from django.db import models, DatabaseError
@@ -642,7 +643,7 @@ class SubmittedFile(UrlMixin, models.Model):
         return guess_type(self.file_object.path)[0]
 
     def is_passed(self):
-        return self.get_mime() in SubmittedFile.PASS_MIME
+        return is_binary(self.file_object.path)
 
 
     ABSOLUTE_URL_NAME = "submission-file"
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,3 +30,4 @@ markdown ~= 3.3.4
 aplus-auth ~= 0.2.0
 celery >= 5.2.1, < 6
 redis >= 4
+binaryornot ~= 0.4.4


ihalaij1:
Note, this change makes it so that, for example, clicking on ``image.png`` on the inspect page downloads the file, instead of opening a preview of the image in the browser. This is because the content type is now ``application/octet-stream``.

Should the ``mimetypes`` module still be used with some common file types, so that this type of browser previewing is still possible?

exercise/views.py
```diff
@@ -477,7 +477,7 @@ def get(self, request, *args, **kwargs):
             return response
 
         if self.file.is_passed():
-            return HttpResponse(bytedata, content_type=self.file.get_mime())
+            return HttpResponse(bytedata, content_type="application/octet-stream")
```
Mankro, line 480: Let's keep the old version of the code here so that the mime type is correct for the commonly used file types (assuming that their filename extension is correct). No need to force everything to octet-stream here.

Mankro (changes requested):
Excellent! But let's keep the old mime type in the download like you said. Then, browsers may still open images and pdfs in the browser's own preview. It has worked well in the past.

exercise/views.py
ihalaij1: Done.

Mankro (approved):
Nice! Is `SubmittedFile.PASS_MIME` used anywhere after this change? It should be removed then.

ihalaij1:
> Nice! Is `SubmittedFile.PASS_MIME` used anywhere after this change? It should be removed then.

Yes, you're right, I'll remove it later. Also, shouldn't this import line be in the first group of imports according to the style guide (I didn't edit it, only added ``binaryornot``)?
https://github.com/apluslms/a-plus/blob/master/exercise/submission_models.py#L15

Mankro:
>  Also, shouldn't this import line be in the first group of imports according to the style guide (I didn't edit it, only added binaryornot)?
> https://github.com/apluslms/a-plus/blob/4ef04727f4eb9c8e84c9ca100036fa3faa463ff3/exercise/submission_models.py#L15

Indeed, that's correct. Mimetypes is a Python builtin module.
https://docs.python.org/3/library/mimetypes.html

PR #978, opened 2022-02-15
Use exists() instead of comparing count() to 0

skulonen:
# Description

**What?**

This commit optimizes queries where the count of some queryset was compared to 0 to check the existence of some objects.

**Why?**

`.exists()` is usually faster than `.count()`. This was discussed in a comment on #973.

**How?**

All occurrences where the result of `.count()` was compared to 0 (either `== 0` or `> 0`), have been replaced with an `.exists()` call.

# Testing

- [x] Django unit tests.
- [x] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox
- [x] This pull request cannot be tested in the browser.

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/exercise_summary.py
+++ b/exercise/exercise_summary.py
@@ -126,10 +126,10 @@ class UserExerciseSummary(object):
     def is_unofficial(self):
         return self.unofficial
 
-    def get_group(self):
+    def get_group(self) -> Optional[StudentGroup]:
         if self.submission_count > 0:
             s = self.submissions[0]
-            if s.submitters.count() > 0:
+            if s.submitters.exists():
                 return StudentGroup.get_exact(
                     self.exercise.course_instance,
                     s.submitters.all()
--- a/exercise/staff_views.py
+++ b/exercise/staff_views.py
@@ -168,7 +168,7 @@ class InspectSubmissionView(SubmissionBaseView, BaseFormView):
     def get_common_objects(self) -> None:
         super().get_common_objects()
         self.get_summary_submissions()
-        self.has_files = self.submission.files.exists()
+        self.files = list(self.submission.files.all())
 
         self.lowest_visible_index = self.index - 10
         self.highest_visible_index = self.index + 10
@@ -202,7 +202,7 @@ class InspectSubmissionView(SubmissionBaseView, BaseFormView):
         self.grading_mode_text = format_lazy(_('GRADING_MODE_TITLE -- {}'), mode)
 
         self.note(
-            'has_files',
+            'files',
             'lowest_visible_index',
             'highest_visible_index',
             'not_final',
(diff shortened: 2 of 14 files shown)


exercise/exercise_summary.py
```diff
@@ -126,10 +126,10 @@ def is_graded(self):
...
 
-    def get_group(self):
+    def get_group(self) -> Optional[StudentGroup]:
         if self.submission_count > 0:
             s = self.submissions[0]
-            if s.submitters.count() > 0:
+            if s.submitters.exists():
                 return StudentGroup.get_exact(
                     self.exercise.course_instance,
                     s.submitters.all()
```
Mankro, line 135: This could be improved based on this:
https://docs.djangoproject.com/en/3.2/topics/db/optimization/#don-t-overuse-count-and-exists

If there are submitters, we could run this with only one database query like this:
```suggestion
            submitters = s.submitters.all()
            if submitters:
                return StudentGroup.get_exact(
                    self.exercise.course_instance,
                    submitters,
```
If there are no submitters, I guess using `exists()` to speed up that case is not really worthwhile.

exercise/staff_views.py
```diff
@@ -168,7 +168,7 @@ class InspectSubmissionView(SubmissionBaseView, BaseFormView):
     def get_common_objects(self) -> None:
         super().get_common_objects()
         self.get_summary_submissions()
-        self.has_files = self.submission.files.count() > 0
+        self.has_files = self.submission.files.exists()
```
Mankro, line 171: If I remember correctly, the html template queries the database again to retrieve the submitted file objects. There would be less queries overall if the files are queried and saved to a variable here. Though, I suppose the template does not query anything if `has_files` is `False`.

Mankro (changes requested):
Excellent! I think we could slightly improve this based on the docs:
https://docs.djangoproject.com/en/3.2/topics/db/optimization/#don-t-overuse-count-and-exists

> Don’t overuse count() and exists()
> If you are going to need other data from the QuerySet, evaluate it immediately.

In this PR, I noticed a couple of places that could apply those ideas. Check the comments below.

exercise/exercise_summary.py
skulonen: That wouldn't do anything (I tried). The submitters of the submission are already prefetched by SubmissionManager. `exists` checks the result cache before executing a new query.

exercise/staff_views.py
skulonen: Good catch. I pushed a new commit that optimizes this.

Mankro (approved):
Nice! Time to merge.

PR #997, opened 2022-03-02
API endpoint for getting submission statistics

### LINKED ISSUE #989: Initial implementation on submission statistics
TBD: description here: per course, over time... temporarily assigned to Pasi for better issue description...

PasiSa:
Returns number of submissions, and number of distinct users
submitting over time window given as query parameters.
These statistics can be queried for a specific course,
for a specific exercise, or through all courses in system.

Closes #989.

# Description

**What?**

Initial implementation of statistics API end point. Idea is to continue this work by developing a separate tool for collecting the data using this API. The tool would then create different visualizations of the data with regular updates. The API is intentionally open for logged in A+ users. In future the API could be extended with other metrics, for example based on user requests.

Not sure if the current way of using ViewSets is good, but I'm sure you will tell me :-)

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/api/views.py
+++ b/course/api/views.py
@@ -16,6 +16,7 @@ from lib.viewbase import BaseMixin
 from lib.api.filters import FieldValuesFilter
 from lib.api.mixins import ListSerializerMixin, MeUserMixin
 from lib.api.constants import REGEX_INT, REGEX_INT_ME
+from lib.api.statistics import BaseStatisticsView
 from userprofile.permissions import IsAdminOrUserObjIsSelf
 from news.models import News
 
@@ -577,3 +578,40 @@ class CourseNewsViewSet(NestedViewSetMixin,
                     Q(audience=AUDIENCE.INTERNAL_USERS)
                 )
         return queryset
+
+
+class CourseStatisticsView(BaseStatisticsView):
+    """
+    Returns submission statistics for a course, over a given time window.
+
+    Returns the following attributes:
+
+    - `submission_count`: total number of submissions.
+    - `submitters`: number of users submitting.
+
+    Operations
+    ----------
+
+    `GET /courses/<course_id>/statistics/`:
+        returns the statistics for the given course.
+
+    - URL parameters:
+        - `endtime`: date and time in ISO 8601 format indicating the end point
+          of time window we are interested in. Default: now.
+        - `starttime`: date and time in ISO 8601 format indicating the start point
+          of time window we are interested in. Default: one day before endtime
+    """
+
+    serializer_class = CourseStatisticsSerializer
+
+    def get_queryset(self) -> QuerySet:
+        queryset = super().get_queryset()
+        course_id = self.kwargs['course_id']
+        return queryset.filter(
+            exercise__course_module__course_instance=course_id,
+        )
+
+    def get_object(self):
+        obj = super().get_object()
+        obj.update({ 'course_id': self.kwargs['course_id'] })
+        return obj
--- a/exercise/api/views.py
+++ b/exercise/api/views.py
@@ -12,10 +12,12 @@ from rest_framework.decorators import action
 from rest_framework.settings import api_settings
 from rest_framework_extensions.mixins import NestedViewSetMixin
 from django.db import DatabaseError
+from django.db.models import QuerySet
 
 from authorization.permissions import ACCESS
 from lib.api.mixins import MeUserMixin, ListSerializerMixin
 from lib.api.constants import REGEX_INT, REGEX_INT_ME
+from lib.api.statistics import BaseStatisticsView
 from userprofile.models import UserProfile, GraderUser
 from course.permissions import (
     IsCourseAdminOrUserObjIsSelf,
@@ -499,3 +501,40 @@ class CoursePointsViewSet(ListSerializerMixin,
         if self.action == 'list':
             return self.instance.students
         return self.instance.course_staff_and_students
+
+
+class ExerciseStatisticsView(BaseStatisticsView):
+    """
+    Returns submission statistics for an exercise, over a given time window.
+
+    Returns the following attributes:
+
+    - `submission_count`: total number of submissions.
+    - `submitters`: number of users submitting.
+
+    Operations
+    ----------
+
+    `GET /exercises/<exercise_id>/statistics/`:
+        returns the statistics for the given exercise.
+
+    - URL parameters:
+        - `endtime`: date and time in ISO 8601 format indicating the end point
+          of time window we are interested in. Default: now.
+        - `starttime`: date and time in ISO 8601 format indicating the start point
+          of time window we are interested in. Default: one day before endtime
+    """
+
+    serializer_class = ExerciseStatisticsSerializer
+
+    def get_queryset(self) -> QuerySet:
+        queryset = super().get_queryset()
+        exercise_id = self.kwargs['exercise_id']
+        return queryset.filter(
+            exercise=exercise_id,
+        )
+
+    def get_object(self):
+        obj = super().get_object()
+        obj.update({ 'exercise_id': self.kwargs['exercise_id'] })
+        return obj
--- a/lib/api/statistics.py
+++ b/lib/api/statistics.py
@@ -1,0 +1,55 @@
+from datetime import datetime, timedelta
+
+from rest_framework import status
+from rest_framework import generics
+from django.db.models import Count, QuerySet
+from django.utils import timezone
+
+from exercise.submission_models import Submission
+from .serializers import StatisticsSerializer
+
+
+class BaseStatisticsView(generics.RetrieveAPIView):
+    """
+    Returns submission statistics for the entire system, over a given time window.
+
+    Returns the following attributes:
+
+    - `submission_count`: total number of submissions.
+    - `submitters`: number of users submitting.
+
+    Operations
+    ----------
+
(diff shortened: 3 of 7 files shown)


course/api/views.py
```diff
@@ -577,3 +580,39 @@ def get_queryset(self) -> QuerySet[News]:
...
+          of time window we are interested in. Default: one day before endtime
+    """
+    def list(
+            self,
+            request: Request,
+            version: Optional[Union[int, str]] = None,
+            course_id: Optional[Union[int, str]] = None,
+            ) -> Response:
+
+        from exercise.submission_models import Submission
```
Mankro, line 612: Was there a problem if the import were at the top of the module as usual?

course/api/views.py
```diff
@@ -577,3 +580,39 @@ def get_queryset(self) -> QuerySet[News]:
...
+          of time window we are interested in. Default: now.
+        - `starttime`: date and time in ISO 8601 format indicating the start point
+          of time window we are interested in. Default: one day before endtime
+    """
+    def list(
+            self,
+            request: Request,
+            version: Optional[Union[int, str]] = None,
+            course_id: Optional[Union[int, str]] = None,
+            ) -> Response:
```
Mankro, line 610: Adding the parameters `*args` and `**kwargs` is recommended. Our DRF API code has followed that pattern. The code won't crash immediately then if the framework has new parameters in some version.

lib/api/statistics.py
```diff
@@ -0,0 +1,82 @@
+from typing import Optional, Union
+from datetime import datetime, timedelta
+from rest_framework import viewsets, status
```
Mankro, line 3: Styleguide suggests separating the import groups with one empty line: built-in modules, third-party-modules
```suggestion

from rest_framework import viewsets, status
```

Mankro (changes requested):
The ViewSets do not indeed follow the patterns we have used in the API.

* `*args` and `**kwargs` are good to have in the method parameters (like in `list()`), though the list method might be removed completely. The `version` parameter seems to be unused in these view sets and you could thus drop it (`**kwargs` will include it).
* I think that the `list()` methods can be replaced with the attributes and methods of `GenericViewSet`:
  - https://www.django-rest-framework.org/api-guide/viewsets/#genericviewset
  - https://www.django-rest-framework.org/api-guide/generic-views/#genericapiview
  - for example, `queryset` attribute for `Submission.objects.all()` or the `get_queryset()` method.
* Usually, responses are formed with serializers. In generic view sets, there are the attribute `serializer_class` and the method `get_serializer_class()`. You could define a new serializer class for statistics.
* My hunch is that you could define attributes `queryset`, `serializer_class` and the method `filter_queryset()` for filtering the submissions. Use attributes when it is not necessary to override the methods with new dynamic behaviour that depends on the input parameters.
* The number of code lines should decrease when you use the Django REST framework like it is supposed to be. The code will be more declarative and it won't repeat the similar lines querying the Submissions.
* Where did you find `self.responsefields`? Is that part of the documented public API?
* Look at existing view sets for examples if there are issues with the less obvious attributes like `lookup_url_kwarg`.
  - `NestedViewSetMixin` is used when the URL has nesting like `/courses/COURSEID/exercises/EXERCISEID`, but these statistics did not seem to have nesting.

PasiSa:
The new commit has the entire thing redesigned to use RetrieveAPIView and serializers, as discussed in the meeting. The two commits will be squashed into one before merging. There is also one comment/question about what to do with invalid query parameters, when they are processed in get_queryset(), currently just ignore if they are not valid. Most earlier comments became irrelevant after redesign.

lib/api/statistics.py
```diff
@@ -0,0 +1,62 @@
...
+        queryset = Submission.objects.all()
+
+        try:
+            endtime = self.request.query_params.get('endtime')
+            if endtime:
+                self.endtime = datetime.fromisoformat(endtime)
+
+            starttime = self.request.query_params.get('starttime')
+            if starttime:
+                self.starttime = datetime.fromisoformat(starttime)
```
Mankro, line 47: The instance variables `self.endtime` and `self.starttime` must be initialized here with the default values if the query parameters are empty. The instance variables are needed in the other method `get_object()`.

course/api/views.py
```diff
@@ -577,3 +578,40 @@ def get_queryset(self) -> QuerySet[News]:
...
+          of time window we are interested in. Default: one day before endtime
+    """
+
+    serializer_class = CourseStatisticsSerializer
+
+    def get_queryset(self) -> QuerySet:
+        queryset = super().get_queryset()
+        course_id = self.kwargs['course_id']
+        return queryset.filter(
+            exercise__course_module__course_instance=course_id
```
Mankro, line 611:
```suggestion
            exercise__course_module__course_instance=course_id,
```
The trailing comma is used in this "one param per line" format so that you can later add new parameters without adding changes (git diff) to the previous line that is otherwise not going to be changed.

exercise/api/views.py
```diff
@@ -505,3 +507,40 @@ def get_queryset(self):
...
+          of time window we are interested in. Default: one day before endtime
+    """
+
+    serializer_class = ExerciseStatisticsSerializer
+
+    def get_queryset(self) -> QuerySet:
+        queryset = super().get_queryset()
+        exercise_id = self.kwargs['exercise_id']
+        return queryset.filter(
+            exercise=exercise_id
```
Mankro, line 540:
```suggestion
            exercise=exercise_id,
```

Mankro (changes requested):
Great! Only minor changes are needed.

lib/api/statistics.py
```diff
@@ -0,0 +1,62 @@
...
+        except ValueError:
+            # Not sure what to do on invalid query params. For now just ignore them.
+            pass
```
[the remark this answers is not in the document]
PasiSa: Yes, 400 is the right code, but my problem was how to generate that from get_queryset(), that should return a QuerySet. Should query_param parsing rather be done some place else?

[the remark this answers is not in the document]
Mankro: @PasiSa You raise an exception: `raise serializers.ValidationError(...)`

If you used a serializer for input validation, then the validation and its errors would be "hidden" inside the serializer class implementation. You would then create a serializer object and call its `is_valid()` method instead of raising exceptions manually. `StatisticsSerializer` could validate the date inputs if the other fields had the attribute `read_only=True`.

* https://www.django-rest-framework.org/api-guide/exceptions/#validationerror
* https://www.django-rest-framework.org/api-guide/fields/#read_only
* https://www.django-rest-framework.org/api-guide/serializers/#validation
* views have built-in exception handling: https://www.django-rest-framework.org/api-guide/views/#handle_exceptionself-exc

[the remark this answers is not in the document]
PasiSa: Right. Indeed, this is now a much cleaner way.

Mankro (changes requested):
Getting better every time! The default values for the input dates could be handled better.

lib/api/statistics.py
```diff
@@ -0,0 +1,61 @@
...
+            timezone.now().isoformat()
+        )
+        starttime = self.request.query_params.get(
+            'starttime',
+            (timezone.now() - timedelta(days=1)).isoformat()
+        )
+        serializer = StatisticsSerializer(data={'starttime': starttime, 'endtime': endtime})
+        serializer.is_valid(raise_exception=True)
+        self.starttime = serializer.validated_data['starttime']
+        self.endtime = serializer.validated_data['endtime']
```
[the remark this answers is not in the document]
PasiSa: Yes. The last suggested option seemed easiest to me (although allow_null seemed to work better in this case). How about now?

lib/api/statistics.py
```diff
@@ -0,0 +1,55 @@
...
+          of time window we are interested in. Default: one day before endtime
+    """
+    serializer_class = StatisticsSerializer
+
+    def get_queryset(self) -> QuerySet:
+        queryset = Submission.objects.all()
+
+        endtime = self.request.query_params.get('endtime')
+        starttime = self.request.query_params.get('starttime')
+        serializer = StatisticsSerializer(data={'starttime': starttime, 'endtime': endtime})
```
Mankro, line 40: Since you have hardcoded `StatisticsSerializer` here instead of calling `self.get_serializer(data={'starttime': starttime, 'endtime': endtime})`, then the child view classes actually also use `StatisticsSerializer` even though those classes declare another serializer class in the class variables. This will cause confusion since this does not behave in the usual Django REST framework way.

When you change this, you will probably need to add `read_only=True` to the child serializers (`ExerciseStatisticsSerializer` and `CourseStatisticsSerializer`).

Mankro (changes requested):
Excellent! I noticed one more inconsistency in the way you use the Django REST framework, but it is easy to fix. After that, you can clean up the git commit history and we will merge this.

lib/api/statistics.py
PasiSa: done

Mankro (approved):

Issue #971, opened 2022-02-10
A+ Security logs, CEF format


PasiSa:
After security audit in spring 2021, some new security-related log events were added, using SecurityLog class. The log output format should be converted to CEF format that can be exported to Aalto ITS logging systems. Also, the current log events should be reviewed: do they contain sufficient information, and should some additional events be added. Note that security log should contain only relevant events, that can be justified from security point of view.

PR #1005, opened 2022-03-18
Apply CEF format in security logs

PasiSa:
Aalto ITS requires ArcSight Common Event Format (CEF) for the
stored logs. Convert the SecurityLog output to follow this format,
using format_cef library.

# Description

**What?**

As said above. We may later need to adjust logger settings and possibly other details based on ITS feedback, but with this commit we would have CEF formatting in place.

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lib/logging.py
+++ b/lib/logging.py
@@ -3,6 +3,12 @@ import logging
 from django.contrib.auth.signals import user_logged_in, user_logged_out, user_login_failed
 from django.dispatch import receiver
 from django.http.request import HttpRequest
+from format_cef import format_cef
+from aplus import VERSION
+
+
+CEF_VENDOR = 'Aalto'
+CEF_PRODUCT = 'aplus'
 
 def skip_unreadable_post(record):
     """Skips log records of unfinished post requests."""
@@ -10,7 +16,8 @@ def skip_unreadable_post(record):
 
 class SecurityLog:
     """
-    Static singleton class used for A+ - wide security logging.
+    Static singleton class used for A+ - wide security logging,
+    to produce ArcSight Common Event Format (CEF) log.
     Django signals are used to track login/logout events.
     """
 
@@ -21,20 +28,32 @@ class SecurityLog:
         return SecurityLog.seclogger
 
     @staticmethod
-    def logevent(request: HttpRequest, type: str, message: str) -> None:
-        # Unit tests do not have user defined in request object.
-        if request and hasattr(request, 'user'):
-            user=request.user
-            ip=request.META.get('REMOTE_ADDR')
-        else:
-            user='?'
-            ip='?'
-        SecurityLog.logger().info("({}/{}): {}: {}".format(
-            user,
-            ip,
-            type,
-            message
-        ))
+    def logevent(
+            request: HttpRequest,
+            type: str,
+            message: str,
+            severity: int = 5,
+            ) -> None:
+        extensions = {}
+        # Unit tests may not have user or source address defined.
+        if request:
+            if hasattr(request, 'user'):
+                extensions['sourceUserName'] = str(request.user)
+                extensions['sourceUserId'] = str(request.user.id)
+            if (addr := request.META.get('REMOTE_ADDR')):
+                extensions['sourceAddress'] = addr
+
+        SecurityLog.logger().info(
+            format_cef(
+                CEF_VENDOR,
+                CEF_PRODUCT,
+                VERSION,
+                type,
+                message,
+                severity,
+                extensions,
+            ).decode("utf-8")
+        )
 
 
 # This example was used as an inspiration for the following functions:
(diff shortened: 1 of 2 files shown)


lib/logging.py
```diff
@@ -21,20 +28,30 @@ def logger() -> logging.Logger:
...
+    def logevent(
+        request: HttpRequest,
+        type: str,
+        message: str,
+        severity: int=5,
+    ) -> None:
+        extensions = {}
+        # Unit tests may not have user or source address defined.
+        if request:
+            if hasattr(request, 'user'): extensions['sourceUserName'] = str(request.user)
```
Mankro, line 40: We could include also `sourceUserId`. No harm in having both username and user id?

lib/logging.py
```diff
@@ -21,20 +28,30 @@ def logger() -> logging.Logger:
...
-            ip,
-            type,
-            message
-        ))
+    def logevent(
+        request: HttpRequest,
+        type: str,
+        message: str,
+        severity: int=5,
+    ) -> None:
```
Mankro, line 36:
```suggestion
    def logevent(
            request: HttpRequest,
            type: str,
            message: str,
            severity: int=5,
            ) -> None:
```
This style of indentation is suggested in the A+ Python style guide. The parameter list is indented one step further than the function body so that they stand out (when parameters are listed one per line).

lib/logging.py
```diff
@@ -21,20 +28,30 @@ def logger() -> logging.Logger:
...
+        extensions = {}
+        # Unit tests may not have user or source address defined.
+        if request:
+            if hasattr(request, 'user'): extensions['sourceUserName'] = str(request.user)
+            if (addr := request.META.get('REMOTE_ADDR')):
+                extensions['sourceAddress'] = addr
+
```
Mankro, line 43: You have unwanted extra whitespace on this empty line.

Mankro (changes requested):
Looks great! There are small stylistic issues and I thought the user id could be added to the log output.

You didn't link the commit to the issue #971. ("Fixes" line.)

A sidenote: `user_login_failed_callback` output is awkward and it would be easy to improve it. It outputs a Python dictionary, but the password is hidden which makes it useless. It wouldn't be good to store attempted passwords in the logs anyway. The log should include just the username.

PasiSa:
Addressed comments, directly force-pushing into the original commit. Also login failed output is now modified, as suggested.

Mankro (changes requested):
Excellent, but please double-check `request.user`. I don't think `hasattr()` works if it points to an `AnonymousUser`.

lib/logging.py
```diff
@@ -21,20 +28,32 @@ def logger() -> logging.Logger:
...
+    def logevent(
+            request: HttpRequest,
+            type: str,
+            message: str,
+            severity: int = 5,
+            ) -> None:
+        extensions = {}
+        # Unit tests may not have user or source address defined.
+        if request:
+            if hasattr(request, 'user'):
```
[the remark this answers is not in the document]
PasiSa: Unit tests do not seem to have user attribute set in HttpRequest, and therefore fail without hasattr check. I don't think any of the security log messages are reachable without logged in user, so couldn't test how it would behave with AnonymousUser.

I now also noticed that the change to failed login caused unit tests to fail because they do not have 'username' key defined, so did one more small fix to that.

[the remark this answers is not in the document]
Mankro: Unit tests may log in with the test client and simulate an authenticated user.

https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Client.login

lib/logging.py
```diff
@@ -51,6 +70,8 @@ def user_logged_out_callback(sender, request, user, **kwargs):
...
 @receiver(user_login_failed)
 def user_login_failed_callback(sender, credentials, **kwargs):
-    SecurityLog.logevent(None, "login-failed","{credentials}".format(
-        credentials=credentials)
-    )
+    try:
+        SecurityLog.logevent(None, "login-failed", f"username: {credentials['username']}")
+    except KeyError:
+        # Unit tests do not have 'username' in credentials, let's not fail them for that
+        pass
```
Mankro, line 77:
```suggestion
    SecurityLog.logevent(None, "login-failed", f"username: {credentials.get('username', 'none')}")
    # Unit tests do not have 'username' in credentials, let's not fail them for that
```
I would have preferred this way so that we do not miss the logs completely in case this could happen in the production environment too.

Mankro (approved):
Thanks!

PR #1019, opened 2022-04-06
Add download buttons to modal dialogs

### LINKED ISSUE #894: File download button needed in submitted file modal dialogs
(By downloading, I mean that the browser downloads the file to the user's computer. Firefox usually shows a dialog while Chrome usually downloads it immediately. The A+ submitted file URL already supports the parameter `?download=yes`.) 

The student can see file contents in some dialogs, but there is no button for downloading the file to the user's computer. A download button would be very good to have in some places: 

* model solution modal dialog in chapter exercises,
* model solution page in the full-view exercise,
* submission template/skeleton code modal dialog in chapter exercises, 
* submission template file in the full-view exercise page,
* submitted file modal dialog in the submission modal in chapter exercises and
* the submitted file modal dialog in full-view submission pages outside chapters

_Originally posted by @Mankro in https://github.com/apluslms/a-plus/pull/893#discussion_r738336787_

PR #893 already has a quick implementation for the download button in the inspect submission page. Other places need the download button too.

joni-toivonen:
Adds download buttons to model answer and template modals

Fixes #894

# Description

**What?**

Adds download buttons to "Show model answer" and "Show exercise template" modals

**Why?**

This allows the user to download the file instead of just seeing the contents

**How?**

Add button to frontend that calls backend with ?download=yes parameter and backend knows then to serve the files with Content-Disposition header which causes browser to initiate downloading

Fixes #894

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/templates/exercise/_model_files.html
+++ b/exercise/templates/exercise/_model_files.html
@@ -1,6 +1,6 @@
 {% load i18n %}
 {% load course %}
-<p>{{ current.name|parse_localization }}</p>
+<p>{{ current.name|parse_localization }}</p>
 {% for entry in models %}
 	{% if entry.content %}
 		{% if entry.html %}
--- a/exercise/templates/exercise/_model_files.html
+++ b/exercise/templates/exercise/_model_files.html
@@ -7,7 +7,7 @@
 			{{ entry.content|safe }}
 		{% else %}
 			<h4>{{ entry.name|parse_localization }}</h4>
-			<pre class="hljs pre-content" data-self-link="{{ exercise|url:'exercise-model' }}?name={{ entry.name }}">{{ entry.content }}</pre>
+			<pre class="hljs" data-url="{{ exercise|url:'exercise-model' }}?name={{ entry.name }}">{{ entry.content }}</pre>
 		{% endif %}
 	{% else %}
 		<h4>{{ entry.name|parse_localization }}</h4>
--- a/exercise/templates/exercise/_model_files.html
+++ b/exercise/templates/exercise/_model_files.html
@@ -7,7 +7,7 @@
 			{{ entry.content|safe }}
 		{% else %}
 			<h4>{{ entry.name|parse_localization }}</h4>
-			<pre class="hljs">{{ entry.content }}</pre>
+			<pre class="hljs pre-content" data-self-link="{{ exercise|url:'exercise-model' }}?name={{ entry.name }}">{{ entry.content }}</pre>
 		{% endif %}
 	{% else %}
 		<h4>{{ entry.name|parse_localization }}</h4>
--- a/exercise/views.py
+++ b/exercise/views.py
@@ -382,6 +382,23 @@ class ExerciseModelView(ExerciseModelBaseView):
                 })
         self.note('models')
 
+    def get(self, request, *args, **kwargs):
+        if request.GET.get("download", False):
+            request_name = request.GET.get("name")
+            for model in self.models:
+                name = model['name']
+                if (name == request_name):
+                    data = model['content']
+            if data is None:
+                return HttpResponseNotFound()
+
+            response = HttpResponse(data,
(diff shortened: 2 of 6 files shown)


exercise/views.py
```diff
@@ -379,9 +379,24 @@ def get_common_objects(self):
...
+            except OSError:
+                return HttpResponseNotFound()
+
+            response = HttpResponse(bytedata,
+                content_type="application/octet-stream")
+            response["Content-Disposition"] = 'attachment; filename="{}"'\
+                .format(self.models[0]['name'])
+            return response
+
+        return super().get(self)
```
Mankro, line 399: You don't pass `self` explicitly to the `super()` parameters in Python 3. Only the normal method parameters are passed. Don't forget to pass the potential `*args` and `**kwargs`.
```suggestion
        return super().get(request, *args, **kwargs)
```

https://docs.python.org/3/library/functions.html#super

exercise/views.py
```diff
@@ -379,9 +379,24 @@ def get_common_objects(self):
...
 
+    def get(self, request, *args, **kwargs):
+        if request.GET.get("download", False):
+            try:
+                bytedata = self.models[0]['data']
+            except OSError:
+                return HttpResponseNotFound()
+
+            response = HttpResponse(bytedata,
+                content_type="application/octet-stream")
```
Mankro, line 394: It doesn't look like it is necessary to send this response as byte data when the text (string type) is also available. `HttpResponse` could just use the text data, i.e., the original `response.text` from `get_common_objects()`.

https://docs.djangoproject.com/en/3.2/ref/request-response/#passing-strings

It looks like the old code has never really handled binary (non-text) files. Rendering binary files as text in the html page outputs garbled nonsensical content. It would be better to avoid that, but that can be another task in the future.

exercise/views.py
```diff
@@ -379,9 +379,24 @@ def get_common_objects(self):
                     'name': name,
                     'content': response.text,
                     'html': 'text/html' in response.headers.get('Content-Type'),
+                    'data': response.content,
```
Mankro, line 382: `data` can be removed if `content` is used in the `get` method instead of `data`.

exercise/views.py
```diff
@@ -405,9 +420,24 @@ def get_common_objects(self):
...
+            except OSError:
+                return HttpResponseNotFound()
+
+            response = HttpResponse(bytedata,
+                content_type="application/octet-stream")
+            response["Content-Disposition"] = 'attachment; filename="{}"'\
+                .format(self.templates[0]['name'])
+            return response
+
+        return super().get(self)
```
Mankro, line 440: All of the comments from ExerciseModelView apply here as well.

Mankro (changes requested):
Nice start! Multiple issues have to be fixed, though. Don't forget that one assignment may have multiple model solution files and/or multiple template files.

exercise/templates/exercise/_model_files.html
```diff
@@ -7,7 +7,7 @@
 			{{ entry.content|safe }}
 		{% else %}
 			<h4>{{ entry.name|parse_localization }}</h4>
-			<pre class="hljs">{{ entry.content }}</pre>
+			<pre class="hljs pre-content" data-self-link="{{ exercise|url:'exercise-model' }}?name={{ entry.name }}">{{ entry.content }}</pre>
```
Mankro, line 10: `pre-content` shall be removed and `data-self-link` renamed to `data-url`.

Mankro (changes requested):
Thanks! I still found severe issues from the code. I will fix these myself next.

Mankro (approved):
I fixed the issues myself and now I will merge this pull request!

PR #1033, opened 2022-05-09
API endpoint for regrading submissions

### LINKED ISSUE #1008: API endpoint for regrading submissions
Like the one in the inspect view https://plus.cs.aalto.fi/o1/2022/w01/ch02/w01_ch02_2/submissions/4936438/inspect/ 

https://plus.cs.aalto.fi/o1/2022/w01/ch02/w01_ch02_2/submissions/4936438/re-submit/

This feature would be used in the A+ Courses IntelliJ plugin and it would be shown only for course assistants. [Here's the relevant PR.](https://github.com/Aalto-LeTech/aplus-courses/pull/851) It would be useful when developing tests for a course, so you don't need to keep making new submissions to try the new tests.

ihalaij1:
# Description

**What?**

Add an API endpoint for resubmitting submissions to grading.
Only course staff can use the resubmit API.

**Why?**

This was requested for O1.

**How?**

See the code.

Fixes #1008

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the new API endpoint works and has the correct permissions.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/api/views.py
+++ b/exercise/api/views.py
@@ -23,6 +23,7 @@ from course.permissions import (
     IsCourseAdminOrUserObjIsSelf,
     JWTSubmissionCreatePermission,
     JWTSubmissionWritePermission,
+    OnlyCourseStaffPermission,
 )
 from course.api.mixins import CourseResourceMixin
 from course.api.serializers import StudentBriefSerializer
@@ -88,7 +89,7 @@ class ExerciseViewSet(mixins.RetrieveModelMixin,
         if request.method in permissions.SAFE_METHODS:
             return self.retrieve(request, *args, **kwargs)
 
-        ## submit and grade new ssubmission
+        ## submit and grade new submission
 
         # Onyl grader is allowed to post to this resource
         user = request.user
@@ -386,6 +387,9 @@ class SubmissionViewSet(mixins.RetrieveModelMixin,
 
     `GET /submissions/<submission_id>/grader/`:
         used by automatic graders when grading a submission.
+
+    `GET /submissions/<submission_id>/re-submit/`:
+        resubmits a submission for grading.
     """
     lookup_field = 'id'
     lookup_url_kwarg = 'submission_id'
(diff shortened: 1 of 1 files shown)


PasiSa:
Looks good to me, but one thought crossed my mind: this has quite much similarity to the original submit API implementation, so I wonder if it would be feasible to share the common parts of code between the two endpoints somehow, instead of having to duplicate code.

exercise/api/views.py
```diff
@@ -424,6 +428,39 @@ def grader_detail(self, request, *args, **kwargs):
...
+        data = None
+
+        page = self.submission.exercise.grade(self.submission, request)
+
+        # Enroll after succesfully resubmitting to the enrollment exercise.
+        if self.submission.exercise.status in (
+            LearningObject.STATUS.ENROLLMENT,
+            LearningObject.STATUS.ENROLLMENT_EXTERNAL,
+        ) and self.submission.status == Submission.STATUS.READY:
+            self.instance.enroll_student(request.user)
```
Mankro, line 451: Why would we need to check enrollment questionnaires in regrading? Well, it could be possible to regrade a somehow temporarily broken submission in the enrollment questionnaire. What was your idea behind this?

exercise/api/views.py
```diff
@@ -424,6 +428,39 @@ def grader_detail(self, request, *args, **kwargs):
...
+
+        status_code = status.HTTP_200_OK
+        headers = {
+            'Location': reverse('api:submission-detail', kwargs={
```
Mankro, line 455: Why did you add the http redirect header to this API endpoint? How is it necessary here?

(I thought APIs usually don't redirect. They can return some success/failure message in the response to the client.)

Mankro (changes requested):
Nice! It looks simple as expected, but you made some choices that confuse me. Answer my questions below and let's see how to continue.

exercise/api/views.py
```diff
@@ -424,6 +428,39 @@ def grader_detail(self, request, *args, **kwargs):
 
         return Response(_post_async_submission(request, self.submission.exercise, self.submission))
 
+    @action(
+        detail=True,
+        url_path='re-submit',
+        url_name='re-submit',
+        get_permissions = lambda: [OnlyCourseStaffPermission()],
+        methods=['get'],
```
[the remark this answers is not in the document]
ihalaij1: I overlooked this part when I copied the function definition from above. I should've changed it to ``post``.

exercise/api/views.py
ihalaij1: I was thinking exactly that, if it's somehow possible to have an enrollment questionnaire stuck in grading, then it could be resubmitted. But is it correct that only submit exercises can get stuck in grading?

exercise/api/views.py
ihalaij1: I added the header because I implemented the ``resubmit`` API endpoint the same way as the above ``submit`` endpoint:
https://github.com/apluslms/a-plus/blob/e832ba304b22392e8acc0e47292a4ccd3dae24b0/exercise/api/views.py#L264-L312

Is there a reason the redirect header is present in the ``submit`` endpoint, but shouldn't be included in ``resubmit``?

exercise/api/views.py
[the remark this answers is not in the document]
ihalaij1: I changed this now.

Mankro (approved):
Thanks! I will make small changes myself and merge this.

It could make sense to return JSON like `{"success": true}` when the regrading is started successfully. Now the server returns no message nor content when there are no errors.

PR #1034, opened 2022-05-09
Add button to regenerate the API access token

### LINKED ISSUE #1022: Add button to regenerate the API access token
Currently it is not possible for the user to re-generate new API token in the user interface. The user profile view should therefore have a button for regenerating one.

ihalaij1:
# Description

**What?**

Previously, it was not possible for the user to regenerate a new API token in the user interface.
The user profile view now has a button for regenerating one.

**Why?**

This makes it easier to generate a new API token.

**How?**

A button was added to the userprofile page for regenerating the token.

Fixes #1022

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that regenerating the API token works.

- [x] Chrome
- [ ] Firefox

# Translation

- [x] Did you modify or add new strings in the user interface?

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -5505,6 +5505,14 @@ msgid "COPY"
 msgstr "Copy"
 
 #: userprofile/templates/userprofile/profile.html
+msgid "REGENERATE_API_ACCESS_TOKEN"
+msgstr "Generate a new API Access Token"
+
+#: userprofile/templates/userprofile/profile.html
+msgid "REGENERATE"
+msgstr "Regenerate"
+
+#: userprofile/templates/userprofile/profile.html
 msgid "AUTOMATIC_REDIRECTIONS"
 msgstr "Automatic redirections"
 
--- a/userprofile/templates/userprofile/profile.html
+++ b/userprofile/templates/userprofile/profile.html
@@ -9,22 +9,22 @@
 
 {% block js-translations %}
 <link
-    data-translation
+	data-translation
 	rel="prefetch"
-    as="fetch"
-    crossorigin="anonymous"
-    type="application/json, */*;"
-    hreflang="fi"
-    href="{{ STATIC_URL }}js-translations/profile.fi.json"
+	as="fetch"
+	crossorigin="anonymous"
+	type="application/json, */*;"
+	hreflang="fi"
+	href="{{ STATIC_URL }}js-translations/profile.fi.json"
 >
 <link
-    data-translation
+	data-translation
 	rel="prefetch"
-    as="fetch"
-    crossorigin="anonymous"
-    type="application/json, */*;"
-    hreflang="fi"
-    href="{{ STATIC_URL }}js-translations/copy-to-clipboard.fi.json"
+	as="fetch"
+	crossorigin="anonymous"
+	type="application/json, */*;"
+	hreflang="fi"
+	href="{{ STATIC_URL }}js-translations/copy-to-clipboard.fi.json"
 >
 {% endblock %}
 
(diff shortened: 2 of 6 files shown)


userprofile/views.py
```diff
@@ -111,6 +99,14 @@ def set_user_language(request):
     return response
 
 
+def regenerate_access_token(request):
+    """Regenerates the API access token."""
+    next = get_next_url(request)
```
Mankro, line 104: I don't understand why the `next` parameter is here. I thought that the profile page would include a button for regenerating the personal API token. Submitting the form sends HTTP POST to the server, which removes the token and sends a normal redirect response back to the browser. The user can be simply redirected back to the same profile page. This does not require any special `next` parameters in the URL or POST data for defining the redirect target.

Are you trying to use AJAX in the form submission? I can't see such code, but I didn't think this feature would need any AJAX anyway.

If the `next` parameter copied from the course language menu is not needed here, as I suspect, then it wasn't necessary to modify `set_user_lang()` or add `get_next_url()`.

userprofile/urls.py
```diff
@@ -17,5 +17,6 @@
     url(r'^accounts/$', views.ProfileView.as_view(),
         name="profile"),
     url(r'^setlang', views.set_user_language, name="set-user-language"),
+    url(r'^regentoken', views.regenerate_access_token, name="regenerate-access-token"),
```
Mankro, line 20: The `setlang` URL has its own history and reasons for its complex implementation. I suspect that `regentoken` does not need such complexity. Refer to the `next` parameter comment in views.py. Basically, I suspect that `regentoken` could include the ending $ in the pattern so that it does not match every URL that starts with `regentoken`.
```suggestion
    url(r'^regentoken/$', views.regenerate_access_token, name="regenerate-access-token"),
```

Mankro (changes requested):
Good! This is mostly what I expected, but I was surprised and confused by the `next` parameter that was copied from `set_user_lang()`. Why is it here? Why would this regen token button need any special parameter for the final redirect target? Compare this to a normal form that is HTTP POSTed to the server. They don't include this `next` parameter that might be in the GET or POST parameters. They just typically redirect to a hardcoded target in the end. 

Note that Django uses the `next` URL GET parameter when the unauthenticated used tries to enter views that require login. The special set user language view has some interaction with that, but I don't really even remember why it was implemented like that. The old set user lang view is surprisingly complex.

locale/en/LC_MESSAGES/django.po
```diff
@@ -5005,6 +5005,11 @@ msgstr ""
 msgid "SHIBBOLETH_ERROR_USER_ACCOUNT_DISABLED"
 msgstr "The user account has been disabled."
 
+#: src/drf-extensions/rest_framework_extensions/exceptions.py
+#, python-brace-format
+msgid "This \"{method}\" request is required to be conditional."
+msgstr ""
+
```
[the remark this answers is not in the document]
ihalaij1: Yeah, I think I used a different command for makemessages that causes this to appear. I'll remove it.

userprofile/views.py
ihalaij1: I didn't really understand the reason for having the ``next`` parameter in ``set_user_lang()``, and I used it as an example for ``regenerate_access_token()``. Therefore, I implemented it the same way with the ``next`` parameter, in case there was some reason it was needed. I'll remove it.

ihalaij1:
I made the suggested changes!

userprofile/templates/userprofile/profile.html
```diff
@@ -82,6 +82,16 @@ <h2>{% blocktranslate trimmed %}BRAND_PREFERENCES -- {{ brand_name }}{% endblock
 						class="glyphicon glyphicon-copy"></span><span class="">{% translate "COPY" %}</span></button>
 			</span>
 		</div>
+		<div>
+			<form action="{% url 'regenerate-access-token' %}" method="post" class="form form-floating">{% csrf_token %}
```
Mankro, line 86: Does the class `form-floating` do anything here? I couldn't find any styles for it in A+ or Bootstrap code.

Mankro (changes requested):
Excellent! There are still a couple of very quick changes that are needed. The regenerate view function has a CSRF security vulnerability.

userprofile/templates/userprofile/profile.html
ihalaij1: You're right, I thought it fixed an issue I had with ``form-horizontal``, but it turns out that it isn't needed at all.

ihalaij1:
I made the changes and cleaned the commit history!

Mankro (approved):
Excellent! I will add the `login_required` decorator to the view myself and merge this pull request.

PR #1035, opened 2022-05-10
Automatic retries of incomplete grading jobs

### LINKED ISSUE #988: Improve the stability and reliability of automatic grading
From #470:
The grader may fail to send the grading results back to A+ because, for example,

*    its disk becomes full during grading
*    there is a network error when the grader tries to send HTTP POST to A+ frontend
*    the grading container fails to send the results to the MOOC-grader
*    MOOC-grader fails to receive the grading results from the container (the container sends an HTTP POST request to the MOOC-grader at the end of the grading)
* In addition, if grader doesn't resolve or it's http queue is full, submission is left in initialized state.
  - To elaborate, when A+ receives the submission, the submission is at first in the initialized state. If A+ fails to send the submission to the MOOC-grader (i.e., MOOC-grader does not respond or responds with an error), then the submission in A+ is stuck in the initialized state.

Submissions shouldn't get stuck so easily in the grading pipeline.

Related to #793.

https://github.com/apluslms/a-plus/issues/470#issuecomment-1040279389
> Indeed, I was more paying attention to the proposed solution than problem description when closing the issue. Not sure how to make the recovery fully automatic (assuming that we can never fully get rid of network/server failure situations). Simple timer-based retries might just potentially congest the (under-failure) system unnecessarily -- what should be the length of timeout in that case, anyway? https://github.com/apluslms/a-plus/issues/793 might be good enough if failures are not very common (e.g., alert on course front page in teacher's view, with some sort of "retry all" button).

https://github.com/apluslms/a-plus/issues/470#issuecomment-1040290542
> Yeah, it is not a trivial question. However, from a teacher's perspective, the student has submitted and the system loses it. It makes the platform look fragile and unreliable. Retrying the grading wouldn't always have to be immediate if there is a risk of overloading the system. One simple idea is that the system could look for lost/stuck submissions every hour and use the new mass regrade feature on them.

https://github.com/apluslms/a-plus/issues/470#issuecomment-1040303787
> Yes. There could also be some sort of intelligence observing if mass-regrade-initiated grading for those submissions actually complete (and if not, maybe cancel mass regrade and try again after another hour). Which reminds me of a couple of additional features that could be added:
> 
> *   Cancel-button for an ongoing mass regrade
> *   I guess submitters would appreciate email notification that the earlier postponed grading has completed

https://github.com/apluslms/a-plus/issues/470#issuecomment-1040382833
> I'm not sure if I documented it anywhere, but I think I was visioning to solve this in the following way:
>
> *   If assessment request fail for an exercise, the exercise is put into maintenance mode or error mode. While in this mode, no assessments are attempted.
> *    Background task will poll backend of an exercise and if that exercise comes back online, then the mode of the exercise is restored and thus background assessment request can continue (assuming workers query DB for submissions waiting to be assessed).
> *    Assuming a single backend/domain handles multiple exercises, then there could be one more level for maintenance mode, where a backend services is put on the maintenance mode, which implies all exercises provided by it are in maintenance mode. Thus pollin is done in domain level and not in an exercise level.
> 
> That said, if amount of error emails doesn't bother, then running that mass regrade hourly might be an acceptable solution for now

PasiSa:
Sometimes it happens that a grading job never completes because of
some technical issue, and it remains in "In Grading" state until
user submits it for regrading. This commit implements automatic
regrading of grading jobs that have not completed within configured
timeout. There is also a simple heuristic to limit the
(likely unsuccesful) automatic retries when it seems that the
grader is more persistently unavailable.

Closes #988.

# Description

**What?**

As described it commit message. There was some earlier discussion on whether the pending submission tracking should be part of Submissions table or a separate table, but my thinking is that separate table is better. Submission table is large, so it would be more inefficient to operate. Also a separate table is more convenient to handle in the admin view. When things are not working well, admin might want to take a look at the state of PendingSubmissions table in the admin view in that case.

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #988

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Made a temporary toggle to grader to drop selected grading requests, and tested with different combinations of tasks, with periods of dropping tasks followed by periods when everything works well, to test the recovery.

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/aplus/celery.py
+++ b/aplus/celery.py
@@ -2,10 +2,14 @@ import os
 import celery
 import datetime
 from datetime import timedelta
+import logging
+from dateutil.relativedelta import relativedelta
 from time import sleep
 
 from django.conf import settings
 
+
+logger = logging.getLogger('aplus.celery')
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'aplus.settings')
 
 app = celery.Celery('aplus')
@@ -16,6 +20,12 @@ app.autodiscover_tasks()
 def setup_periodic_tasks(sender, **kwargs):
     if hasattr(settings, 'SIS_ENROLL_SCHEDULE'):
         sender.add_periodic_task(settings.SIS_ENROLL_SCHEDULE, enroll.s(), name='enroll')
+
+@app.on_after_configure.connect
+def setup_periodic_tasks(sender, **kwargs):
+    if settings.SUBMISSION_EXPIRY_TIMEOUT:
+        # Run timed check twice in timeout period, for more timely retries
+        sender.add_periodic_task(settings.SUBMISSION_EXPIRY_TIMEOUT/2, retry_submissions.s(), name='retry_submissions')
 
 @app.task
 def enroll():
@@ -34,3 +44,26 @@ def enroll():
         i.enroll_from_sis()
         if settings.SIS_ENROLL_DELAY:
             sleep(settings.SIS_ENROLL_DELAY)
+
+@app.task
+def retry_submissions():
+    from exercise.submission_models import PendingSubmission, PendingSubmissionManager
+
+    # Recovery state: only send one grading request to probe the state of grader
+    # We pick the one with most attempts, so that total_retries comes down more quickly
+    # when things are back to normal, and we can return to normal state
+    if not PendingSubmissionManager.is_grader_stable:
+        pending = PendingSubmission.objects.order_by('-num_retries').first()
+        logging.info(f"Recovery state: retrying expired submission {pending.submission}")
+        pending.submission.exercise.grade(pending.submission)
+        return
+
+    # Stable state: retry all expired submissions
+    expiry_time = datetime.datetime.now(datetime.timezone.utc) - relativedelta(seconds=settings.SUBMISSION_EXPIRY_TIMEOUT)
+    expired = PendingSubmission.objects.filter(submission_time__lt=expiry_time)
+
+    for pending in expired:
+        if pending.submission.exercise.can_regrade:
+            logger.info(f"Retrying expired submission {pending.submission}")
+            pending.submission.exercise.grade(pending.submission)
+            sleep(0.5)  # Delay 500 ms to avoid choking grader
--- a/course/templatetags/base.py
+++ b/course/templatetags/base.py
@@ -3,8 +3,9 @@ from datetime import datetime
 from django import template
 from django.conf import settings
 from django.utils.safestring import mark_safe
-from django.utils.translation import get_language
+from django.utils.translation import get_language, gettext_lazy as _
 from lib.helpers import remove_query_param_from_url, settings_text, update_url_params
+from exercise.submission_models import PendingSubmissionManager
 
 
 register = template.Library()
@@ -44,6 +45,9 @@ def site_alert():
     if message:
         return mark_safe('<div class="alert alert-danger">{}</div>'
                          .format(pick_localized(message)))
+    if not PendingSubmissionManager.is_grader_stable():
+        # Prefer configured alert text, if one is set
+        return mark_safe('<div class="alert alert-danger">{}</div>'.format(_('GRADER_PROBLEMS_ALERT')))
     return ''
 
 
--- a/exercise/submission_models.py
+++ b/exercise/submission_models.py
@@ -4,16 +4,18 @@ import logging
 from mimetypes import guess_type
 import os
 from typing import IO, Dict, Iterable, List, Tuple, TYPE_CHECKING
+from urllib.parse import urlparse
 
 from binaryornot.check import is_binary
 from django.conf import settings
 from django.db import models, DatabaseError
+from django.db.models import F
 from django.db.models.signals import post_delete
 from django.http.request import HttpRequest
 from django.utils import timezone
 from django.utils.translation import get_language, gettext_lazy as _
-from exercise.protocol.exercise_page import ExercisePage
 
+from exercise.protocol.exercise_page import ExercisePage
 from authorization.models import JWTAccessible
 from authorization.object_permissions import register_jwt_accessible_class
 from lib.fields import DefaultForeignKey, JSONField, PercentField
@@ -26,6 +28,7 @@ from lib.helpers import (
 )
 from lib.models import UrlMixin
 from userprofile.models import UserProfile
+from aplus.celery import retry_submissions
 from . import exercise_models
 
 
@@ -565,9 +568,11 @@ class Submission(UrlMixin, models.Model):
 
     def set_waiting(self):
         self.status = self.STATUS.WAITING
+        self.mark_pending()
 
     def set_ready(self, approve_unofficial=False):
         self.grading_time = timezone.now()
+        self.clear_pending()
         if self.status != self.STATUS.UNOFFICIAL or self.force_exercise_points or approve_unofficial:
             self.status = self.STATUS.READY
 
@@ -581,11 +586,18 @@ class Submission(UrlMixin, models.Model):
                 "site": settings.BASE_URL,
             })
 
+        if not PendingSubmissionManager.is_grader_stable:
+            # We have a successful grading task in the recovery state. It may be a sign that problems
+            # have been resolved, so immediately retry the next pending submission, to speed up recovery
+            retry_submissions()
+
     def set_rejected(self):
         self.status = self.STATUS.REJECTED
+        self.clear_pending()
 
     def set_error(self):
         self.status = self.STATUS.ERROR
+        self.clear_pending()
 
     @property
     def is_graded(self):
@@ -612,6 +624,22 @@ class Submission(UrlMixin, models.Model):
 
     def get_inspect_url(self):
         return self.get_url("submission-inspect")
+
+    def mark_pending(self):
+        grading_host = urlparse(self.exercise.service_url).netloc
+        if grading_host in settings.SUBMISSION_RETRY_SERVICES:
+            pending, created = PendingSubmission.objects.get_or_create(submission=self)
+            if not created:
+                pending.num_retries = F('num_retries') + 1
+            pending.submission_time = timezone.now()
+            pending.save()
+
+    def clear_pending(self):
+        try:
+            pending = PendingSubmission.objects.get(submission=self)
+            pending.delete()
+        except PendingSubmission.DoesNotExist:
+            pass
 
 
 class SubmissionDraft(models.Model):
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -1429,6 +1429,12 @@ msgstr ""
 "<p>Are you sure you want to ban this user from the course?</p><p>Banned "
 "users cannot re-enrol in the course.</p>"
 
+#: course/templatetags/base.py
+msgid "GRADER_PROBLEMS_ALERT"
+msgstr ""
+"Automatic grading has currently issues. "
+"You may experience delay in receiving feedback."
+
 #: course/views.py
 msgid "COURSE_INSTANCES_NOT_VISIBLE_TO_STUDENTS"
 msgstr "The course instances of this course are not visible to students."
--- a/locale/fi/LC_MESSAGES/django.po
+++ b/locale/fi/LC_MESSAGES/django.po
@@ -1437,6 +1437,12 @@ msgstr ""
 "<p>Haluatko varmasti estää tämän käyttäjän pääsyn kurssille?</p><p>Estetyt "
 "käyttäjät eivät pysty ilmoittautumaan kurssille uudelleen.</p>"
 
+#: course/templatetags/base.py
+msgid "GRADER_PROBLEMS_ALERT"
+msgstr ""
+"Automaattisessa arvioinnissa on tällä hetkellä ongelmia. "
+"Palautteen saamisessa saattaa esiintyä viivettä."
+
 #: course/views.py
 msgid "COURSE_INSTANCES_NOT_VISIBLE_TO_STUDENTS"
 msgstr "Kurssikerrat eivät ole opiskelijoiden nähtävissä."
--- a/aplus/settings.py
+++ b/aplus/settings.py
@@ -237,6 +237,24 @@ SOCIAL_AUTH_USERNAME_IS_FULL_EMAIL = True
 SIS_ENROLL_DELAY = 15
 
 ##########################################################################
+# Settings related automatic retries of unfinished grading tasks
+
+# Number of seconds after which a submission can be resent to grader, if not completed
+# If set to None, retries are disabled.
+SUBMISSION_EXPIRY_TIMEOUT = 30 * 60
+
+# List of services with automatic grading where retries are allowed
+# Network location is sufficient, e.g. "localhost:8080" or "grader.cs.aalto.fi"
+SUBMISSION_RETRY_SERVICES = []
+
+# Number of unresponded retries beyond which we move to recovery state.
+# In recovery state there likely is more persistent problem with the grader
+# or network that needs fixing.
+# In recovery state A+ periodically probes the state of the grader, sending only one
+# grading request out every SUBMISSION_EXPIRY_TIMEOUT seconds.
+# We do not want to congest the potentially broken system unnecessarily with several
+# requests in this case.
+GRADER_STABLE_THRESHOLD = 5
 
 ## Celery
 CELERY_ACCEPT_CONTENT = ['json']
--- a/exercise/admin.py
+++ b/exercise/admin.py
@@ -17,6 +17,7 @@ from exercise.models import (
     RevealRule,
     ExerciseTask,
     LearningObjectDisplay,
+    PendingSubmission,
 )
 from exercise.exercisecollection_models import ExerciseCollection
 from lib.admin_helpers import RecentCourseInstanceListFilter
@@ -440,6 +441,14 @@ class LearningObjectDisplayAdmin(admin.ModelAdmin):
     readonly_fields = ('timestamp',)
 
 
+class PendingSubmissionAdmin(admin.ModelAdmin):
+    list_display = (
+        'submission',
+        'submission_time',
+        'num_retries',
+    )
+
+
 admin.site.register(CourseChapter, CourseChapterAdmin)
 admin.site.register(BaseExercise, BaseExerciseAdmin)
 admin.site.register(StaticExercise, StaticExerciseAdmin)
@@ -452,3 +461,4 @@ admin.site.register(ExerciseCollection, ExerciseCollectionAdmin)
 admin.site.register(RevealRule, RevealRuleAdmin)
 admin.site.register(ExerciseTask, ExerciseTaskAdmin)
 admin.site.register(LearningObjectDisplay, LearningObjectDisplayAdmin)
+admin.site.register(PendingSubmission, PendingSubmissionAdmin)


Mankro (changes requested):
Nice work! There are a few small issues to fix. Then, we will need to test this more carefully in a server environment with a higher number of incoming submissions.

locale/en/LC_MESSAGES/django.po
```diff
@@ -1420,6 +1420,10 @@ msgstr ""
 "<p>Are you sure you want to ban this user from the course?</p><p>Banned "
 "users cannot re-enrol in the course.</p>"
 
+#: course/templatetags/base.py
+msgid "GRADER_PROBLEMS_ALERT"
+msgstr "Automatic grading has problems at the moment. You may experience delay in receiving feedback."
```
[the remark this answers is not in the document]
PasiSa: Works for me, although I think "Automatic grading has issues currently" is better word ordering.

exercise/submission_models.py
```diff
@@ -686,3 +710,31 @@ def _delete_file(sender, instance, **kwargs):
...
+    )
+    num_retries = models.IntegerField(
+        verbose_name=_('LABEL_NUMBER_OF_RETRIES'),
+        default=0,
+    )
+
+    class Meta:
+        verbose_name = _('MODEL_NAME_PENDING_SUBMISSION')
+        verbose_name_plural = _('MODEL_NAME_PENDING_SUBMISSION_PLURAL')
+        constraints = [models.UniqueConstraint(fields=['submission'], name='unique_submission')]
```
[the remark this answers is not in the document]
PasiSa: Yes, I tried that initially, but ran into some problems (don't remember what, since it was three months ago, maybe migrations related), but for some reason the foreign key approach appeared to work. However, now that I retried OneToOne, it works properly. Don't know what it was back then, but OneToOne is indeed how it should be.

course/templatetags/base.py
```diff
@@ -44,6 +45,9 @@ def site_alert():
     if message:
         return mark_safe('<div class="alert alert-danger">{}</div>'
                          .format(pick_localized(message)))
+    if not grader_stable():
```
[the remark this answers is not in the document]
PasiSa: If exercise tests are broken, wouldn't grader still be able to reply with some error (maybe after timeout), instead of just silently dropping the response? So not getting any response would be an some error in grader itself (or in the network, etc.)? But we could of course make this also course specific.

In other context we have also discussed making the sitewide alert adjustable during runtime instead of through settings.py. Maybe at the same time we could implement system for course-specific alerts, aside with run-time sitewide alerts. Then also this kind of "automatic" alerts could be implemented using this new approach. But if we would do something like this, I guess it would be a separate PR and commit. So not doing anything about this yet, pending follow-up discussion.

PasiSa:
force pushed a new version, hopefully addressing comments properly

locale/fi/LC_MESSAGES/django.po
```diff
@@ -1427,6 +1427,10 @@ msgstr ""
 "<p>Haluatko varmasti estää tämän käyttäjän pääsyn kurssille?</p><p>Estetyt "
 "käyttäjät eivät pysty ilmoittautumaan kurssille uudelleen.</p>"
 
+#: course/templatetags/base.py
+msgid "GRADER_PROBLEMS_ALERT"
+msgstr "Automaattisessa arvioinnissa on tällä hetkellä ongelmia. Palautteen saamisessa saattaa esiintyä viivettä."
```
ihalaij1, line 1432: This message and the English version seem too long, which means that the next time someone runs ``makemessages`` it's going to shorten these lines automatically. It happened to me too...

course/templatetags/base.py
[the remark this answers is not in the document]
Mankro: So if the grader responds and sets the submission to the error status in A+, then this PR won't consider the grader unstable and there won't be any warnings in the site? That's an important distinction, indeed.

Submission status in the grader feedback meta attributes:
https://github.com/apluslms/mooc-grader/blob/1f5b16365930f5b18de4c39e427ce2be7e0884d6/access/templates/access/exercise_head.html#L13

exercise/submission_models.py
```diff
@@ -533,6 +544,19 @@ def get_url_kwargs(self):
     def get_inspect_url(self):
         return self.get_url("submission-inspect")
 
+    def mark_pending(self):
+        pending, created = PendingSubmission.objects.get_or_create(submission=self)
+        if not created:
+            pending.num_retries += 1
```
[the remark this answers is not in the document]
Mankro: Did you see this comment in the first review?

exercise/submission_models.py
```diff
@@ -766,3 +789,34 @@ def _delete_file(sender, instance, **kwargs):
...
+
+
+class PendingSubmissionManager(models.Manager):
+
+    def is_grader_stable():
+        total_retries = PendingSubmission.objects.aggregate(sum=models.Sum('num_retries'))['sum']
+        if total_retries and total_retries > settings.GRADER_STABLE_THRESHOLD:
+            return False
+        
+        return True
```
Mankro, line 801:
```suggestion
        return not (total_retries and total_retries > settings.GRADER_STABLE_THRESHOLD)
```

Mankro (changes requested):
Nice work! Only a few small changes are needed! It also feels like that you never looked at some of the comments in the first review, particularly in the file submission_models.py.

aplus/celery.py
```diff
@@ -16,6 +21,10 @@ def setup_periodic_tasks(sender, **kwargs):
     if hasattr(settings, 'SIS_ENROLL_SCHEDULE'):
         sender.add_periodic_task(settings.SIS_ENROLL_SCHEDULE, enroll.s(), name='enroll')
 
+@app.on_after_configure.connect
+def setup_periodic_tasks(sender, **kwargs):
+    sender.add_periodic_task(settings.SUBMISSION_EXPIRY_TIMEOUT, retry_submissions.s(), name='retry_submissions')
```
PasiSa, line 25: I could do one more modification to this PR: possibility to disable this feature from settings, e.g. by setting SUBMISSION_EXPIRY_TIMEOUT to None, so that it would be possible to turn this off, if some unexpected problems show up after deployment.

Mankro: Good idea!

aplus/celery.py
```diff
@@ -32,3 +40,25 @@ def enroll():
...
+        return
+
+    # Stable state: retry all expired submissions
+    expiry_time = datetime.datetime.now(datetime.timezone.utc) - relativedelta(seconds=settings.SUBMISSION_EXPIRY_TIMEOUT)
+    expired = PendingSubmission.objects.filter(submission_time__lt=expiry_time)
+
+    for pending in expired:
+        logger.info(f"Retrying expired submission {pending.submission}")
+        pending.submission.exercise.grade(pending.submission)
+        sleep(0.5)  # Delay 500 ms to avoid choking grader
```
Mankro, line 64:
```suggestion
        if pending.submission.exercise.can_regrade:
            logger.info(f"Retrying expired submission {pending.submission}")
            pending.submission.exercise.grade(pending.submission)
            sleep(0.5)  # Delay 500 ms to avoid choking grader
```
LTI exercises can not be regraded since the submission is not stored in A+; the student sends it directly to the LTI Tool. Though, LTI exercise submissions should never have any PendingSubmission. However, A+ uses the LTIExercise model for Rubyric exercises, but since there is a field `aplus_get_and_post`, then Rubyric exercises still behave almost like normal exercises in grading. But regrading manually graded exercises in Rubyric would not make sense. We don't want to automatically send multiple copies of the same submission to Rubyric until some teacher has manually graded it.
https://github.com/apluslms/a-plus/blob/25e6a22f43fdb283e9a4d6330dc70b68f179630b/exercise/exercise_models.py#L1013-L1016

aplus/celery.py
```diff
@@ -32,3 +40,25 @@ def enroll():
...
+    # We pick the one with most attempts, so that total_retries comes down more quickly
+    # when things are back to normal, and we can return to normal state
+    if not PendingSubmissionManager.is_grader_stable():
+        pending = PendingSubmission.objects.order_by('-num_retries').first()
+        logging.info(f"Recovery state: retrying expired submission {pending.submission}")
+        pending.submission.exercise.grade(pending.submission)
+        return
+
+    # Stable state: retry all expired submissions
+    expiry_time = datetime.datetime.now(datetime.timezone.utc) - relativedelta(seconds=settings.SUBMISSION_EXPIRY_TIMEOUT)
```
Mankro, line 58: I started to think about the default `SUBMISSION_EXPIRY_TIME` in settings.py. Should it be at least 10 minutes or even higher, 30 minutes? There are courses that use much heavier graders. Their execution times are always more than five minutes. There has been a system-wide timeout for the maximum duration. It might be around 30 minutes. (It is configured directly in Kubernetes outside A+.)

Mankro (changes requested):
I got new ideas today. There are assignments that always take a while to finish grading. Some graders are heavy and they don't normally finish in less than 5-10 minutes.

There are manually graded assignments. Their submissions are supposed to stay in the WAITING status for even multiple weeks. Rubyric is one example. However, A+ does not really know which assignment services are fast, slow, or manually graded. At least, we can separate LTI assignments from the normal assignments. LTI assignments can not be regraded since the submission never exists in A+, but an empty submission may be created afterwards for storing the points.

--------------------------------------------------------

I started to think that it could be possible to avoid the whole extra model `PendingSubmission`. If we just query submissions that are in the status WAITING and were submitted a long time ago, then that would reveal stuck submissions in the same way as `PendingSubmission`. It would work without juggling extra states in another table. How important is it to count the number of retries, which is now easier in the `PendingSubmission` model?

PasiSa:
Now addressed the comments I missed on the first time. Also made it possible to disable this feature.

Mankro (commented):
Good thus far! What about the couple of old comments that you haven't replied to? What about the concerns about manually graded assignments (yesterday's comments)?

aplus/celery.py
Mankro: Another example of manually graded exercises: Jutut feedback questionnaires. In the A+ database, they are BaseExercises, no LTI there. We can't regrade Jutut questionnaires since we don't know if the teacher intends to manually grade them at all.

course/templatetags/base.py
```diff
@@ -44,6 +45,9 @@ def site_alert():
     if message:
         return mark_safe('<div class="alert alert-danger">{}</div>'
                          .format(pick_localized(message)))
+    if not PendingSubmissionManager.is_grader_stable():
```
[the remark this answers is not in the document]
PasiSa: Yep, I thought that the user-set alert should have higher priority, if we do not want too many alerts at the same time. Ideally 
(and hopefully) the grading problem alert should be very rare. But perhaps we should rethink the alert system in near future, and at the same time design this more adequately?

PasiSa:
Sorry, I don't know why I missed some of the comments. I use the thumb emoji for my bookkeeping to keep track which comments were addressed, but I nevertheless missed those.
ly
Good point about manual exercises, those indeed should be excluded. I don't remember why we ended up with PendingSubmissions table last spring. There are many years old submissions in the production DB, so at least the query efficiency might be a question.

Amended the easy changes to the last commit, will do can_regrade as separate commit, after daily.

course/templatetags/base.py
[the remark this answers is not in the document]
Mankro: Yes, we will return to this in the future.

Mankro:
@PasiSa Since this has been implemented this far with the PendingSubmissions table, we could at least try this in production. We might learn something of its performance and stability. If necessary, the whole feature could be removed later. Removing is easy, though removing the whole table will require migrations. The feature could be first removed from the code even if the table exists in the database (basically, unused table waiting to be removed later).

However, the manually graded assignments are still an open issue. I don't think we can deploy this at all until we have a solution for manual grading. Do you have any ideas? 

One simple idea is to add Django settings that list the exercise services (their URLs) and the expected grading times. For example, rubyric.cs and jutut.cs - no automatic regrading; grader.cs - results should arrive in at most 30 minutes. The `BaseExercise` `service_url` can be compared to the listed service URLs before sending it to regrading. The regrading could be then skipped for the specified "manual grading services". However, I am interested in if we could avoid manually configuring all these details since it feels awkward and annoying at first sight.

PasiSa:
> One simple idea is to add Django settings that list the exercise services (their URLs) and the expected grading times. For example, rubyric.cs and jutut.cs - no automatic regrading; grader.cs - results should arrive in at most 30 minutes. The `BaseExercise` `service_url` can be compared to the listed service URLs before sending it to regrading. The regrading could be then skipped for the specified "manual grading services". However, I am interested in if we could avoid manually configuring all these details since it feels awkward and annoying at first sight.

Ok. let's do this. So the SUBMISSION_EXPIRY_TIMEOUT setting could consist of ("service hostname", timeout) pairs. If a service is not mentioned, it will not use this feature. I guess hostname part of the URL is sufficient for configuration?

Mankro:
> Ok. let's do this. So the SUBMISSION_EXPIRY_TIMEOUT setting could consist of ("service hostname", timeout) pairs. If a service is not mentioned, it will not use this feature. I guess hostname part of the URL is sufficient for configuration?

OK! Yes, host name (such as grader.cs.aalto.fi) is enough.

> Ideally, there could be a setting in exercise configuration that (optionally) specifies exercise-specific timeout because there can be much variation between exercises. But this is for some later time...

Nice, that would make sense (in the future).

PasiSa:
> One simple idea is to add Django settings that list the exercise services (their URLs) and the expected grading times. For example, rubyric.cs and jutut.cs - no automatic regrading; grader.cs - results should arrive in at most 30 minutes. The `BaseExercise` `service_url` can be compared to the listed service URLs before sending it to regrading. The regrading could be then skipped for the specified "manual grading services". However, I am interested in if we could avoid manually configuring all these details since it feels awkward and annoying at first sight.

Ugh, I propose even further simplification to this idea: for now, let's use one global deadline for all submissions that are automatically graded. Currently the retried submissions are filtered from pending table with this simple filter:
`PendingSubmission.objects.filter(submission_time__lt=expiry_time)`

If different submissions had different timeouts, then the query would become more complicated. It might even be best to add deadline field to the pending submissions table in that case. Maybe, if later we wanted to implement exercise specific timeouts, that would be needed in any case.

But for now, maybe we would go with global timeout value in one setting, and list of service hosts automatically graded as another setting?

Mankro:
> > One simple idea is to add Django settings that list the exercise services (their URLs) and the expected grading times. For example, rubyric.cs and jutut.cs - no automatic regrading; grader.cs - results should arrive in at most 30 minutes. The `BaseExercise` `service_url` can be compared to the listed service URLs before sending it to regrading. The regrading could be then skipped for the specified "manual grading services". However, I am interested in if we could avoid manually configuring all these details since it feels awkward and annoying at first sight.
> 
> Ugh, I propose even further simplification to this idea: for now, let's use one global deadline for all submissions that are automatically graded. Currently the retried submissions are filtered from pending table with this simple filter: `PendingSubmission.objects.filter(submission_time__lt=expiry_time)`
> 
> If different submissions had different timeouts, then the query would become more complicated. It might even be best to add deadline field to the pending submissions table in that case. Maybe, if later we wanted to implement exercise specific timeouts, that would be needed in any case.
> 
> But for now, maybe we would go with global timeout value in one setting, and list of service hosts automatically graded as another setting?

Yes, that is fine!

PasiSa:
Latest push now includes the option to list services to retry.

exercise/submission_models.py
```diff
@@ -613,6 +625,22 @@ def get_url_kwargs(self):
...
+                pending.num_retries = F('num_retries') + 1
+            pending.submission_time = timezone.now()
+            pending.save()
+
+    def clear_pending(self):
+        try:
+            pending = PendingSubmission.objects.get(submission=self)
+            pending.delete()
+        except PendingSubmission.DoesNotExist:
+            logger.warn(f"Submission {self} completed, but was not on PendingSubmissions table.")
```
Mankro, line 642: Isn't this going to print warnings for all submissions in exercise services that are not included in `settings.SUBMISSION_RETRY_SERVICES`?

Mankro (approved):
Thanks! I had one question, which may require a small code change. Then, you may rebase and clean up the git history. Remember to mention the previously merged commit/pull request about the `PendingSubmission` model in the commit message!

exercise/submission_models.py
PasiSa: Oops, indeed. Perhaps it is best just to silently ignore this exception. Changed this.

Mankro (approved):
@PasiSa Nice, this is finished now. You should rebase and clean up the git history. Remember to mention the previously merged commit/pull request about the PendingSubmission model in the commit message! (The PendingSubmission model was already merged some time ago in another PR and it would be best to link these together.)

PasiSa:
commit history cleaned, message edited. Ready for merging, I suppose.

exercise/submission_models.py
```diff
@@ -581,11 +586,18 @@ def set_ready(self, approve_unofficial=False):
                 "site": settings.BASE_URL,
             })
 
+        if not PendingSubmissionManager.is_grader_stable:
```
Mankro, line 589: I must have originally missed this. I think this line is very broken currently. Without the parentheses `()`, `is_grader_stable` refers to the function object, which I assume is **always truthy** in the `if` condition.

In addition, manager instances are used via the `objects` field of the model (or other field similarly, but `objects` is the de facto standard name for the default manager). The method `is_grader_stable` has not been marked as a static method even though you are now trying to use it as if it were static.

```python
# normal method, not static
if not PendingSubmission.objects.is_grader_stable():

# if the method were static, Python still allows calling it via the object instance
if not PendingSubmission.objects.is_grader_stable():

# but static methods could also be called like this.
# If we used this case, then it was not very useful to move
# the method definition to the PendingSubmissionManager class.
if not PendingSubmissionManager.is_grader_stable():
```

Mankro: In the method definition, note that normal methods MUST have at least one parameter `self` (the name does not technically matter, but it would be a dire breach of the best practices to use any other name than `self`).

If we used static methods, they would need the decorator `@staticmethod`. No `self` parameter is used in a static method.
https://docs.python.org/3/library/functions.html#staticmethod

Manager objects can (should even?) refer to the query methods via `self`. It is not necessary to hardcode the partner model class's name `PendingSubmission` when you can use `self`.

NEW SOLUTION
```python
class PendingSubmissionManager(models.Manager):
    def is_grader_stable(self):
        total_retries = self.aggregate(sum=models.Sum('num_retries'))['sum']
        return not (total_retries and total_retries > settings.GRADER_STABLE_THRESHOLD)
```
https://docs.djangoproject.com/en/3.2/topics/db/managers/

OLD broken version, currently in the master branch. No `self` parameter nor `@staticmethod`.
```python
class PendingSubmissionManager(models.Manager):
    def is_grader_stable():
        total_retries = PendingSubmission.objects.aggregate(sum=models.Sum('num_retries'))['sum']
        return not (total_retries and total_retries > settings.GRADER_STABLE_THRESHOLD)
```

Mankro (commented):
I just found severe issues in the `is_grader_stable()` method. It's best to fix these soon.

exercise/submission_models.py
Mankro: Fixed in https://github.com/apluslms/a-plus/pull/1103

PR #1036, opened 2022-05-19
Add improvements to resultsdata db query

### LINKED ISSUE #918: Make resultsdata DB query significantly faster (for improved All Results performance)
The All results page still loads too slowly. Loading the page may take 5-10 minutes on a medium-sized course.

This needs to be improved. It is best to start from scratch instead of trying to modify the current code. The details of the page can be changed compared to the current version. E.g., it is not necessary to include all the same data and the UI could be very different.

This requires a lot of designing first. The design shall be discussed and brainstormed together.

Questions:

* what data/information is really needed?
* how should the data be presented and structured? (Probably some kind of tables are still needed.)
* what kind of backend API endpoints support this page in the best way so that the page loads remain fast.
* what technologies to use in the implementation? The backend is still Django, but the frontend could be something new. New JS libraries, React.js?

It is IMPORTANT to be able to download the data in CSV/Excel file format.
Filtering based on student tags and text searches (for student's name or score range) are probably needed.

-----------------------------

The current implementation: resultsdata API endpoint and a huge JavaScript code file for building the table and calculating some statistics.

* https://github.com/apluslms/a-plus/blob/v1.11.3/exercise/api/csv/views.py#L251-L259
* https://github.com/apluslms/a-plus/blob/v1.11.3/exercise/static/exercise/results_staff.js

Previous work:

* #797 
* #678 
* #450

joni-toivonen:
# Description

**What?**

This commit adds separate resultsdata query which ignores last-only grading rule in exercises. This query is used by default in all results view and the previous query can be enabled with a checkbox.

**Why?**

Previously used resultsdata query was very slow so a new one was made that is much faster

**How?**

Previous resultsdata contained a certain case which made the query very slow. New one is basically the same without the slow part.

Fixes #918 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested the resultsdata performance from API interface with django debug tools, also tested the All results page and made sure that the checkbox which alternates between the two works correctly.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/api/urls_v2.py
+++ b/api/urls_v2.py
@@ -59,8 +59,8 @@ with api.register(r'courses',
     courses.register(r'resultsdata',
                      exercise.api.csv.views.CourseResultsDataViewSet,
                      basename='course-resultsdata')
-    courses.register(r'resultsdatabest',
-                     exercise.api.csv.views.CourseResultsDataViewSetBest,
+    courses.register(r'bestresultsdata',
+                     exercise.api.csv.views.CourseBestResultsDataViewSet,
                      basename='course-resultsdata-best')
     courses.register(r'mygroups',
                      course.api.views.CourseOwnStudentGroupsViewSet,
--- a/api/urls_v2.py
+++ b/api/urls_v2.py
@@ -59,6 +59,9 @@ with api.register(r'courses',
     courses.register(r'resultsdata',
                      exercise.api.csv.views.CourseResultsDataViewSet,
                      basename='course-resultsdata')
+    courses.register(r'resultsdatabest',
+                     exercise.api.csv.views.CourseResultsDataViewSetBest,
+                     basename='course-resultsdata-best')
     courses.register(r'mygroups',
                      course.api.views.CourseOwnStudentGroupsViewSet,
                      basename='course-mygroups')
--- a/exercise/api/csv/views.py
+++ b/exercise/api/csv/views.py
@@ -317,7 +317,7 @@ class CourseResultsDataViewSet(NestedViewSetMixin,
     def retrieve(self, request, version=None, course_id=None, user_id=None):
         return self.serialize_profiles(request, [self.get_object()])
 
-    def get_aggr(self, ids, profiles, exclude_list, revealed_ids, show_unofficial):
+    def get_submissions_query(self, ids, profiles, exclude_list, revealed_ids, show_unofficial):
         return (
             Submission.objects
             .filter(exercise__in=ids, submitters__in=profiles)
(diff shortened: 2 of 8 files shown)


exercise/api/csv/views.py
```diff
@@ -353,6 +353,102 @@ def get_renderer_context(self):
         context['header'] = getattr(self, 'renderer_fields', None)
         return context
 
+class CourseResultsDataViewSetBest(NestedViewSetMixin,
```
PasiSa, line 356: I know we discussed that we should start with a quickly deployable (even if ugly) fix to make the results page work again ASAP, but we should nevertheless try to avoid unnecessary code duplication, when it is easily doable: there is only line that differs from the other function as far as I can tell. For example, no need to copy-paste comments word-to-word. Instead the description text could tell that this is a quick fix that will be improved later. If this class would inherit the original, at least the completely duplicate definitions could be avoided, right? Also, if inheritance relationship was applied, splitting the serialize_profiles function into common parts and to the differing part that collects 'aggr' would perhaps be another way of avoiding completely duplicate code? (e.g. a 'aggr = collect_submissions' function that will be separately implemented in the two classes, otherwise the serialize_profiles implementation seems common)

exercise/submission_models.py
```diff
@@ -124,6 +124,81 @@ def annotate_submitter_points(
...
         )
 
+    def annotate_submitter_points_best(
+            self,
+            field_name: str = 'total',
+            revealed_ids: Iterable[int] = None,
+            include_unofficial: bool = False,
+            ) -> 'SubmissionQuerySet':
+        """
+        Annotates the total points earned by the submitter in the exercise to
```
PasiSa, line 134: Also here the comments could highlight that this is a quick fix for an immediate need that should be improved later.

api/urls_v2.py
```diff
@@ -59,6 +59,9 @@ class AplusRouter(ExtendedDefaultRouter):
     courses.register(r'resultsdata',
                      exercise.api.csv.views.CourseResultsDataViewSet,
                      basename='course-resultsdata')
+    courses.register(r'resultsdatabest',
```
PasiSa, line 62: Could add a comment that this is a temporary experimental fix, and the API endpoint should not be used by external applications.

Mankro (approved):
Great! I think some variable names and code comments are lacking or misleading. I will fix them myself and merge this pull request.

PR #1040, opened 2022-05-30
Submission Duplicate Check [DRAFT]

### LINKED ISSUE #1006: Detect and ignore duplicate submissions
User may accidentally make a submission that is exact duplicate of an earlier submission for the same exercise. A+ could detect and ignore these submissions, to avoid unnecessary consuming available submissions for the exercise.

muhammadfaiz12:
# Description

**What?**
This feature check if a user will make a duplicate submission. If the system detects that the user will make a duplicate submission it will prompt a confirmation from the user. User would be able to submit the submission if they indicate so in the confirmation. No changes of behaviour happened if the user make a non-duplicate submission.

**Why?**

To prevent user to make accidental duplicate submission.

**How?**

The system create a submission hash for submitted submission. this hash was stored in both Database and Cache. This hashes are rendered to Front end when user open exercise page. When user want to submit a submission, the front end system create hash of the submission and compare it to the old submissions hashes. Front end  side create hash to compare new submission hash with the previous submissions hashes, Back end side then recompute the hash to store the hash in the database. The FE created hash and BE created hash has to be the same (as per now, it still different)

Fixes #1006 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -318,14 +318,7 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     'url': submission.get_url('submission-plain'),
                     'feedback_revealed': True,
                     'feedback_reveal_time': None,
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
-                    'submission_hash' : submission.meta_data.get('submission_hash', None),
->>>>>>> 71300be6... complete backend
-=======
-                    'hash' : submission.meta_data.get('submission_hash', None),
->>>>>>> bc9e7e58... feat: add hash storage
+                    'submission_hash' : submission.meta_data.get('submission_hash', None), #Duplicate Check: We insert the submission hash to Cache
                 })
                 # Update best submission if exercise points are not forced, and
                 # one of these is true:
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -318,6 +318,10 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     'url': submission.get_url('submission-plain'),
                     'feedback_revealed': True,
                     'feedback_reveal_time': None,
+<<<<<<< HEAD
+=======
+                    'submission_hash' : submission.meta_data.get('submission_hash', None),
+>>>>>>> 71300be6... complete backend
                 })
                 # Update best submission if exercise points are not forced, and
                 # one of these is true:
--- a/exercise/cache/points.py
+++ b/exercise/cache/points.py
@@ -319,9 +319,13 @@ class CachedPoints(ContentMixin, CachedAbstract):
                     'feedback_revealed': True,
                     'feedback_reveal_time': None,
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
                     'submission_hash' : submission.meta_data.get('submission_hash', None),
 >>>>>>> 71300be6... complete backend
+=======
+                    'hash' : submission.meta_data.get('submission_hash', None),
+>>>>>>> bc9e7e58... feat: add hash storage
                 })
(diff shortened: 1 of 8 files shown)


exercise/submission_models.py
```diff
@@ -154,6 +156,10 @@ def create_from_post(self, exercise, submitters, request):
         if 'lang' not in meta_data_dict:
             meta_data_dict['lang'] = get_language()
 
+        #create Hash for the new submission
+        new_hash = helpers.get_hash(submission_data_list, request.FILES)
```
Mankro, line 160: This computes the hash without the `__aplus__` key in the POST data. I think the JS code chapter.js includes those keys in its hash computation. This is one reason why the hashes become different in the frontend and backend.

The `get_hash()` function definition uses `*params`, but this line calls the function with a normal list without unpacking it.

exercise/views.py
```diff
@@ -342,7 +344,31 @@ def _load_exercisecollection(self, request, submission_disabled):
             }
 
         return loaded_content
-
+    
+    #This code is unused. Kept as this is the original backend code for duplicate check
```
Mankro, line 348: Note: this unused code could be removed in the end before merging.

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,68 @@
...
+          });
+      } ;
+
+        //The code for duplicate check
+        //first, we retrieve all the hashes of the previous submissions. The hashes are rendered in the HTML 
+        //You could run the  line 613-622 on browser console to see what each line does. Open a exerise page then submit a quizz.
+
+        var isDuplicate = true;
+        let hashes = new Set()
+        let datahash_elem = Object.entries($("[data-hash]"))
```
Mankro, line 618: This line queries all html elements with the attribute `data-hash` from the whole document. It should include hashes only for the currently active exercise.

exercise/cache/points.py
```diff
@@ -318,6 +318,7 @@ def check_reveal_rule() -> None:
                     'url': submission.get_url('submission-plain'),
                     'feedback_revealed': True,
                     'feedback_reveal_time': None,
+                    'submission_hash' : submission.meta_data.get('submission_hash', None), #Duplicate Check: We insert the submission hash to Cache
```
Mankro, line 321: Now that we have moved the hash comparisons to the frontend JavaScript, we should think again if the hashes are even needed in the cache. Does the current code use the cache when rendering the hashes in the html template (exercise_plain.html, data attributes)?

lib/helpers.py
```diff
@@ -54,6 +57,19 @@ def get_random_string(length=32, choices=None):
...
+def get_hash(*params):
+    """
+    This function creates a hash of the given params.
+
+    @param length: The params that want to be hashed.
+    @return: Single hash that represent all the params given, a string
+    """
+    dhash = hashlib.md5()
+    for p in params:
+        param_encoded = json.dumps(p, sort_keys=True, default=str).encode()
```
Mankro, line 69: At first sight, I think the current backend code here and the frontend code chapter.js serialize the submission data differently to JSON. If the JSON string becomes different for the same submission data, then the output hash is also different. (They should be computed identically in the frontend and backend.)

Mankro (commented):
Thanks! This is a good start. You have documented what the missing parts are. I noticed some potential causes for the different hashes when they are computed in the frontend and backend for the same submission data.

Mankro:
Closing this pull request. The newer PR #1048 replaces this one.

PR #1041, opened 2022-06-01
Changes to relative URLs on static content

### LINKED ISSUE #974: Add feature for mapping static file URLs to another domain than just the exercise service URL
The O1 course wants to build the RST course without the `static_host` setting in conf.py because they commit the built html to the course git repo. O1 doesn't want to use `static_host` since in their case, its value should be different in local testing versus the production server. In order to make this work without `static_host`, new URL manipulation in the  A+ front is required (`a-plus/lib/remote_page.py fix_relative_urls()`):

A+ could set relative static URLs to some other domain/netloc than the original backend service (e.g., images in grader.cs assignment pointing to gitmanager.cs). It would probably be easy since it would be just a mapping from a domain to another. A+ would need to check the mappings for every relative URL in all loaded chapters and assignments, though.

O1 concerns in aplusguru: https://rt.cs.aalto.fi/Ticket/Display.html?id=20318

## Background

With the new Git Manager in v1.12, the course static files are hosted in gitmanager.cs instead of grader.cs (MOOC-Grader). Images and static file URLs were broken inside assignment descriptions, i.e., the embedded exercise frame in an A+ chapter page, if RST courses were compiled without the `static_host` setting in conf.py

Aplusguru tickets:

* O1: https://rt.cs.aalto.fi/Ticket/Display.html?id=20412
* https://rt.cs.aalto.fi/Ticket/Display.html?id=20318
* AI 2022: https://rt.cs.aalto.fi/Ticket/Display.html?id=20496

The image URLs pointed to grader.cs when they should point to gitmanager.cs. The domain grader.cs was taken from the exercise service URL (in `a-plus/lib/remote_page.py fix_relative_urls()`).

a-plus-rst-tools compiles links relatively because that's how Sphinx works. The setting `static_host = os.environ.get('STATIC_CONTENT_HOST)` can be used to make static URLs absolute with the netloc/domain already in the course build.

The new Git Manager automatically sets the build environment variable `STATIC_CONTENT_HOST="https://gitmanager.cs.aalto.fi/static/COURSEKEY"`

EDIT Jira: https://jira.aalto.fi/browse/EDIT-795

PasiSa:
Previously the code fixing relative URLs pointing to static content
assumed that they are located on grader server. These days the
static content is on Gitmanager, so the URLs should be fixed
accordingly.

Closes #974.

# Description

**What?**

As described in commit message. Am I thinking too simply about this? Assuming that GITMANAGER_URL only specifies the scheme and network location, as comment in recent commit by Jimmy suggests.

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #974

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lib/remote_page.py
+++ b/lib/remote_page.py
@@ -237,6 +237,16 @@ class RemotePage:
         # URLs /plain, /info, /info/model, /info/template.
         exercise_info = re.compile(r'/((plain)|(info(/model|/template)?))/?(#.+)?$')
 
+        # If Gitmanager is in use, fix relative static URLs to that host
+        if settings.GITMANAGER_URL:
+            gmurl = urlparse(settings.GITMANAGER_URL)
+            if settings.REMOTE_PAGE_HOSTS_MAP:
+                domain = settings.REMOTE_PAGE_HOSTS_MAP.get(gmurl.netloc, gmurl.netloc)
+                gmurl = gmurl._replace(netloc=domain)
+            staticurl = url._replace(scheme=gmurl.scheme, netloc=gmurl.netloc)
+        else:
+            staticurl = url
+
         for element in self.soup.find_all(tag_name, {attr_name:True}):
             value = element[attr_name]
             if not value:
@@ -312,15 +322,16 @@ class RemotePage:
 
                 # url points to the exercise service, e.g., MOOC-Grader.
                 # This fixes links to static files (such as images) in RST chapters.
-                # The image URL must be absolute and refer to the grader server
-                # instead of the A+ server. A relative URL with only path
-                # "/static/course/image.png" would target the A+ server when
-                # it is included in the A+ page. The value should be a relative
+                # The image URL must be absolute and refer to the server with static content
+                # instead of the A+ server. Traditionally this has been the grader server, but
+                # recently gitmanager is used. If GITMANAGER_URL is specified, we assume gitmanager.
+                # A relative URL with only path "/static/course/image.png" would target the A+ server
+                # when it is included in the A+ page. The value should be a relative
                 # path in the course build directory so that it becomes the full
                 # correct URL to the target file.
                 # E.g., urljoin('http://localhost:8080/static/default/module1/chapter.html', "../_images/image.png")
                 # -> 'http://localhost:8080/static/default/_images/image.png'
-                element[attr_name] = urljoin(url.geturl(), value)
+                element[attr_name] = urljoin(staticurl.geturl(), value)
 
     def find_and_replace(self, attr_name, list_of_attributes):
         l = len(list_of_attributes)


Mankro (approved):
Excellent! I still need to test this myself before merging. I think this simple approach is adequate for the issue. Another option could be to use a configurable mapping so that the admin would define in the settings that only grader URLs are manipulated like this (to point to gitmanager). However, we don't need that now. The URL manipulations in the A+ front are rather hacky and it could be nice to remove them completely some day. If we could prepare all the links/URLs in the course build phase, then the A+ front wouldn't need to do anything for them.

The commit message doesn't follow the style guide. The message heading should usually start with a verb, e.g., "Fix the static URLs in exercise descriptions"

lib/remote_page.py
```diff
@@ -312,14 +312,18 @@ def _fix_relative_urls(self, url, tag_name, attr_name):
...
+                # recently gitmanager is used. If GITMANAGER_URL is specified, we assume gitmanager.
+                # A relative URL with only path "/static/course/image.png" would target the A+ server
+                # when it is included in the A+ page. The value should be a relative
                 # path in the course build directory so that it becomes the full
                 # correct URL to the target file.
                 # E.g., urljoin('http://localhost:8080/static/default/module1/chapter.html', "../_images/image.png")
                 # -> 'http://localhost:8080/static/default/_images/image.png'
+                if settings.GITMANAGER_URL:
+                    gmurl = urlparse(settings.GITMANAGER_URL)
+                    url = url._replace(scheme=gmurl.scheme, netloc=gmurl.netloc)
```
Mankro, line 326: What do you think of these?

* This `url` variable is re-assigned inside the for loop. You lose the original method parameter value. If `url` were used elsewhere inside the loop, you would need to be careful about changing the value here so that it does not affect the next loop iteration in unwanted ways.
* This repeats the same `url` modification on multiple loop iterations. Only one time could be enough.
* In local testing, `GITMANAGER_URL` is `http://gitmanager:8070`. However, that address is not intended for the web browser as the browser should use the address `http://localhost:8070`. This always uses the incorrect address that does not work in the web browser unless you set a mapping `127.0.0.1 gitmanager` in `/etc/hosts` (requires root privileges in the host machine).
  * A+ already has the workaround: setting `REMOTE_PAGE_HOSTS_MAP`. However, this new code ignores that setting and overrides the `url` netloc with `GITMANAGER_URL`.
  * https://github.com/apluslms/a-plus/blob/9c7e6073b9554e959140c6f125ee2faedc9c7e60/lib/remote_page.py#L133-L136
  * https://github.com/apluslms/run-aplus-front/blob/b14596727258a221c039f178df80670edac1dcc7/rootfs/srv/aplus-cont-settings.py#L102-L104


```
REMOTE_PAGE_HOSTS_MAP = {
    "grader:8080": "localhost:8080",
    "gitmanager:8070": "localhost:8070",
}
```

Mankro (changes requested):
While testing this locally in the containers, I found out some new concerns.

Mankro:
I made this change in the run-aplus-front container so that it could be used with gitmanager and without it (so that grader serves the static content):
https://github.com/apluslms/run-aplus-front/commit/b14596727258a221c039f178df80670edac1dcc7

A+ `GITMANAGER_URL` is defined only when the environment variable declares that gitmanager is being used.

PasiSa:
Thanks for comments! How about now?

Mankro (approved):
Excellent, thanks!

PR #1048, opened 2022-06-21
Add duplicate submission checks

### LINKED ISSUE #1006: Detect and ignore duplicate submissions

ihalaij1:
# Description

**What?**

When submitting a duplicate submission a modal dialog now asks for confirmation from the user whether or not they want to submit the identical submission.

**Why?**

Students sometimes accidentally submit identical code / questionnaire answers.

**How?**

An md5 hash based on the submission data is created in the frontend JavaScript (chapter.js) and appended to the submitted formData.
In the case of a file submit exercise, the hash is created from the files concatenated into a string.
In the case of a questionnaire, a JSON string from the questionnaire fields and answers is created, and this string is hashed.

The hash is rendered in the exercise HTML the next time, and a check is performed to see if the subsequent submissions have an identical hash. A modal dialog is shown if the same hash already exists in the HTML.

Fixes #1006

This PR is based on the draft #1040

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that duplicate submissions are detected correctly and that the modal dialog works.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


ihalaij1:
I couldn't figure out an easy way to add information to the modal dialog about which submissions are the detected duplicates.
Do you have any ideas or should it be left like this?

exercise/templates/exercise/exercise.html
```diff
@@ -74,6 +76,7 @@
 <script src="{% static 'exercise/group.js' %}"></script>
 <script src="{% static 'exercise/autosave.js' %}"></script>
 <script src="{% static 'exercise/chapter.js' %}"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/blueimp-md5/2.19.0/js/md5.min.js"></script>
```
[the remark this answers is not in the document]
ihalaij1: This cdn is from the draft. Should be changed to something else?

exercise/submission_models.py
```diff
@@ -234,6 +234,8 @@ def create_from_post(self, exercise, submitters, request):
         if 'lang' not in meta_data_dict:
             meta_data_dict['lang'] = get_language()
 
+        meta_data_dict["submission_hash"] = dict(submission_data_list).get("__submission_hash", "")
```
Mankro, line 237: Would it make sense to move the submission hash under the `__aplus__` key in the POST data? Most of the POST data should be part of the submission. The special fields could be all under the `__aplus__` key (dictionary) when the client sends the submission POST message. Now there is the `__aplus__` key and also another special field `__submission_hash` in the POST data.

The HTTP POST data could look like this under the special `__aplus__` field:
```
"__aplus__": {
    "lang": "en",
    "hash": "xyzabc",
}
```

exercise/templates/exercise/_duplicate_submission_modal.html
```diff
@@ -0,0 +1,48 @@
...
+                    <span aria-hidden="true">&times;</span>
+                </button>
+                <h4 class="modal-title" id="duplicate-submission-modal-label">
+                    {% translate "DUPLICATE_SUBMISSION_TITLE" %}
+                </h4>
+            </div>
+            <div class="modal-body">
+                {% blocktranslate trimmed %}
+                    DUPLICATE_SUBMISSION
+                {% endblocktranslate %}
```
Mankro, line 27: If the string contains no parameters, then use `{% translate %}` instead of `blocktranslate`.

exercise/templates/exercise/_duplicate_submission_modal.html
```diff
@@ -0,0 +1,48 @@
+{% load i18n %}
+<div
+    class="modal"
```
Mankro, line 3: Does this file use spaces for indentation? New HTML code should use tabs.

exercise/templates/exercise/exercise.html
```diff
@@ -58,6 +58,8 @@
 
 {% include "exercise/_exercise_wait.html" %}
 
+{% include "exercise/_duplicate_submission_modal.html" %}
```
Mankro, line 61: Bootstrap docs recommend that the modal HTML is not nested inside other HTML elements (except the `<body>)`. Nesting causes some risk that inherited CSS styles would affect the modal in unintended ways. At least, you could move the modal include to the bottom of this file outside the `<div>` elements.

exercise/templates/exercise/exercise.html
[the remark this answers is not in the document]
Mankro: Let's host the JS file locally in the A+ server and avoid new references to the cloud. At least in exams, it is better if the static files are hosted locally. We already have a simple build pipeline for including NPM JavaScript packages in the A-plus git repo. Look at Bootstrap3 for an example.

* https://github.com/apluslms/a-plus/tree/df64e846d83436bccd92e310662f449a4ff74210/assets_src/bootstrap3
* Running `drone exec` is supposed to download the JS packages.
* https://github.com/apluslms/a-plus/blob/df64e846d83436bccd92e310662f449a4ff74210/.drone.yml#L6
* https://github.com/apluslms/a-plus/blob/df64e846d83436bccd92e310662f449a4ff74210/doc/README.md#developing-the-code-of-a-this-repository

exercise/templates/exercise/exercise_plain.html
```diff
@@ -80,7 +80,7 @@
 								<ul class="dropdown-menu">
 									{% for submission in submissions %}
 										<li>
-											<a href="{{ submission|url:'submission-plain' }}">
+											<a href="{{ submission|url:'submission-plain' }}" data-hash="{{ submission.meta_data.submission_hash }}">
```
Mankro, line 83: `exercise_plain.html` affects only exercises embedded in chapters. A+ also supports full-view exercises without chapters. Is the duplicate submission check supported in full-view exercises yet? If I remember correctly, the JS code in chapter.js also does not affect full-view exercises, in which case it could be better to write reusable JS functions outside chapter.js so that they can be called outside chapter.js too.

Full-view exercise html template:
https://github.com/apluslms/a-plus/blob/df64e846d83436bccd92e310662f449a4ff74210/exercise/templates/exercise/exercise_base.html#L59-L63

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
         });
       } else {
-        var exercise = this;
-        if (exercise.quiz) {
-          exercise.quizSubmission("submit");
-        } else {
-          /* TODO: migrate to overlay */
-          chapter.openModal(chapter.messages.submit);
+        // Callback for submitting an exercise
+        var submitCallback = function(exercise, hash) {
```
Mankro, line 583: Why not `const`?

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
-            exercise.dom_element.dispatchEvent(
-              new CustomEvent("aplus:submission-finished",
-                {bubbles: true, detail: {type: exercise.exercise_type}}));
+        var isDuplicateCallback = function(exercise, hash) {
+          isDuplicate = hashes.has(hash);
+          if (isDuplicate) {
+            var continueBtn = document.getElementById("duplicate-submission-modal-button");
+            continueBtn.onclick = () => {
+              submitCallback(exercise, hash);
+            }
```
Mankro, line 628: `onclick` is the old outdated API. Let's use the jQuery API here to attach the click event handler:
https://api.jquery.com/click/

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
-        });
+        }
+
+        // Compute a hash for the current submission
+        const inputFileElements = form_element.querySelectorAll('input[type="file"]');
+
+        if (inputFileElements.length > 0) {
+          const reader = new FileReader();
+          reader.onerror = function() {
+            alert("File cannot be opened");
```
Mankro, line 641: Opening alert popups is not great for usability. Perhaps the duplicate check could be skipped if the hash can not be computed at all?

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
-            exercise.updateSubmission(input);
+            submitCallback(exercise, hash);
           }
-        });
+        }
+
+        // Compute a hash for the current submission
+        const inputFileElements = form_element.querySelectorAll('input[type="file"]');
+
+        if (inputFileElements.length > 0) {
```
Mankro, line 638: Note that the exercise form may include both normal inputs and file inputs at the same time, even though often file upload exercises do not include any other visible form fields.

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
+              exercise.dom_element.dispatchEvent(
+                new CustomEvent("aplus:submission-finished",
+                  {bubbles: true, detail: {type: exercise.exercise_type}}));
+            } else {
+              exercise.updateSubmission(input);
+            }
+          });
+        };
+
+        let isDuplicate = false;
```
Mankro, line 611: Looks like there is no need at all to define the variable `isDuplicate` outside the function `isDuplicateCallback` that actually uses it. Let's keep the variable's scope as small as it really needed.

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
+          const formProps = Object.fromEntries(formData);
+
+          // Grader language shouldn't affect the hash
+          if (formProps.hasOwnProperty('__grader_lang')) {
+            delete formProps['__grader_lang'];
+          }
+
+          const formDataJsonString = JSON.stringify(formProps, Object.keys(formProps).sort());
+          const hash = md5(formDataJsonString);
+          isDuplicateCallback(this, hash);
```
Mankro, line 674: Most of the new code here in chapter.js should be moved to a new JS file. Define functions and call the functions here in chapter.js. It will make the code cleaner and chapter.js won't look so bloated. The duplicate check functions would also be more reusable and they could be used outside the chapter.js code too.

locale/en/LC_MESSAGES/django.po
```diff
@@ -2441,7 +2443,9 @@ msgstr "Batch create and assess submissions"
 
 #: edit_course/templates/edit_course/batch_assess.html
 #, python-format
-msgid "BATCH_CREATE_AND_ASSESS_JSON_FORM_INSTRUCTIONS -- %(id)s, %(exercises_api_url)s"
+msgid ""
+"BATCH_CREATE_AND_ASSESS_JSON_FORM_INSTRUCTIONS -- %(id)s, "
+"%(exercises_api_url)s"
```
Mankro, line 2448: How did these changes appear in this pull request?

exercise/static/exercise/chapter.js
```diff
@@ -579,31 +579,100 @@
...
+          });
+        };
+
+        let isDuplicate = false;
+        // Retrieve all the hashes of the previous submissions from the HTML
+        const hashes = new Set();
+        const datahashElements = this.element.find("[data-hash]");
+        for (let elem of datahashElements) {
+          try {
+            hashes.add(elem.dataset["hash"]);
```
Mankro, line 617: If the hashes are stored in a list/array and the indexes match the order of the submissions, then you would later know which submission is the duplicate one. It would be best to show the duplicate submission number to the user so that they can more easily check the submission contents.

Mankro (changes requested):
Nice, this is a good start! 

* I think the JavaScript code should be reorganized and cleaned up a lot.
* The full-view exercise pages don't seem to have any duplicate checks yet (only chapter pages have it now).
* git commit history: is Muhammad's code used any more? His commits could be removed if his code isn't used in the final version. At any rate, the commit message may acknowledge that the work started from his draft pull request.
If Muhammad's code is still used, then we can squash his commits into one. Since a git commit has only one author, we can avoid squashing commits from different authors into a combined commit.

Mankro:
> I couldn't figure out an easy way to add information to the modal dialog about which submissions are the detected duplicates. Do you have any ideas or should it be left like this?

If the hashes are stored in a list/array instead of a set and the indexes match the order of the submissions, then you would later know which submission is the duplicate one. It would be best to show the duplicate submission number to the user so that they can more easily check the submission contents.

exercise/templates/exercise/exercise.html
ihalaij1: I don't think this include is inside any other HTML element, so I left it there.

locale/en/LC_MESSAGES/django.po
ihalaij1: I think the line lengths weren't short enough when those changes were made, so running `makemessages` shortened them automatically now.

exercise/static/exercise/chapter.js
ihalaij1: Yes, I had this idea too, but the problem I was facing was that I wasn't sure how I can pass the variable containing the duplicate submission number to the HTML modal template's `blocktranslate`. How can this be done?

exercise/templates/exercise/exercise_plain.html
ihalaij1: I have now attempted to add the duplicate check to the full-view exercises too.

I couldn't get it to work properly, mainly because I would somehow need the full-view exercise pages to support the `AplusExercise` prototype from `chapter.js`, since I need to be able to check if the exercise is a quiz or active element exercise.

Do you have any ideas how to make this work or some workarounds?

ihalaij1:
I made improvements based on your suggestions and replied to some of the comments.

Mankro (changes requested):
Great, it's getting better!

### Duplicate submission check on the full-view exercise page and using chapter.js

One problem is that the full-view page has slightly different HTML structure than the chapter pages. There are some top-level div elements that do not exist in both views (such as `<div id="exercise-all">`). The attributes `data-aplus-quiz` and `data-aplus-active-element` might not exist in the full view, but I didn't check that now. The navigation selectors in chapter.js might not work in the full view. In the full view, we don't want to use the AJAX form submission that is used in chapter.js. Submitting the form is supposed to redirect to the separate submission feedback page.

We need to be careful in case using chapter.js in the old full-view exercise pages introduces some new surprises or incompatibilities (if chapter.js worked at all there).

I have a feeling chapter.js is not going to work in the full-view exercise page without major changes. You may check this more carefully.

**If setting up duplicate checks in the full-view exercise page becomes too hard or too laborious, then we ignore that in this pull request. Don't spend a ton of time with it!**

Instead of using chapter.js in the full-view exercise page, writing a new similar JS script could be much easier. That's why most of the work is done in reusable functions that can be called anywhere.

### Adding the duplicate submission number to the modal dialog for improved usability

The duplicate submission is identified only when the user tries to submit again. Thus, the submission number can not be known when the html templates are rendered in the Django backend. However, you can modify the html in the browser JavaScript and add the submission number there.

The translated string `DUPLICATE_SUBMISSION` could include some html inside the string (not optimal, but easy to do; the same html is repeated in both the English and Finnish strings). E.g. `<span data-dup-submission=""></span>`.

This could be called in duplicate_check.js - function `modalOrSubmit`.
```
// Set the number of the existing duplicate submission.
$("#duplicate-submission-modal").find("[data-dup-submission]").text("123");
```

### Setting up external libraries in the A-plus git repo - blueimp-md5

blueimp-md5 JS should not be copied to `a-plus/assets_src` as a whole directory. Use `package.json` and `install.sh` that are compatible with our `drone exec` setup (build pipeline).
I wrote more details in a comment in the file `assets_src/blueimp-md5/README.md`.

exercise/templates/exercise/exercise_plain.html
Mankro: I wrote my answer in the review comment today:
https://github.com/apluslms/a-plus/pull/1048#pullrequestreview-1058583782

exercise/static/exercise/chapter.js
Mankro: I answered in the review today:
https://github.com/apluslms/a-plus/pull/1048#pullrequestreview-1058583782

Basically, you can use client-side JS to update the duplicate submission number in the modal contents. The number is not known at the time when the html template is rendered in the Django backend.

exercise/static/exercise/duplicate_check.js
```diff
@@ -0,0 +1,63 @@
+function modalOrSubmit(exercise, hashes, hash, submitCallback) {
+  const isDuplicate = hashes.has(hash);
+  if (isDuplicate) {
+    $("#duplicate-submission-modal-button").click(function() {
```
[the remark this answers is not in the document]
ihalaij1: Yes, you were right, that was possible. I fixed it.

locale/en/LC_MESSAGES/django.po
```diff
@@ -3693,6 +3697,16 @@ msgstr "submitted files"
...
 
+#: exercise/templates/exercise/_duplicate_submission_modal.html
+msgid "DUPLICATE_SUBMISSION_TITLE"
+msgstr "Identical submission"
+
+#: exercise/templates/exercise/_duplicate_submission_modal.html
+msgid "DUPLICATE_SUBMISSION"
+msgstr ""
+"This submission is identical to a previous submission. Are you sure you want "
+"to submit it?"
```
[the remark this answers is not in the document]
ihalaij1: Thanks!

ihalaij1:
The attributes ``data-aplus-quiz`` and ``data-aplus-active-element`` do not exist in the full-view exercise pages.

Adding an event listener to the form submit button on full-view exercise pages might work (since active element exercises don't have those?) And I can differentiate between a quiz and a submit exercise by checking the amount of files included in the submission.

But then should I create a function that redirects to the separate submission feedback page and use that as the ``submitCallback`` function?

Mankro:
> The attributes `data-aplus-quiz` and `data-aplus-active-element` do not exist in the full-view exercise pages.
> 
> Adding an event listener to the form submit button on full-view exercise pages might work (since active element exercises don't have those?) And I can differentiate between a quiz and a submit exercise by checking the amount of files included in the submission.
> 
> Active element exercises also have an `iframe` with `class="acos-iframe"`, which could be used to identify them.
> 
> But then should I create a function that submits the submission and redirects to the separate submission feedback page, and use that as the `submitCallback` function in full-view exercise pages?

I'm not sure what you are thinking of when you say "active elements". A+ active elements have never worked on the full-view exercise pages because they are very dependent on the chapter.js code.

Active elements are also not related to Acos server exercises.
The iframe with `class="acos-iframe"` comes from Acos server.

https://plus.cs.aalto.fi/aplus-manual/master/active_elements/introduction/

Questionnaire exercises could also include file upload fields. They are very rarely used in practice, but for example, the teaching assistant pool questionnaire uses file uploads.

**Nonetheless, we shall just ignore the full-view exercises now!** We can return to them at some point if there is a real demand for it.

Thanks for taking a deeper look at this! :+1:

locale/en/LC_MESSAGES/django.po
```diff
@@ -3693,6 +3698,16 @@ msgstr "submitted files"
...
 
+#: exercise/templates/exercise/_duplicate_submission_modal.html
+msgid "DUPLICATE_SUBMISSION_TITLE"
+msgstr "Identical submission"
+
+#: exercise/templates/exercise/_duplicate_submission_modal.html
+msgid "DUPLICATE_SUBMISSION"
+msgstr ""
+"This submission is identical to the previous submission <span data-dup-"
+"submission=></span>. Are you sure you want to submit it?"
```
Mankro, line 3709: This is not valid HTML. There is `=` without any `"` in the data attribute. The syntactically valid alternatives are the following:
```
<span data-dup-submission></span>
<span data-dup-submission=""></span>
```

locale/fi/LC_MESSAGES/django.po
```diff
@@ -3701,6 +3707,16 @@ msgstr "palautetut tiedostot"
...
 
+#: exercise/templates/exercise/_duplicate_submission_modal.html
+msgid "DUPLICATE_SUBMISSION_TITLE"
+msgstr "Identtinen palautus"
+
+#: exercise/templates/exercise/_duplicate_submission_modal.html
+msgid "DUPLICATE_SUBMISSION"
+msgstr ""
+"Tämä palautus on identtinen aikaisemman palautuksen <span data-dup-"
+"submission=></span> kanssa. Oletko varma, että haluat lähettää palautuksen?"
```
Mankro, line 3718: Note the HTML syntax comment in the English file.

exercise/static/exercise/duplicate_check.js
```diff
@@ -0,0 +1,67 @@
+function openDuplicateModalOrSubmit(exercise, hashes, hash, submitCallback) {
+  const isDuplicate = hashes.includes(hash);
+  if (isDuplicate) {
+    // Set the number of the existing duplicate submission
+    $("#duplicate-submission-modal").find("[data-dup-submission]").text(hashes.reverse().indexOf(hash) + 1);
+    // Unbind previous event handlers and bind a new one
+    $("#duplicate-submission-modal-button").unbind('click');
```
Mankro, line 7: > As of jQuery 3.0, .unbind() has been deprecated. It was superseded by the [.off()](https://api.jquery.com/off/) method since jQuery 1.7, so its use was already discouraged.

https://api.jquery.com/unbind/

Let's use the new API `off` instead of `unbind`.

ihalaij1:
Oh yeah, I mixed up active elements and Acos exercises, probably because they are both interactive and start with the letter A.
Questionnaires with file uploads wouldn't cause issues with the current duplicate check code.

I think the commit messages can be squashed into one, since Muhammad's code is pretty much all gone.
Do you agree?

exercise/static/exercise/duplicate_check.js
```diff
@@ -0,0 +1,67 @@
...
+  const formProps = Object.fromEntries(formData);
+  // Grader language shouldn't affect the hash
+  if (formProps.hasOwnProperty('__grader_lang')) {
+    delete formProps['__grader_lang'];
+  }
+  let hashThis = JSON.stringify(formProps, Object.keys(formProps).sort());
+
+  // File contents are not included in formProps, so we read the files separately
+  const inputFileElements = form_element.querySelectorAll('input[type=file]');
+  if (inputFileElements.length > 0) {
```
Mankro, line 39: If the same form contains both file upload fields and other form fields like text inputs and radio buttons, then does this code compute the md5 hash only from the files or from all inputs? I just want to verify it. I get the impression that all fields are included and that is the goal.

Mankro (changes requested):
Wonderful, this is finished after you fix a couple of tiny issues!

locale/en/LC_MESSAGES/django.po
ihalaij1: I have no idea where the ``"`` characters disappeared. I'll add them.

Mankro:
> Oh yeah, I mixed up active elements and Acos exercises, probably because they are both interactive and start with the letter A. Questionnaires with file uploads don't cause issues with the current duplicate check code.
> 
> I think the commit messages can be squashed into one, since Muhammad's code is pretty much all gone. Do you agree?

Yes, I agree on squashing the commits. If Muhammad's code has been replaced, then we don't need those original commits in the history. However, you should write in the final commit message that this work started from Muhammad's pull request. You can link his old pull request too.

exercise/static/exercise/duplicate_check.js
ihalaij1: Yes, all form fields are added to the variable ``hashThis`` first, and if there are file fields present, the files are read and concatenated with the JSON string of the other fields.

locale/en/LC_MESSAGES/django.po
ihalaij1: Actually, it's probably because the translation file is already using those characters for strings. I'll remove the characters from the HTML.

ihalaij1:
This should be ready for merging!

locale/en/LC_MESSAGES/django.po
Mankro: As far as I remember, django.po supports escaping too: `\"`

Mankro (approved):
Excellent!

PR #1061, opened 2022-08-19
Add hook URL in the webhook instruction when creating course instance

ihalaij1:
# Description

**What?**

Add a mention of the hook URL in the info message when creating a new course instance.
This is because it should also be copied to GitLab webhook settings, along with the webhook secret.


Mankro (approved):
Good catch!

PR #1062, opened 2022-08-22
Option to send email to students about news items

### LINKED ISSUE #377: Feature request: send email when teacher adds news
Hi

It would be very nice to be able to notify students via email when teacher adds news. This should be an option so the teacher could decide on case-by-case basis whether to send the email or not.

What do you think?

Thanks!

PasiSa:
Adds checkbox to Edit news view to request sending email.

Closes #377.

# Description

**What?**

As described above. Tested locally using MailHog as mail server, to internal/external student audiences.

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lib/email_messages.py
+++ b/lib/email_messages.py
@@ -1,9 +1,10 @@
 import logging
 import traceback
 from django.conf import settings
-from django.core.mail import send_mail
-from django.urls import reverse
-from .helpers import build_aplus_url
+from django.core.mail import send_mail, send_mass_mail
+
+from .helpers import Enum, build_aplus_url
+from course.models import CourseInstance
 
 
 logger = logging.getLogger('aplus.lib.email_messages')
@@ -61,3 +62,30 @@ def email_course_error(request, exercise, message, exception=True):
         email_course_instance(instance, subject, body)
     except:
         pass
+
+
+def email_course_students(
+        instance: CourseInstance,
+        subject: str,
+        message: str,
+        audience: Enum = CourseInstance.ENROLLMENT_AUDIENCE.ALL_USERS,
+        ) -> int:
+    """
+    Sends an email to students on the course. Audience parameter controls whether the mail goes
+    to all (default), just internal, or just external students.
+    Returns number of emails sent, or -1 in case of error.
+    """
+    students = instance.students
+    if audience == CourseInstance.ENROLLMENT_AUDIENCE.INTERNAL_USERS:
+        students = students.filter(organization=settings.LOCAL_ORGANIZATION)
+    elif audience == CourseInstance.ENROLLMENT_AUDIENCE.EXTERNAL_USERS:
+        students = students.exclude(organization=settings.LOCAL_ORGANIZATION)
+
+    recipients = students.exclude(user__email='').values_list("user__email", flat=True)
+    emails = tuple(map(lambda x: (subject, message, settings.SERVER_EMAIL, [x]), recipients))
+
+    try:
+        return send_mass_mail(emails)
+    except:
+        logger.exception('Failed to send course instance emails.')
+        return -1
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-08-19 08:42+0300\n"
+"POT-Creation-Date: 2022-09-14 21:50+0300\n"
 "PO-Revision-Date: 2021-05-27 14:47+0300\n"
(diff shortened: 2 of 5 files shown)


PasiSa:
I discussed about this with Juha earlier this week, and he had a preference that email would be triggered using a separate button, instead of checkbox as in current PR. Reviews are welcome, but I could try how the button approach would work out before we merge this.

lib/email_messages.py
```diff
@@ -61,3 +62,28 @@ def email_course_error(request, exercise, message, exception=True):
         email_course_instance(instance, subject, body)
     except:
         pass
+
+
+def email_course_students(
+        instance: CourseInstance,
+        subject: str,
+        message: str,
+        audience: CourseInstance.ENROLLMENT_AUDIENCE=CourseInstance.ENROLLMENT_AUDIENCE.ALL_USERS
```
Mankro, line 71: Is `CourseInstance.ENROLLMENT_AUDIENCE` really accepted as a type here? I thought that the type is our own `Enum` class and `CourseInstance.ENROLLMENT_AUDIENCE` is just one instance of the `Enum`. I haven't tried this myself, so I am curious.

Our own A+ `Enum` is a normal class with no connection to the built-in enumerations in the newer Python versions.

The usual Python style for keyword arguments: 
* no whitespace when there is no static typing
* use whitespace when there static type is set
`audience: CourseInstance.ENROLLMENT_AUDIENCE = CourseInstance.ENROLLMENT_AUDIENCE.ALL_USERS,`

The comma in the end is also good to have when the parameters are already listed one per line.

lib/email_messages.py
```diff
@@ -61,3 +62,28 @@ def email_course_error(request, exercise, message, exception=True):
...
+        ) -> bool:
+    """
+    Sends an email to students on the course. Audience parameter controls whether the mail goes
+    to all (default), just internal, or just external students.
+    """
+    students = instance.students
+    if audience == instance.ENROLLMENT_AUDIENCE.INTERNAL_USERS:
+        students = students.filter(organization=settings.LOCAL_ORGANIZATION)
+    if audience == instance.ENROLLMENT_AUDIENCE.EXTERNAL_USERS:
```
Mankro, line 80:
```suggestion
    elif audience == instance.ENROLLMENT_AUDIENCE.EXTERNAL_USERS:
```
Both if conditions can not be True in this case.

lib/email_messages.py
```diff
@@ -61,3 +62,28 @@ def email_course_error(request, exercise, message, exception=True):
...
+        students = students.exclude(organization=settings.LOCAL_ORGANIZATION)
+
+    recipients = students.exclude(user__email='').values_list("user__email", flat=True)
+
+    try:
+        return send_mail(subject, message, settings.SERVER_EMAIL, recipients, True) == 1
```
Mankro, line 86: Did you notice that Django has a function `send_mass_mail` that is specifically intended for mass mailing, e.g., 1000 students at a time? `send_mass_mail` is slightly more efficient.

With `send_mass_mail`, we can also easily avoid including all students in the "To" header, which would be visible to all students in the course (leaking all student email addresses to all students).

Note that `send_mass_mail` may return larger integers than just `1`.

https://docs.djangoproject.com/en/3.2/topics/email/#send-mass-mail

--------------------

`recipients` 
* in the docs, `recipient_list` parameter: A list of strings, each an email address. Each member of `recipient_list` will see the other recipients in the “To:” field of the email message.
* If there are 1000 students, I assume it is problematic to show 1000 recipients in the "To" field in any email client.
* At any rate, students' emails must not be revealed to all other students in the course. It is personal data and we can't leak it to everyone like this. It is not useful to list the emails to everyone like this. **This is a privacy breach.** Imagine if someone "replies to all" in the email: not very nice.

`fail_silently`
* Did you intend to use `fail_silently=False` (the default)?
* Why would we hide the raised exception? The exception can be used to log more details about the error. It will help the server admins.
* You already have the `except` block, but you specifically silenced the exceptions that could be raised.
* Let's use `fail_silently=False`, which is the default.

https://docs.djangoproject.com/en/3.2/topics/email/#send-mail

Mankro (changes requested):
It is a good start and the structure is what we wanted, but you didn't notice the issues in `send_mail`. There is also a better-suited `send_mass_mail`. We must not include all student email addresses in the email "To" header since all students would see the emails of all other students.

It looks like you copied the `send_mail` part from the old function `email_course_instance`, but you also copied the old bugs with it. Mass emailing all students has a bit different requirements than just emailing the teachers of one course instance.

PasiSa:
Thanks for the comments! I think they are now addressed.

Mankro (approved):
Excellent! I will rebase, force-push and merge this now.

PR #1066, opened 2022-08-25
Submitters aren't lost when saving deviations

### LINKED ISSUE #998: Submitters field cleared on missing input in deadline deviations
If I forget to enter some required input (e.g. the number of days, or the assignment) for a deadline deviation and mistakenly click 'Save', the field of submitters are cleared. It would be nice if the contents of this field could be kept on reload just as it is for the module/assignment fields. (Or if 'Save' was greyed out until all required input has been entered if submitters can't be held in state.)

EerikSaksi:
The list of submitters are cached in localStorage whenever the
save button is clicked. On page load, the localStorage is loaded,
and the list of submitters are added to the form.

Fixes #998

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

I tested different situations, such as refreshing the page, making sure a selected submitter was present after a save, and that no submitter was present after a page. I also made sure that users could still be deleted and wouldn't appear after a save.

- [ ] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.
I added comments as this was quite a small change
# Is it Done?

### DIFF
--- a/deviations/templates/deviations/add_dl.html
+++ b/deviations/templates/deviations/add_dl.html
@@ -13,6 +13,7 @@
 {% endblock %}
 
 {% block columns %}
+
 <div class="col-md-12">
 	<form method="post" class="well form-horizontal">
 		{% csrf_token %}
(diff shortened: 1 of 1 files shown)


ihalaij1 (changes requested):

ihalaij1 (approved):
Nice!

Mankro (changes requested):
Thanks! I don't really like this hackish JavaScript solution. It should be possible to implement this in the backend that validates the forms. The JavaScript solution attached to the deadline deviation form page does not extend to other forms that have the same issue with `UsersSearchSelectField`. Additionally, this solution breaks the click event handlers in the buttons representing the selected users.

The commit message could be made a bit clearer. It mentions "saving deviations", but the problem is specific to not passing the validation (required field missing), not just saving generally.

I wrote down ideas about the backend solution in the issue:
https://github.com/apluslms/a-plus/issues/998#issuecomment-1294953902

Mankro:
I tried to modify the `clean()` method of class `SearchSelectField` so that it updates the custom `initial_queryset` attribute. It worked so well that I have already finished the issue. Since this pull request tried to use a different approach via JavaScript and it wasn't optimal, I closed this pull request. It will not be merged.

https://github.com/apluslms/a-plus/commit/b16e75e300bb36b2323297770375c88e309cba24

PR #1067, opened 2022-08-31
_course_cards.html doesn't hide label when condensed

### LINKED ISSUE #727: Course archive page does not show "hidden from students" labels
The course archive page does not show the label "hidden from students" on a course card if the course instance has the setting `visible_to_students = False`. Those labels are useful if the user (like a teacher) can see hidden courses in the list.

The course archive page can be accessed from the A+ front page. It has the URL `/archive`.

EerikSaksi:
Before, when the _course_cards.html was created by a parent and condensed was set to true, the hidden label would not be visible. Now condensed cards also show the label. This was done through removing the "and not condensed" condition from the card template

# Description

**What?**

The archive now shows the red "Hidden from students" label on course cards

**Why?**

It is useful for those with the permissions to see hidden courses (such as admins) to know that their courses are hidden

**How?**

Before, condensed cards would hide the label. I removed the condensed check (and not condensed) to prevent this from happening.

Fixes #727 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I added a course instance which was hidden, and ensured that this had the "Hidden From Students" label on it, and courses which weren't hidden didn't have this label (from the root account). I also ensured that this change did not make hidden instances visible to students.

- [ ] Chrome
- [X ] Firefox

# Programming style

# Is it Done?

### DIFF
--- a/course/templates/course/_course_cards.html
+++ b/course/templates/course/_course_cards.html
@@ -26,7 +26,7 @@
 					</p>
 					<p class="card-text">
 						{{ instance.course.code }}
-						{% if not instance.visible_to_students and not condensed %}
+						{% if not instance.visible_to_students %}
    							<span class="label label-danger">{% translate "HIDDEN_FROM_STUDENTS" %}</span>
 						{% endif %}
 						<br />


ihalaij1:
The commit message should also include ``Fixes #727`` before the longer explanation text.

Mankro (approved):
Excellent! Before merging, I will modify the commit message so that its heading describes the original problem being resolved more clearly. The current commit message focuses too much on how the code was changed, but it should start from the higher-level goal of the change.

PR #1068, opened 2022-09-01
The redirect now considers whether there are public modules

### LINKED ISSUE #919: Fix the current course instance redirection when the newest instance is not open yet
The current course instance redirection (URL contains course without any course instance identifier): https://plus.cs.aalto.fi/o1

https://github.com/apluslms/a-plus/blob/6a7704d22469cb0e2c0f1f52ba420cb3a4cf3c7c/course/views.py#L110

From @jsorva:

In a meeting some time ago, I promised to return to the issue of the 
“current instance link” that points to the most recent version of a 
course. So, here’s how I’d prefer like it to work:

1) While a course is running, the link points to the active instance. 
[For O1, this is in during the fall term.]

2) After a course is over [in December], the link continues to points to 
the most recent instance. This continues even after next year’s instance 
has been created in A+ but isn’t yet open to the public [typically early 
in the next calendar year]. And it even continues after the next 
instance is technically visible in some form but there’s only the front 
page and maybe some other materials, but the actual course contents 
aren’t visible yet [this is often the case during the summer].

3) Only when the actual course contents are visible [ideally some time 
in August, but might be September] does the link start pointing at the 
new instance.

To summarize, I’d like the link to always point to the newest instance 
whose actual contents are actually visible; the mere existence of the 
instance isn’t enough. (I’m not sure how close this wish is to the 
current behavior.)

Internal ticket: https://rt.cs.aalto.fi/Ticket/Display.html?id=20168

-----------------------------------

My first hunch is that this should be easy to accomplish by tweaking the latest course instance selection in the LastInstanceView.

EerikSaksi:
By joining the instance with the module, filtering all modules which are ready, and checking if any exist, we can check if there are any public modules. This means that we select the most recent instance with public modules.

# Description

When entering /course/ the user should be redirected to an instance /course/instance/. Before, this redirect would just point to the most recent instance. This pull request amends this by redirecting the user to the most recent instance with public modules. 

Why?

To prevent redirecting users to an instance created for next year/semester.

How?

We check if there are any modules which are public for course instance for it to be a viable candidate.

Fixes #919 

# Testing

- [X] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

In addition to the unit tests I ensured that /def/ behaved as it did.

- [X] Chrome
- [ ] Firefox

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.
Added comments to added code.

# Is it Done?


ihalaij1:
The commit message should also include ``Fixes #919`` after the longer explanation text.

ihalaij1 (approved):
Looks good to me!

PasiSa:
The commit message is still missing "Fixes #919"

EerikSaksi:
Should have the correct Fixes message now @PasiSa

Mankro (changes requested):
Good start and great to see some unit tests too! This does not yet fix the given issue. See the comments below. I already started to fix this myself so you don't need to.

EerikSaksi:
Thanks for the feedback @Mankro

Mankro (approved):
I have made changes myself and I will merge this pull request now.

PR #1069, opened 2022-09-01
Fix duplicate submission check not hashing all duplicate form fields

ihalaij1:
# Description

**What?**

Duplicate submission modal was sometimes popping up when submitting non-duplicate submissions.
This was because the submitted forms may include duplicate field names, and `duplicate_check.js` only hashed the last one of the fields with a duplicate name.

**Why?**

This bug was found by a student and pointed out by a course teacher.

**How?**

`duplicate_check.js` now uses `serializeArray()` instead of `Object.fromEntries()` when extracting the form fields. `serializeArray()` does not get rid of the duplicate fields, therefore fixing the bug.

### DIFF
--- a/exercise/static/exercise/duplicate_check.js
+++ b/exercise/static/exercise/duplicate_check.js
@@ -26,15 +26,12 @@ function duplicateCheck(exercise, form_element, submitCallback) {
   }
 
   // Compute a hash for the current submission
-  const formData = new FormData(form_element);
-  const formProps = Object.fromEntries(formData);
+  const formAsArray = $(form_element).serializeArray();
   // Grader language shouldn't affect the hash
-  if (formProps.hasOwnProperty('__grader_lang')) {
-    delete formProps['__grader_lang'];
-  }
-  let hashThis = JSON.stringify(formProps, Object.keys(formProps).sort());
+  const formAsArray2 = formAsArray.filter(function(f) { return f.name !== '__grader_lang'; });
+  let hashThis = JSON.stringify(formAsArray2);
 
-  // File contents are not included in formProps, so we read the files separately
+  // File contents are not included in the JSON string, so we read the files separately
   const inputFileElements = form_element.querySelectorAll('input[type=file]');
   if (inputFileElements.length > 0) {
     const reader = new FileReader();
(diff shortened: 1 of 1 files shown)


exercise/static/exercise/duplicate_check.js
```diff
@@ -26,15 +26,14 @@ function duplicateCheck(exercise, form_element, submitCallback) {
   }
 
   // Compute a hash for the current submission
-  const formData = new FormData(form_element);
-  const formProps = Object.fromEntries(formData);
+  const formAsArray = $(form_element).serializeArray()
   // Grader language shouldn't affect the hash
-  if (formProps.hasOwnProperty('__grader_lang')) {
-    delete formProps['__grader_lang'];
+  if (formAsArray.at(-1).name === '__grader_lang') {
```
raphendyr, line 31: Hi, ex-dev here. A friend linked me the PR and I got inteterest of the current stuff, thus I commenting.

Have you considered using a filter here? Like,

```
const formAsArray2 = formAsArray.filter(function(f) { return f.name !== '__grader_lang'; })
```

Of course it's unlikely that the field is at any other position than the last, but this would protect a bit against regression bugs, if Grader is edited without updating this part of the code. Which is a benecifial pattern overall, allowing different microservices to be updated/deployed at a different time (or with a tiered update). Of course, not a priority to my knowledge.

Few other things consider (although, these can be stupid as I have no idea about the overall design of this feature):
- Maybe all hidden fields could be filtered out? (doesn't work if the grader adds e.g. seeds for randomisation).
- Maybe all fields starting with `__` (two underscores) could be filtered out? (pros: grader can add similar meta fields without needing a change in A+; cons: grader doesn't know that this filtering is done and could add field, which shouldn't be filtered out).

ihalaij1: Thank you Jaakko for the ideas!

I edited the code to use the filter like you suggested, because it turned out that the `__grader_lang` field wasn't always the last field, especially in randomized questionnaires.

We decided not to filter out all hidden fields, because randomization and some custom exercises depend on them.

Filtering out all fields starting with `__` might break duplicate check in custom exercises, if the programmer decided to use double underscores for their own hidden fields.

Mankro:
The duplicate submission check was a new feature in #1048. (Linking this bug to the related PR)

Mankro (approved):
Thanks, this looks great!

PR #1071, opened 2022-09-02
Clicking the label in a SearchSelectField focuses input

### LINKED ISSUE #934: Search select form input do not connect the visible label to the search input
Search select form inputs (like add assistants in the edit course form) do not connect the visible label to the visible search input. The label `<label for="id_assistants">` points to the hidden `<select id="id_assistants">` element, but for accessibility and usability, the label should connect to the visible search text input with the id. For example, when you click the label, it should focus the text input. This already works with the open and close time fields in the edit course form,

_Originally posted by @Mankro in https://github.com/apluslms/a-plus/pull/926#pullrequestreview-828816969_

EerikSaksi:
# Description

**What?**

This PR edits the ajax_search_select.js by adding the id the label refers to (<label for="id".../>) to the input of the field (<input id="id".../>) 

**Why?**

Before, when the label of a SearchSelectField was clicked, the input wouldn't be focused, although other types of fields had this behavior, which is bad user experience.

**How?**

We access the hidden select's id through the inner_element attribute. We then find the child input of the element and set it's id to the hidden select's id.

Fixes #934 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Double clicking the label and ensuring that this focused the input, analyzing the DOM and ensuring that only the input was changed and that the input had a new id and nothing else

- [ ] Chrome
- [X] Firefox

# Style 

# Is it Done?

### DIFF
--- a/assets/js/ajax_search_select.js
+++ b/assets/js/ajax_search_select.js
@@ -140,6 +140,9 @@ $(function() {
         self.paste($(this).data("field"));
       });
 
+      //access the inner id of the hidden select, and set it as the input's id, so that
+      //the label refers to the input and clicking the label focuses the input
+      this.element.find('input')[0].id = this.inner_element[0].id
       this.inner_element.removeAttr('required');
     },
 


ihalaij1 (changes requested):
The commit message should be reworded to be more concise and should also include ``Fixes #934``.

ihalaij1 (approved):
Looks good to me now!

Mankro (changes requested):
I thought about this and I decided that it is better to implement this in the backend. You switched the ids in the frontend JS in this pull request. I finished the new implementation myself and thus, I will close this pull request.

PR #1077, opened 2022-09-09
Added support for MathJax 3.x

### LINKED ISSUE #871: Add support for MathJax.js v3
A+ JavaScript code has three places where the MathJax API is called in order to typeset mathematics in dynamically updated content. The MathJax API has changed in v3, thus A+ JS code must be updated. We must support both v2 and v3.

Two functions in chapter.js and one in aplus.js use MathJax:

* https://github.com/apluslms/a-plus/blob/73a3e667ccab1ed958314e381149005b9e92f410/exercise/static/exercise/chapter.js#L110
* https://github.com/apluslms/a-plus/blob/73a3e667ccab1ed958314e381149005b9e92f410/exercise/static/exercise/chapter.js#L841
* https://github.com/apluslms/a-plus/blob/73a3e667ccab1ed958314e381149005b9e92f410/assets/js/aplus.js#L413

A+ source code does not include the MathJax.js library. Some courses include it in their materials. A+ code calls the MathJax API so that maths work in dynamically updated content too, such as the submission feedback modal dialog (its content is downloaded with AJAX).

MathJax 3:
* https://docs.mathjax.org/en/latest/web/typeset.html#typeset-async
* when typeset content is removed, we should call MathJax.typesetClear()  
  https://docs.mathjax.org/en/latest/web/typeset.html#updating-previously-typeset-content
* new API uses  ES6 promises, `MathJax.typesetPromise()`. The API says that promises should be chained if multiple typeset calls are done at the same time, but we could maybe assume that only one is run at a time. Only one submission feedback dialog can be opened at a time or only one exercise can be submitted at a time.

It might be cleaner to make a new jQuery plugin for this. Something like this perhaps:

```
(function(window, document, $) {
  $.fn.aplusTypesetMath = function(options) {
    if (typeof window.MathJax === "undefined") return;
    return this.each(function() {
      if (window.MathJax.VERSION == 2) {
        window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, this]);
      } else {
        // typeset in MathJax v3 API
      }
    });
  };
})(window, document, jQuery);
```

EerikSaksi:
This commit does not break the support for MathJax 2.x. Rather than unconditionally using the 2.x API (Hub.Queue([])) we instead check the version of MathJax and either use the old Hub.Queue implementation or the new promise based implementation of 3.x (MathJax.typesetPromise()). This conditional typesetting is defined once as a jQuery plugin in assets/js/aplus.js. This plugin expects the HTML element which should be typeset.

Fixes #871

# Description

**What?**

Added support for MathJax 3.x without breaking compatibility 2.x.

**Why?**
According to the MathJax documentation, 2.x will only be maintained and there will be no further development. 

"MathJax v2 will continue to be maintained as we work to move more features into version 3, but MathJax v2 likely will not see much further development, just maintenance, once MathJax v3 is fully converted."
https://docs.mathjax.org/en/v3.2-latest/upgrading/v2.html

**How?**

Added a jQuery plugin in assets/js/aplus.js which typesets conditionally based on the version of MathJax which was loaded in the html of the build.

Fixes #871 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I edited an HTML file which used MathJax in aplus-manual (_build/programming_exercises/hello_world.html) and changed the version of MathJax to 3.2.2 by editing the src attribute of the script tag which loads MathJax. This allowed me to test whether a 3.x version of MathJax worked. I also changed the script tag back to 2.7.7 to ensure that 2.x version of MathJax still work with the new changes.

- [ ] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -503,7 +503,7 @@ $(function() {
                         modal.trigger("opened.aplus.modal");
                         // render math in the modal
                         if (typeof window.MathJax !== "undefined") {
-                            window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, modal.get(0)]);
+                          modal.aplusTypesetMath();
                         }
                     }
                 }).fail(function() {
@@ -520,6 +520,30 @@ $(function() {
             }
         });
     };
+})(jQuery, window);
+
+
+(function($, window) {
+  "use strict";
+
+  const pluginName = "aplusTypesetMath";
+  $.fn[pluginName] = function() {
+    if (typeof window.MathJax === "undefined") {
+      return;
+    }
+    switch (window.MathJax.version[0]) {
+      case "2":
+        window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, this.get()]);
+        break;
+      case "3":
+        window.MathJax.typesetPromise(this.get()).catch((err) =>
+          console.error(err)
+        );
+        break;
+      default:
+        console.error("Unsupported version of MathJax, use 2.x or 3.x!");
+    }
+  };
 })(jQuery, window);
 
 /**
--- a/exercise/static/exercise/chapter.js
+++ b/exercise/static/exercise/chapter.js
@@ -104,10 +104,7 @@
     },
 
     renderMath: function() {
-      if (typeof window.MathJax === "undefined") {
-        return;
-      }
-      window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, this.modalElement.get(0)]);
+      this.modalElement.aplusTypesetMath();
     },
   });
 
@@ -859,10 +856,7 @@
     },
 
     renderMath: function() {
-      if (typeof window.MathJax === "undefined") {
-        return;
-      }
-      window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, this.element.get(0)]);
+      this.element.aplusTypesetMath();
     },
   });
 


exercise/static/exercise/chapter.js
```diff
@@ -104,10 +104,7 @@
     },
 
     renderMath: function() {
-      if (typeof window.MathJax === "undefined") {
-        return;
-      }
-      window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, this.modalElement.get(0)]);
+      $.fn.aplusTypesetMath(this.modalElement.get(0))
```
Mankro, line 107:
```suggestion
      this.modalElement.aplusTypesetMath();
```
After you have fixed the plugin definition, I assume this line would be changed to this.

exercise/static/exercise/chapter.js
```diff
@@ -859,10 +856,7 @@
     },
 
     renderMath: function() {
-      if (typeof window.MathJax === "undefined") {
-        return;
-      }
-      window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, this.element.get(0)]);
+      $.fn.aplusTypesetMath(this.element.get(0))
```
Mankro, line 859:
```suggestion
      this.element.aplusTypesetMath();
```
Something like that.

assets/js/aplus.js
```diff
@@ -503,7 +503,7 @@ $(function() {
                         modal.trigger("opened.aplus.modal");
                         // render math in the modal
                         if (typeof window.MathJax !== "undefined") {
-                            window.MathJax.Hub.Queue(["Typeset", window.MathJax.Hub, modal.get(0)]);
+                            $.fn.aplusTypesetMath(modal.get(0));
```
Mankro, line 506:
```suggestion
                            modal.aplusTypesetMath();
```
After changing the plugin definition, this line would be something like that.

Mankro (changes requested):
Nice, this looks mostly good, but I think you have misunderstood how to define and use jQuery plugins. This tutorial is useful:
https://learn.jquery.com/plugins/basic-plugin-creation/

About testing:
you said you have manually modified the MathJax.js URL in the built HTML file. You could also set the URL in the script element in the file `_templates/layout.html` in the aplus-manual repo and compile the course again.
It is also possible to set the URL in the A+ course settings, the head URLs field. It can be changed manually in the A+ Edit course page, Course tab. If you set "head URLs", you should remove the `data-aplus` attribute from the script element in `_templates/layout.html` so that the MathJax.js library is not included twice in the same HTML document.

Did you test submission feedback that includes math notations? The feedback modal dialogue in the chapter pages (in the My submissions dropdown on an assignment frame) can change its contents multiple times when you open different submissions one after another in the modal. Is there any problem there? The MathJax documentation suggests that the removal of math from the html document should be announced to MathJax.js too by calling `MathJax.typesetClear()`.

https://docs.mathjax.org/en/latest/web/typeset.html#updating-previously-typeset-content

Mankro (approved):
Great! I had some concerns about the MathJax 3 promise API, but the implementation works anyway at the moment at least in small-scale testing. I made minor changes myself corresponding to the comments below.

PR #1080, opened 2022-09-14
Added enrollment question check to CourseModulePermissionBase

### LINKED ISSUE #679: Enrollment questionnaire is affected by the module opening time
Continued from https://github.com/apluslms/a-plus/issues/462

Enrollment questionnaires should not be affected by the module opening time. The course already has its own enrollment opening/closing time settings and they are enough. Enrollment questionnaires should not be affected by the module opening/closing times. Fixing this is difficult because the `CourseModulePermission` does not directly have access to the exercise and it can't check if the exercise is an enrollment exercise. Perhaps the exercise could be defined in the attributes of the `view` object and it would always be defined if the user is on an exercise page. If the exercise is defined and it is an enrollment exercise, then the course module opening time is ignored in the permission.

https://github.com/apluslms/a-plus/blob/cd10e04963557444efeb11b06b2834bf47ace903/course/permissions.py#L140

EerikSaksi:
Before, the permission check would work by simply checking whether the module had opened yet. With this check, the module opening time is only checked if the exercise is not an enrollment question

Fixes #679

# Description

**What?**

This change adds a check for if the exercise is an enrolment question to CourseModulePermissionBase.

**Why?**

Enrolment questions should be accessible before the module is open

**How?**

Before the check for if the module is open, we check whether the exercise is an enrolment question. If this is the case, then we do not care if the module is open.

Fixes #679 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Ensuring that students can open the URL to enrolment exercises in modules which have not opened yet, making sure that no error is thrown when the view supplied to CourseModulePermissionBase does not have a exercise attribute (when called from ModuleView and not ExerciseView)

- [ ] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/permissions.py
+++ b/course/permissions.py
@@ -15,7 +15,7 @@ from authorization.permissions import (
     FilterBackend,
 )
 from exercise.cache.points import CachedPoints
-from exercise.models import BaseExercise
+from exercise.models import BaseExercise, LearningObject
 from exercise.submission_models import Submission
 from userprofile.models import GraderUser, UserProfile
 from .models import (
@@ -247,6 +247,13 @@ class CourseModulePermissionBase(MessageMixin, Permission):
 
         if module.status == CourseModule.STATUS.HIDDEN:
             return False
+
+        # Enrollment questionnaires are not affected by the module opening time.
+        if hasattr(view, 'exercise') and view.exercise.status in (
+                LearningObject.STATUS.ENROLLMENT,
+                LearningObject.STATUS.ENROLLMENT_EXTERNAL,
+                ):
+            return True
 
         if not module.is_after_open():
             # FIXME: use format from django settings
--- a/exercise/permissions.py
+++ b/exercise/permissions.py
@@ -39,7 +39,13 @@ class ExerciseVisiblePermissionBase(ObjectVisibleBasePermission):
             self.error_msg(_('EXERCISE_VISIBILITY_ERROR_NOT_VISIBLE'))
             return False
 
-        if exercise.is_submittable and not exercise.course_module.have_exercises_been_opened():
+        if (
+            exercise.is_submittable
+            and not exercise.course_module.have_exercises_been_opened()
+            and not exercise.status in (
+                LearningObject.STATUS.ENROLLMENT,
+                LearningObject.STATUS.ENROLLMENT_EXTERNAL,
+                )):
             self.error_msg(
                 format_lazy(
                     _('EXERCISE_VISIBILITY_ERROR_NOT_OPEN_YET -- {}'),
--- a/exercise/tests.py
+++ b/exercise/tests.py
@@ -1184,3 +1184,80 @@ class ExerciseTest(TestCase):
         self.assertEqual(draft1.submission_data, [["key", "value4"]])
         draft2 = self.base_exercise.get_submission_draft(self.user2.userprofile)
         self.assertEqual(draft2.submission_data, [["key", "value2"]])
+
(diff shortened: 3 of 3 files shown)


course/permissions.py
```diff
@@ -248,7 +248,8 @@ def has_object_permission(self, request: HttpRequest, view: Any, module: CourseM
         if module.status == CourseModule.STATUS.HIDDEN:
             return False
 
-        if not module.is_after_open():
+        #if this isn't an enrollment question and the module hasn't opened yet
+        if not (hasattr(view, 'exercise') and view.exercise.status in ('enrollment', 'enrollment_ext')) and not module.is_after_open():
```
PasiSa, line 252: Should we rather use
```
view.exercise.status in (LearningObject.STATUS.ENROLLMENT, LearningObject.STATUS.ENROLLMENT_EXTERNAL)
```
Perhaps could also be split into multiple lines, as this line is getting rather long

PasiSa (changes requested):
One style-related proposal, otherwise looks good to me. Didn't test this PR (or my proposal), though.

EerikSaksi:
Changed the formatting and added LearningObject import @PasiSa

PasiSa (approved):

Mankro (changes requested):
Good start! When I tested this, I noticed that it really does not work yet because there is also another permission (`ExerciseVisiblePermissionBase`) that checks the module opening time. Fixing that other permission is easy. I also thought that some details could be formatted better. `CourseModulePermissionBase.has_object_permission()` has the part `module.requirements.exists()` in the end that should not be used in enrollment questionnaires.

I made some changes myself and pushed them to git so that you may finalize this PR.

**git commit message**
* commit message heading: use the imperative mood
* line length: at most 80 characters
* the heading could be reworded better. Emphasize the goal ("enrollment questionnaires should not be affected by the module opening time") instead of the implementation details on the code lines ("check in CourseModulePermissionBase").

EerikSaksi:
Thanks @Mankro for the feedback, I will get to it!

EerikSaksi:
I tested the new changes, it seems like the enrolment questionnaire is accessible before the course is open, but the course is not. I am not sure if this is the desired behaviour.

https://user-images.githubusercontent.com/46625399/191976229-87631e78-d401-4e5f-bb4a-ef38aaccd975.mp4

Mankro:
> > it seems like the enrolment questionnaire is accessible before the course is open, but the course is not. I am not sure if this is the desired behaviour.
> 
> I haven’t studied this issue / PR properly at all, but I’ll just mention that AFAIC, it shouldn’t be possible to fill in or even see the enrollment questionnaire before the course is actually open.

If I remember correctly, the enrollment questionnaire is supposed to be open if the course settings (enrollment starting time and enrollment ending time) define so. If the teacher wants to set enrollment starting time before the course opening time, then it is possible.

If the enrollment opening times are not defined, they default to the course opening times.

If the enrollments are fetched from Sisu, we should check what happens. Is access restricted by the course opening times?

EerikSaksi:
I added a test for accessing the enrollment exercise which tests combinations of starting times and enrollment starting times which passes. I might have misunderstood what the expected behavior is but if the test describes the expected behavior then this should work.

Mankro (approved):
Good job with the unit tests! I added some more tests, rebased and now I am merging this.

PR #1081, opened 2022-09-19
Anchor links work even when page height changes

### LINKED ISSUE #1028: Chapter exercise links using anchors #chapter-exercise-1 do not land on the correct scroll position because the exercise frame heights change after the page load
Using a toc-link to navigate to an exercise uses an anchor to the exercise element id, which seems to create a "data race" between the automatic scrolling and exercise views opening to full height. The web page does not end up on the correct point (at the exercise frame) since the exercises move around after they have settled on the correct height for their frames.

Solution 1:
record exercise height after loading and send it to server. Server adds element height to the html based on average of above information. Possible to fool the server with false data? 

Solution 2: 
JS updates location after loading exercise to update the scrolling position. If user creates scroll event, above logic is disabled.

----------------------------

I created this issue from the previous combined issue list (#438). The old description here is rather messy. Solution 1 sounds very strange. Solution 2 makes more sense. The view port could be set to the correct scroll position after the exercise frames have set their heights.

EerikSaksi:
By using an ajaxComplete exercises maintain position on the exercise even as content loads. The scroll listener allows the user to scroll freely and prevent the repeated focusing on the exercise as content loads.

Fixes #1028

# Description

**What?**

This change ensures that when the user clicks on a anchor link referencing an exercise (such as #chapter-exercise-1) the focus remains on the referenced exercise even when content loads and the page height changes.

*No user scrolling*
no_scroll

*With user scrolling (focusing on the exercise disables as soon as scrolling is detected)*
scroll

**Why?**

Before, when a user clicked on an anchor link and content above loaded, the page would scroll and the exercise which was referenced by the link would move down either off screen or out of focus of the user which can cause confusion.

**How?**

As exercises load, the ajaxComplete listener is triggered. Whenever ajaxComplete triggers,  we set the users scroll position to the new position of the referenced exercise. We also created a scroll listener, which prevents repeated focusing on the referenced exercise if the user scrolls manually. We detect manual scrolling if there are 3 consecutive scroll events with no ajaxComplete events in between, as exercises loading triggers the scroll event.

Fixes #1028 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I tested the O1 course locally, which had issues with focusing on exercises. I ensured that on both Firefox and Chrome that with no scrolling, clicking on a link with a chapter exercise anchor reference resulted in the browser constantly focusing the correct referenced exercise. I also ensured that the moment the user tries to scroll that the browser stops trying to maintain focus on the exercise. 

- [X] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/templates/exercise/exercise.html
+++ b/exercise/templates/exercise/exercise.html
@@ -11,13 +11,13 @@
 
 {% block js-translations %}
 <link
-    data-translation
-    rel="preload"
-    as="fetch"
-    crossorigin="anonymous"
-    type="application/json"
-    hreflang="fi"
-    href="{{ STATIC_URL }}js-translations/exercise.fi.json"
+	data-translation
+	rel="preload"
+	as="fetch"
+	crossorigin="anonymous"
+	type="application/json"
+	hreflang="fi"
+	href="{{ STATIC_URL }}js-translations/exercise.fi.json"
 >
 {% endblock %}
 
@@ -28,32 +28,32 @@
 {% endif %}
 
 <div
-    id="exercise-page-content"
-    {% if exercise.is_submittable %}
-        data-aplus-chapter="{{ exercise|url }}"
-        data-aplus-group="{{ exercise.min_group_size }}-{{ exercise.max_group_size }}"
-        {% if summary.get_submission_count > 0 %}
-            data-aplus-group-fixed="{{ summary.get_group_id }}"
-        {% endif %}
-    {% endif %}
+	id="exercise-page-content"
+	{% if exercise.is_submittable %}
+		data-aplus-chapter="{{ exercise|url }}"
+		data-aplus-group="{{ exercise.min_group_size }}-{{ exercise.max_group_size }}"
+		{% if summary.get_submission_count > 0 %}
+			data-aplus-group-fixed="{{ summary.get_group_id }}"
+		{% endif %}
+	{% endif %}
 >
-  {% if latest_enrollment_submission_data %}
-    {{ latest_enrollment_submission_data|json_script:"enroll-submission-data" }}
-  {% endif %}
-  <div class="overlay-parent">
-    {% if issues and not submission_allowed %}
-        {% include 'exercise/_warnings_overlay.html' %}
-    {% endif %}
-    <div {% if disable_submit %}data-aplus-disable-submit="true"{% endif %}>
-      {{ page.content|safe }}
-    </div>
-  </div>
+	{% if latest_enrollment_submission_data %}
+		{{ latest_enrollment_submission_data|json_script:"enroll-submission-data" }}
+	{% endif %}
+	<div class="overlay-parent">
+		{% if issues and not submission_allowed %}
+			{% include 'exercise/_warnings_overlay.html' %}
+		{% endif %}
+		<div {% if disable_submit %}data-aplus-disable-submit="true"{% endif %}>
+			{{ page.content|safe }}
+		</div>
+	</div>
 </div>
 
 <div id="submit-progress" class="hide progress">
-  <div class="progress-bar progress-bar-striped active" role="progressbar" style="width:100%;">
-    {% translate "POSTING_SUBMISSION" %}
-  </div>
+	<div class="progress-bar progress-bar-striped active" role="progressbar" style="width:100%;">
+		{% translate "POSTING_SUBMISSION" %}
+	</div>
 </div>
 
 {% include "exercise/_exercise_wait.html" %}
@@ -61,10 +61,10 @@
 {% include "exercise/_duplicate_submission_modal.html" %}
 
 <div id="loading-indicator" class="hide progress"
-  data-msg-load="{% translate 'LOADING_EXERCISE' %}"
-  data-msg-submit="{% translate 'POSTING_SUBMISSION' %}"
-  data-msg-error="{% translate 'EXERCISE_ERROR_COMMUNICATION' %}">
-  <div class="progress-bar progress-bar-striped active" role="progressbar" style="width:100%;"></div>
+	data-msg-load="{% translate 'LOADING_EXERCISE' %}"
+	data-msg-submit="{% translate 'POSTING_SUBMISSION' %}"
+	data-msg-error="{% translate 'EXERCISE_ERROR_COMMUNICATION' %}">
+	<div class="progress-bar progress-bar-striped active" role="progressbar" style="width:100%;"></div>
 </div>
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


ihalaij1 (changes requested):
Good work! This scrolling issue with anchor links has been very annoying.

The whole file ``exercise.html`` should be updated to use tabs instead of spaces for indentation (in a separate commit from your other changes).

Also, ``const`` and ``let`` should be used with variables instead of ``var`` when ``var`` is not required.

Could you add the missing semicolons at the end of the statements (like in the script above), so that the coding style is consistent?

exercise/templates/exercise/exercise.html
```diff
@@ -114,16 +114,16 @@
...
 	window.addEventListener('scroll', () => {
-		consecutiveScrollEvents += 1
+		consecutiveScrollEvents += 1;
 	})
 
 	$( document ).ajaxComplete(() => {
 		if (consecutiveScrollEvents < 3) {
-			location.hash = hash
-			consecutiveScrollEvents = 0
+			location.hash = hash:
```
ihalaij1, line 125: Nice, just change this one ``:`` to ``;`` and its good.

ihalaij1 (changes requested):
One more typo to fix, see my comment.

ihalaij1 (changes requested):
I just noticed that the indentation is very inconsistent in this file.
Sometimes the code is indented multiple times when it really shouldn't be.

Could you go through the file and fix the issues?
It looks very bad when the code is indented unnecessarily deep.

ihalaij1 (approved):
Looks good to me!

exercise/templates/exercise/exercise.html
```diff
@@ -79,28 +79,54 @@
...
+			url = $stats.data("url");
+		$stats.load(url, function() {
+			$stats.find('[data-toggle="tooltip"]').tooltip();
+		});
+	}
 });
 </script>
+
+<script>
+var hash = window.location.hash
```
Mankro, line 107:
```suggestion
var hash = window.location.hash;
```

exercise/templates/exercise/exercise.html
```diff
@@ -79,28 +79,54 @@
...
+	//it isn't enough to just register a scroll event which disables setting the scroll
+	//location to the hash id. The scroll event fires every time the page loads, and
+	//whenever an exercise loads (as this changes the height of the page). To identify
+	//if a user manually scrolls we should receive 3 consecutive scroll events with no
+	//ajaxComplete events in between. An ajaxComplate event occuring signifies that the
+	//page scroll was caused by an exercise loading, and not the user.
+	var consecutiveScrollEvents = 0
+
+	window.addEventListener('scroll', () => {
+		consecutiveScrollEvents += 1
```
Mankro, line 120:
```suggestion
		consecutiveScrollEvents += 1;
```

exercise/templates/exercise/exercise.html
```diff
@@ -79,28 +79,54 @@
...
+	})
+
+	$( document ).ajaxComplete(() => {
+		if (consecutiveScrollEvents < 3) {
+			location.hash = hash
+			consecutiveScrollEvents = 0
```
Mankro, line 126:
```suggestion
			location.hash = hash;
			consecutiveScrollEvents = 0;
```

exercise/templates/exercise/exercise.html
```diff
@@ -79,28 +79,54 @@
...
+		}
+	})
```
Mankro, line 128:
```suggestion
	});
```

exercise/templates/exercise/exercise.html
```diff
@@ -79,28 +79,54 @@
...
+	//location to the hash id. The scroll event fires every time the page loads, and
+	//whenever an exercise loads (as this changes the height of the page). To identify
+	//if a user manually scrolls we should receive 3 consecutive scroll events with no
+	//ajaxComplete events in between. An ajaxComplate event occuring signifies that the
+	//page scroll was caused by an exercise loading, and not the user.
+	var consecutiveScrollEvents = 0
+
+	window.addEventListener('scroll', () => {
+		consecutiveScrollEvents += 1
+	})
```
Mankro, line 121:
```suggestion
	});
```

exercise/templates/exercise/exercise.html
```diff
@@ -79,28 +79,54 @@
...
+
+	$( document ).ajaxComplete(() => {
```
Mankro, line 123: Did you consider adding an event handler like this to chapter.js? In chapter.js, you could identify which exercise is targeted by the `window.location.hash` and only react when that exercise is loaded. Though, when the exercises listed before the targeted exercise are loaded, that already changes the window's scroll position. Chapter.js loads the exercises in order from the first to the last.

exercise/templates/exercise/exercise.html
Mankro, line 128: Could the `ajaxComplete` handler recognize that all exercises have been loaded and then remove the scroll event handler from the window? It feels unnecessary that the scroll event handler will always be called even long after the initial page load. Scroll events tend to happen a lot.

Mankro (commented):
I have some questions first. There are a few comments about the semicolons, but also real topical questions.
* code in chapter.js?
* permanent scroll event handler?

See below for details.

Mankro:
@EerikSaksi Beware, I already made minor changes to add semicolons and fix some spelling mistakes. I force-pushed due to rebasing, thus you shouldn't git-pull and merge this branch. It might be best to "git fetch" and "git reset --hard origin/chapter-exercise-anchor-links". Note that a hard reset basically wipes out the local version of the current branch completely.

exercise/templates/exercise/exercise.html
EerikSaksi: I think a good way of removing the scroll listener would be to remove it when manual scrolling is detected. At that point the scroll listener is useless. 

```
function incrementScrollEvents() {
	consecutiveScrollEvents += 1;
}

window.addEventListener('scroll', incrementScrollEvents);

$( document ).ajaxComplete(() => {
	if (consecutiveScrollEvents < 3) {
		location.hash = hash;
		consecutiveScrollEvents = 0;
	}
	else {
		window.removeEventListener('scroll', incrementScrollEvents);
	}
})
```

exercise/templates/exercise/exercise.html
EerikSaksi: If by the time exercise 4 loads it's guaranteed that exercise 1, 2 and 3 have already been loaded then you could remove the ajax and scroll listener when the exercise referenced by the URL has loaded, as the page won't be scrolled further by loading exercises.

Mankro: It doesn't seem absolutely necessary to move the code to chapter.js. The code is related to what is done in chapter.js, though. Inside `exercise.html`, the JavaScript affects also full-view exercise pages outside chapters, but I don't think it is harmful. Those exercise pages usually don't use any hashes/anchors in the URL (the `#` part in the end of the URL).

exercise/templates/exercise/exercise.html
Mankro: That sounds like a good idea! I would like you to implement it.

Mankro (approved):
Great! I will rebase and squash this and merge the pull request. I will also slightly adjust the whitespace in the code and the commit messages.

PR #1083, opened 2022-09-20
Fix crash: use redirect instead of self.redirect in CourseInstanceMixin

### LINKED ISSUE #1074: Error: view object has no attribute 'redirect'
We have seen this error in the server logs a lot recently. Any internal server error is bad and implies that users end up suddenly on a crash page (HTTP 500 Internal server error). This must be looked into.

```
Internal Server Error: /COURSEKEY/COURSEINSTANCEKEY/toc/

AttributeError at /COURSEKEY/COURSEINSTANCEKEY/toc/
'TableOfContentsView' object has no attribute 'redirect'
```

This occurs on multiple courses (at least, I saw many errors on the courses Y1 and TRAK Y) and on multiple views. `TableOfContentsView` has been common there, but several other view classes have been named in the error logs, at least the following:

* SubmissionView
* ResultsView
* LanguageView
* ModuleView

EerikSaksi:
Edit: this description does not match the updated version after the discussion.

Views deriving from CourseInstanceBaseMixin use self.redirect which is not defined. To fix this, This change defines self.redirect() which is equal to django.shortcuts.redirect

Fixes #1074

# Description

**What?**

This change defines self.redirect in CourseInstanceBaseMixin.

**Why?**

Multiple views were getting errors from self.redirect being undefined. All these views derive from CourseInstanceBaseMixin. 

**How?**

self.redirect is set to django.shortcuts.redirect

Fixes #1074 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I first reproduced the error by singing in as unenrolled and trying to open /def/current/toc. This would try to redirect which would fail. I then ensured that after making the change self.redirect is defined and redirects the user to the correct enrollment exercise.

- [ ] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/viewbase.py
+++ b/course/viewbase.py
@@ -167,7 +167,7 @@ class CourseInstanceMixin(CourseInstanceBaseMixin, UserProfileMixin):
                 exercise = LearningObject.objects.find_enrollment_exercise(
                     self.instance, self.profile.is_external)
                 if exercise:
-                    return self.redirect(exercise.get_absolute_url())
+                    return redirect(exercise.get_absolute_url())
             return redirect(self.instance.get_url('enroll'))
         return super().handle_no_permission()
 


Mankro (changes requested):
You got the correct idea, but it makes more sense to avoid calling `self.redirect()` when Django's `redirect()` should be called directly. I wrote more details in the issue:
https://github.com/apluslms/a-plus/issues/1074#issuecomment-1252486231

To summarize, change this line to use `redirect()` instead of `self.redirect()`:
https://github.com/apluslms/a-plus/blob/a236226539c7a37f601e78a3bb39ec9f76c843fa/course/viewbase.py#L170

-------------------------

About the git commit message: you have a compact message title, but the message body uses very long lines. It is advisable and preferred to use at most 80-character long lines in the commit message (unless you copy-paste long URLs that shouldn't be divided to multiple lines). That is to say, you need to manually hit the Enter key on the keyboard to divide the lines so that you don't write the whole message body on a single very long line.

course/viewbase.py
```diff
@@ -5,6 +5,7 @@
 from django.utils import translation
 from django.utils.translation import gettext_lazy as _
 from django.utils.translation import get_language, get_language_info
+from django.shortcuts import redirect
```
Mankro, line 8: This was already imported on line 4. You don't need this line.

Mankro (changes requested):
Good! You should remove the unnecessary import and update the git commit message. The message does not match the new version at all. I also previously commented on the line length in the git commit message body.

Mankro (approved):
Excellent! Though, note that the heading in git commit message is normally started in the imperative form: "Change" instead of "Changed".

PR #1090, opened 2022-10-06
Add caching to /courses/<course_id>/exercises/

### LINKED ISSUE #1084: Exercise list API /api/v2/courses/<course_id>/exercises/ should use caching to boost performance
This course exercise list API endpoint seems to be too heavy for the database during the busy periods with heavy traffic (e.g., many students working right before course deadlines). 
https://github.com/apluslms/a-plus/blob/a236226539c7a37f601e78a3bb39ec9f76c843fa/course/api/views.py#L178-L211

It does not currently use the cache at all. Try to use the cache so that it would not always query all the course modules and assignments from the database.

`CourseExerciseTreeViewSet` already uses the cache through the `self.content` variable. This API endpoint has been clearly faster.
https://github.com/apluslms/a-plus/blob/a236226539c7a37f601e78a3bb39ec9f76c843fa/course/api/views.py#L214-L241

`self.content` is inherited from `CourseInstanceBaseMixin`:
https://github.com/apluslms/a-plus/blob/a236226539c7a37f601e78a3bb39ec9f76c843fa/course/viewbase.py#L61
It uses `CachedContent`: https://github.com/apluslms/a-plus/blob/a236226539c7a37f601e78a3bb39ec9f76c843fa/exercise/cache/content.py#L14

EerikSaksi:
Fixes #1084 

# Description
This change adds caching to the CourseExercisesViewSet endpoint. The outputs of the endpoint (with caching) are shown below.

/api/v2/courses/1/exercises/
https://gist.github.com/EerikSaksi/fef85032e20e45b09f98f2851cc780b5

/api/v2/courses/1/exercises/1/
https://gist.github.com/EerikSaksi/caaeca53237b08f0acf8c16ee316582e

/api/v2/courses/1/exercises/123456/
https://gist.github.com/EerikSaksi/63378e65211c39f89cc2c4aaa400d924

**Why?**
It is inefficient to fetch the exercises for every user from the database separately. This change saves the exercises in memory to make this endpoint more efficient, and to reduce load on the database.

**How?**
Below are details of all the changes this request introduces, as well as possible improvements which could be considered before this is merged.

***exercise/content/cache.py***
_generate_data now returns a new variable, exercises. exercises is generated from the modules variable. Although this works this could possibly be done better. This implementation even requires a temporary modification of the modules variable (adding is_open to every module which is then removed once exercises is generated). It might be better to not derive exercises from modules for maintainability and for readability, but this was an easy way to do it as modules was well structured and easy to use.

***course/api/views.py***
In this file we modify CourseExercisesViewSet to instead use the cache, rather than the database. We also add support for fetching just one module's exercises with the retrieve function.

***course/api/full_serializers.py***
Here we modify CourseModuleSerializer to process the dictionary data and ModuleExerciseSerializer to serializer the exercises of each module. This works, but there might be a better way to do this which makes better use of existing serializers.

Fixes #1084 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

I ran diff on the output of the old and new API output. The only difference I found was that the order of some of the exercises were different. The output using the caching seems to order the exercises correclyt (5.1.1, 5.1.2 ... 5.1.9).

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/api/views.py
+++ b/course/api/views.py
@@ -1,3 +1,5 @@
+import re
+
 from rest_framework import filters, viewsets, status, mixins
 from rest_framework.decorators import action
 from rest_framework.exceptions import ParseError
@@ -5,19 +7,21 @@ from rest_framework.response import Response
 from rest_framework.settings import api_settings
 from rest_framework_extensions.mixins import NestedViewSetMixin
 from rest_framework.permissions import IsAdminUser
-from django.db.models import Prefetch, Q, QuerySet
+from django.db.models import Q, QuerySet
+from django.http import Http404
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
-from exercise.exercise_models import BaseExercise
-from lib.email_messages import email_course_instance
 
+from aplus.api import api_reverse
+from edit_course.operations.configure import configure_from_url
+from lib.api.constants import REGEX_INT, REGEX_INT_ME
 from lib.api.filters import FieldValuesFilter
 from lib.api.mixins import ListSerializerMixin, MeUserMixin
-from lib.api.constants import REGEX_INT, REGEX_INT_ME
 from lib.api.statistics import BaseStatisticsView
+from lib.email_messages import email_course_instance
+from lib.helpers import build_aplus_url
 from news.models import News
 
-from edit_course.operations.configure import configure_from_url
 from ..models import (
     Enrollment,
     USERTAG_EXTERNAL,
@@ -40,7 +44,6 @@ from ..permissions import (
 )
 from .serializers import CourseBriefSerializer, CourseStudentGroupBriefSerializer, StudentBriefSerializer
 from .full_serializers import (
-    CourseModuleSerializer,
     CourseNewsSerializer,
     CourseSerializer,
     CourseStatisticsSerializer,
@@ -182,7 +185,7 @@ class CourseViewSet(ListSerializerMixin,
 class CourseExercisesViewSet(NestedViewSetMixin,
                              CourseModuleResourceMixin,
                              CourseResourceMixin,
-                             viewsets.ReadOnlyModelViewSet):
+                             viewsets.ViewSet):
     """
     The `exercises` endpoint returns information about the course modules
     defined in the course instance, and the exercises defined in those modules.
@@ -199,20 +202,71 @@ class CourseExercisesViewSet(NestedViewSetMixin,
     lookup_url_kwarg = 'exercisemodule_id'
     lookup_value_regex = REGEX_INT
     parent_lookup_map = {'course_id': 'course_instance.id'}
-    serializer_class = CourseModuleSerializer
 
-    def get_queryset(self) -> QuerySet[CourseModule]:
-        return (
-            CourseModule.objects
-            .get_visible(self.request.user)
-            .prefetch_related(
-                Prefetch('learning_objects', BaseExercise.objects.all(), 'exercises')
-            )
-            .all()
-        )
+    def __recurse_exercises(self, module, exercises):
+        for child in filter(lambda ex: ex['type'] == 'exercise', module['children']):
+            if child['submittable']:
+                # check if there is a xx.yy pattern at the start of the name (e.g. 6.2 Hello Python). Also matches 1.
+                # Also matches roman numerals
+                # https://stackoverflow.com/questions/267399/how-do-you-match-only-valid-roman-numerals-with-a-regular-expression
+                hierarchical_name = (
+                    child['name']
+                    if re.match(
+                        r"^([0-9]+\.[0-9.]* )|(M{0,4}(CM|CD|D?C{0,3})(XC|XL|L?X{0,3})(IX|IV|V?I{0,3})) ",
+                        child['name'],
+                    )
+                    else f"{child['number']} {child['name']}"
+                )
+                exercise_dictionary = {
+                    'id': child['id'],
+                    'url': build_aplus_url(
+                        api_reverse('exercise-detail', kwargs={'exercise_id': child['id']}),
+                        True,
+                    ),
+                    'html_url': build_aplus_url(child['link'], True),
+                    'display_name': child['name'],
+                    'max_points': child['max_points'],
+                    'max_submissions': child['max_submissions'],
+                    'hierarchical_name': hierarchical_name,
+                    'difficulty': child['difficulty'],
+                }
+                exercises.append(exercise_dictionary)
 
-    def get_object(self):
-        return self.get_member_object('module', 'Exercise module')
+            # Both exercises and chapters may have children.
+            # (Chapters are "non-submittable exercises" in the cache.)
+            exercises = self.__recurse_exercises(child, exercises)
+
+        return exercises
+
+    def __module_to_dict(self, module, **kwargs):
+        kwargs['exercisemodule_id'] = module['id']
+        module_dictionary = {
+            'id': module['id'],
+            'url': build_aplus_url(api_reverse("course-exercises-detail", kwargs=kwargs), True),
+            'html_url': build_aplus_url(module['link'], True),
+            'display_name': module['name'],
+            'is_open': CourseModule.check_is_open(
+                module['reading_opening_time'],
+                module['opening_time'],
+                module['closing_time'],
+            ),
+        }
+        module_dictionary['exercises'] = self.__recurse_exercises(module, [])
+        return module_dictionary
+
+    def list(self, request, *args, **kwargs):
+        modules = []
+        for module in self.content.data['modules']:
+            modules.append(self.__module_to_dict(module, **kwargs))
+        return Response({"count": len(modules), "next": None, "previous": None, 'results': modules})
+
+    def retrieve(self, request, *args, **kwargs):
+        # try to get the module list index
+        idx = self.content.data['module_index'].get(int(kwargs['exercisemodule_id']))
+        if idx is None:
+            raise Http404()
+
+        return Response(self.__module_to_dict(self.content.data['modules'][idx[0]], **kwargs))
 
 
 class CourseExerciseTreeViewSet(CourseResourceMixin,
--- a/course/models.py
+++ b/course/models.py
@@ -1319,11 +1319,15 @@ class CourseModule(UrlMixin, models.Model):
         if errors:
             raise ValidationError(errors)
 
-    def is_open(self, when=None):
+    @staticmethod
+    def check_is_open(reading_opening_time, opening_time, closing_time, when=None):
         when = when or timezone.now()
-        if self.reading_opening_time:
-            return self.reading_opening_time <= when <= self.closing_time
-        return self.opening_time <= when <= self.closing_time
+        if reading_opening_time:
+            return reading_opening_time <= when <= closing_time
+        return opening_time <= when <= closing_time
+
+    def is_open(self, when=None):
+        return self.check_is_open(self.reading_opening_time, self.opening_time, self.closing_time, when=when)
 
     def is_after_open(self, when=None):
         """


course/api/full_serializers.py
```diff
@@ -36,29 +36,35 @@
...
+    derive from `AplusModelSerializer` because the items are `dict`s instead of
+    model objects.
+    """
+    id = serializers.IntegerField()
+    url = serializers.CharField()
+    html_url = serializers.CharField()
+    display_name = serializers.CharField()
+    is_open = serializers.BooleanField()
+    exercises = serializers.SerializerMethodField()
+    def get_exercises(self, obj):
```
ihalaij1, line 62: Type hinting is missing here.

exercise/cache/content.py
```diff
@@ -183,6 +185,44 @@ def add_to(target: Dict[str, Any], exercise: BaseExercise) -> None:
...
+                            'max_submissions': exercise['max_submissions'],
+                            'hierarchical_name': exercise['name'],
+                            'difficulty': exercise['difficulty']
+                    }
+                    #add this exercise to the parent module
+                    exercises[-1]['exercises'].append(ed)
+
+        #remove the temporary is_open field from modules used in exercises
+        for module in modules:
+            module.pop('is_open', None)
```
ihalaij1, line 225: A similar for-loop already exists above, so this line should be moved there. That way all modules don't have need to be looped over again.

exercise/cache/content.py
```diff
@@ -183,6 +185,44 @@ def add_to(target: Dict[str, Any], exercise: BaseExercise) -> None:
...
+            exercises.append(m)
+            for submodule in module['children']:
+                children = []
+                #loop over each exercise in the module
+                for exercise in filter(lambda ex: ex['type'] == 'exercise', submodule['children']):
+                    #extract all the relevant fields of the exercise
+                    api_url = f"{settings.BASE_URL}/api/v2/exercises/{exercise['id']}/"
+                    html_url = f"{module_html_url}{exercise['get_path']}/"
+                    ed = {
+                            'id': exercise['id'],
```
ihalaij1, line 211: The key-value pairs in ``ed`` dict are indented too deep.

course/api/full_serializers.py
```diff
@@ -36,29 +36,35 @@
...
         )
-
-    def get_exercises(self, obj: CourseModule) -> OrderedDict:
-        # List only exercises derived from BaseExercise, thus no Chapter texts (/exercise/<id> returns 404 for those)
-        # Check if the exercises were prefetched in the view
-        if hasattr(obj, 'exercises'):
-            exercises = obj.exercises
-        else:
-            exercises = BaseExercise.objects.filter(course_module=obj).all()
-        serializer = ExerciseBriefSerializer(instance=exercises, many=True, context=self.context)
```
ihalaij1, line 61: ``BaseExercise`` and ``ExerciseBriefSerializer`` imports are now unused and should be removed.

ihalaij1 (changes requested):
I can see what you mean with possibly better ways to implement some of the changes, but I couldn't come up with a significantly better way to do it myself.

I commented some parts though, and was also wondering that if caching ``ìs_open`` for exercises is going to cause the API to provide inaccurate data when it is fetched from cache and the exercise has closed recently?

exercise/cache/content.py
EerikSaksi: So the reason I decided to iterate over the modules list rather than this loop (which I'm assuming that you're referring to)
https://github.com/EerikSaksi/a-plus/blob/cd6deb3b9d7347efc7793837780a2fc9da9ba0ac/exercise/cache/content.py#L104

is because the module in the scope of this loop does not have access to the exercises that belong to this module. This loop edits the module list and adds the corresponding exercises to each module. I could try to build the data in this loop a well but looping through the structured data felt far easier than building it through the recursion in the loop.

course/api/full_serializers.py
```diff
@@ -36,29 +34,35 @@
...
+    id =  serializers.IntegerField()
+    url = serializers.CharField()
+    html_url = serializers.CharField()
+    display_name = serializers.CharField()
+    max_points = serializers.IntegerField()
+    max_submissions = serializers.IntegerField()
+    hierarchical_name = serializers.CharField()
+    difficulty = serializers.CharField()
 
+class CourseModuleSerializer(serializers.Serializer):
```
Mankro, line 48: Why would you change the parent class from `AplusModelSerializer` to the basic `Serializer`? Model serializers enable DRY ("don't repeat yourself") and you replaced it with manually repeated fields.

course/api/full_serializers.py
```diff
@@ -36,29 +34,35 @@
...
+    exercises = serializers.SerializerMethodField()
     def get_exercises(self, obj: CourseModule) -> OrderedDict:
-        # List only exercises derived from BaseExercise, thus no Chapter texts (/exercise/<id> returns 404 for those)
-        # Check if the exercises were prefetched in the view
-        if hasattr(obj, 'exercises'):
-            exercises = obj.exercises
-        else:
-            exercises = BaseExercise.objects.filter(course_module=obj).all()
-        serializer = ExerciseBriefSerializer(instance=exercises, many=True, context=self.context)
+        serializer = ModuleExerciseSerializer(
```
Mankro, line 61: At first sight, it looks like this file doesn't change any functionality. The old DRY code was replaced with duplicated field definitions that do basically the same thing as the old code. What was the purpose of this?

course/api/views.py
```diff
@@ -175,10 +176,8 @@ def send_mail(self, request, *args, **kwargs):
                 return Response(_("SEND_EMAIL_FAILED"))
 
 
-class CourseExercisesViewSet(NestedViewSetMixin,
-                             CourseModuleResourceMixin,
-                             CourseResourceMixin,
-                             viewsets.ReadOnlyModelViewSet):
+class CourseExercisesViewSet(CourseResourceMixin, viewsets.ViewSet):
```
Mankro, line 179: I would have thought that removing the super class `NestedViewSetMixin` breaks all the views that depend on nesting. The URL used to nest exercises under the course, e.g., `/api/v2/courses/COURSEID/exercises/EXERCISEID/`. Why did you do this?

`CourseModuleResourceMixin`: how could this be removed?
`viewsets.ReadOnlyModelViewSet`: this used to guarantee that the view does not allow write operations such as HTTP POST and PUT. Why would you use `viewsets.ViewSet`?

exercise/cache/content.py
```diff
@@ -9,6 +9,7 @@
 from lib.cache import CachedAbstract
 from ..models import LearningObject, BaseExercise
 from .hierarchy import ContentMixin
+from django.conf import settings
```
Mankro, line 12: coding style: import statements are grouped
1) Python builtin modules
2) 3rd-party modules like Django
3) the project's internal modules

exercise/cache/content.py
```diff
@@ -183,6 +185,43 @@ def add_to(target: Dict[str, Any], exercise: BaseExercise) -> None:
         if total['min_group_size'] > total['max_group_size']:
             total['min_group_size'] = 1
 
+        exercises = []
+        for module in modules:
+            #add the module fields to cached exercise endpoint
+            api_url = f"{settings.BASE_URL}/api/v2/courses/{instance.id}/exercises/{module['id']}/"
```
Mankro, line 191: Use the `api_reverse` function to build URLs.

https://github.com/apluslms/a-plus/blob/d7dc06549856b10306732a892735cb0d9d66ed0b/aplus/api/__init__.py#L4

exercise/cache/content.py
```diff
@@ -183,6 +185,43 @@ def add_to(target: Dict[str, Any], exercise: BaseExercise) -> None:
         if total['min_group_size'] > total['max_group_size']:
             total['min_group_size'] = 1
 
+        exercises = []
```
Mankro, line 188: Why is it necessary to add a new `exercises` field to the cache?

course/api/views.py
```diff
@@ -192,23 +191,30 @@ class CourseExercisesViewSet(NestedViewSetMixin,
...
+            context={ 'request': request }
         )
+        response_data = { 'results': serializer.data }
+        return Response(response_data)
 
-    def get_object(self):
-        return self.get_member_object('module', 'Exercise module')
+    #get the exercises of a specific course module (has O(n) complexity, should this scale better?)
+    def retrieve(self, request, version = None, course_id = None, pk=None):
+        for module in self.content.data['exercises']:
```
Mankro, line 208: Not ideal if we have to loop over the whole course contents in the cache just to find the given exercise or module...

Mankro (changes requested):
Most of the code changes seem far from optimal since they increase code duplication/repetition and the choices made feel questionable.

We need to figure out a better way to implement this.

Mankro:
I would prefer to avoid duplicating the exercise data inside `CachedContent`. Some of the cache entries (overall in the A+ cache) have already grown too large and that may cause issues with Memcached. I don't want to significantly increase the cache item size in this PR/issue.

I was thinking it would be cleaner to read the existing cache in the methods of `CourseExercisesViewSet` and build the correct output from that data. Eerik said that the output used to include the `is_open` field, which currently does not exist in `CachedContent`. Perhaps `is_open` should be just manually added to the output data in `CourseExercisesViewSet`. It will require looping over the data and using the existing open/close time values to compute `is_open`, but that should still be faster than reading all the data from the database as is currently done in `CourseExercisesViewSet`.

EerikSaksi:
Took another stab at this, this new commit does not implement new serializers, does not add a new giant cache entry, doesn't remove the inheritance from NestedViewSetMixin and CourseModuleResourceMixin and uses module_index to retrieve modules in O(1) time.


Sample outputs from the new commit

/api/v2/courses/1/exercises/
https://gist.github.com/EerikSaksi/6c308f60c3a9bccf491350e1b547ebe6

/api/v2/courses/1/exercises/1/
https://gist.github.com/EerikSaksi/cab270fed61f17db5bc3e26baadf25b8

course/api/views.py
```diff
@@ -197,18 +201,55 @@ class CourseExercisesViewSet(NestedViewSetMixin,
     parent_lookup_map = {'course_id': 'course_instance.id'}
     serializer_class = CourseModuleSerializer
```
Mankro, line 202: If the serializer really isn't used, then it shouldn't be declared in the class at all. But I am curious if this class could still be implemented with the serializer instead of writing all the output manually in the class methods.

course/api/views.py
```diff
@@ -195,20 +198,60 @@ class CourseExercisesViewSet(NestedViewSetMixin,
...
+            'url': api_url,
+            'html_url': module_html_url,
+            'display_name': module['name'],
+            'is_open': CourseModule.is_open_helper(module['reading_opening_time'], module['opening_time'],
+                                                   module['closing_time']
+                                                   ),
+            'exercises': [],
+        }
+        # add this module to cached exercises
+        for exercise in filter(lambda ex: ex['type'] == 'exercise', module['children']):
```
Mankro, line 219: The original implementation excluded chapters from the output; only the (submittable) exercises were listed. This new implementation includes the chapters. I think `CachedContent` includes a field `submittable: True/False` for each `type: "exercise"` object. Chapters have `submittable False` and the exercises `True`.

Note that the exercises may be nested under the chapters as their children, but the exercises may also be listed directly under the module's children. A+ in theory supports unlimited nesting for the learning objects, thus you should traverse all levels of the "children" until the leaf objects have no children. You can see that the class `CachedContent` uses recursion for that.

course/api/views.py
```diff
@@ -195,20 +198,60 @@ class CourseExercisesViewSet(NestedViewSetMixin,
...
+            # extract all the relevant fields of the exercise
+            api_url = build_aplus_url(api_reverse(
+                'exercise-detail', kwargs={'exercise_id': exercise['id']}))
+
+            # check if there is a xx.yy pattern at the start of the name (e.g. 6.2 Hello Python). Also matches 1.
+            hierarchical_name = exercise['name'] if re.match(
+                "^[0-9]+.[0-9]*", exercise['name']) else f"{exercise['number']} {exercise['name']}"
```
Mankro, line 226: Note that course instance settings include the possibility for Roman numerals (I, II, III, IV, V, VI, ...).

This regex does not cover the Roman numerals yet.

https://github.com/apluslms/a-plus/blob/9ed509f62eaaf7d7b87c1238d54e04b8a9bd0ae4/exercise/exercise_models.py#L230-L261

Mankro (changes requested):
There are still a few critical errors. I am also interested in the idea that I presented previously (using the serializer `CourseModuleSerializer`), but let's fix the current implementation first in the way you have started.

course/api/views.py
```diff
@@ -195,20 +198,60 @@ class CourseExercisesViewSet(NestedViewSetMixin,
...
-        return self.get_member_object('module', 'Exercise module')
+    def list(self, request, *args, **kwargs):
+        l = []
+        for module in self.content.data['modules']:
+            l.append(self.__module_to_dict(module, **kwargs))
+        return Response({"count": len(l), "next": None, "previous": None, 'results': l})
+
+    def retrieve(self, request, *args, **kwargs):
+        # try get the module list index
+        i = self.content.data['module_index'].get(int(kwargs['exercisemodule_id']))
```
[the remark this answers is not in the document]
EerikSaksi: I might be wrong but based on the regex for the CourseExerciseViewSet
`^courses/(?P<course_id>[0-9]+)/exercises/(?P<exercisemodule_id>[0-9]+)/$`
the retrieve method doesn't seem to be called unless the supplied id is an integer.

I also tested this by adding a raise Exception to the start of the retrieve method, the exception was only raised when an integer parameter was supplied, and when foo was supplied instead a "page not found" was supplied with a list of URL patterns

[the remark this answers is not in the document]
Mankro: Sounds great, you got it right!

EerikSaksi:
The output is essentially the same as the old output, with a few differences.

The cache stores the exercises in the correct order, whilst the old JSON output exercises could be out of order. I used this python code to sort the old json output by ID so that it would not show up in the diff. For these exercises sorting by ID leads to the correct exercise order.
```
import json
with open('old.json', 'r') as openfile:
    # Reading from json file
    j = json.load(openfile)
    for module in j['results']:
        module['exercises'].sort(key=lambda exercise: exercise['id'])
```

Another difference is that the cache uses `plus:8000` and not `localhost:8000`. I used a simple find and replace for this to not who up on the diff. This shouldn't be a problem in production.

One difference which has not been fixed is the value of the `html_url`. These point to the correct exercises, but generally use the hash in the url (such as `#chapter-exercise-1`) which directs the user to the exercise in the chapter, rather than opening the exercise itself.

Mankro (approved):
Thanks, I will merge this now. I have made some fixes myself. I rebased the branch, fixed old conflicts and force-pushed.

There are still some issues left, but we can merge the pull request now and continue working on the remaining tasks.

Mankro:
**Remaining issues**

* `CachedContent` includes hidden modules too. Students must not see hidden modules in the course exercise list API, but now they do. Perhaps the hidden modules should be filtered out here:
  https://github.com/apluslms/a-plus/blob/ebf3bc012f61f219893d1e00a86acb00f08aa84e/course/api/views.py#L259-L260
* The `list` method does not use the pagination of the REST framework properly, but hardcodes the structure (count, next, previous, results). Is it possible to avoid hardcoding the structure of the paginated response ? We have a default setup for DRF API pagination, or perhaps the method should call the paginator manually.

  https://www.django-rest-framework.org/api-guide/pagination/
* Could `hierarchical_name` be added to `CachedContent`? `__recurse_exercises` now uses an awful regex hack to deal with hierarchical names. If the cache contained the hierarchical name directly in a new field, we could remove the regex hack.
  https://github.com/apluslms/a-plus/blob/ebf3bc012f61f219893d1e00a86acb00f08aa84e/course/api/views.py#L212-L219
* `hierarchical_name` is formed incorrectly for exercises when the course uses Roman content numbering. `__recurse_exercises` uses `f"{child['number']} {child['name']}"` but in this case, the number is Roman while the hierarchical name should always use Arabic numbers "1.2.3".

**Testing:**

* Must still test that the API works when submittable exercises have other exercises as children (rare in practice but possible to do, usually only chapters have children). Can be configured in the course index.yaml by hand since RST does not have a method for setting up exercises with their own children. Our Databases course, for example, uses index.yaml without any RST materials. You can also copy the `_build/yaml/index.yaml` file from any RST course and modify it for testing.
* A+ All results page (course points page) downloads the exercise list from the API. Test that the All results page works as expected.

**Other comments:**

* `html_url`: in the course exercises list, exercise `html_url` has changed to point to the chapter page and the URL ends with the anchor `#chapter-exercise-1`. In the exercise detail view, the `html_url` has not changed. We nowadays actually want to use the chapter URL for the embedded exercises in the `html_url` field since users are supposed to view the exercise in the chapter page.
* The exercises in the list are sorted according to the course hierarchy. In the past, the list was not sorted.
* `build_aplus_url` function has a parameter that in practice chooses between `plus:8000` and `localhost:8000` addresses in the local testing containers. Since the old API had localhost URLs, I set those in the new API code too.

**Potential for major refactoring:**

* Could the output be formed with `CourseModuleSerializer` like the old version of the code used to do when it queried everything from the database? This would replace the methods `__recurse_exercises` and `__module_to_dict`.
* Old, long comment in this PR (and Github likes to fold/collapse it out of sight): https://github.com/apluslms/a-plus/pull/1090#discussion_r1030346488

Mankro:
Today's comments above for your information, @EerikSaksi, @lainets and @etanttila.

> Potential for major refactoring:
>
> Could the output be formed with CourseModuleSerializer like the old version of the code used to do when it queried everything from the database? This would replace the methods __recurse_exercises and __module_to_dict.

The old comment seems to go missing in Github. Here is a copy.

> I wonder if the existing `CourseModuleSerializer` could be used to parse the data from the cache instead of listing all the fields here manually. I don't know if it could work since the cached data is only a dictionary, not the model class instance. But serializers also parse the HTTP POST request parameters that match the model fields, so they are not limited to just rendering model instances in JSON.
> 
> Though, now that you have already written this code, it is maybe not very worthwhile to try to smash it into the serializer format. If the serializer worked easily like we want, then you would just insert the cached dictionary to the serializer in one line of code and the serializer would output the correct JSON to the HTTP response.
> 
> It is worth noting that if the serializer required a model instance as input in order to do what we want here, then the model instance wouldn't have to be read from the database. I think you could just create a new `CourseModule()` with the normal constructor and enter the fields from the cached values. You wouldn't save that object to the database as it is only needed temporarily in memory. The CourseModule instance would be supplied to the `CourseModuleSerializer` instance and the serializer would do its magic. The `get_exercises()` method under `CourseModuleSerializer` seems to even support cached exercises in the module object so that the exercises wouldn't be read from the database, either.
> 
> The output of the API endpoint was previously created with the `CourseModuleSerializer`. If the serializer were still used, it would guarantee that the output structure is identical to the past.
> 
> https://www.django-rest-framework.org/api-guide/serializers/

PR #1104, opened 2022-12-02
Use A+ as an LTI 1.3 tool

murhum1:
# Description
This PR adds functionality to A+ that enables it to be used as an LTI 1.3 tool in Moodle and possibly other LTI platforms.

**Why?**
Historically, the [A+ Astra moodle plugin](https://github.com/apluslms/moodle-mod_astra) that enables A+ exercises to be used in Moodle, required updates whenever the deployed Moodle version changed. By using the LTI standard to add the functionality instead, we ensure compatibility through version updates.

**How?**
The PR adds LTI protocol specific urls, views, and templates that allow embedding A+ material in an LTI context, and allow users to submit exercises to be graded in A+, with grading results being sent back to the Moodle gradebook. LTI deep linking is also supported, allowing teachers to select content via a UI.

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested in Moodle:
* Adding A+ as an LTI tool via Moodle's admin user interface
* Selecting content to embed in an activity as a teacher
* Browsing embedded material as student and teacher
* Submitting exercises as student and teacher
* Verifying that grades are recorded in Moodle's gradebook
  * Teacher grades are not sent to Moodle, though

- [X] Chrome
- [X] Firefox

Mostly Firefox; Chrome only occasionally during development.

Most of the functionality is standalone and should not affect main A+ functionality. Some templates have been modified, mainly adding overwritable blocks, which shouldn't break things. Some code was added to submitting exercise solutions and receiving grades, but the additions are only used in LTI contexts.

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

Documentation regarding the local development setup should be added later.

# Is it Done?

**Known issues**

ACOS exercises do not have their grading results sent to Moodle - this still requires extra thought.

### DIFF
--- a/aplus/settings.py
+++ b/aplus/settings.py
@@ -139,6 +139,7 @@ INSTALLED_APPS = (
     'bootstrapform',
     'rest_framework',
     'rest_framework.authtoken',
+    'pylti1p3.contrib.django.lti1p3_tool_config',
 
     # First party applications
     'inheritance',
@@ -155,6 +156,7 @@ INSTALLED_APPS = (
     'diploma',
     'apps',
     'redirect_old_urls',
+    'lti_tool',
 
     'js_jquery_toggle',
     'django_colortag',
--- a/exercise/protocol/aplus.py
+++ b/exercise/protocol/aplus.py
@@ -9,6 +9,8 @@ from lib.email_messages import email_course_error
 from lib.remote_page import RemotePage, RemotePageException
 from .exercise_page import ExercisePage
 
+from lti_tool.utils import send_lti_points
+
 
 if TYPE_CHECKING:
     from exercise.models import LearningObject
@@ -111,6 +113,8 @@ def load_feedback_page(request, url, exercise, submission, no_penalties=False):
                 exercise.service_url)
             page.errors.append(_('ASSESSMENT_SERVICE_ERROR_RESPONDED_ERROR'))
         submission.save()
+        if page.is_graded and page.is_sane() and submission.lti_launch_id:
+            send_lti_points(request, submission)
 
     return page
 
--- a/lti_tool/templates/lti_tool/lti_submission.html
+++ b/lti_tool/templates/lti_tool/lti_submission.html
@@ -1,0 +1,20 @@
+{% extends "exercise/submission.html" %}
+
+{% block header %}{% endblock %}
+{% block course_sidebar %}
+<div class="col-sm-1"></div>
+{% endblock %}
+{% block breadcrumb %}{% endblock %}
+{% block breadcrumblist %}{% endblock %}
+{% block siblings %}{% endblock %}
+{% block siblings_bottom %}{% endblock %}
+{% block footer %}{% endblock %}
+
+{% block exercise_wait %}
+{% include "lti_tool/_lti_exercise_wait.html" %}
+{% endblock %}
+
+{% block exerciseinfo %}
+{{ block.super }}
+{% include 'exercise/_submission_info.html' %}
+{% endblock %}
--- a/lti_tool/urls.py
+++ b/lti_tool/urls.py
@@ -1,0 +1,59 @@
+from django.conf.urls import url
+from django.views.decorators.csrf import csrf_exempt
+
+from . import views
+from exercise.views import SubmissionPollView
+from course.urls import INSTANCE_URL_PREFIX, MODULE_URL_PREFIX
+from exercise.urls import EXERCISE_URL_PREFIX, SUBMISSION_URL_PREFIX
+
+INSTANCE_URL_FILL = INSTANCE_URL_PREFIX[1:]
+MODULE_URL_FILL = MODULE_URL_PREFIX[1:]
+EXERCISE_URL_FILL = EXERCISE_URL_PREFIX[1:]
+SUBMISSION_URL_FILL = SUBMISSION_URL_PREFIX[1:]
+
+LTI_PREFIX = r'lti/'
+LTI_SELECT_CONTENT_PREFIX = LTI_PREFIX + "select-content/"
+LTI_INSTANCE_URL_PREFIX = LTI_PREFIX + INSTANCE_URL_FILL
+LTI_EDIT_URL_PREFIX = LTI_INSTANCE_URL_PREFIX + r'teachers/'
+LTI_USER_URL_PREFIX = LTI_INSTANCE_URL_PREFIX + r'user/'
+LTI_MODULE_URL_PREFIX = LTI_PREFIX + MODULE_URL_FILL
+
+LTI_EXERCISE_URL_PREFIX = LTI_PREFIX + EXERCISE_URL_FILL
+LTI_SUBMISSION_URL_PREFIX = LTI_PREFIX + SUBMISSION_URL_FILL
+
+urlpatterns = [
+    url(LTI_PREFIX + r'login/$',
+        views.lti_login),
+    url(LTI_SELECT_CONTENT_PREFIX + '$',
+        views.LtiSelectContentView.as_view(),
+        name="lti-select-content"),
+    # csrf_exempt these for now, as proper csrf handling
+    # for these views proved too difficult so far
+    url(LTI_SELECT_CONTENT_PREFIX + INSTANCE_URL_FILL +  '$',
+        csrf_exempt(views.LtiSelectCourseView.as_view()),
+        name="lti-select-course"),
+    url(LTI_SELECT_CONTENT_PREFIX + MODULE_URL_FILL  + '$',
+        csrf_exempt(views.LtiSelectModuleView.as_view()),
+        name="lti-select-module"),
+    url(LTI_SELECT_CONTENT_PREFIX + EXERCISE_URL_FILL + '$',
+        csrf_exempt(views.LtiSelectExerciseView.as_view()),
+        name="lti-select-exercise"),
+    url(LTI_PREFIX + r'launch/$',
+        csrf_exempt(views.LtiLaunchView.as_view()),
+        name="lti-launch"),
+    url(LTI_SUBMISSION_URL_PREFIX + r'$',
+        views.LtiSubmissionView.as_view(),
+        name="lti-submission"),
+    url(LTI_SUBMISSION_URL_PREFIX + r'poll/$',
+        SubmissionPollView.as_view(),
+        name="lti-submission-poll"),
+    url(LTI_EXERCISE_URL_PREFIX + r'$',
+        views.LtiExerciseView.as_view(),
+        name="lti-exercise"),
+    url(LTI_MODULE_URL_PREFIX + r'$',
+        views.LtiModuleView.as_view(),
+        name="lti-module"),
+    url(LTI_INSTANCE_URL_PREFIX + r'$',
+        views.LtiInstanceView.as_view(),
+        name="lti-course"),
+]
--- a/lti_tool/utils.py
+++ b/lti_tool/utils.py
@@ -1,0 +1,74 @@
+import logging
+from datetime import datetime
+from django.conf import settings
+from django.contrib.auth.models import User
+from pylti1p3.grade import Grade
+from pylti1p3.contrib.django import DjangoMessageLaunch, DjangoCacheDataStorage, DjangoDbToolConf
+from pylti1p3.lineitem import LineItem
+from pylti1p3.exception import LtiServiceException
+
+logger = logging.getLogger('aplus.lti_tool')
+
+def get_launch_url(request):
+    target_link_uri = request.POST.get('target_link_uri', request.GET.get('target_link_uri'))
+    if not target_link_uri:
+        logger.error('Missing "target_link_uri" param')
+        raise Exception('Missing "target_link_uri" param')
+    return target_link_uri
+
+def get_tool_conf():
+    # Cannot default to DjangoDbToolConf() in settings.py due to models not being initialized yet
+    if hasattr(settings, "LTI_TOOL_CONF"):
+        return settings.LTI_TOOL_CONF
+    else:
+        return DjangoDbToolConf()
+
+def get_launch_data_storage():
+    return DjangoCacheDataStorage()
+
+def send_lti_points(request, submission):
+    from exercise.exercise_summary import UserExerciseSummary
+    exercise = submission.exercise
+    request.COOKIES['lti1p3-session-id'] = submission.meta_data.get('lti-session-id')
+    launch = DjangoMessageLaunch.from_cache(
+        submission.lti_launch_id,
+        request,
+        get_tool_conf(),
+        launch_data_storage=get_launch_data_storage()
+    )
+
+    try:
+        username = launch.get_launch_data()['https://purl.imsglobal.org/spec/lti/claim/ext']['user_username']
+    except (KeyError, TypeError):
+        username = launch.get_launch_data()['email']
+
+    try:
+        user = User.objects.get(username=username)
+    except User.DoesNotExist:
+        logger.warn(f"Tried to send LTI points for a non-existing user '{username}'")
+        return
+
+    # Moodle does not have gradebook entries for teachers - don't send result if submitter is a teacher
+    is_course_teacher = exercise.course_instance.is_teacher(user)
+    if not is_course_teacher:
+        summary = UserExerciseSummary(exercise, user)
+        best_submission = summary.best_submission
+        ags = launch.get_ags()
+        grade = Grade()
+        (grade.set_score_given(best_submission.grade)
+            .set_timestamp(best_submission.submission_time.strftime('%Y-%m-%dT%H:%M:%S+0000'))
+            .set_score_maximum(exercise.max_points)
+            .set_activity_progress('Completed')
+            .set_grading_progress('FullyGraded')
+            .set_user_id(launch.get_launch_data().get('sub')))
+        line_item = LineItem()
+        line_item.set_tag(str(submission.exercise.id))
+        try:
+            ags.put_grade(grade, line_item)
+        except LtiServiceException as e:
+            # At least Moodle sends a 409 when trying to save
+            # a grade with same timestamp as an existing grade
+            if e.response.status_code == 409:
+                logger.info("Grade for submission has already been saved through LTI; continuing")
+            else:
+                raise e
--- a/lti_tool/views.py
+++ b/lti_tool/views.py
@@ -1,0 +1,325 @@
+import logging
+from django.http import HttpResponse
+from django.views.decorators.clickjacking import xframe_options_exempt
+from django.views.decorators.csrf import csrf_exempt
+from django.shortcuts import redirect, get_object_or_404
+from django.contrib.auth.models import User
+from django.contrib.auth import login
+from django.urls import reverse
+from pylti1p3.contrib.django import DjangoOIDCLogin, DjangoMessageLaunch
+from pylti1p3.deep_link_resource import DeepLinkResource
+from pylti1p3.lineitem import LineItem
+from pylti1p3.exception import LtiException
+
+from course.views import InstanceView, ModuleView
+from course.models import CourseInstance, Enrollment, CourseModule
+from course.viewbase import CourseInstanceBaseView, CourseModuleBaseView
+from exercise.viewbase import ExerciseBaseView
+from exercise.models import LearningObject, BaseExercise
+from exercise.views import ExerciseView, SubmissionView
+from userprofile.models import UserProfile
+from lib.viewbase import BaseTemplateView, BaseRedirectView, BaseMixin
+from authorization.permissions import ACCESS
+from .utils import get_tool_conf, get_launch_data_storage, get_launch_url
+
+logger = logging.getLogger('aplus.lti_tool')
+
+# CSRF exempts due to difficulty in proper implementation. To re-review later.
+@csrf_exempt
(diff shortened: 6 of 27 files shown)


lti_tool/utils.py
```diff
@@ -0,0 +1,57 @@
...
+    return target_link_uri
+
+def get_tool_conf():
+    return DjangoDbToolConf()
+
+def get_launch_data_storage():
+    return DjangoCacheDataStorage()
+
+def send_lti_points(request, submission):
+    from exercise.exercise_summary import UserExerciseSummary
```
PasiSa, line 23: This could be at the beginning of file with other imports, no?

lti_tool/urls.py
```diff
@@ -0,0 +1,61 @@
...
+from django.views.decorators.clickjacking import xframe_options_exempt
+from django.views.decorators.csrf import csrf_exempt
+
+
+from . import views
+from exercise.views import SubmissionPollView
+
+from course.urls import INSTANCE_URL_PREFIX, MODULE_URL_PREFIX
+from exercise.urls import EXERCISE_URL_PREFIX, SUBMISSION_URL_PREFIX
+from django.views.decorators.csrf import csrf_exempt
```
PasiSa, line 11: This django import could be together with the others earlier

lti_tool/utils.py
```diff
@@ -0,0 +1,57 @@
+from django.conf import settings
+from datetime import datetime
+from pylti1p3.grade import Grade
+from pylti1p3.contrib.django import DjangoMessageLaunch, DjangoCacheDataStorage, DjangoDbToolConf
+from pylti1p3.lineitem import LineItem
+from django.contrib.auth.models import User
```
PasiSa, line 6: Group django imports together, separate with empty line from A+ imports.

lti_tool/utils.py
```diff
@@ -0,0 +1,57 @@
...
+from userprofile.models import UserProfile
+from course.models import Enrollment
+
+def get_launch_url(request):
```
PasiSa, line 10: Recently we have applied type hints at method definitions. Could do so also in these new lti_tool modules, e.g.
`def get_launch_url(request: HttpRequest) -> str:`

lti_tool/views.py
```diff
@@ -0,0 +1,309 @@
+from django.http import HttpResponse
+from django.views.decorators.clickjacking import xframe_options_exempt
+from django.views.decorators.csrf import csrf_exempt
+from django.shortcuts import redirect
+from django.contrib.auth.models import User
+from django.contrib.auth import login
+from course.views import InstanceView, ModuleView
```
PasiSa, line 7: Group and order the imports into django imports, pylti1p3 imports and A+ imports.

lti_tool/views.py
```diff
@@ -0,0 +1,309 @@
...
+from exercise.models import LearningObject, BaseExercise
+from exercise.views import ExerciseView, SubmissionView
+from userprofile.models import UserProfile
+from course.models import CourseInstance, Enrollment, CourseModule
+from course.viewbase import CourseInstanceBaseView, CourseModuleBaseView
+from lib.viewbase import BaseTemplateView, BaseRedirectView, BaseMixin
+from authorization.permissions import ACCESS
+from .utils import get_tool_conf, get_launch_data_storage, get_launch_url
+
+@csrf_exempt
```
PasiSa, line 23: These exempts could be supplied with a comment explaining why.  There seems to be others also later in this file, so a general comment in the beginning is probably good enough.

lti_tool/views.py
```diff
@@ -0,0 +1,309 @@
...
+                url_params = {
+                    k: launch_params[k] for k in ["course_slug", "instance_slug"]
+                }
+                return redirect("lti-course", **url_params)
+            else:
+                pass # invalid params
+        elif message_launch.is_deep_link_launch():
+            return redirect("lti-select-content")
+        else:
+            pass # invalid launch type - should not reach
```
PasiSa, line 106: Or should an exception be raised instead? Or log message at least?

PasiSa (changes requested):
Made some minor stylistic comments. There are quite a lot of commits, I think many of them could still be combined together (just based on a quick glance on commit messages).

lti_tool/utils.py
murhum1: That caused a circular import error that I was unable to resolve otherwise - if this is undesirable, do you have any tips that could fix the error by other means?
```
  File "/srv/aplus/exercise/models.py", line 2, in <module>
    from .exercise_models import *
  File "/srv/aplus/exercise/exercise_models.py", line 46, in <module>
    from .cache.exercise import ExerciseCache
  File "/srv/aplus/exercise/cache/exercise.py", line 11, in <module>
    from ..protocol.aplus import load_exercise_page
  File "/srv/aplus/exercise/protocol/aplus.py", line 12, in <module>
    from lti_tool.utils import send_lti_points
  File "/srv/aplus/lti_tool/utils.py", line 9, in <module>
    from exercise.exercise_summary import UserExerciseSummary
  File "/srv/aplus/exercise/exercise_summary.py", line 9, in <module>
    from .cache.content import CachedContent
  File "/srv/aplus/exercise/cache/content.py", line 10, in <module>
    from ..models import LearningObject, BaseExercise
ImportError: cannot import name 'LearningObject' from partially initialized module 'exercise.models' (most likely due to a circular import) (/srv/aplus/exercise/models.py)
```

lti_tool/templates/lti_tool/lti_submission.html
```diff
@@ -0,0 +1,20 @@
...
+{% endblock %}
+{% block breadcrumb %}{% endblock %}
+{% block breadcrumblist %}{% endblock %}
+{% block siblings %}{% endblock %}
+{% block siblings_bottom %}{% endblock %}
+{% block footer %}{% endblock %}
+
+{% block exercise_wait %}
+	{% include "lti_tool/_lti_exercise_wait.html" %}
+{% endblock %}
```
ihalaij1, line 15: The include shouldn't be indented.

lti_tool/views.py
```diff
@@ -0,0 +1,309 @@
...
+                launch_id,
+                self.request,
+                tool_conf,
+                launch_data_storage=get_launch_data_storage()
+            )
+            self.message_launch_data = self.message_launch.get_launch_data()
+
+class LtiInstanceView(InstanceView, LtiSessionMixin):
+    
+    access_mode = ACCESS.ENROLLED
```
ihalaij1, line 128: There should be two empty lines between each of the different classes in this file.

ihalaij1 (changes requested):
I requested some minor changes.

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+        
+        # Get or create user
+        username = self.message_launch_data["email"]
+        try:
+            # Moodle sends username in this parameter
+            username = self.message_launch_data['https://purl.imsglobal.org/spec/lti/claim/ext']['user_username']
+        except (KeyError, TypeError):
+            # Email always received, use as fallback
+            username = self.message_launch_data['email']
+        users = User.objects.filter(username=username)
```
Mankro, line 56: `username` is unique. You should use `User.objects.get()` and catch the exception when the user does not exist.

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+        if len(users) == 1:
+            self.user = users[0]
+        elif len(users) == 0:
+            self.user = User.objects.create(
+                username=username,
+                email=self.message_launch_data["email"],
+                first_name=self.message_launch_data["given_name"],
+                last_name=self.message_launch_data["family_name"]
+            )
+            profile = UserProfile.objects.get(user=self.user)
```
Mankro, line 66:
```suggestion
            profile = self.user.userprofile
```

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+            self.profile = profile
+        login(self.request, self.user)
+        self.request.session["lti-launch-id"] = message_launch.get_launch_id()
+
+        if message_launch.is_resource_launch():
+            launch_params = self.message_launch_data["https://purl.imsglobal.org/spec/lti/claim/custom"]
+            instance = CourseInstance.objects.get(
+                course__url=launch_params["course_slug"],
+                url=launch_params["instance_slug"]
+            )
```
Mankro, line 80: `get_object_or_404()` would be nice here.

https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#get-object-or-404

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+            roles = self.message_launch_data["https://purl.imsglobal.org/spec/lti/claim/roles"]
+            is_course_teacher = Enrollment.objects.filter(
+                    course_instance=instance,
+                    user_profile=UserProfile.objects.get(user=self.user),
+                    role=Enrollment.ENROLLMENT_ROLE.TEACHER
+                ).exists()
```
Mankro, line 86:
```suggestion
            is_course_teacher = instance.is_teacher(self.user)
```

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+            if "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner" in roles and not is_course_teacher:
+                enrolment_success = instance.enroll_student(self.user)
+            if "exercise_path" in launch_params:
```
Mankro, line 89: One empty line before the `if "exercise_path" ...` line would be nice!

exercise/protocol/aplus.py
```diff
@@ -111,6 +113,8 @@ def load_feedback_page(request, url, exercise, submission, no_penalties=False):
                 exercise.service_url)
             page.errors.append(_('ASSESSMENT_SERVICE_ERROR_RESPONDED_ERROR'))
         submission.save()
+        if submission.meta_data.get('lti-launch-id') != None:
+            send_lti_points(request, submission)
```
Mankro, line 117: If the exercise is asynchronously graded, then there are no points at this time yet. This sends zero points to the LTI Platform/Moodle at this time. Is this intentional?

Mankro (changes requested):
Nice work! I have plenty of comments, but most of them are just small changes/improvements.

You asked about the Acos exercises. I think their points are received in the following API function and the submission is also created in the A+ database only at that time:
https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/exercise/api/views.py#L125-L132

lti_tool/templates/lti_tool/lti_submission.html
Mankro: Why not? It is easier to read when the code inside a block or tag is indented more than the surrounding block/tag.

aplus/settings.py
```diff
@@ -552,7 +553,8 @@
 
 APLUS_AUTH.update(APLUS_AUTH_LOCAL)
 
-
+# pylti1p3 admin interface
+INSTALLED_APPS += ('pylti1p3.contrib.django.lti1p3_tool_config',)
```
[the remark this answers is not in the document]
murhum1: Not really - I'll move it to the initial INSTALLED_APPS declaration under 3rd party apps. Unsure why I added it here in the first place.

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+class LtiSelectContentMixin(LtiSessionMixin):
+
+    access_mode = ACCESS.TEACHER
+
+    def get_resource_objects(self):
+        super().get_resource_objects()
+        user = self.request.user
+        profile = UserProfile.objects.get(user=user)
+        self.is_teacher = (Enrollment.objects
+            .filter(user_profile=profile,role=Enrollment.ENROLLMENT_ROLE.TEACHER).exists())
```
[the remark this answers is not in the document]
murhum1: Yes - it's a check to see whether a user should be able to access the content selection dialog at all. But the suggested change makes sense.

lti_tool/views.py
```diff
@@ -0,0 +1,320 @@
...
+
+    template_name = "lti_tool/exercise_list.html"
+
+    def post(self, request, *args, **kwargs):
+        # Create gradebook items
+        ags = self.message_launch.get_ags()
+        lineitems = ags.get_lineitems()
+        exercises = BaseExercise.objects.filter(course_module=self.module)
+        for e in exercises:
+            if e.max_points > 0 and len([li for li in lineitems if li['tag'] == str(e.id)]) == 0:
```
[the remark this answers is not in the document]
murhum1: The expression is supposed to check if the lineitem already exists in the LTI platform in order to avoid unnecessary requests - in testing these sometimes slowed things down a lot. But you're right, it's confusing. How about the following?

```suggestion
            lineitem_exists = [li for li in lineitems if li['tag'] == str(e.id)]
            if e.max_points > 0 and not lineitem_exists:
```

exercise/protocol/aplus.py
murhum1: No, this was only intended to be a way of sending results of synchronously graded exercises, as async results are handled in `exercise/async_views.py`. What would be the proper place of handling synchronous results?

lti_tool/views.py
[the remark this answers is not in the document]
Mankro: That looks great! 

An alternative in the functional style could look like the following, but in Python it is not very pretty. Yours looks nicer!

```python
any(map(lambda li: li['tag'] == str(e.id), lineitems))
```

exercise/protocol/aplus.py
Mankro: This file is the correct place to handle synchronous exercises (questionnaires). You could just move the LTI part (`send_lti_points()`) to the `if` branches so that `send_lti_points()` is only called here when the points really are ready. Though, it is better to save the submission first (`submission.save()`). You could just add more `if` conditions like

```python
if  page.is_graded and page.is_save() and submission.meta_data.get('lti-launch-id'):
    send_lti_points(request, submission)
```

Mankro:
@murhum1 after your git force push, the commit history got broken in the same way what happened once before. Your branch contains a few unwanted copies of commits that already exist in the master branch. Though, you have rebased them so that their commit checksums have changed.

murhum1:
> @murhum1 after your git force push, the commit history got broken in the same way what happened once before. Your branch contains a few unwanted copies of commits that already exist in the master branch. Though, you have rebased them so that their commit checksums have changed.

Sorry, seems like I accidentally included extra commits in the rebase. This should be fixed now.

lti_tool/views.py
```diff
@@ -0,0 +1,317 @@
...
+            status=Enrollment.ENROLLMENT_STATUS.ACTIVE,
+        ).exists()
+
+    def get_common_objects(self):
+        super().get_common_objects()
+        self.parse_lti_session_params()
+        self.teacher_enrollments = Enrollment.objects.select_related("course_instance").filter(
+            user_profile=self.request.user.userprofile,
+            role=Enrollment.ENROLLMENT_ROLE.TEACHER
+        )
```
Mankro, line 198:
```suggestion
        self.teacher_enrollments = Enrollment.objects.select_related("course_instance").filter(
            user_profile=self.request.user.userprofile,
            role=Enrollment.ENROLLMENT_ROLE.TEACHER,
            status=Enrollment.ENROLLMENT_STATUS.ACTIVE,
        )
```
Must check that the status is ACTIVE.

lti_tool/views.py
```diff
@@ -0,0 +1,317 @@
...
+        return super().get(request, *args, **kwargs)
+
+
+class LtiSelectExerciseView(ExerciseBaseView, LtiSelectContentMixin):
+
+    def post(self, request, *args, **kwargs):
+        # Create gradebook items
+        ags = self.message_launch.get_ags()
+        children = self.exercise.children.all()
+        if len(children) > 0:
```
Mankro, line 292: Does the existence of children imply that this is a chapter with embedded exercises in it? Then, the gradebook items are created for the child exercises, right?

This is not important now, but in theory, any learning object may have several layers of children. I think `exercise.children` includes only the direct children, not the "grandchildren" etc. We don't really have any courses that use such extra layers of children.

Mankro (changes requested):
Great! I have a couple of more comments that are quick to address. Are you able to finish them before the holidays?

Question about the git commit history: it looks like you have squashed all the git commits into one. Do you prefer to keep it that way? I think you previously said that one large commit is not optimal, but I can accept the squashed history too. Separating git commits is very time-consuming and we don't have much time left.

lti_tool/utils.py
Mankro: Now that we are running out time, I think we have to skip the static type hinting.

lti_tool/views.py
```diff
@@ -0,0 +1,317 @@
...
+            for e in children:
+                if e.max_points > 0:
+                    li = LineItem()
+                    li.set_tag(str(e.id)).set_score_maximum(e.max_points).set_label(str(e))
```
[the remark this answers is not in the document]
murhum1: It does contain both languages - let's check up on that later, then.

lti_tool/views.py
murhum1: Yea, right now it only creates gradebook items for children and not grandchildren - was not aware there could be more layers.

murhum1:
Thanks for the comments! I should've fixed the last issues now.

I figured squashing the commits to one (+ PR change requests separately) was the least painful way of doing this for now.

Mankro (approved):
Thanks! I will squash the git commits and merge this pull request now. I will soon also write more comments that we could discuss after the holidays. Further fixing and tweaking may continue in new pull requests.

Mankro:
Remaining issues that I know thus far.

* [ ] Acos and JSAV assignments missing in the Moodle gradebook.  
  A+ code where Acos grading is received and the submission is created: https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/exercise/api/views.py#L125-L132

* [ ] https://github.com/apluslms/a-plus/issues/1114  
  Polling of the grading results after submitting. Is the polling widget and JavaScript included in the LTI exercise and chapter pages?  
  The `_lti_exercise_wait.html` template is a copy of the normal `_exercise_wait.html` template with a different hardcoded polling URL.
  
  https://github.com/apluslms/a-plus/pull/1104#discussion_r1044834028  
  > It looks like this file is a copy of the old existing file `exercise/templates/exercise/_exercise_wait.html`, but this one line with the submission poll URL is slightly different. It would be nice to avoid such repetition, but at least this file is very small.

* [ ] csrf exempt in both urls.py and views.py. Is it necessary to set it twice?  
  Why is there a `csrf_exempt` in `LtiSessionMixin.dispatch()`?  
  https://github.com/apluslms/a-plus/pull/1104#discussion_r1054412338

* [ ] multilingual courses and Moodle gradebook item names: messy titles when both languages are dumped there.  
  https://github.com/apluslms/a-plus/issues/1116

* [ ] `exercise/async_views.py` extra try-except.  
  https://github.com/apluslms/a-plus/pull/1104#discussion_r1054436131

* [ ] write instructions about the server setup for admins in the READMEs (e.g. how to generate RSA keys and where to put them).  
  Document new `settings.py` and `local_settings.py` variables: `get_tool_conf()` may use `settings.LTI_TOOL_CONF`).

* [ ] Write teacher's instructions in the Aplus-manual.

* [ ] No static type hinting in the Python code.

* [ ] gradebook items are not created if assignments use more nesting than usually in the course hierarchy.  
  https://github.com/apluslms/a-plus/pull/1104#discussion_r1053638823

* [ ] iframe links and error http status: https://github.com/apluslms/a-plus/issues/1115  
  (See also the LTI view class discussion below.)


**Bigger design question:**

Is it necessary to make new view classes for LTI: `LtiExerciseView` etc.? What if they used the same normal views like `ExerciseView` and the same URLs accordingly? New `/lti/` URLs cause extra work in changing the URLs in the views. The new LTI view classes manipulate the URLs saved in the cache on the fly and it looks hacky.

The LTI classes mainly just set different HTML templates and otherwise, they are the same as the normal views. A view class could also select HTML templates dynamically in the `get_template_names()` method: 
https://docs.djangoproject.com/en/3.2/ref/class-based-views/mixins-simple/#django.views.generic.base.TemplateResponseMixin.get_template_names

Pseudo-code about the idea: normal views could use different LTI templates if the request includes LTI parameters. Otherwise, the normal template is rendered.
```
def get_template_names(self, ...):
    if request includes LTI parameters:
        # "lti" should maybe be inserted in the middle so that the directory paths would make more sense, but this is just a sketch.
        return "lti/" + self.template_name
    else:
        return self.template_name
```

Is this related at all? Iframe links and error http status: https://github.com/apluslms/a-plus/issues/1115

lti_tool/views.py
```diff
@@ -0,0 +1,317 @@
...
+                url=launch_params["instance_slug"]
+            )
+            roles = self.message_launch_data["https://purl.imsglobal.org/spec/lti/claim/roles"]
+            is_course_teacher = instance.is_teacher(self.user)
+            if "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner" in roles and not is_course_teacher:
+                enrolment_success = instance.enroll_student(self.user)
+            
+            if "exercise_path" in launch_params:
+                url_params = {
+                    k: launch_params[k] for k in ["exercise_path", "module_slug", "course_slug", "instance_slug"]
```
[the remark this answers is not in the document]
Mankro: https://github.com/apluslms/a-plus/issues/1155

PR #1105, opened 2022-12-05
Add column granter and time when multiple for same user

### LINKED ISSUE #1046: deviations: columns Granter and Grant time are missing info
Kun monta saman henkilön poikkeamaa on "taiteltu plussan taakse", ei Granter- ja Grant time -sarakkeissa lue mitään edes siinä tapauksessa, että aika ja myöntäjä olisivat kaikilla piilotetuilla riveillä samat. 
Jos tiedot ovat kaikille saman plussan takana olevilla identtiset, voisivat ne näkyä.
(e.g. https://plus.cs.aalto.fi/o1/2021/teachers/deadline-deviations/)

--------------------------------------

In English
> It is extremely common that a student gets a deadline extension for multiple exercises at once. In such cases, the DL extensions table folds the exercises into one entry so that the view doesn’t get cluttered. That’s great, but it would be nice if the metadata — who granted the extension? when? — was shown for the folded items as well (assuming it’s one person and a single time, which should almost always be the case). This does not currently happen. The columns on the right are blank here:

https://github.com/apluslms/a-plus/issues/1121#issue-1517628848

EerikSaksi:
Fix column granter and time not being visible when the deadline deviation applies to multiple exercises for the same user

Fixes #1046 

# Description

**What?**

Fix not being able to see the granter and grant time when multiple exercise have deviations (at least this is how I understood it, my Finnish isn't as good as my English).
 
Screenshot from 2022-12-05 14-43-51

**Why?**
This is important metadata which should always be visible.

**How?**
I copied these lines:
https://github.com/apluslms/a-plus/blob/20d2ed61beb0b8201746f8e374374cefef0c36f5/deviations/templates/deviations/list_dl.html#L126-L139

And then replaced deviation with deviations.0 and replaced the empty `<td>` tags with this.

Fixes #1046 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Ensuring that the granter and grant time were always visible and corresponded to the database values.

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/deviations/templates/deviations/list_dl.html
+++ b/deviations/templates/deviations/list_dl.html
@@ -54,7 +54,7 @@
 				</thead>
 				<tbody>
 					{% translate "UNKNOWN" as unknown %}
-					{% for deviations, can_group, group_id in deviation_groups %}
+					{% for deviations, can_group, group_id, show_granter in deviation_groups %}
 					{% if can_group %}
 					<tr data-group-parent="{{ group_id }}">
 						<td>
@@ -83,8 +83,24 @@
 								{% translate "NO" %}
 							{% endif %}
 						</td>
-						<td></td>
-						<td></td>
+						<td>
+							{% if show_granter %}
+								{% if deviations.0.granter %}
+								{{ deviations.0.granter.name_with_student_id }}
+								{% else %}
+								{{ unknown }}
+								{% endif %}
+							{% endif %}
+						</td>
+						<td data-datetime="{{ deviations.0.grant_time|date:'Y-m-d H:i:s' }}">
+							{% if show_granter %}
+								{% if deviations.0.grant_time %}
+								{{ deviations.0.grant_time|date:'SHORT_DATETIME_FORMAT' }}
+								{% else %}
+								{{ unknown }}
+								{% endif %}
+							{% endif %}
+						</td>
 						<td>
 							<input type="checkbox">
 						</td>
--- a/deviations/templates/deviations/list_submissions.html
+++ b/deviations/templates/deviations/list_submissions.html
@@ -46,7 +46,7 @@
 				</thead>
 				<tbody>
 					{% translate "UNKNOWN" as unknown %}
-					{% for deviations, can_group, group_id in deviation_groups %}
+					{% for deviations, can_group, group_id, _ in deviation_groups %}
 					{% if can_group %}
 					<tr data-group-parent="{{ group_id }}">
 						<td>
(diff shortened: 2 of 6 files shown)


Mankro (changes requested):
Great start, but there are some issues!

* Let's add the same fix for the submission attempt deviations in `list_submissions.html`:
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/templates/deviations/list_submissions.html#L66-L67
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/views.py#L67
* The deviations are grouped together even if the granters and granter times are not identical. Since the user sees only one granter at the top of the group, it is misleading that the same granter does not necessarily apply to all rows inside the group. **If the granters are not identical inside the group, then we should leave the "granter" cell empty in the group header** (like it always used to be before this pull request). Similarly for the grant time.
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/viewbase.py#L30
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/viewbase.py#L267
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/models.py#L202-L206
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/models.py#L235-L236
  * The `is_groupable` methods linked above ignore the granters and grant times. It makes sense to ignore them since usually the granter is not as important piece of information and grouping makes the table easier to read. We just should not render an incorrect shared granter for the whole group at the top if the granter does not really apply to all of the rows inside the group.

Alternative ideas for the implementation:
1. create a new template tag/filter and use the tag to check (in an `if` condition in the html template) whether the granter may be rendered in the group header (or if the granter should be empty in the group header). (Basically, in the place of the new `if` branches you added in the first version.)
  * https://docs.djangoproject.com/en/3.2/howto/custom-template-tags/#writing-custom-template-filters
  * https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#filters
2. modify the output of the `get_deviation_groups()` function so that it yields additional boolean values indicating whether the granter and granter time columns may be grouped or not. Those booleans would be then used in the `if` condition in the HTML templates.
  * https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/deviations/viewbase.py#L267

Mankro (approved):
Great! I will add the missing part in the submission deviation list and merge this.

PR #1151, opened 2023-01-17
Fix X-Frame-Options Header issues in LTI

### LINKED ISSUE #1115: LTI Tool: Some links not working properly when embedded in iframe
When run as embedded iframe, some of the links in A+ bluntly show (in my case) "minus.cs.aalto.fi refused to connect", but when opened as separate window they work, or at least show something more sensible. Something should be done about them so that they show something more informative also when opened in iframe. For example, these buttons/links have the issue:

- [ ] "Edit assignment" button when operating as teacher (the button could be removed in LTI mode?)
- [ ] "View all submissions" button (the button could be removed in LTI mode?)
- [ ] When submitting solution to an exercise with expired deadline (probably same happens also when submission count is full -- have not yet tested)

Also the legendary "403 Gandalf" page (see issue #858 ) causes the same, for example when A+ course is hidden from students, or the user has another reason for not allowed to see the page.

EerikSaksi:
Fixes #1115

# Description

**What?**

Add X-Frame-Options excemptions to urls which do not work in LTI

**Why?**

When errors and links do not display and instead simply refuse to display, this makes LTI much less usable in an iframe
**How?**

Fixes #1115 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Made sure all links worked locally with DEBUG = False, and tested on mycourses test and minus.cs

- [X] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


Mankro (changes requested):
Some clean-up needed. Did you test all the cases listed in the issue #1115?

authorization/views.py
```diff
@@ -67,7 +68,7 @@ def handle_no_permission(self):
 
         # For authenticated users, raise exception.
```
Mankro, line 69: "raise exception" in the comment is not correct anymore.

Is the response status code now 200 instead of 403? Then, the code wouldn't describe the real condition. I wonder if we could keep the 403 status and still avoid the X-Frame-Options DENY response header.

But this isn't very critical if you can't find simple solutions.

authorization/views.py
```diff
@@ -2,6 +2,7 @@
 from django.contrib.auth.views import redirect_to_login
 from django.contrib.messages import error as error_message
 from django.core.exceptions import PermissionDenied
```
Mankro, line 4:
```suggestion
```
`PermissionDenied`: this import is now unused, right? It can be then removed.

Mankro (commented):
I have some questions below for you to answer!

authorization/views.py
EerikSaksi: Thanks for pointing out those issues, I changed the comment and removed the import. 

The render method has an optional argument for the status code, so I added a status=403 to the render call.

```
def render(request, template_name, context=None, content_type=None, status=None, using=None):
    """
    Return a HttpResponse whose content is filled with the result of calling
    django.template.loader.render_to_string() with the passed arguments.
    """
    content = loader.render_to_string(template_name, context, request, using=using)
    return HttpResponse(content, content_type, status)
```

authorization/views.py
```diff
@@ -65,9 +65,9 @@ def handle_no_permission(self):
         # on login and error pages
         error_message(request, message)
 
-        # For authenticated users, raise exception.
+        # For authenticated users, render the exception template.
         if request.user.is_authenticated:
-            raise PermissionDenied(message)
+            return render(template_name="403.html", request=request, status=403)
```
Mankro, line 70: The parameters of `render()`: since the `request` parameter is mandatory and not defined as a keyword argument in the `render` function, it would be better coding style to pass the `request` parameter as a positional parameter. In addition, `template_name` is a mandatory positional parameter too.

Note also that the default 403 template name could be read from a constant:
`ERROR_403_TEMPLATE_NAME` (its value is "403.html")
(the constant is defined in django/views/defaults.py)

```python
return render(request, ERROR_403_TEMPLATE_NAME, status=403)
```

https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#render

I was thinking carefully about why `raise PermissionDenied` does not work here. It causes the HTTP response to use the header `X-Frame-Options DENY`. The header is added even though the `LtiExerciseView` uses the `xframe_options_exempt` decorator. However, Django handles the exception `PermissionDenied` through another view, by default `django.views.defaults.permission_denied`. Since it is another view, the original decorators from the first view do not apply. When this `AuthenticationMixin` class returns a HTTP response directly through `render()`, the code stays inside the original `LtiExerciseView` (that inherits `AuthenticationMixin`) and its decorators apply. The decorator prevents the HTTP response header `X-Frame-Options DENY`.

Mankro (changes requested):
I thought more about this and tested it manually too.

* I think the git commit should be split into two separate commits. One commit removes the course staff buttons in the LTI exercise views and the other commit changes `AuthenticationMixin`.
* There are more staff buttons that should be removed like the ones you have handled in LtiExerciseView. 
  - In the single-exercise view (LtiExerciseView), the submission feedback page includes buttons "view all submissions" and "inspect submission". 
  - When you open a chapter that has embedded exercises, the exercise frames contain the staff button "view all submissions".

lti_tool/views.py
```diff
@@ -184,6 +184,10 @@ def dispatch(self, request, *args, **kwargs):
     def post(self, request, *args, **kwargs):
         return super().post(request, *args, **kwargs, redirect_view='lti-submission')
 
+    def get_common_objects(self):
```
Mankro, line 187: I think this method could be moved to the class `LtiSessionMixin` so that it is not necessary to duplicate it to multiple classes.

Mankro (approved):
Great! I will make slight adjustments myself and merge this pull request.

Mankro:
> The pull request #1151 did not fix one part. 
> 
> When you open a chapter that has embedded exercises, the exercise frames contain the staff button "view all submissions". The button is still there. Fixing this is a bit more complicated since the embedded exercises are loaded from the normal `ExerciseView`, not the LTI Tool -specific `LtiExerciseView`. The normal `ExerciseView` does not set the template context variable `disable_staff_nav` that could be used to remove the button. It requires some thinking if we wanted to change the URLs of embedded exercises in the chapter page so that they use `LtiExerciseView`.

https://github.com/apluslms/a-plus/issues/1115#issuecomment-1531703688

Issue #1078, opened 2022-09-09
Course build performance: avoid unnecessary writing to the database in the aplus-json course import


Mankro:
Importing the aplus-json course definition to A+ is slowed down by unnecessary writing to the database. All course assignments, modules and categories are re-written to the database even if very little has been changed to the previous version. Django's `save()` method has the parameter `update_fields` which could speed up the operation by writing only the necessary fields. In addition, if an assignment has not been modified at all, then it is not necessary to save it again to the database at all.

Chapters and assignments are always saved here:
https://github.com/apluslms/a-plus/blob/f0ef3db6385c242a4ba92d701b1c413d3ccc194c/edit_course/operations/configure.py#L322
Course modules and categories are always saved as well.

https://docs.djangoproject.com/en/3.2/ref/models/instances/#ref-models-update-fields
Note: `update_fields` can not be used when creating a new model instance to the database.

PR #1161, opened 2023-02-09
Optimize course configuring

lainets:
# Description

**What?**

Optimizes database access in course configuring.
Adds simple unit tests to test the course structure after configure call.

**Why?**

Course configuring has been extremely slow.

**How?**

By fetching objects in a single query and only updating objects that have changes.

# Testing

- [X] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/api/views.py
+++ b/course/api/views.py
@@ -17,7 +17,7 @@ from lib.api.constants import REGEX_INT, REGEX_INT_ME
 from lib.api.statistics import BaseStatisticsView
 from news.models import News
 
-from edit_course.operations.configure import configure_content
+from edit_course.operations.configure import configure_from_url
 from ..models import (
     Enrollment,
     USERTAG_EXTERNAL,
@@ -136,7 +136,7 @@ class CourseViewSet(ListSerializerMixin,
     @action(detail=True, methods=["post"], get_permissions=lambda: [JWTInstanceWritePermission()])
     def notify_update(self, request, *args, **kwargs):
         try:
-            success, errors = configure_content(self.instance, self.instance.configure_url)
+            success, errors = configure_from_url(self.instance, self.instance.configure_url)
         except Exception as e:
             success = False
             errors = [format_lazy(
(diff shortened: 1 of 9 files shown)


lainets:
This should close #1078. While there still are improvement possibilities, I do not think they are necessary or worth the effort.

Mankro (changes requested):
Great, this looks good! I have some questions and very minor changes.

* I don't understand all parts of the code well, e.g., comparing the new and old configs and using the difference in updating the database. We could discuss them in the office so that I get the big picture.
* If the teacher changes the type of an exercise to another, then is the A+ database updated correctly? E.g., if an LTIExercise with some key turns into a BaseExercise using the same key in the new version of the course configs (aplus-json).
* I assume this works like it used to be: if the new aplus-json from the course git repo does not define some exercise keys that used to be in the course, then does the A+ front delete those exercises from the database assuming that there are no submissions attached to them?
* If the aplus-json config used to define some setting with a non-default value, and then the teacher wants to reset it to the default by removing the whole setting from the course git repo, then does the A+ front update the setting or is it stuck in the old value? This tends to be a problem since our code changes values in the database only when the setting is defined in the new downloaded configurations. (All the `if key in o:` statements.)

lainets:
> If the teacher changes the type of an exercise to another, then is the A+ database updated correctly? E.g., if an LTIExercise with some key turns into a BaseExercise using the same key in the new version of the course configs (aplus-json).

Yes.

> I assume this works like it used to be: if the new aplus-json from the course git repo does not define some exercise keys that used to be in the course, then does the A+ front delete those exercises from the database assuming that there are no submissions attached to them?

Yes, the behaviour should be the same.

> If the aplus-json config used to define some setting with a non-default value, and then the teacher wants to reset it to the default by removing the whole setting from the course git repo, then does the A+ front update the setting or is it stuck in the old value? This tends to be a problem since our code changes values in the database only when the setting is defined in the new downloaded configurations. (All the `if key in o:` statements.)

I tried to keep the behaviour the same as before, so this is still a problem. Fixing this requires adding some kind of 'deleted' flag to the diff for missing fields (e.g. set the field equal to a custom singleton `Deleted`), and then resetting the fields to default if the flag is present.

Mankro (approved):
Excellent!

PR #1162, opened 2023-02-15
Fix LTI submission polling

### LINKED ISSUE #1114: LTI Tool: Problems with submission received view
When running an A+ task as LTI activity from MyCourses, after submitting solution to a task, "Submission received ... Your submission is being graded ..." is shown, but the view is not refreshed to show the submission result after grading completes. Also the "Submission Info" box is shown two times on right, once is enough.

EerikSaksi:
Fixes #1114

**What?**
Fix not polling the grader for results in the LTI exercise view

**Why?**
The user might think the grading is stuck, or would have to constantly refresh to get the results

**How?**
The `_lti_exercise_wait.html` template wasn't being loaded, but now loads in the super block.

Fixes #1114

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Added sleep to exercise to simulate grader delay in local environment, ensured that the grader is being polled, and that the UI is updated when the grading result is ready.

- [ ] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/lti_tool/templates/lti_tool/lti_submission.html
+++ b/lti_tool/templates/lti_tool/lti_submission.html
@@ -10,11 +10,8 @@
 {% block siblings_bottom %}{% endblock %}
 {% block footer %}{% endblock %}
 
-{% block exercise_wait %}
-{% include "lti_tool/_lti_exercise_wait.html" %}
-{% endblock %}
-
 {% block exerciseinfo %}
 {{ block.super }}
 {% include 'exercise/_submission_info.html' %}
+{% include "lti_tool/_lti_exercise_wait.html" %}
 {% endblock %}


lti_tool/templates/lti_tool/lti_submission.html
```diff
@@ -10,11 +10,8 @@
...
 {% block footer %}{% endblock %}
 
-{% block exercise_wait %}
-{% include "lti_tool/_lti_exercise_wait.html" %}
-{% endblock %}
-
 {% block exerciseinfo %}
 {{ block.super }}
 {% include 'exercise/_submission_info.html' %}
+{% include "lti_tool/_lti_exercise_wait.html" %}
```
Mankro, line 16: I don't think `_lti_exercise_wait.html` belongs to the `exerciseinfo` block. The `exerciseinfo` block is intended for the column on the right side of the exercise and submission pages.

Block `exercise_wait` is confusing since it is not defined at all in `exercise_base.html` or `exercise.html`. The purpose of the block `exercise_wait` needs to be checked. Perhaps, we will remove it completely also from `submission.html`.

Mankro (changes requested):
We could reorganize this slightly better still.

lti_tool/templates/lti_tool/lti_submission.html
```diff
@@ -10,11 +10,8 @@
 {% block siblings_bottom %}{% endblock %}
 {% block footer %}{% endblock %}
 
-{% block exercise_wait %}
-{% include "lti_tool/_lti_exercise_wait.html" %}
-{% endblock %}
-
 {% block exerciseinfo %}
 {{ block.super }}
 {% include 'exercise/_submission_info.html' %}
```
Mankro, line 15:
```suggestion
```
Since this template inherits `submission.html`, `{{ block.super }}` already includes `submission_info.html`. Including it twice causes the same box to appear twice in the rendered html page.

EerikSaksi:
The loading bar seems to go into the correct place by moving it inside the exercisecontent block and calling super.
Screenshot from 2023-04-19 12-05-02

```
{% block exercisecontent %}
	{% include "lti_tool/_lti_exercise_wait.html" %}
	{{ block.super }}
{% endblock %}
```

Should I find a better solution or is this viable?

Mankro:
I chose a different approach and finished it in the commit https://github.com/apluslms/a-plus/commit/ad95bc93b2640713da4dfa20adc7784bf0398918.

It is important to note that the submission feedback polling has been broken since #1104 in the **normal full-view exercise pages**, not only in the LTI Tool exercise pages. I fixed them all in my commit.

About what Eerik suggested:
> ```
> {% block exercisecontent %}
> 	{% include "lti_tool/_lti_exercise_wait.html" %}
> 	{{ block.super }}
> {% endblock %}
> ```
> Should I find a better solution or is this viable?

The template block `exercise_wait` was broken. When I removed it, the `include _exercise_wait.html` part was again back inside the block `exercisecontent`. Then, `lti_submission.html` does not need to separately include `_exercise_wait.html` any longer.

Your suggestion also moved the waiting progress bar to the top, even though it has been located under the other content.

PR #1163, opened 2023-02-16
Allows sending mail from the participants view

oseppala:
Copying mail addresses from A-plus for sending mail is cumbersome because mail is a link to the students page. Changing this into a mailto link makes it a bit more convenient.

This is a minor change to participants.js that changes the email address to link to mailto: email instead of the participant page.


Mankro (approved):
Thanks! I will change the wording of the commit message before merging so that it would match the normal format.

PR #1166, opened 2023-03-16
Implement way of sending Acos grading results to LTI platform

### LINKED ISSUE #1154: LTI Tool v1.3: add LTI grading support for Acos exercises (AJAX submission)
Pull request #1104 added the LTI Tool v1.3 feature to A+. However, the LTI gradebook integration does not yet include assignments that submit the final (graded) solution directly to the grader backend (via AJAX), which then forwards it to the A+ front (including the submission and grading data as well as the feedback and points). For example, Acos and JSAV. The A+ front creates the submission in the database only in the end and attaches the points at the same time. 

A+ code where Acos grading is received and the submission is created: 
https://github.com/apluslms/a-plus/blob/9b530cb00d0594d8d9f49b0a0c84c4239a039791/exercise/api/views.py#L125-L132

The API endpoint above matches the `submission_url` argument of the A+ grader protocol when A+ retrieves the assignment description from the grader (HTTP GET):
https://github.com/apluslms/a-plus/blob/6f7da8cbc52ef01da1799097f7cf737e99b05862/doc/GRADERS.md

murhum1:
# Description

This, combined with https://github.com/acos-server/acos-aplus/pull/8, fixes an issue where A+ was unable to send points from ACOS back to the LTI platform when A+ was used as an LTI tool.

Fixes #1154

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

- [ ] Chrome
- [x] Firefox

### DIFF
--- a/exercise/async_views.py
+++ b/exercise/async_views.py
@@ -52,6 +52,12 @@ def _post_async_submission(request, exercise, submission, errors=None):
         submission.feedback = form.cleaned_data["feedback"]
         submission.grading_data = post_data
 
+        # Acos-submissions initialize this as an empty string for some reason; fix here
+        if submission.meta_data == "":
+            submission.meta_data = {}
+        if form.cleaned_data["lti_launch_id"]:
+            submission.meta_data["lti-launch-id"] = form.cleaned_data["lti_launch_id"]
+
         if form.cleaned_data["error"]:
             submission.set_error()
         else:
--- a/exercise/exercise_models.py
+++ b/exercise/exercise_models.py
@@ -955,7 +955,7 @@ class BaseExercise(LearningObject):
             )
             return self._build_service_url(
                 language, students,
-                ordinal, url_name, submission_url
+                ordinal, url_name, submission_url, lti_launch_id=request.session.get("lti-launch-id")
             )
         return super().get_load_url(language, request, students, url_name, ordinal)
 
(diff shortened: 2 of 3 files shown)


exercise/async_views.py
```diff
@@ -52,6 +52,12 @@ def _post_async_submission(request, exercise, submission, errors=None):
         submission.feedback = form.cleaned_data["feedback"]
         submission.grading_data = post_data
 
+        # Acos-submissions initialize this as an empty string for some reason; fix here
```
Mankro, line 55: The `Submission.meta_data` field uses our old custom `JSONField` that is simply a normal `TextField` with some extra JSON parsing. Thus, its default value is an empty string. In the case of Acos exercises, no value is set to the `submission.meta_data` field before this point. The submission is created here:

https://github.com/apluslms/a-plus/blob/b36e145cbb91b5a3b47204c122b081571d1bcee8/exercise/api/views.py#L128

Mankro (approved):
Great, thanks! I will add the "fixes #1154" link to the git commit message and merge this pull request.

Short description of the implementation as a reminder:
- when the A+ front retrieves the exercise description from the exercise service, the HTTP GET request includes an extra parameter `lti_launch_id` if the user's current session includes LTI parameters from an LTI v1.3 launch (to A+ as the LTI Tool).
- The Aplus protocol at the Acos server (https://github.com/acos-server/acos-aplus/pull/8/) is modified so that it will store the LTI launch id parameter in the user's browser in a hidden HTML  `<input>` element.
- Acos server forwards the LTI launch parameter to the A+ front when the exercise submission is sent to the Acos server.
- When A+ receives the graded submission from the Acos server, A+ stores the LTI launch id in the submission `meta_data` field. The LTI launch id is then used to send the grade to the LTI Platform.

PR #1172, opened 2023-04-04
Add separate vertical scroll bar to exercise submission inspect

### LINKED ISSUE #1111: Separate vertical scroll bars for the submitted file and the grader feedback in the inspect submission view
The inspect submission view shows submitted files on the left and the grader feedback on the right. When the contents are long, **it would be useful to scroll both sides independently vertically**, that is to say, you could scroll down the submitted file without scrolling/moving the grader feedback part.

This is useful in manual assessment, e.g., in the Y1 course exam.

EerikSaksi:
Fixes #1111

# Description

**What?**

The exercise submission can now be scrolled separately in the submission inspect view.  

https://user-images.githubusercontent.com/46625399/229806869-1b67961a-8cc0-47f8-b049-eff04d7a6f26.mp4

**Why?**
This makes viewing long submissions easier

**How?**
Add the same scroll separately button to the submission as well, and create saas rules which applies the sticky scrolling correctly.

Fixes #1111 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Ensuring the sticky scrolling buttons works, and persists in `LocalStorage`

- [ ] Chrome
- [X] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


exercise/static/exercise/assessment.js
```diff
@@ -32,7 +36,29 @@ $(function () {
...
+          };
+          text.highlightCode({extraButtons: [stickyButton]});
+
+          if (canSticky) {
+            const panelHeading = $('.assessment-panel .panel-heading');
+            const observer = new ResizeObserver(function () {
+              element.parent().css('--sticky-top', panelHeading.outerHeight() + 'px');
+            });
+            observer.observe(panelHeading.get(0));
+          }
```
ihalaij1, line 61: This code is similar to the code below that is used for grader feedback. Can we make it into a function?

ihalaij1 (changes requested):
Looks good, but I think the code in assessment.js could be improved by making the duplicated code into a function. See the comment I left.

exercise/static/exercise/assessment.js
```diff
@@ -1,4 +1,37 @@
 $(function () {
+    $.fn.addStickyButton = function (localStorageKey, actionCallback) {
+      // The "skicky" feature is enabled if ResizeObserver is available and there
```
Mankro, line 3: "skicky" -> "sticky"

exercise/static/exercise/assessment.js
```diff
@@ -1,4 +1,37 @@
 $(function () {
+    $.fn.addStickyButton = function (localStorageKey, actionCallback) {
```
Mankro, line 2: This line and the function definition are indented by 4 spaces, but they should only be indented one level. In JS, we have used only 2 spaces per level as you can see in the line 35:
`  $(document).on('aplus:translation-ready', function() {`

The function body is of course indented more than the outer `function(...)` line.

Mankro, line 2: If the `actionCallback` is never used, why does the parameter exist?

Mankro (changes requested):
Nice, it looks good indeed! I requested a few minor stylistic changes.

We also need to test this more. The implementation does not seem to change anything concerning HTML-formatted grader feedback (when the feedback is not just plain text in `<pre>`). Need to check what happens then.

PasiSa:
It looks like Eerik has pushed a new version to address the comments. Maybe reviewers can check if their comments have been satisfied (and mark this as approved if so)?

ihalaij1 (changes requested):
There seems to be some problems still.

For example, try to submit two long Python files to aplus-manual exercise 6.3.6 Wallet.

After toggling/scrolling/toggling the left-hand side `scroll separately` button a few times, you will notice that it does not always function the way it should (sometimes it scrolls separately when the checkmark isn't checked etc.).

Additionally, the HTML feedback on the right-hand side is not scrollable at all.

EerikSaksi:
This should work now hopefully

EerikSaksi:
The scroll separately button should be found based on parameters and not the text of the button. Maybe this is possible when we feed the extrabuttons parameter to the highlightCode function?

PasiSa:
The functionality of this pull request will be covered in an later PR #1286 , that is planned to be merged soon. Therefore closing this pull request now.

PR #1175, opened 2023-04-14
LTI 1.3 platform implementation

PasiSa:
# Description

**What?**

This pull requests implements basic LTI 1.3 platform functionality according to the [Core specification](https://www.imsglobal.org/spec/lti/v1p3) and [Security Framework specification](https://www.imsglobal.org/spec/security/v1p0/). It also includes parts of the [Assignment and Grade Services specification](http://www.imsglobal.org/spec/lti-ags/v2p0), allowing the LTI tool to submit points to A+ platform.

Changes have been distributed in the following sections of in the code:

- External_services contains most of the core functionality needed for A+ authentication and resource link launch. This involves a new table in database for LTI 1.3 Service.
- Exercise models have been updated to support LTI 1.3 exercises. This involves a new table in database for LTI 1.3 Exercise.
- The lineitems and scores API endpoints needed by the Assignments and Grade services specification are implemented using the Django Rest Framework extending the courses API. It works otherwise similarly than existing routes in the API, except that it applies OAuth2 Bearer tokens for access control. Therefore this seemed a logical place for this functionality, even though it is LTI 1.3 related. In principle the API endpoints could also be used outside LTI, if valid bearer tokens are obtained somehow.
- ~~Authorization contains additions related to bearer token management. Also there is a new database table for storing active bearer tokens. Expired bearer tokens are cleaned up from the database once an hour.~~ This was later changed to use signed JWT tokens, that do not need separate table in database.

Notes:

- The scores API requires that the URL does not end with slash. This deviates from the earlier Django default functionality that automatically appended a slash at the end of the URL, and therefore the API routing needed to be changed to support URLs that either end or not with a slash. This also caused some of the existing tests to break, so they needed to be modified (this is a separate commit). However, it seems that existing URLs and functionalities continue to work as before from user's perspective.
- The DB model verbose texts were deliberately left non-translated. Some of the terms are such that trying to translate them to Finnish would only confuse the admin user in admin view.
- We reuse the private key from APLUS_AUTH settings for the OAuth2 JWT exchanges. I didn't see benefit in specifying a separate keyset for this use.

Small changes are needed also in RST tools and gitmanager to parse the LTI 1.3 related exercise parameters.

This is tested against Moodle 4 as LTI 1.3 tool, and Koodisäilö's recent LTI 1.3 modifications.

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


external_services/lti1p3.py
```diff
@@ -0,0 +1,350 @@
...
+        return False
+
+    def validate_scopes(self,
+                        client_id: str,
+                        scopes: List[str],
+                        client: LTI1p3Client,
+                        request,
+                        *args,
+                        **kwargs
+    ) -> bool:
```
ihalaij1, line 136: This method definition does not follow our style guide.

external_services/lti1p3.py
```diff
@@ -0,0 +1,350 @@
...
+        return False
+
+    def validate_grant_type(self,
+                            client_id,
+                            grant_type: str,
+                            client: LTI1p3Client,
+                            request,
+                            *args,
+                            **kwargs,
+    ) -> bool:
```
ihalaij1, line 228: This method definition does not follow our style guide.

ihalaij1 (commented):
I did not find any real issues in the code just from looking at it. But I did add some minor comments.

authorization/models.py
```diff
@@ -55,3 +56,37 @@ def has_access(self, user: AnyUser, payload: Payload, permission: Permission, in
         Check that <payload> has <permission> access to <instance>.
         """
         return False
+
+
+class OAuth2BearerToken(models.Model):
```
lainets, line 61: I _think_ the bearer token can be a signed JWT, so it wouldn't need to be saved to the database: https://oauthlib.readthedocs.io/en/latest/oauth2/tokens/bearer.html

course/api/lti_views.py
```diff
@@ -0,0 +1,127 @@
...
+            lineitem = LineItem(request, exercise, kwargs['course_id'])
+            serializer = LtiLineItemSerializer(lineitem)
+        except LTI1p3Exercise.DoesNotExist as e:
+            logger.error("LTI 1.3/lineitem: Exercise (%d) not found", kwargs['exercise_id'])
+            raise Http404("LTI exercise not found") from e
+        return Response(serializer.data)
+
+    @action(methods=['post'], detail=True, url_path='scores')
+    def scores(self, request: Request, *args, **kwargs):
+        jsondata = json.loads(request.body)
```
Mankro, line 87: Does this crash if the request body is not valid JSON?

course/api/lti_views.py
```diff
@@ -0,0 +1,127 @@
...
+        data = serializer.validated_data
+
+        try:
+            exercise = LTI1p3Exercise.objects.get(pk=kwargs['exercise_id'])
+        except LTI1p3Exercise.DoesNotExist as e:
+            logger.error("LTI 1.3/scores: Exercise (%d) not found", kwargs['exercise_id'])
+            raise Http404("LTI exercise not found") from e
+
+        try:
+            user = User.objects.get(pk=int(data.get('userId')))
```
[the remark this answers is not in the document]
Mankro: I think Django allows numeric strings like `"99"` for the `pk` or `id` in the queryset `get()` and `filter()` methods. Calling `int()` wouldn't then be strictly required, though invalid strings will then pass.

course/api/lti_views.py
```diff
@@ -0,0 +1,127 @@
...
+        except UserProfile.DoesNotExist:
+            logger.error("LTI 1.3/scores: User (%s) does not exist", data.get('userId'))
+            return Response({'UserProfile does not exist'}, status=400)
+
+        logger.info("Received LTI 1.3 'scores' request for exercise %s, user %s", str(exercise), str(user))
+        sub = Submission.objects.filter(exercise=exercise, submitters=user.userprofile)
```
Mankro, line 105: Does this match submissions that have multiple submitters and `user.userprofile` is one of them?

course/api/lti_views.py
```diff
@@ -0,0 +1,127 @@
...
+        if sub.exists():
+            # When points come from LTI, we assume there is only one submission per exercise per user,
+            # that can be updated if LTI delivers larger points than previously.
+            # If points are not larger than previously, we will just ignore the message.
+            # Some LTI tools seem to send scores updates quite frequently.
+            sub = sub.first()
+            adjusted = (1.0 * exercise.max_points * data.get('scoreGiven') / data.get('scoreMaximum'))
```
Mankro, line 112: Have `scoreGiven` and `scoreMaximum` been validated at this point so that they are floats and not any nonsensical types?

Mankro (changes requested):
I have reviewed most of the code thus far. I have a few fairly small comments.

external_services/lti1p3.py
```diff
@@ -0,0 +1,350 @@
...
+
+        return True
+
+    def validate_silent_authorization(self, request) -> bool:
+        return True
+
+    def validate_silent_login(self, request) -> bool:
+        return True
+
+    def validate_user_match(self, id_token_hint, scopes, claims, request) -> bool:
```
Mankro, line 162: The method parameters are not used here. Is it intentional and not a mistake?

external_services/lti1p3.py
```diff
@@ -0,0 +1,350 @@
...
+            id_token.update({
+                "email": self.user.email,
+                "name": self.user.get_full_name(),
+                "given_name": self.user.first_name,
+                "family_name": self.user.last_name,
+                "https://purl.imsglobal.org/spec/lti/claim/ext": {
+                    "user_username": self.user.username,
+                },
+                "https://purl.imsglobal.org/spec/lti/claim/lis": {
+                    "person_sourcedid": self.user.userprofile.student_id,
```
Mankro, line 337: Is this supposed to be unique like `user.id`? `userprofile.student_id` is not unique since different universities may use the same student ids for different people.

external_services/lti1p3.py
```diff
@@ -0,0 +1,350 @@
...
+from authorization.models import OAuth2BearerToken
+
+
+logger = logging.getLogger('aplus.external_services')
+
+def _add_claim(id_token: dict, key: str, value: Any) -> None:
+    id_token.update({f"https://purl.imsglobal.org/spec/lti/claim/{key}": value})
+
+
+class LTI1p3InitiateLogin:
```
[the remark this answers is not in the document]
Mankro: I don't think the class looks that bad. Even though only one method is currently used, classes combine the data together and it is possible to add more methods (if there were needs for that).

Mankro (changes requested):
Additional comments from me. I have now reviewed all the code.

requirements.txt
```diff
@@ -18,6 +18,7 @@ lxml >= 4.6.3, < 5
 cachetools ~= 4.2.1
 icalendar ~= 4.0.7
 oauthlib ~= 3.2.1
+jwcrypto ~= 1.4.2
```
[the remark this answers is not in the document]
PasiSa: jwcrypto is used for generating JSON Web Keys. I couldn't find if PyJWT supports this.

external_services/templates/external_services/authlogin.html
```diff
@@ -0,0 +1,9 @@
+<form action="{{ action }}" name="ltiAuthForm" id="ltiAuthForm"
+      method="post" enctype="application/x-www-form-urlencoded">
+  {% for key, value in params.items %}
```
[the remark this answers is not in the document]
PasiSa: Changed to tabs. The OpenID Connect protocol should take care of CSRF mitigation using the state variable which is one of the parameters here.

external_services/admin.py
```diff
@@ -33,6 +33,11 @@ class LTIServiceAdmin(LinkServiceAdmin):
     list_display = LinkServiceAdmin.list_display + ('access_settings',)
 
 
+class LTI1p3ServiceAdmin(LinkServiceAdmin):
+    search_fields = LinkServiceAdmin.search_fields + ('login_url',)
+    list_display = LinkServiceAdmin.list_display + ('client_id',)
```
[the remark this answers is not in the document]
PasiSa: There shouldn't be any select elements in those fields added by LTI (which inherits from LinkService).

external_services/models.py
```diff
@@ -166,6 +166,48 @@ def get_url(self, replace=None, kwargs={}): # pylint: disable=dangerous-default-
...
+        max_length=128,
+    )
+    jwks_url = models.CharField(
+        verbose_name="JWKS URL for tool's public key",
+        max_length=256,
+    )
+
+    # Note that setting share_name to False is not the same as LTI's "anonymous mode",
+    # because the A+ user ID will still be provided to the tool.
+    # Otherwise we would not be able to connect incoming scores with the correct user.
```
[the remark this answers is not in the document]
PasiSa: LTI spec uses term "anonymous mode" for use case where the user ID (as JWT 'sub' claim) is left entirely out from the messages. Apart from that, user name and email are optional fields that may or may not be included in the messages, which this option controls.

course/api/lti_views.py
PasiSa: What would be the most convenient way of handling the multiple submitters case? On the other hand, with LTI 1.3 the submissions always have just single submitter, and I don't think LTI 1.3 protocol even supports multi-user use.

external_services/lti1p3.py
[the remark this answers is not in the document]
PasiSa: Yep, it is quite simple, but follows the same approach as the old LTI, as called from views.py

external_services/views.py
```diff
@@ -78,23 +94,127 @@ def get_resource_objects(self):
...
+            credentials=credentials,
+        )
+
+        logger.info("LTI 1.3: OAuth2 login request from user: %d, client_id: %s",
+                    request.user.pk,
+                    credentials['client_id'])
+
+        # Template creates an auto-submitted POST form from create_authorization_response output
+        context = super().get_context_data(**kwargs)
+        context['action'] = headers['Location']
```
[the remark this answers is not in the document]
PasiSa: It comes from oauthlib's create_authorization_response. I think we can trust it exists, or if not, then this function has other issues, too.

external_services/lti1p3.py
PasiSa: It is not specifically said how different implementations should use this attribute, and including it is optional, but according to Markus' tool code, MyCourses seems to send student number here. Nevertheless, I now made it "{organization}:{student_id}" for uniqueness. This is now also more similar to the style applied in examples of LTI spec.

external_services/lti1p3.py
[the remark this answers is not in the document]
lainets: Here is a long text about class philosophy and how it applies to this particular class. It might read a bit harsh but don't mind that, it is just how text naturally comes to me. TLDR is that it isn't clear to me what the class represents. The name should't be a verb (e.g. `LTI1p3InitiateLoginRequest`, copying the old LTIRequest, would make more sense), and the doc string needs to change because it makes it seem like the class itself is handling sending the authentication request back. With those changes I'd find the class acceptable but I'd still rather it just be function: I'd find the code clearer if it just called a function (e.g. `get_lti1p3_parameters` or something similar) and it can be made into a class later if more data or methods need to be added.

> Even though only one method is currently used, classes combine the data together and it is possible to add more methods

In general I would agree with this but a class needs to be (1) a clear representation of some concept and (2) either encapsulate data related to the concept together (i.e. multiple attributes) or contains a set of methods that operates on the same data (the object's contents). Without those requirements, you would end up making every function into a class. Making a class because it might fulfill these conditions later doesn't make sense: just make the class later if needed instead of potentially having a class that doesn't fulfill the requirements. A rare exception for the first requirement to be enough is that if it makes the code easier to understand. The reason why this is rare is that without the second requirement, it is basically a function masquerading as a class which typically makes the code harder to understand.

In this case, the class only has a single useful attribute (I don't count self.service as that is not used anywhere) and a single method that just works as a getter for that attribute. This clearly fails the second requirement.

As for the exceptional case of making the code more understandable, I just find this class does the opposite. The name makes it sound like the class initiates a login but all it does is extract parameters into a list. It is also weird that a class name is a verb as it is the class' methods that are supposed to do things, not the class itself. At the very least the doc string should change. It makes it sound like the class itself handles sending the authentication request back. 

As for the old LTI (I presume you mean LTIRequest), it is a bit questionable but it does have a much clearer name and multiple methods that do things, so I can somewhat understand what it represents and why it exists.

course/api/lti_views.py
lainets: Based on the django tests https://github.com/django/django/blob/531f557f9238b8f2e5032e569cf36f0c05bb4043/tests/many_to_many/tests.py#L207-L210, the code indeed matches all submission that have `user.userprofile` in `submitters`.

external_services/views.py
```diff
@@ -78,23 +94,127 @@ def get_resource_objects(self):
...
+            request.build_absolute_uri(),
+            request.method,
+            request.body,
+            request.headers,
+            None,  # Additional credentials, maybe can omit entirely?
+        )
+
+        response = HttpResponse(content=body, status=status)
+        for k, v in headers.items():
+            response[k] = v
```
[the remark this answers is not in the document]
PasiSa: update(headers) didn't work for HttpResponse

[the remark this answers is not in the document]
lainets: `response.update(headers.items())` should work

external_services/models.py
```diff
@@ -166,6 +166,48 @@ def get_url(self, replace=None, kwargs={}): # pylint: disable=dangerous-default-
...
+class LTI1p3Service(LinkService):
+    '''
+    Settings for a remote LTI v1.3 Tool.
+    '''
+    # Verbose names are intentionally non-translated, because they would not make sense in Finnish.
+    login_url = models.CharField(
+        verbose_name="Login initiation URL",
+        max_length=256,
+    )
+    client_id = models.CharField(
```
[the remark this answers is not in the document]
PasiSa: client_id should. Not sure about deployment_id, we do not use it yet for much anything, but it is included because specification wants it, and maybe in future there will be some use for multiple deployments.

requirements.txt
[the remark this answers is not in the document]
Mankro: OK. Let's keep jwcrypto up-to-date.

authorization/permissions.py
```diff
@@ -165,3 +171,48 @@ def has_object_permission(self, request, view, obj):
...
+
+
+class OAuth2TokenPermission(Permission):
+    '''
+    Permission controlled by OAuth2 Bearer tokens.
+    '''
+    def has_permission(self, request: HttpRequest, view: OAuth2ScopeChecker):
+        token = request.headers.get('Authorization')
+        if not token:
+            logger.warning("Missing authorization token")
```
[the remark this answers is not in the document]
Mankro: OK!

course/api/lti_views.py
Mankro: Since LTI doesn't support groups, let's just assume that we don't need do anything special about them. Leave the code as is.

external_services/models.py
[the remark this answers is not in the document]
Mankro: So let's add the uniqueness constraint to the `client_id`. I found this Q&A, and I'm not sure what to make out of it since I'm not that familiar with the terminology.

https://www.imsglobal.org/forum/learning-tools-interoperability-public-forum/197896

> 1) For a given issuer, there can be multiple client_id values. Can the service URLs vary based on the client_id, or are they always linked only to the issuer?
> 
> Yes there can be multiple client_ids for an issuer pointing to the same tool. The way that you would make a distinction between them is by having an id in the in the OIDC login initiation url as a path or query parameter.

external_services/lti1p3.py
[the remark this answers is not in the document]
PasiSa: I don't have strong preferences, either. Made it a function, then.

edit_course/operations/configure.py
```diff
@@ -213,7 +222,16 @@ def update_learning_objects( # noqa: MC0001
...
 
-        if lobject_cls in (LTIExercise, BaseExercise):
+        if lobject_cls == LTI1p3Exercise:
+            lobject.lti_service = LTI1p3Service.objects.filter(menu_label=str(o["lti1p3"])).first()
+            obj_key = "lti_custom"
+            if obj_key in o:
+                setattr(lobject, key, o[obj_key])
+            obj_key = "lti_open_in_iframe"
+            if obj_key in o:
+                setattr(lobject, key, parse_bool(o[obj_key]))
```
lainets, line 232: This isn't right. The `key` is never assigned here. Something like
```
key = "open_in_iframe"
if f"lti_{key}" in o:
    setattr(lobject, key, parse_bool(o[f"lti_{key}"]))
```
or just
```
if "lti_open_in_iframe" in o:
    setattr(lobject, "open_in_iframe", parse_bool(o["lti_open_in_iframe"]))
```
would work (same with lti_custom)

external_services/views.py
```diff
@@ -78,23 +94,127 @@ def get_resource_objects(self):
...
+    """
+    LTI 1.3 Authentication request from LTI Tool. This can be either a GET or POST
+    message according to the OIDC specification.
+    """
+    access_mode = ACCESS.ANONYMOUS
+    template_name = "external_services/authlogin.html"
+
+    @xframe_options_exempt
+    def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
+        validator = LTI1p3Validator(request.user, request.GET, request.GET['lti_message_hint'])
```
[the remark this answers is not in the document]
lainets: This still seems to be an issue?

lainets (changes requested):
Great changes! There are still a couple things left

edit_course/operations/configure.py
PasiSa: Oops... Now it should be right

external_services/lti1p3.py
```diff
@@ -0,0 +1,370 @@
...
+    Implementation of oauthlib Auhtentication request validator for incoming OAuth2
+    requests used in LTI 1.3.
+
+    @param user: Django user associated with HTTP requests, when used in login
+    authorization request triggered by A+ user.
+    None, if request has originated from LTI tool.
+    
+    @param params: URL query parameters from OAuth request.
+    """
+    def __init__(self, user: User, params: dict, message_hint: str):
```
lainets, line 82: `message_hint` type should now be `Optional[str]` after the changes to `LTI1p3AuthRequestView`

PasiSa:
@lainets @Mankro : now the bearer token to jwt tokens change is pushed, if you want to take a look. I made it a separate commit for clarity, but if everything looks ok, I can squash everything, and then it is ready for merging on my behalf.

PasiSa:
I don't have strong opinions about the expiry time, but the LTI relationships are quite static and long-term (you configure it once, and then the service is accessible probably for at least several months), so I don't see one-hour expiry time as big issue. The expiry time would be a new settings variable, then?

lainets:
The concern with the expiry time is not the service itself but any bad actors that may get their hands on the token, e.g. by hacking said service. While we can remove the service from our platform, they would still have a valid token that can be used until the expiry time. Technically, we could check if the token's owner is still in the lti service database when checking the permissions but I'm not sure if that is worth it.

Yes, it would be a new settings variable.

external_services/lti1p3.py
```diff
@@ -0,0 +1,360 @@
...
+        user: User,
+        instance: CourseInstance,
+        exercise: LTI1p3Exercise = None
+        ) -> List[Tuple[str, str]]:
+    """
+    Prepares LTI 1.3 initiate login request sent from A+ to LTI Tool, which is expected
+    to respond by sending an Authentication request back. Returns parameters included in
+    the request.
+    """
+    #def __init__(self, service: LTI1p3Service, user: User, instance: CourseInstance, exercise: LTI1p3Exercise = None):
```
Mankro, line 41: Do we need the commented out `def __init__()` anymore?

external_services/lti1p3.py
```diff
@@ -0,0 +1,360 @@
...
+
+def _add_claim(id_token: dict, key: str, value: Any) -> None:
+    id_token.update({f"https://purl.imsglobal.org/spec/lti/claim/{key}": value})
+
+
+def prepare_lti1p3_initiate_login(
+        service: LTI1p3Service,
+        user: User,
+        instance: CourseInstance,
+        exercise: LTI1p3Exercise = None
```
Mankro, line 34:
```suggestion
        exercise: LTI1p3Exercise = None,
```

external_services/lti1p3.py
```diff
@@ -0,0 +1,360 @@
...
+            return False
+        return True
+
+    def validate_silent_authorization(self, _request) -> bool:
+        return True
+
+    def validate_silent_login(self, _request) -> bool:
+        return True
+
+    def validate_user_match(self, _id_token_hint, _scopes, _claims, _request) -> bool:
```
Mankro, line 155: Does this method override a method from the super class? The method parameter names should not be changed compared to the original super class since the parameter names are also part of the interface. Remember that even normal positional parameters can be supplied in the keyword argument style in function calls.

```python
# calling the method in the keyword style
myobj.validate_user_match(
    _id_token_hint=my_hint,
    _scopes=my_scopes,
    _claims=my_claims,
    _request=my_request,
)
```

Mankro (changes requested):
Looks good overall. I still have a few small comments.

PasiSa:
> The concern with the expiry time is not the service itself but any bad actors that may get their hands on the token, e.g. by hacking said service. While we can remove the service from our platform, they would still have a valid token that can be used until the expiry time. Technically, we could check if the token's owner is still in the lti service database when checking the permissions but I'm not sure if that is worth it.
> 
> Yes, it would be a new settings variable.

I could not not figure out how to adjust the expiry time in this case with this API. It is possible to override the "exp" claim, but the OAuth2 also has the "expires_in" parameter in message header that still is 3600. Note that all examples in OAuth2 specification also use 3600 in their examples, so apparently they don't consider that such dangerous value.

lainets:
Based on https://github.com/oauthlib/oauthlib/blob/d4b6699f8ccb608152b764919e0bd3d38a7b171f/oauthlib/openid/connect/core/endpoints/pre_configured.py#LL35C12-L35C12, the expiry time is set with `token_expires_in` in the `OIDCServer` constructor. But I supposed it isn't such a big issue; you can just leave it be.

PasiSa:
Yes, that was it. Thanks! Now updated according to latest comments. Squash now, maybe?

PasiSa:
By the way, I don't know why this linter message comes up here but not in master:

```
selenium_test/test/test_initializer.py
  Line: 3[9](https://github.com/PasiSa/a-plus/actions/runs/5277166108/jobs/9544839929#step:5:10)
    pylint: unexpected-keyword-arg / Unexpected keyword argument 'chrome_options' in constructor call (col 21)
```

I haven't touched selenium tests in any way.

lainets:
Seems like `chrome_options` had been deprecated and now was finally removed. It needs to be changed to `options`.

Mankro (approved):
Wonderful, I will merge this now!

PR #1176, opened 2023-04-20
Fix publish_url not being called in configure

lainets:
# Description

**What?**

Change config variable name in a loop.

**Why?**

It was redefining another variable used later.


Mankro (approved):
Great fix, thanks!

PR #1178, opened 2023-05-09
Add some comments and type hints for configuring

lainets:
Add clarifying comments and missing type hints for configure() related things.

Could probably just push directly to master but let's do it the "right" way in case there is something more to clarify.

### DIFF
--- a/edit_course/operations/configure.py
+++ b/edit_course/operations/configure.py
@@ -408,7 +408,10 @@ class ConfigParts:
     the lobject class type.
     - *_keys fields contain whatever the <new> config contained. I.e. they
     contain the set of all such keys. E.g. module_keys contains the keys of
-    all modules in the <new> config.
+    all modules in the <new> config. They can be used to determined whether
+    a module, exercise, etc. still exists in the course config even if nothing
+    has changed (e.g. module config disappears from the modules field in the
+    .diff() result if it had no changes).
     """
     config: Dict[Any, Any]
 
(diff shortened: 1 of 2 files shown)


edit_course/operations/configure.py
```diff
@@ -408,7 +408,10 @@ class ConfigParts:
     the lobject class type.
     - *_keys fields contain whatever the <new> config contained. I.e. they
     contain the set of all such keys. E.g. module_keys contains the keys of
-    all modules in the <new> config.
+    all modules in the <new> config. They can be used to determined whether
+    a module, exercise, etc. still exists in the course config even if nothing
+    has changed (e.g. module config dissappears from the modules field in the
```
Mankro, line 413: typo dissappears -> disappears

edit_course/operations/configure.py
```diff
@@ -490,7 +493,8 @@ def get_children(config, n = 0):
         )
 
     @staticmethod
-    def diff(old: Optional["ConfigParts"], new: "ConfigParts"):
+    def diff(old: Optional["ConfigParts"], new: "ConfigParts") -> "ConfigParts":
+        """Returns a new ConfigParts objects containing the differences between the old and new configs"""
```
Mankro, line 497: Python prefers the imperative mood in docstrings. Note also the grammatical error "a new objects" and the missing period in the end.
```suggestion
        """Return a new ConfigParts object containing the differences between the old and new configs."""
```

https://peps.python.org/pep-0257/#one-line-docstrings

edit_course/operations/configure.py
```diff
@@ -502,6 +506,7 @@ def diff(old: Optional["ConfigParts"], new: "ConfigParts"):
             keep_unchanged=True
         )
 
+        # Get the changes between the configs
```
Mankro, line 509: Full sentences could end with the period `.`.

Mankro (changes requested):
Good! I found a couple of typos.

Mankro (approved):

PR #1181, opened 2023-05-17
Protect against buggy participants reports from Aalto SISU API

### LINKED ISSUE #1180: Protect against buggy participants reports from Aalto SISU API
Course enrollments can optionally be fetched from Student Information System such as SISU in Aalto. SISU responds with list of student numbers. Based on the list, student is enrolled in A+ if she did not exist on the course yet. On the other hand, if the student number is not on the list, the A+ enrollment is removed, if it happened as a result of this SISU message.

It has turned out that Aalto's SISU API gateway occasionally returns an empty list for all courses queried using this API because of some malfunction. This causes all students to be removed from course enrollments (and a major hassle as a result). Therefore we should add a check to the code to protect against this situation: if the response contains an empty participant list, it should be ignored.

The relevant code is in [enroll_from_sis method](https://github.com/apluslms/a-plus/blob/8ab057c5e05f1bea2320a6c61e6bee5c296faae5/course/models.py#L918).

Aalto-specific SISU plugin is used to carry out the API messaging in Aalto. The plugin is in [separate repository](https://github.com/apluslms/aalto-sisu). As a side note, It was noted that when SISU API returns an error code in the plugin, it would ne nice if the error handling reported the error code more clearly in logs.

ihalaij1:
# Description

**What?**

It has turned out that Aalto's SISU API gateway occasionally returns an empty list for all courses queried using this API because of some malfunction. This causes all students to be removed from course enrollments (and a major hassle as a result). Therefore we have now added a check to the code to protect against this situation: if the response contains an empty participant list, it is ignored.

Additionally, the error handling now reports the error code more clearly in logs.

Fixes #1180

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

This fix was not tested, because of the difficulty of doing that, but I assume there are no problems.

- [ ] Chrome
- [ ] Firefox

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -24,6 +24,7 @@ from django.utils.functional import cached_property
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
 from django_colortag.models import ColorTag
+from requests.exceptions import HTTPError
 
 from apps.models import BaseTab, BasePlugin
 from authorization.models import JWTAccessible
@@ -935,9 +936,11 @@ class CourseInstance(UrlMixin, models.Model):
         delcount = 0
         try:
             participants = sis.get_participants(self.sis_id)
+        except HTTPError as exc:
+            logger.exception("%s: Error %d when getting participants from SIS.", self, exc.response.status_code)
+            return -1, -1
         except Exception:
-            # pylint: disable-next=logging-fstring-interpolation
-            logger.exception(f"{self}: Error in getting participants from SIS.")
+            logger.exception("%s: Error in getting participants from SIS.", self)
             return -1, -1
 
         from exercise.models import LearningObject # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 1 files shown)


course/models.py
```diff
@@ -935,6 +936,11 @@ def enroll_from_sis(self) -> Tuple[int, int]:
         delcount = 0
         try:
             participants = sis.get_participants(self.sis_id)
+        except HTTPError as exc:
+            code = exc.response.status_code
+            # pylint: disable-next=logging-fstring-interpolation
+            logger.exception(f"{self}: Error {code} when getting participants from SIS.")
```
PasiSa, line 942: Why not do this in the linter-preferred way using %s and %d in the string, and then avoid linter exception? Also, since we are editing this function in any case, also fixing the below similar old log message.

course/models.py
```diff
@@ -954,16 +960,18 @@ def enroll_from_sis(self) -> Tuple[int, int]:
...
-        # unenrolled themselves.
-        students = self.all_students.filter(enrollment__from_sis=True)
-        to_remove = students.exclude(student_id__in=participants)
-        qs = Enrollment.objects.filter(user_profile__in=to_remove, course_instance=self)
-        qs.update(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
-        for e in qs:
-            invalidate_content(Enrollment, e)
-            delcount += 1
+        # Ignore empty participants list caused by a rare SIS API gateway malfunction
+        if participants:
```
PasiSa, line 964: How about adding an else branch, with a (non-severe) log message about the empty participants list, and in that case not showing the "enrolled X students" message (not forgetting to return (0,0) in that case)? Later in the course it indicates a likely problem at SISU end, but for courses that have not yet started it may be a valid situation.

PasiSa (changes requested):
Couple of change proposal related to logging. What do you think?

ihalaij1:
> Couple of change proposal related to logging. What do you think?

I made the suggested changes!

PasiSa (approved):
Thanks, looks good to me.

Mankro (approved):
Well done! I have one preference, though. I like to include a bit more background in the git commit message. If there is only a link to the issue, then you have to open the issue to remember what the problem really was.

I will merge this now.

PR #1184, opened 2023-05-29
Fix the removed counts in SIS logging

PasiSa:
SIS logging incorrectly counted earlier removed students as removed on every updated. Now it counts only those participants who were not removed earlier.

# Description

**What?**

[ANSWER HERE]

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -964,11 +964,13 @@ class CourseInstance(UrlMixin, models.Model):
             # unenrolled themselves.
             students = self.all_students.filter(enrollment__from_sis=True)
             to_remove = students.exclude(student_id__in=participants)
-            qs = Enrollment.objects.filter(user_profile__in=to_remove, course_instance=self)
-            qs.update(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
+            qs = (Enrollment.objects
+                  .filter(user_profile__in=to_remove, course_instance=self)
+                  .exclude(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
+            )
             for e in qs:
                 invalidate_content(Enrollment, e)
-                delcount += 1
+            delcount = qs.update(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
         else:
             logger.warning("%s: Received an empty participants list from SIS.", self)
             return 0, 0


course/models.py
```diff
@@ -964,11 +964,13 @@ def enroll_from_sis(self) -> Tuple[int, int]:
...
-            qs = Enrollment.objects.filter(user_profile__in=to_remove, course_instance=self)
-            qs.update(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
+            qs = (Enrollment.objects
+                  .filter(user_profile__in=to_remove, course_instance=self)
+                  .exclude(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
+            )
             for e in qs:
                 invalidate_content(Enrollment, e)
-                delcount += 1
+            delcount = qs.update(status=Enrollment.ENROLLMENT_STATUS.REMOVED)
```
Mankro, line 973: Would it still make more sense to remove the students before invalidating the cache? `qs.update()` used to come before `invalidate_content()`.

Mankro (changes requested):
Good. I just think it is best to run the operation and then clear the cache, not the other way around.

course/models.py
PasiSa: I tried that first, but it seems that after the update operation the qs becomes empty here. Apparently the update can affect the qs contents if the same attribute that is updated is also used for filtering.

Mankro: Right, I didn't think of that. At least, one possibility is to copy the original list before running `update()`. 
```
qs_before_update = list(qs)
delcount = qs.update(status=...)
for e in qs_before_update:
    invalidate_content(...)
```

Do you prefer to change this or merge the pull request without further changes?

PasiSa: I don't have strong preference, so either option is fine with me, but is the order between invalidate and update that important? Compared to having a new variable for a temporary list?

Mankro (approved):

course/models.py
Mankro: It is not a big deal. We can keep it like you originally did.

PR #1187, opened 2023-06-02
Replace nosetests with unittest + small selenium change

### LINKED ISSUE #1030: Replace the nose library in testing. Required for Python 3.10.
The nose library used in (unit) testing is unmaintained (since 2015) and broken on Python 3.10. Replace it in the testing procedures so that we can run tests on Python 3.10 too.

* https://github.com/apluslms/a-plus/blob/b6472dc6b1cdcca264140b492ca182d6a4f2837a/requirements_testing.txt#L6
* https://github.com/apluslms/a-plus/blob/b6472dc6b1cdcca264140b492ca182d6a4f2837a/aplus/settings.py#L398 (is this related?)
* https://github.com/apluslms/a-plus/blob/b6472dc6b1cdcca264140b492ca182d6a4f2837a/selenium_test/run_servers_and_tests.sh#L13

* https://pypi.org/project/nose/#history
* https://nose.readthedocs.io/en/latest/

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using [Nose2](https://github.com/nose-devs/nose2), [py.test](http://pytest.org/), or just plain unittest/unittest2.

Nose2 is not a direct replacement and it didn't work when I just replaced the library in requirements_testing.txt.

https://docs.nose2.io/en/latest/differences.html

image
https://docs.nose2.io/en/latest/differences.html#test-discovery-and-loading

If that means that all tests.py files in A+ should be renamed, that is a bit awkward. That also does not quite follow the Django conventions.

lainets:
# Description

**What?**

Replace nosetests with unittest.

**Why?**

#1030 

**How?**

Remove XML output from selenium tests and use unittest instead of nosetests.

Closes #1030

# Testing

Checked that the tests are still run correctly (and pass).

# Is it Done?

### DIFF
--- a/doc/jenkins/README.md
+++ b/doc/jenkins/README.md
@@ -1,11 +1,4 @@
-
-A+ is using a continuous integration server running Jenkins.
-The Jenkins is currently running at http://plustest.cs.hut.fi
-
-* The Jenkins UI at port 8080 is limited by firewall at Aalto University.
-* Test reports and test site are public to internet.
-* Pull requests at GitHub are automatically tested and commented by Jenkins.
-
+Jenkins is not currently used but the instructions and configurations are left here in case it will be used in the future. Note that the configurations want a `selenium_test_report.xml` file from the selenium tests but it is not produced anymore.
 
 Jenkins setup
 =============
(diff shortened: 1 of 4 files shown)


Mankro (approved):
Excellent!

PR #1189, opened 2023-06-05
Make course module folding optional and implement automatic scrolling

### LINKED ISSUE #1125: Improve/remove folding on points pages
Points pages now fold all modules whose deadline has passed. I believe this is suboptimal/annoying for both students and teachers. 

People often look at points pages after a deadline has passed. I expect that people are mostly concerned with three things: the previous module that just passed, the module that is relevant next, and the overall points totals. The first of the three is now folded away for no good reason. 

Three alternatives for improvement:
1. Get rid of the folding. Auto-scroll the page to the section that is expected to be relevant. Users can scroll up and down as they like.
2. Keep the folding but fold one module less than now. (I.e., leave the most recently closed module unfolded.)
3. Just get rid of the folding; users scroll down from the top.

I’d personally prefer #1 but would be fine with any of the three.

ihalaij1:
# Description

**What?**

The automatic folding and hiding of expired course modules on the user results page was suboptimal and annoying for some students/teachers.

Old course modules are no longer hidden behind a button and folding is made optional with the use of a new "Expand all modules" button. Automatic scrolling takes the user to the first open module, unless the page contains previously unseen course news.

**Why?**

Improvements to course module folding were requested by a teacher.

**How?**

Local storage is used to save the state of the "Expand all modules" button and the course news strings for each course.

If the "Expand all modules" button is checked, all course modules will be expanded upon page load, even if they are expired or inaccessible.

If the page contains course news that aren't found in local storage, the page will not be automatically scrolled to the first open module, so that the student has a chance of noticing the news. 

Fixes #1125

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the folding and scrolling work as expected.

- [x] Chrome
- [x] Safari

# Translation

- [x] Did you modify or add new strings in the user interface?

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/static/exercise/user_results.js
+++ b/exercise/static/exercise/user_results.js
@@ -1,0 +1,146 @@
+function checkButton(button) {
+  button
+    .attr("aria-pressed", "true")
+    .find("span")
+    .removeClass("glyphicon-unchecked")
+    .addClass("glyphicon-check");
+}
+
+function uncheckButton(button) {
+  button
+    .attr("aria-pressed", "false")
+    .find("span")
+    .removeClass("glyphicon-check")
+    .addClass("glyphicon-unchecked");
+}
+
+function expandModules(skipAnimation) {
+  if (skipAnimation) {
+    $(".module-panel > div").addClass("in"); // addClass("in") skips the expand animation
+  }
+  $(".module-panel > div").collapse("show");
+  checkButton($("#toggle-expand-all-modules"));
+}
+
+function collapseModules() {
+  $(".module-panel, .module-to-collapse > div").collapse("hide");
+  uncheckButton($("#toggle-expand-all-modules"));
+}
+
+function startListen(course) {
+  $(".filter-categories button").on("click", function(event) {
+    const button = $(this);
+    const id = button.attr("data-category");
+    if (button.attr("aria-pressed") === "false") {
+      checkButton(button);
+      $('.module-panel tr[data-category="' + id + '"]').removeClass("hide");
+    } else {
+      uncheckButton(button);
+      $('.module-panel tr[data-category="' + id + '"]').addClass("hide");
+    }
+    $('.module-panel').each(function(index, panel) {
+      const mod = $(panel);
+      if (mod.find("tr:not(.hide)").length > 0) {
+        mod.removeClass("hide");
+      } else {
+        mod.addClass("hide");
+      }
+    });
+  });
+
+  if ($(".module-to-collapse").length > 0) {
+    // Show the "Expand all modules" button if some modules are collapsed
+    $("#toggle-expand-all-modules").show();
+    // Check local storage and expand all modules if the variable "expandAllModules" is set to true
+    const expandAllModules = localStorage.getItem("expandAllModules") === "true";
+    if (expandAllModules) {
+      expandModules(true);
+    }
+
+    // Register an event handler to the "Expand all modules" button
+    $("#toggle-expand-all-modules").on("click", function(event) {
+      const expandAllModules = localStorage.getItem("expandAllModules") === "true";
+      if (expandAllModules) {
+        collapseModules();
+        localStorage.setItem("expandAllModules", false);
+      } else {
+        expandModules(false);
+        localStorage.setItem("expandAllModules", true);
+      }
+    });
+
+    // Check if automatic scrolling is enabled and initialize the local storage variable
+    $("#toggle-auto-scroll").show();
+    const autoScrollEnabled = localStorage.getItem("autoScrollEnabled");
+    if (autoScrollEnabled === "false") {
+      uncheckButton($("#toggle-auto-scroll"));
+    } else if (!autoScrollEnabled) {
+      localStorage.setItem("autoScrollEnabled", true); // Scroll automatically by default
+    }
+
+    // Check if automatic scrolling behavior is set to "instant" and initialize the local storage variable
+    if (autoScrollEnabled !== "false") {
+      $("#auto-scroll-behavior-instant").show();
+    }
+    const autoScrollBehavior = localStorage.getItem("autoScrollBehavior");
+    if (autoScrollBehavior === "instant") {
+      checkButton($("#auto-scroll-behavior-instant"));
+    } else if (!autoScrollBehavior) {
+      localStorage.setItem("autoScrollBehavior", "smooth"); // Scroll smoothly by default
+    }
+
+    // Register event handlers to both buttons related to automatic scrolling
+    $("#toggle-auto-scroll").on("click", function(event) {
+      const autoScrollEnabled = localStorage.getItem("autoScrollEnabled");
+      if (autoScrollEnabled === "false") {
+        checkButton($("#toggle-auto-scroll"));
+        $("#auto-scroll-behavior-instant").show();
+        localStorage.setItem("autoScrollEnabled", "true");
+      } else {
+        uncheckButton($("#toggle-auto-scroll"));
+        $("#auto-scroll-behavior-instant").hide();
+        localStorage.setItem("autoScrollEnabled", "false");
+      }
+    });
+
+    $("#auto-scroll-behavior-instant").on("click", function(event) {
+      const autoScrollBehavior = localStorage.getItem("autoScrollBehavior");
+      if (autoScrollBehavior === "instant") {
+        uncheckButton($("#auto-scroll-behavior-instant"));
+        localStorage.setItem("autoScrollBehavior", "smooth");
+      } else {
+        checkButton($("#auto-scroll-behavior-instant"));
+        localStorage.setItem("autoScrollBehavior", "instant");
+      }
+    });
+  }
+
+  // Get current course news from the document and previously shown course news from local storage
+  const news = [];
+  $(".news-panel > .list-group > .list-group-item").each(function() {
+    const title = $(this).children(".list-group-item-heading").children(".list-group-item-title")[0].innerText || "";
+    const body = $(this).children(".list-group-item-text")[0].innerText || "";
+    news.push({ [title]: body });
+  });
+  const allPreviousNews = JSON.parse(localStorage.getItem("courseNews")) || {};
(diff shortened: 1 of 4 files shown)


ihalaij1:
We agreed to add another button to select the scrolling behavior, because some people might want to disable it or want it to scroll immediately without animation.

ihalaij1:
I pushed some updates, ready for a review.

exercise/static/exercise/user_results.js
```diff
@@ -0,0 +1,102 @@
...
+			$("#auto-scroll-behavior-off").show();
+			localStorage.setItem("autoScrollBehavior", "off");
+		} else if (autoScrollBehavior === "off") {
+			$("#auto-scroll-behavior-smooth").show();
+			localStorage.setItem("autoScrollBehavior", "smooth");
+		} else {
+			$("#auto-scroll-behavior-instant").show();
+			localStorage.setItem("autoScrollBehavior", "instant");
+		}
+	});
```
lainets, line 53: This has the same duplication issue as the expand all button.

exercise/static/exercise/user_results.js
```diff
@@ -0,0 +1,102 @@
...
+			}
+		});
+	});
+
+	// Get current course news from the document and previously shown course news from local storage
+	const news = [];
+	$(".news-panel > .list-group > .list-group-item > .list-group-item-text").each(function() {
+		news.push($(this)[0].innerText);
+	});
+	const course = "{{course|safe}}";
```
lainets, line 80: `{{course|safe}}` doesn't work in a .js file: they aren't django templates.

exercise/static/exercise/user_results.js
```diff
@@ -0,0 +1,102 @@
...
+			} else {
+				mod.addClass("hide");
+			}
+		});
+	});
+
+	// Get current course news from the document and previously shown course news from local storage
+	const news = [];
+	$(".news-panel > .list-group > .list-group-item > .list-group-item-text").each(function() {
+		news.push($(this)[0].innerText);
```
lainets, line 78: This only considers the body of the news item but not the title. I think it would be better to also include the title. For example, the title may have been edited or the news item only has a title and no body.

exercise/templates/exercise/_user_results.html
```diff
@@ -16,22 +19,37 @@
...
 
-<p id="toggle-expired" class="hide">
-	<button class="aplus-button--secondary aplus-button--xs">
-		<span class="glyphicon glyphicon-time" aria-hidden="true"></span>
-		{% translate "SHOW_OLDER_ASSIGNMENTS" %}
+<p>
+	<button id="toggle-expand-all-modules" class="aplus-button--secondary aplus-button--xs" style="display:none">
+		<span class="glyphicon glyphicon-unchecked" aria-hidden="true"></span>
+		{% translate "EXPAND_ALL_MODULES" %}
 	</button>
```
lainets, line 26: I believe the toggle button needs an aria-pressed attribute for accessibility https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#toggle_buttons.

exercise/templates/exercise/_user_results.html
```diff
@@ -16,22 +19,37 @@
...
+		</button>
+		<button id="auto-scroll-behavior-instant" class="aplus-button--secondary aplus-button--xs" style="display:none">
+			<span class="glyphicon glyphicon-flash" aria-hidden="true"></span>
+			{% translate "AUTO_SCROLL_BEHAVIOR_INSTANT" %}
+		</button>
+		<button id="auto-scroll-behavior-off" class="aplus-button--secondary aplus-button--xs" style="display:none">
+			<span class="glyphicon glyphicon-unchecked" aria-hidden="true"></span>
+			{% translate "AUTO_SCROLL_BEHAVIOR_OFF" %}
+		</button>
+	</span>
```
lainets, line 40: It doesn't seem right to have a simple button that cycles through different options (the scroll behaviour button). I'm not sure what is the best way to do it but the two options I came up with are (1) two checkbox buttons: one for on/off and one for instant/smooth (hide/show the second button based on the first), and (2) three radio buttons (Does this fit the A+ style? Are they used anywhere?).

exercise/static/exercise/user_results.js
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	function autoScroll() {
+		const autoScrollBehavior = localStorage.getItem("autoScrollBehavior");
+		if (autoScrollBehavior !== "off") {
+			// Scroll to the first open module if there are no new news items
+			const openModules = $(".module-panel:not(.module-to-collapse)");
+			if (openModules.length > 0 && JSON.stringify(currentNews) === JSON.stringify(previousNews)) {
+				const firstHeadingId = openModules[0].querySelector(".panel-heading").id;
+				document.getElementById(firstHeadingId).scrollIntoView({ behavior: autoScrollBehavior });
```
lainets, line 96: Why does this query the element, take its id, and then query it again? Seems to me like `openModules[0].querySelector(".panel-heading").scrollIntoView(...` would work fine. This also makes the heading id unneccesary in the html.

lainets (changes requested):
Great work! One proposed behavioural change: I don't think the page should scroll automatically for people new to the course. We probably want any introductory text at the top of the course be visible in that case. The simple fix is to not scroll if the first module (or, equivalently, all modules) is visible. Scrolling in that case doesn't seem necessary as the module would be visible anyway (unless the intro text is really long).

exercise/static/exercise/user_results.js
```diff
@@ -0,0 +1,102 @@
...
+			const expandAllModules = localStorage.getItem("expandAllModules") === "true";
+			if (expandAllModules) {
+				$(".module-panel, .module-to-collapse > div").collapse("hide");
+				button.find("span").removeClass("glyphicon-check").addClass("glyphicon-unchecked");
+				localStorage.setItem("expandAllModules", false);
+			} else {
+				$(".module-panel > div").collapse("show");
+				button.find("span").removeClass("glyphicon-unchecked").addClass("glyphicon-check");
+				localStorage.setItem("expandAllModules", true);
+			}
```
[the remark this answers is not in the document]
ihalaij1: Fixed!

exercise/static/exercise/user_results.js
ihalaij1: Fixed by adding a separate button for enabling/disabling auto-scrolling and setting the scroll behavior.

exercise/static/exercise/user_results.js
ihalaij1: Fixed!

exercise/static/exercise/user_results.js
ihalaij1: You're right. I forgot to change this when I moved the JavaScript code from the HTML file to a separate .js file.

exercise/templates/exercise/_user_results.html
ihalaij1: I don't think this is needed since the buttons have the ``aria-hidden`` class.

From Mozilla:
"Adding ``aria-hidden="true"`` to an element removes that element and all of its children from the accessibility tree"

The ``filter-categories`` buttons above don't have ``aria-pressed`` class either when toggled.

exercise/templates/exercise/_user_results.html
ihalaij1: Fixed by adding a separate button for enabling/disabling auto-scrolling and setting the scroll behavior.

exercise/static/exercise/user_results.js
ihalaij1: You're right! Fixed.

lainets (changes requested):
Nice work! In the future, it'd be preferable if any new changes came in a separate commit to make it easier to see what has changed.

exercise/static/exercise/user_results.js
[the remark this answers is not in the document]
ihalaij1: Added another function for collapsing modules.

exercise/templates/exercise/_user_results.html
ihalaij1: Okay, I added this to all buttons now (even the older ones).

ihalaij1:
> Nice work! In the future, it'd be preferable if any new changes came in a separate commit to make it easier to see what has changed.

You can still see the what changed since the last force-push if you click the ``Compare`` button next to the force-push activity.
It's easier not having to squash the commits later if everything is ready for merging.

lainets (commented):
I otherwise approve this but I'm wondering whether the previous module should be uncollapsed as stated in the issue #1125 or wheter it becomes a non-issue after these changes. I also don't see any hurry in merging this, so we could wait in case Markku or Pasi has some comments about the exact behaviour.

> You can still see the what changed since the last force-push if you click the `Compare` button next to the force-push activity. It's easier not having to squash the commits later if everything is ready for merging.

Oh! I didn't realize. However, there is still the issue that review comments cannot be added through that compare page. In this case it wouldn't have mattered as I had no new comments to make but in general at least I would prefer having the corrections in separate commits.

lainets (approved):

ihalaij1:
> I otherwise approve this but I'm wondering whether the previous module should be uncollapsed as stated in the issue #1125 or wheter it becomes a non-issue after these changes. I also don't see any hurry in merging this, so we could wait in case Markku or Pasi has some comments about the exact behaviour.

I previously talked about expanding the most recently closed module with Markku, and I think we came to the conclusion it would be a bit confusing. For example, it might look like to the student that the module is still open, since the older closed modules are collapsed.

Mankro (approved):

exercise/static/exercise/user_results.js
```diff
@@ -0,0 +1,146 @@
+function checkButton(button) {
+	button
+	  .attr("aria-pressed", "true")
```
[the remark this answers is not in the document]
Mankro: I changed it to spaces myself.

Issue #1188, opened 2023-06-05
Active elements output empty


lainets:
The manual's active elements example is supposed to echo back the input but the output is left empty. The problem is not yet on production but can be found on minus: https://minus.cs.aalto.fi/aplus-manual/master/active_elements/introduction/.

Commit #ad95bc93b2640713da4dfa20adc7784bf0398918 changes the URL being polled to be configured through a HTML attribute but the URL hasn't been configured for active elements. Furthermore, active elements wants to poll the submission but the submission doesn't exist when the page is loaded, so the URL isn't known until after the page is loaded when the submission is made.

PR #1190, opened 2023-06-05
WIP using JSON responses to AJAX requests

### LINKED ISSUE #487: Active elements get somehow broken after the module close time
Active elements usually have no points, and they are supposed to be used like interactive visualisations as part of the material. Therefore, it is problematic that students can not submit them after the module close time. In addition, if the module has late submissions enabled, then submissions to active elements seem to have some strange issues even though submitting should be possible during the late period.

lainets:
I created this PR to discuss the idea.

I do not think it is a good idea to load pages using ajax and then use DOM manipulation to get the needed information out of the loaded page. This will easily cause issues (that are also hard to notice and debug) if the page HTML changes. There is also no indication in the HTML itself (or whatever creates it) that the JS is dependent on a specific structure. I propose we add JSON versions of the pages that would otherwise need such DOM introspection to move the logic to python where it will be much easier to see if changes have an effect on something else.

# Description

**What?**

[ANSWER HERE]

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/static/exercise/chapter.js
+++ b/exercise/static/exercise/chapter.js
@@ -453,14 +453,16 @@
     },
 
     // Submit the formData to given url and then execute the callback.
-    submitAjax: function(url, formData, callback) {
+    submitAjax: function(url, formData, ajaxParams, callback) {
+      ajaxParams = ajaxParams ?? {};
+      ajaxParams.dataType = ajaxParams.dataType ?? "html";
       var exercise = this;
       $.ajax(url, {
         type: "POST",
         data: formData,
         contentType: false,
         processData: false,
-        dataType: "html"
+        ...ajaxParams,
       }).fail(function(xhr, textStatus, errorThrown) {
         //$(form_element).find(":input").prop("disabled", false);
         //exercise.showLoader("error");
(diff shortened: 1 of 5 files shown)


Mankro:
I thought it would be faster to fix the bug with small changes. You are suggesting large refactoring, but we should fix the active element submission feedback polling quickly. If we at least first finish a quick and small solution, then there is more time later on to think about major refactoring.

For example, the `data-poll-url` attribute is already used here:
https://github.com/apluslms/a-plus/blob/564ee0985050288b4b09bcf99deca6b22f50c091/exercise/static/exercise/chapter.js#L605
If `data-ready-url` were added there similarly, would that lead to a quick and simple fix to the urgent issue (#1188)?

lainets:
#1192 implements a quick fix to the issue as we need it fixed ASAP. Let's have this PR solely about the ajax refactoring

lainets:
I now made an implementation of JSON responses to AJAX calls for active elements submitting.

The more I think about this, the more I want to try swapping to using React because then most of the rendering happens on the client, simplifying updating of the site considerably.

PR #1191, opened 2023-06-05
WIP submissions after module close (needed for active elements)

lainets:
I created this draft PR to discuss the implementation details. The commit has nothing to do with this.

The current code will block any submissions past the deadline (which makes sense). To get around this issues, I came up with two options: (1) always allow submissions but set the point multiplier to 0 past the deadline, and (2) allow submissions past the deadline if the max points are 0.

# Description

**What?**

[ANSWER HERE]

**Why?**

[ANSWER HERE]

**How?**

[ANSWER HERE]

Fixes #<issue number>

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/static/exercise/chapter.js
+++ b/exercise/static/exercise/chapter.js
@@ -598,24 +598,25 @@
           var url = exercise.url;
           exercise.submitAjax(url, formData, function(data) {
             const content = $(data);
+
+            const site_alerts = content.find('.site-messages');
+            output.find(".site-messages").replaceWith(site_alerts);
+
             // Look for error alerts in the feedback, but skip the hidden element
             // that is always included: <div class="quiz-submit-error alert alert-danger hide">
-            const alerts = content.find('.alert-danger:not(.hide)');
-            if (!alerts.length) {
+            const alerts = content.find(':not(.site-messages) .alert-danger:not(.hide)');
+            output.find(".site-messages").append(alerts);
+
+            if (content.find(".exercise-wait").length != 0) {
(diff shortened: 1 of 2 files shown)


Mankro:
Interesting ideas.

(1) I think there are situations when it is important to prevent new submissions. Even if the student gains no points, they still see the feedback from the grader. Questionnaires are also prone to trial and error (guessing): the feedback could reveal the correct solutions that could be forwarded to other students. (Well, students have many ways to share correct solutions with each other.)

A+ also has the feature "allow unofficial submissions". Unofficial submissions can be made after the deadline but they gain no points. However, teachers decide if they want to enable unofficial submissions (in the learning object category). Most seem to keep it disabled (which is also the default).

(2) This seems better, but I am not sure yet if we want to enable this for all zero-point assignments. We need to think more of the consequences.

If (2) turns out to be unfeasible, then it is still possible to add a new exercise setting that would make the exercise skip the deadline completely.

lainets:
> A+ also has the feature "allow unofficial submissions". Unofficial submissions can be made after the deadline but they gain no points. However, teachers decide if they want to enable unofficial submissions (in the learning object category). Most seem to keep it disabled (which is also the default).

We could add all active element exercises to a separate category with unofficial submissions enabled in a-plus-rst-tools.

Another option would be to add a new boolean database field to LearningObject that allows submission after the deadline (and presumably after the submission limit is up) but with zero points given.

PR #1192, opened 2023-06-07
Fix active elements output not showing

### LINKED ISSUE #1188: Active elements output empty

lainets:
# Description

**What?**

Fix active elements output not showing

**How?**

Add the missing data-ready-url attribute.

Fixes #1188

# Testing

Did a quick check that the output updates now.

- [X] Chrome
- [ ] Firefox

# Is it Done?


Mankro (approved):
Great, this works well!

PR #1198, opened 2023-06-12
Fix ae submissions when alerts are present

lainets:
# Description

**What?**

Fix ae submissions when alerts are present

**Why?**

Danger alerts overrode the output and caused the submission to be skipped. For example, deadline having passed or being staff without enrolling in the course would cause issues.

**How?**

Always show the alerts (replace the previous alerts above the exercise output), and poll the submission whenever one is created (even if a danger alert was included).

Part of #487

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested submitting with and without alerts to see that they show up but can still submit.

- [X] Chrome
- [ ] Firefox

# Is it Done?

### DIFF
--- a/exercise/static/exercise/chapter.js
+++ b/exercise/static/exercise/chapter.js
@@ -598,24 +598,25 @@
           var url = exercise.url;
           exercise.submitAjax(url, formData, function(data) {
             const content = $(data);
+
+            const site_alerts = content.find('.site-messages');
+            output.find(".site-messages").replaceWith(site_alerts);
+
             // Look for error alerts in the feedback, but skip the hidden element
             // that is always included: <div class="quiz-submit-error alert alert-danger hide">
-            const alerts = content.find('.alert-danger:not(.hide)');
-            if (!alerts.length) {
+            const alerts = content.find(':not(.site-messages) .alert-danger:not(.hide)');
+            output.find(".site-messages").append(alerts);
+
+            if (content.find(".exercise-wait").length != 0) {
               const poll_url = content.find(".exercise-wait").attr("data-poll-url");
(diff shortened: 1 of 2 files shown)


lainets:
There is one issue: for some reason commit 83ed12416dcd646af569bde0513ec756561ddf4f makes it so that all issues are marked as errors when doing a post request to ExerciseView while they are warnings when doing a get request. This annoyingly causes the alert style to change from warning to danger after a submission is made. It also doesn't make sense that every issue is given the same severity: e.g. "Staff can submit assignments without enrolling." should be info instead of warning or error.

Mankro (changes requested):
This looks good! You could fix the two very minor issues that I commented below. The git commit message is missing the issue link `Fixes #487`.

Mankro:
> There is one issue: for some reason commit [83ed124](https://github.com/apluslms/a-plus/commit/83ed12416dcd646af569bde0513ec756561ddf4f) makes it so that all issues are marked as errors when doing a post request to ExerciseView while they are warnings when doing a get request. This annoyingly causes the alert style to change from warning to danger after a submission is made. It also doesn't make sense that every issue is given the same severity: e.g. "Staff can submit assignments without enrolling." should be info instead of warning or error. Do we want to fix this now or leave it as is?

To my understanding, this does not affect the fix you have made in this pull request. We don't need to tackle the warning/error style now since it is not a critical problem. We are already running out of time before September so we can focus on the more important issues now.

We could make a new issue about the warning styles.

lainets:
I fixed the problems, created a new issue and rebased on top of master

Mankro (approved):
Excellent!

Issue #31, opened 2012-11-04
Re-submitting exercise form in submission view is not possible


Rubinous:
It is intuitive to try to re-submit exercise form in the submission view in cases where the submission feedback includes the original form. But the form action in that case would be the submissions url instead of exercise url and it won't be handled as the user would expect.

teemulehtinen:
The original POST URL is given as a GET parameter now. The grader may include a POST form in the feedback for resubmitting. Of course the exercise should be designed for resubmitting.

PR #1213, opened 2023-07-05
Add consecutive_order, embedded_in and status to exercise API

etanttila:
# Description

**What?**

Add a consecutive numbering to indicate the order of learning objects (chapters and exercises). 

Add a field "embedded_in" to the exercise API, which contains the name of the chapter an exercise is embedded in (if it is embedded in a chapter).

Also add the status attribute of the exercise to the API. This for example, indicates if the exercise is an enrollment questionnaire.

**Why?**

Distinguishing the order of exercises is difficult, as the ids don't correspond to the order of the exercises in the material. This has caused issues, e.g. in MOOC-Jutut.
The `consecutive_order` field is used to fix Jutut issues [#43](https://github.com/apluslms/mooc-jutut/issues/43) and [#44](https://github.com/apluslms/mooc-jutut/issues/44).

The name of the chapter can provide more understandable context than just the exercise path or the title of the exercise (e.g. "Feedback"). Accessing the details of a parent chapter of an exercise through the API is quite difficult, so this provides a shortcut.
The `embedded_in` field is used to fix Jutut issue [#46](https://github.com/apluslms/mooc-jutut/issues/46).

The `status` can be used to identify enrollment questionnaires, needed for fixing Jutut issue [#31](https://github.com/apluslms/mooc-jutut/issues/31).

**How?**

Add a consecutive_order to learning object models and, which is included in the Exercise serializer. Add functionality to update the numbering when content configuration is updated and when numbering is updated (in case new exercises or chapters have been added).
Add `embedded_in` and `status` to the exercise serializer as well.

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

I manually checked that the consecutive_order numerals were in the correct order in the exercise API, also after changing the order (and renumbering the contents) in the Edit Course view as well as after importing changes in config.
I also manually checked, that the "embedded_in" and "status" fields appear (with the correct values) in the exercise API.

I did not implement Django unit tests for the changes.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


exercise/exercise_models.py
```diff
@@ -177,6 +177,10 @@ class LearningObject(UrlMixin, ModelWithInheritance):
         verbose_name=_('LABEL_CONTENT'),
         blank=True,
     )
+    consecutive_order = models.IntegerField(
```
PasiSa, line 180: This feels a little redundant extra DB field in addition to the old 'order' field, although counted differently. I'm wondering if anything would break badly if the behavior of 'order' would be changed to represent "consecutive order". I didn't notice 'order' being used in aplus logic much at all, but of course we don't know if something uses it through API.

edit_course/views.py
```diff
@@ -122,6 +124,22 @@ def renumber_recursion(self, parent, n=1):
             n += 1
         return n
 
+    def number_learning_objects_consecutively(self) -> None:
```
PasiSa, line 127: This function gets called only when the form is updated through A+ UI, right? Is it a problem that consecutive_order is not having correct values until the user does submits the form? For reference, the old 'order' field is set also in managers.py, when new objects are created. Should consecutive_order be set there, too?

PasiSa (commented):
I didn't test this, just reading through the diff, and it looks generally good to me. A couple thoughts inline anyways, although I'm not sure if they require actions (they are more like questions).

edit_course/views.py
```diff
@@ -122,6 +124,22 @@ def renumber_recursion(self, parent, n=1):
...
+                model.consecutive_order = n
+                model.save()
+                n += 1
+                if not model.is_submittable:
+                    n = inner(entry, n)
+            return n
+
+        n = 1
+        for module in self.content.modules():
+            n = inner(module, n)
```
lainets, line 141: General comment about cache (`self.content` is a CachedContent) usage: the idea behind it is to avoid querying the database. If you are going to query the database anyway and not use any precomputed values within the cache, there is no point in using it.

The above function is said to take a `CourseModule` but `module` here isn't actually a `CourseModule` object but a dict. The `get_common_objects` method above should set self.modules to be the CourseModule objects, so you can use that instead.

exercise/exercise_models.py
PasiSa: Ok, with closer reading, we don't want to change the order behavior. Stupid comment from me :-)

edit_course/views.py
```diff
@@ -122,6 +124,22 @@ def renumber_recursion(self, parent, n=1):
             n += 1
         return n
 
+    def number_learning_objects_consecutively(self) -> None:
+        """Number all learning objects (i.e. chapters and exercises)"""
+        def inner(parent: Union[CourseModule, CourseChapter], n: int):
+            for entry in parent['children']:
+                model = LearningObject.objects.get(id=entry['id'])
```
lainets, line 131: Model objects (CourseModule, CourseChapter) should be indexed using `.attribute` instead of `['attribute']`. The objects.get is also unnecessary after the function actually takes model objects because iterating `parent.children.all()` will give you actual objects.

exercise/exercise_models.py
etanttila: Yeah, the "order" seems to provide "weights", which determine the ordering; the name of the old "order" is quite misleading. (When I heard about it, I was excited that it might provide a solution, but it didn't take long to realize the value is 0 for nearly everything. :( )

exercise/api/full_serializers.py
```diff
@@ -60,18 +65,26 @@ def get_post_url(self, obj):
             return request.build_absolute_uri(url)
         return None
 
+    def get_embedded_in_name(self, obj: BaseExercise) -> str:
+        parent = obj.parent
+        if isinstance(parent, LearningObject):
+            return str(parent)
+        return ''
+
```
lainets, line 73: `obj.parent` is always either None or a LearningObject, so it is enough to check whether the parent is None. `embedded_in_name` should also be renamed to `parent_name` to match the naming convention everywhere else. I believe you can also return `None` if the parent is None which would be preferable. Is there a reason why this returns `str(parent)` instead of `parent.id` (and name it `parent_id` instead of `parent_name`)? Note that `str(parent)` isn't necessarily unique to the parent learning object.

exercise/exercise_models.py
PasiSa: One more question: if consecutive_order is only needed for the API, specifically for Jutut. Could it be calculated dynamically when the API response is composed, and not save it to database at all?

edit_course/views.py
etanttila: This is a good question, since as you mentioned, if a new object is created, it is given the next order number, and the order numbers can also be manually edited for each learning object in the Edit Course view.
Probably most often new objects are created by editing the course rst files and importing them (in which case the consecutive_order is set in edit_course/operations/configure). Therefore this implementation is somewhat more of a backup allowing a teacher to renumber the learning objects consecutively while renumbering the learning objects. However, it's true that a teacher might not renumber the learning objects using the button if they just add a new learning object. However, when the consecutive_order is updated, it should be updated for all learning objects. I'm not sure how easy it would be to update all of them through managers.py (although it might be less bad to have some duplicates than have some missing the field entirely).

So good question, and I'm not sure.

exercise/exercise_models.py
lainets: @PasiSa Why can't we make the order field actually represent the order? I don't see anything blocking that

exercise/migrations/0048_learningobject_consecutive_order.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.20 on 2023-07-05 07:46
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
lainets, line 6: Any existing courses would have the consecutive_order set to 0 for all learning objects. If this field is added to the database, this should probably be fixed.

edit_course/views.py
```diff
@@ -111,6 +112,7 @@ def form_valid(self, form):
                 nn = self.renumber_recursion(module, n)
                 if module['status'] != 'hidden':
                     n = nn
+        self.number_learning_objects_consecutively()
```
lainets, line 115: Is this always required? It doesn't seem like the order can change unless either of the renumber buttons are pressed.

lainets (changes requested):
NOTE: You can ignore most of the issues raised here if the order is calculated dynamically instead.

I'm wondering if we can just make the order atribute work like the consecutive order here. But if not: The consecutive ordering code is now duplicated in the configure code and in the view code. These should be combined. You can make the `number_learning_objects_consecutively` method take a list of modules, move it outside the view class and call it from `form_valid` method and the configure code after the modules and learning objects have been otherwise configured.

I don't _think_ the consecutive_order is updated if course modules/exercises are removed or added.

lainets:
I think Pasi's suggestion of calculating the order dynamically is the best solution. The order can be calculated using `ContentCache` without the need to query the database, so it should be a pretty quick calculation. We avoid adding data to the database (and therefore a migration), and it saves us a lot of trouble trying to keep the `consecutive_order` field updated.

edit_course/views.py
lainets: If the order is calculated dynamically, you can just use the cache directly, so there are no need to use model objects here.

edit_course/views.py
lainets: If the order is calculated dynamically, you can just use the cache directly, so there are no need to use model objects here.

exercise/api/full_serializers.py
etanttila: The name of the parent is nearly impossible to get from the API (as the API doesn't have pages for the chapters, only exercises), and at least in Jutut. It doesn't need to be unique, one just needs the title of the chapter. But good points about changing the logic and the name of the field, I'll change those.

edit_course/views.py
etanttila: The ordering can also change if the user manually edits the order of the learning objects in the Edit course view.

lainets: I don't see how one edits the ordering through EditContentView. AFAIK that would happen through EditModelView. Though, now that I think about this a bit more, I don't think this logic should be in any of the views. If the ordering is added to the database, the numbering would need to be updated whenever the course content is updated, and that could technically happen from anywhere in the code. The numbering logic would then have to be in the model and manager classes. This complexity is probably the reason the existing order attribute doesn't actually contain the absolute order.

etanttila:
I suppose calculating the consecutive order dynamically could be an option, and I suppose in that case one could consider that it's calculated just for the exercises rather than exercises and chapters. @Mankro thought that it might be convenient to have also the chapters be a part of the consecutive ordering, so that if need arises, it could be easy to compare their orders and not just the order of exercises (the idea being that there might be other uses for comparing the order of exercises (and possibly chapters) than just in Jutut.

However, if the order were calculated dynamically, I have a few questions about practicalities:
- Would it be only calculated for the API? If so, how and where? How could one get the proper values for the serializers? And would it be a problem if it were calculated each and every time the details of an exercise are fetched from the exercise API? Or should/could it be cached somehow to decrease the amount of required calculations?
- Or should it be calculated dynamically somewhere else and stored, e.g. as a (cached?) property?

@lainets what is this `ContentCache` you mention and how can it be used?

exercise/exercise_models.py
lainets: That protocol/aplus.py usage seems to assume that the order actually is the absolute order: it is used to set the html tag id for the exercise. There can now be the same html id on multiple exercises if the order is the same (which is possible e.g. by manually setting the order)

edit_course/views.py
etanttila: That's true, yes, the actual changes are made in the EditModelView (I misleadingly called the page in the UI a view), I unfortunately didn't think very far about where the renumbering should be calculated; rather just thought that if it's done whenever the Edit course page is saved (or the renumbering buttons are pressed), that's probably good enough, typically the changes are imported anyway by rebuilding the course. And good point about the course content being possible to happen from nearly anywhere in the code. and probably the numbering logic should be done somewhere else.

lainets:
> I suppose in that case one could consider that it's calculated just for the exercises rather than exercises and chapters. @Mankro thought that it might be convenient to have also the chapters be a part of the consecutive ordering, so that if need arises, it could be easy to compare their orders and not just the order of exercises (the idea being that there might be other uses for comparing the order of exercises (and possibly chapters) than just in Jutut.

I'd still give chapters their own numbers in case we want to do that in the future. For example, we might make the order attribute actually work like an absolute order and replace the need for dynamically computing it. One should also note that an exercise can technically also contain other exercises as its children, in which case you'd have to number the faux-chapter anyway. There might also be use for the numbered chapters in Jutut at some point in the future, who knows. So, unless there is a reason to omit the chapters in the numbering, I would include them.

> Would it be only calculated for the API? If so, how and where? How could one get the proper values for the serializers? And would it be a problem if it were calculated each and every time the details of an exercise are fetched from the exercise API? Or should/could it be cached somehow to decrease the amount of required calculations? Or should it be calculated dynamically somewhere else and stored, e.g. as a (cached?) property?

There are two options:
1. Calculate the ordering on-demand. In this case there probably should be some function that can be called with appropriate arguments (probably either a CourseInstance or LearningObject object). This could be a method on `ContentCache`. E.g. it could be a method that takes a LearningObject object (or maybe its id? whichever is easier for the API) and returns the order for that object.
2. Cache the ordering somewhere. The most natural place would be `ContentCache` as that is roughly the cached equivalent of a course instance's contents (though that would mean the ordering is computed for every course and not only the ones that need it).

In general, whether some data should be cached or not depends on how often it is needed and how long it takes to compute. If it only takes a few milliseconds to compute, it isn't really necessary. On the other hand, if it takes hundreds of milliseconds or more, it should probably be cached as long as the data is accessed multiple times before it needs to be recalculated.

In this case, I believe the computation to be so fast that caching isn't required. It'd be nice if you implemented option (1) above and then measured the execution time e.g. by using time.perf_counter to get the time before and after the computation and logging the difference (take multiple samples as it might vary quite a bit). Caching the result of option (1) later should be trivial if needed.

> @lainets what is this `ContentCache` you mention and how can it be used?

It can be found in `exercise.cache.content`. It is initialized using a CourseInstance, after which it has a .data attribute that is a dict representing the course instance content. The dict has a "modules" key which contains a list of dicts that represent modules, and those modules have a "children" key that contains a list of dicts that represent the child learning objects, and those learning objects have a "children" key...

The `self.content` you use in `number_learning_objects_consecutively` is an instance of `ContentCache`. You can almost use `number_learning_objects_consecutively` as is to compute the ordering. All you have to do is return the ordering of the LearningObject we are interested in as discussed above in option (1) (and it probably should recurse down to the children even if the exercise is submittable, as discussed in the first paragraph).

Mankro:
Note that the old LearningObject.order field is used to render numbered titles to the assignments and chapters. It is also used in sorting the objects in the table of contents or course hierarchy so that the content is displayed in the order that the teacher has defined in the git repo. If you change the semantics of the order field, you have to change how the learning object titles are rendered. It would be awkward if an assignment were not able to define its own title without querying all its sibling assignments. But it must know its siblings to count the order without the old order field. 

My first thought is that changing the order field is not great. Though, CachedContent exists now unlike when ordering was originally implemented. Does caching make the rendering of titles convenient even without the old order field?

lainets:
> Note that the old LearningObject.order field is used to render numbered titles to the assignments and chapters. It is also used in sorting the objects in the table of contents or course hierarchy so that the content is displayed in the order that the teacher has defined in the git repo. If you change the semantics of the order field, you have to change how the learning object titles are rendered. It would be awkward if an assignment were not able to define its own title without querying all its sibling assignments. But it must know its siblings to count the order without the old order field.

I was thinking the order would be otherwise the same as it is now but we make sure that (a) the numbers are unique and (b) the numbers are consecutive. This would otherwise have the exact behaviour as now but you can't have titles where the order number is the same (is there any reason why someone would want to do this?) nor can you increase the order number over the number of learning objects. The whole behaviour where the number in the title might jump to a completely different number or be the same for multiple exercises is a bit odd. Though one use case is to have a different version of an exercise for different user groups with the same numbering.

> My first thought is that changing the order field is not great. Though, CachedContent exists now unlike when ordering was originally implemented. Does caching make the rendering of titles convenient even without the old order field?

Yes, you could just get the title from the cache. However, I'm not sure it is worth it to make the order field have the actual order. It seems like a lot of work for very little gain.

etanttila:
There might be a chance of misunderstanding of what is meant by changing how the order field works. I believe two different alternatives have come up in the discussion:
1. Making the order field work like consecutive_order, where each exercise has a unique order number. (This was initially brought up by Pasi, who immediately scrapped the idea.)
At the moment, this would cause problems with displaying titles, as the hierarchical titles are basically _'module order'.'parent learning object order (if there is a parent)'.'learning object order' Name_
I believe no one actually wants to change the order field to work like this.

2. Making sure that siblings do not have the same order number or that order numbers are skipped if the user manually changes the order number (through the EditModelView via the 'Edit course" -> 'Content' tab).
At the moment, if one changes the order of a chapter, e.g. from 1 to 2, the previous learning object with the order of 2 still has the order 2, and both have titles _X.2 Name_. If one deletes a learning object, the other learning objects' order numbers are not updated, so an order number is skipped, and when one creates a new learning object, that will get the order number of how many learning objects there are. So if initially there were chapters 1.1, 1.2, and 1.3; chapter 1.2 was deleted and a new chapter is added, you'll end up with chapters 1.1., 1.3 and 1.3. This is rather strange behaviour, which I suspect @lainets was referring to in the previous message. It indeed might be worth fixing, but would merit a new issue. Also it probably wouldn't be of a very high priority, as probably most users don't change the order, delete or add new contents through this page but rather directly via the rst-files.

Mankro:
1. Ok, if nobody was changing the old order to a new consecutive order, then nevermind.

2. The old implementation allows duplicate orders etc. in the titles like you said. It looks bad, but it is mostly irrelevant in practice since the auto import from git resets the order numbers. It is uncommon to delete content manually in the Edit course page except during testing and such. There is a renumeration button in the bottom of the page for resetting the order numbers so there is no practical need to force-change the order numbers during the deletion of objects.

We don't really need to support duplicate order numbers but "fixing" them is rather useless since resetting numbers is easy. Note that the git import (aplus-json configure) sets the order to 9999 for every lost object that becomes hidden (removed in git but old submissions exist in the A+ database).

lainets:
Great! Seems good to me.

lainets (approved):

PR #1219, opened 2023-08-07
Fix sticky alerts that hide information when using anchor links

### LINKED ISSUE #1207: Past-course warning banner hides relevant information on page
### Background:

- It is sometimes useful or necessary to link to materials on pages of past course instances. Sometimes, those links are to anchors within pages. 
- A+ shows a floating banner in case the user is viewing a past course instance. This is useful, but...

### Example 1

If I share a link like this:

https://plus.cs.aalto.fi/o1/2022/wNN/faq/?hl=en#may-i-use-a-different-programming-environment-than-intellij

I get the following view, with the relevant section’s title and first lines hidden behind the floating banner. The link’s recipient will presumably be a bit disoriented as to what they’re supposed to be focusing on.

image

### Example 2:

In some cases, the entire relevant section will be hidden. For example, if I share this link:

https://plus.cs.aalto.fi/o1/2022/wNN/faq/?hl=en#can-i-substitute-for-o1-with-the-other-aalto-course-basics-in-programming-y1

I get this:

image

Note: the link’s recipient is actually supposed to be looking at the short section behind the banner, not at the text that is visible.

Mikael-Lenander:
# Description

**What?**

If a page has a sticky alert, scrolls down the height of the alert automatically.

**Why?**

When opening materials of past courses with anchor links, the "This course has already ended." alert hides the information the link is supposed to show.

**How?**

Adds a quick JQuery animation that scrolls the page down.

Fixes #1207

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Opened links and verified that the relevant information is not hidden behind the alert. The animation is triggered only if there is an element with the sticky-alert class on the page which is only used with the course ended-alert. Hence, no other alerts are affected.

- [x] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


ihalaij1 (changes requested):
I tested this and it does not appear to work on long pages with a lot of content, such as aplus-manual page http://localhost:8000/def/current/programming_exercises/graderutils/#iotester.

I tried a few different things, but the only way I was able to make the code work was to add a 1 second delay before executing the new function. That is not a pretty solution.

See https://developer.mozilla.org/en-US/docs/Web/CSS/:target.
@Mankro suggested that the margin-top of the anchor-URL target could be increased using that pseudo-class. It would push the target to be visible from under the banner.

Also, it is considered a good habit to use ``const`` by default for JavaScript variables, unless they require ``let`` or ``var``. Only use ``var``, if it is necessary.

The commit message should include ``Fixes #1207`` on a separate line, with an additional empty line above it.

Mikael-Lenander:
I don't think this is possible with pure css, since we need the height of the alert. Insead, I created another script that doesn't rely on scrolling or animations but simply adds padding to the achor-URL target.

course/templates/course/course_base.html
```diff
@@ -202,6 +202,19 @@ <h4 class="modal-title" id="unenroll-modal-title">
...
 {% endcomment %}
+
+<script>
+$(function() {
+    const hash = window.location.hash;
+    const target = $(hash);
+    const stickyAlert = $('.sticky-alert');
+    if (target.length > 0 && stickyAlert.length > 0) {
+      const alertHeight = stickyAlert[0].offsetHeight;
+      target.css('padding-top', alertHeight + 'px')
```
ihalaij1, line 213: Missing ``;``

ihalaij1 (changes requested):
One more tiny change, see below. Otherwise, looks good!

ihalaij1 (approved):
Ready for merge!

course/templates/course/course_base.html
```diff
@@ -202,6 +202,19 @@ <h4 class="modal-title" id="unenroll-modal-title">
 	</div>
 </div>
 {% endcomment %}
+
+<script>
+$(function() {
```
Mankro, line 207: This could have been added to a separate JavaScript file instead of an inline script. Separate JS files are generally easier to maintain. But anyway, I will merge this. It is a small script connected to the course pages like this html template.

Mankro (approved):

Issue #503, opened 2020-03-04
No public cloud CDN for static CSS/JS files (e-exams)


jaguarfi:
Currently, A+ uses some public cloud CDN for Bootstrap CSS/JS and courses may use the public cloud for MathJax JS. However, in exam scenarios, we should disseminate such files from our servers instead of depending on public Internet access. Exams should have very limited Internet access.

---

**Conversation**

>**Jaakko Kantojärvi**
Solution a) download CDN files and add to the repo (I don't like that idea much)

>Solution b) implement django app, which extends collectstatic command to download the asset to local storage and replaces `{% static "cdn://file.css" %} with reference to a local file.

>Therre are many tools for the b) which also merges all the assets to a a single file, but we don't need that. Specially as HTTP/2 solves that problem.

>This django app could then be used to download jquery toggle javascript used by colortag app in A+ and Jutut.

>Preferably the download/use CDN behaviour would be behind setting variable.

>**Jaakko Kantojärvi**
It's probably a good time to check if there are small updates to basic CDN libs :)

Mankro:
The Drone CI/CD pipeline currently mainly takes care of Bootstrap v3.4. The rest of the assets like jQuery are still linked to the public CDNs.

PR #1222, opened 2023-08-15
Set programming language manually for highlight.js

### LINKED ISSUE #1221: Set the programming language manually to code snippets to avoid bad automatic language detection in highlight.js
The library highlight.js is used to highlight the syntax of programming source code in submissions, model solutions and assignment template/skeleton code. Highlight.js was updated in #1108.

Highlight.js is particularly used in our custom jQuery plugin `hightlightCode`:
https://github.com/apluslms/a-plus/blob/28f2735979eda3229d10ab1ddc034d775ac84d25/assets/js/aplus.js#L312

Highlight.js may automatically detect the programming language whose syntax is highlighted, but the latest version of highlight.js fails to detect the language correctly more often than expected. For example, some Python files I tried were flagged as Kotlin code. The highlighting is incorrect when it uses the wrong language.

This can be circumvented by setting the programming language manually to the code element.

https://highlightjs.readthedocs.io/en/latest/readme.html#in-the-browser

>  If automatic detection doesn’t work for you, or you simply prefer to be explicit, you can specify the language manually in the using the class attribute:
>
> `<pre><code class="language-html">...</code></pre>`

(We don't use the recommended tags `<pre><code>`.)

https://highlightjs.readthedocs.io/en/latest/api.html#highlightelement

> The function uses language detection by default but you can specify the language in the class attribute of the DOM node. See the [scopes reference](https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html) for all available language names and scopes.

https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md

So the class language attribute can be set to `codeBlock` before `highlightElement` is called:
https://github.com/apluslms/a-plus/blob/28f2735979eda3229d10ab1ddc034d775ac84d25/assets/js/aplus.js#L312

Usually, `highlightCode` is called on an element that already has the `data-url` attribute. The source code is loaded via AJAX from the URL. The URL ends in the filename, thus we can use the filename extension (.py, .scala, .c etc.) for manually detecting the programming language.

If the `data-url` attribute is missing, we let highlight.js detect the programming language automatically.

If the filename extension is also recognized by highlight.js as an alias, then we could set the filename extension directly to the html class language attribute instead of making some mapping from filename extensions to the full programming language names. For example, highlight.js should recognize both `language-python` and `language-py` values in the class attribute.

https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md

ihalaij1:
# Description

**What?**

Highlight.js may automatically detect the programming language whose syntax is highlighted, but the latest version of highlight.js fails to detect the language correctly more often than expected.

This fix sets the language manually based on the URL file extension. If the URL does not end with a file extension or highlight.js does not know the file extension, the language is detected automatically.

Matlab and Dockerfile language detection is not included in highlight.js by default, so language support for those was also added.

Fixes #1221

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the programming language is now highlighted correctly in previously incorrectly highlighted code, including Python, Matlab, and Dockerfiles.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/assets/js/aplus.js
+++ b/assets/js/aplus.js
@@ -309,6 +309,26 @@ $(function() {
                 copyTargetCounter += 1;
             }
 
+            // If the URL ends with a file extension or "Dockerfile", set language for highlight.js manually.
+            // The programming language is detected automatically if highlight.js does not know the file extension or
+            // the URL does not end with a file extension.
+            const url = codeBlock.attr('data-url');
+            if (url) {
+              let fileExtOrAlias = "";
+              const splitUrl = url.split('.');
+              if (splitUrl.length > 1) { // Has file extension
+                fileExtOrAlias = splitUrl.pop();
+              } else if (splitUrl.pop().endsWith("Dockerfile")) {
+                fileExtOrAlias = "dockerfile";
+              }
+              if (fileExtOrAlias === "m") {
+                fileExtOrAlias = "matlab";
+              }
+              if (fileExtOrAlias) {
+                codeBlock.addClass('hljs language-' + fileExtOrAlias);
+              }
+            }
+
             hljs.highlightElement(codeBlock[0]);
 
             // Add line numbers.
--- a/templates/base.html
+++ b/templates/base.html
@@ -43,6 +43,12 @@
 	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.8.0/languages/scala.min.js"
 			integrity="sha512-F7HBSjTYDsAreqced8cggvF5AuB+j2TXktnLYpCrpWF2fhQJJCc5KimR3LYLs0AppTBecbPN9M9Hjf8yC2WstA=="
 			crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.8.0/languages/matlab.min.js"
+			integrity="sha512-5ZGiuxBBzpJ0aH15H1I850TRlyLKnJz4oUehftIWUpEGv4k/Yq7Iva5m7yqfLQYjMjZLcQ8eFCkPQR5JL5CUrQ=="
+			crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.8.0/languages/dockerfile.min.js"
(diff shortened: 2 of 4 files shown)


templates/base.html
```diff
@@ -43,6 +43,12 @@
 	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.8.0/languages/scala.min.js"
 			integrity="sha512-F7HBSjTYDsAreqced8cggvF5AuB+j2TXktnLYpCrpWF2fhQJJCc5KimR3LYLs0AppTBecbPN9M9Hjf8yC2WstA=="
 			crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+	<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.8.0/languages/matlab.min.js"
```
Mankro, line 46: Now that we have already three extra languages added to the default bundle, it starts to make sense to build highlight.js locally and include the used languages there. Then, we only need to add one highlight.js file in the HTML templates.

This is a different task for the future. We already have an issue for it: #503 (that is, extend the drone.io pipeline with more libraries; Bootstrap is already there).

assets/js/aplus.js
```diff
@@ -309,6 +309,26 @@ $(function() {
...
+            // The programming language is detected automatically if highlight.js does not know the file extension or
+            // the URL does not end with a file extension.
+            const url = codeBlock.attr('data-url');
+            if (url) {
+              let fileExtOrAlias = "";
+              const splitUrl = url.split('.');
+              if (splitUrl.length > 1) { // Has file extension
+                fileExtOrAlias = splitUrl.pop();
+              } else if (splitUrl.pop().endsWith("Dockerfile")) {
+                fileExtOrAlias = "dockerfile"
```
Mankro:
```suggestion
                fileExtOrAlias = "dockerfile";
```

assets/js/aplus.js
```diff
@@ -309,6 +309,26 @@ $(function() {
...
+              }
+              if (fileExtOrAlias === "m") {
+                fileExtOrAlias = "matlab"
```
Mankro:
```suggestion
                fileExtOrAlias = "matlab";
```

Mankro (approved):
Excellent; I added some missing semicolons to the JS;

PR #1230, opened 2023-08-18
Add hierarchical name to CachedContent and remove regex hack

### LINKED ISSUE #1228: Add the learning object's hierarchical name to CachedContent and remove the regex hack in CourseExercisesViewSet
From the comment https://github.com/apluslms/a-plus/pull/1090#issuecomment-1615351930

* Could `hierarchical_name` be added to `CachedContent`? `__recurse_exercises` now uses an awful regex hack to deal with hierarchical names. If the cache contained the hierarchical name directly in a new field, we could remove the regex hack.
  https://github.com/apluslms/a-plus/blob/ebf3bc012f61f219893d1e00a86acb00f08aa84e/course/api/views.py#L212-L219
* `hierarchical_name` is formed incorrectly for exercises when the course uses Roman content numbering. `__recurse_exercises` uses `f"{child['number']} {child['name']}"` but in this case, the number is Roman while the hierarchical name should always use Arabic numbers "1.2.3".
  https://github.com/apluslms/a-plus/blob/ebf3bc012f61f219893d1e00a86acb00f08aa84e/course/api/views.py#L218
  This is easy to fix if the correct hierarchical name can be read directly from the cache. Then,  `CourseExercisesViewSet` does not need to concatenate any string components together.

ihalaij1:
# Description

**What?**

Add the learning object's hierarchical name to ``CachedContent`` and remove the regex hack in ``CourseExercisesViewSet``.

This fixes the hierarchical name for Roman content numbering, since the hierarchical name should always use Arabic numbering.

Fixes #1228

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the API now returns correct hierarchical names for all numbering styles.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/cache/content.py
+++ b/exercise/cache/content.py
@@ -59,6 +59,7 @@ class CachedContent(ContentMixin, CachedAbstract):
                     'order': o.order,
                     'status': o.status,
                     'name': str(o),
+                    'hierarchical_name': o.hierarchical_name(),
                     'number': module['number'] + '.' + o.number(),
                     'link': o.get_display_url(),
                     'submittable': False,
--- a/course/api/views.py
+++ b/course/api/views.py
@@ -1,5 +1,3 @@
-import re
-
 from rest_framework import filters, viewsets, status, mixins
 from rest_framework.decorators import action
 from rest_framework.exceptions import ParseError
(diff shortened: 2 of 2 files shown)


exercise/cache/content.py
```diff
@@ -79,6 +79,14 @@ def recursion(
                     'allow_assistant_viewing': False,
                     'children': [],
                 }
+                if self.instance.content_numbering in (
+                        CourseInstance.CONTENT_NUMBERING.NONE,
+                        CourseInstance.CONTENT_NUMBERING.HIDDEN,
+                        ):
+                    entry['hierarchical_name'] = f"{entry['number']} {entry['name']}"
+                else:
+                    entry['hierarchical_name'] = f"{entry['number']} {' '.join(entry['name'].split(' ')[1:])}"
```
Mankro, line 88: Would it be simpler to do this? `entry['name']` is annoying to use here since it may or may not have the number at the start. `o.name` never has any numbering prefixed.
```python
entry['hierarchical_name'] = f"{entry['number']} {o.name}"
```

Mankro (changes requested):
Nice. I think I found a simpler way to implement this. Note that the new PR #1231 changes all the cache code so we will need to think about how to merge these.

ihalaij1:
Simplified the code by using ``LearningObject.hierarchical_name()``. It doesn't run extra database queries.

Mankro (approved):
Great!
@ihalaij1 how did you verify that `o.hierarchical_name()` doesn't run any new database queries? Django Debug Toolbar? I was somewhat concerned since computing the number requires loading all the parents of the learning object and the course module too.

Luckily, the `LearningObject.parent_list()` method has had its own "internal caching" for a long time:
https://github.com/apluslms/a-plus/blob/512906647b5689ae6d7655e17b40e9906995447e/exercise/exercise_models.py#L263-L273

PR #1236, opened 2023-08-25
Make 'grader is unstable' banner course specific

### LINKED ISSUE #1146: Automatic regrade: make "grader is unstable" - banner course specific
Currently, when there are enough many unfinished grading tasks in PendingSubmissions table, a globally visible "Grader is unstable" banner is shown in A+ on all pages. However, oftentimes the problem may concern a particular exercise and course, and the alert unnecessarily bothers teachers and students on other courses. The banner should therefore be made course-specific, and only show on courses that have more than GRADER_STABLE_THRESHOLD (current settings.py parameter) retries pending.

ihalaij1:
# Description

**What?**

- Previously, when there were many unfinished grading tasks in ``PendingSubmissions`` table, a globally visible "Grader is unstable" banner was shown in A+ on all pages. However, oftentimes the problem may have concerned a particular exercise and course, and the alert unnecessarily bothered teachers and students on other courses. The banner is therefore made course-specific, and only shows on courses that have more than ``GRADER_STABLE_THRESHOLD`` (``settings.py`` parameter) retries pending.

- Additionally, the banner identifies the exercises that have retries pending, since it seems that when problems start occurring, typically it has been a specific exercise that raises this condition.

Fixes #1146

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the banner now appears only on courses that have more than ``GRADER_STABLE_THRESHOLD`` retries pending.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/course/templatetags/base.py
+++ b/course/templatetags/base.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from django import template
 from django.conf import settings
 from django.utils.safestring import mark_safe
+from django.utils.text import format_lazy
 from django.utils.translation import get_language, gettext_lazy as _
 from lib.helpers import remove_query_param_from_url, settings_text, update_url_params
 from exercise.submission_models import PendingSubmission
@@ -31,13 +32,28 @@ def get_date(cont, key):
 def brand_name():
     return mark_safe(settings.BRAND_NAME)
 
+
 @register.simple_tag
 def brand_name_long():
     return mark_safe(settings.BRAND_NAME_LONG)
 
+
 @register.simple_tag
 def brand_institution_name():
     return mark_safe(settings_text('BRAND_INSTITUTION_NAME'))
+
+
+@register.simple_tag
+def course_alert(instance):
+    exercises = PendingSubmission.objects.get_exercise_names_if_grader_is_unstable(instance)
+    if exercises:
+        message = format_lazy(
+            _('GRADER_PROBLEMS_ALERT -- {exercises}'),
+            exercises=exercises,
+        )
+        return mark_safe(format_lazy('<div class="alert alert-danger">{message}</div>', message=message))
+    return ''
+
 
 @register.simple_tag
 def site_alert():
@@ -45,9 +61,6 @@ def site_alert():
     if message:
         return mark_safe('<div class="alert alert-danger">{}</div>'
                          .format(pick_localized(message)))
-    if not PendingSubmission.objects.is_grader_stable():
-        # Prefer configured alert text, if one is set
-        return mark_safe('<div class="alert alert-danger">{}</div>'.format(_('GRADER_PROBLEMS_ALERT')))
     return ''
 
 
--- a/exercise/submission_models.py
+++ b/exercise/submission_models.py
@@ -829,6 +829,24 @@ class PendingSubmissionManager(models.Manager):
         return not (total_retries and total_retries > settings.GRADER_STABLE_THRESHOLD)
 
 
+    def get_exercise_names_if_grader_is_unstable(self, instance):
+        total_retries_per_exercise = self.values(
+            'submission__exercise__name',
+        ).filter(
+            submission__exercise__course_module__course_instance=instance.id,
+        ).annotate(
+            num_retries=models.Sum('num_retries'),
+        ).order_by(
+            '-num_retries',
+        )[:10]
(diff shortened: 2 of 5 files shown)


course/templatetags/base.py
```diff
@@ -31,23 +32,35 @@ def get_date(cont, key):
...
+
+@register.simple_tag
+def course_alert(instance):
+    exercises = PendingSubmission.objects.get_exercise_names_if_grader_is_unstable(instance)
+    if exercises:
+        message = format_lazy(
+            _('GRADER_PROBLEMS_ALERT -- {exercises}'),
+            exercises=exercises,
+        )
+        return mark_safe('<div class="alert alert-danger">{}</div>'.format(message))
```
Mankro, line 54: I have a feeling that calling `.format()` in the end forces evaluation of the lazy string `message`. You should use `format_lazy()` here as well.

> A version of [str.format()](https://docs.python.org/3/library/stdtypes.html#str.format) for when **format_string, args, and/or kwargs** contain lazy objects. The first argument is the string to be formatted.

https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.text.format_lazy

`mark_save()` should not force evaluation of the lazy string. In the Django source code, its function definition includes the decorator `keep_lazy`. So using `mark_safe()` is OK.
https://docs.djangoproject.com/en/4.2/_modules/django/utils/safestring/#mark_safe

exercise/submission_models.py
```diff
@@ -829,6 +829,22 @@ def is_grader_stable(self):
         return not (total_retries and total_retries > settings.GRADER_STABLE_THRESHOLD)
 
 
+    def get_exercise_names_if_grader_is_unstable(self, instance):
+        total_retries_per_exercise = self.values(
+            'submission__exercise__name',
```
Mankro, line 834: If multiple exercises in the course instance have identical names (for example, some courses use the name "feedback" a lot), then does this query separate those exercises or does it combine them into one due to the identical name?

exercise/submission_models.py
```diff
@@ -829,6 +829,22 @@ def is_grader_stable(self):
...
+        ).filter(
+            submission__exercise__course_module__course_instance=instance.id
```
Mankro, line 836:
```suggestion
            submission__exercise__course_module__course_instance=instance.id,
```

exercise/submission_models.py
```diff
@@ -829,6 +829,22 @@ def is_grader_stable(self):
...
+        ).annotate(
+            num_retries=models.Sum('num_retries')
```
Mankro, line 838:
```suggestion
            num_retries=models.Sum('num_retries'),
```

exercise/submission_models.py
```diff
@@ -829,6 +829,22 @@ def is_grader_stable(self):
...
+        )
+        total_retries = sum(entry['num_retries'] for entry in total_retries_per_exercise)
+        # Check if the grader can be considered unstable on this course instance
+        if total_retries > settings.GRADER_STABLE_THRESHOLD:
+            exercises = ", ".join(f"'{entry['submission__exercise__name']}'" for entry in total_retries_per_exercise)
```
Mankro, line 843: If the course has 100 exercises and all of them are not working, then is the list of exercise names going to grow too much? If all of them are rendered in the alert, it becomes a mess. What if the database query is limited to take only the 10 first exercises? Maybe the user does not need an exact list of all the exercises when there are clearly many problematic exercises.

Mankro (changes requested):
This looks good. There are small details you should still check.

exercise/submission_models.py
ihalaij1: Done! It now only queries for the first 10 exercises with the most retries!

exercise/submission_models.py
ihalaij1: The query combines exercises with identical names under one name along with their combined number of retries.

Mankro: I was thinking that it should separate each exercise, but maybe in this context it does not really matter. After all, the function is only counting a sum from the exercises and listing the most problematic exercises (without the more descriptive identifiers like hierarchical names). So this could be just fine like this.

Mankro (approved):
Good to merge!

PR #1274, opened 2023-09-23
Fix ModuleRevealState crashing if it contains no exercises

Mikael-Lenander:
# Description

**What?**

ModuleRevealState stores the exercises of the module. If the module has none, an empty list is passed to max function, which throws an exception.

**Why?**

It is unlikely that any chapter should be hidden before a module with no exercises is "completed". However, if a teacher does set a model solution to this kind of module, the site will crash.

**How?**

Check if the list is empty before feeding it to the max function.

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

The site does not crash anymore if a module with no exercises has a model solution.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/exercise/reveal_states.py
+++ b/exercise/reveal_states.py
@@ -12,10 +12,12 @@ if TYPE_CHECKING:
     from .cache.points import LearningObjectPoints, ModulePoints, ExercisePoints
 
 
-def _get_exercise_common_deadlines(exercise: ExercisePoints) -> List[datetime.datetime]:
-    deadlines = [exercise.closing_time]
-    if exercise.late_allowed and exercise.late_percent > 0:
-        deadlines.append(exercise.late_time)
+def _get_exercise_common_deadlines(
+        exercise_or_module: Union[ExercisePoints, ModulePoints],
+        ) -> List[datetime.datetime]:
+    deadlines = [exercise_or_module.closing_time]
+    if exercise_or_module.late_allowed and exercise_or_module.late_percent > 0:
+        deadlines.append(exercise_or_module.late_time)
     return deadlines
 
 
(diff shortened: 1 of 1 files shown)


exercise/reveal_states.py
```diff
@@ -148,6 +148,8 @@ def __init__(self, module: Union[CourseModule, ModulePoints], student: Optional[
         self.max_deviation: Optional[DeadlineRuleDeviation] = None
 
     def get_deadline(self) -> Optional[datetime.datetime]:
+        if len(self.exercises) == 0:
```
Mankro, line 151: Empty lists are falsy in `if` conditions. We can avoid the `len()` call and comparison to zero by doing `if not self.exercises:`.

exercise/reveal_states.py
```diff
@@ -167,6 +169,8 @@ def get_latest_deadline(self) -> Optional[datetime.datetime]:
             deadlines.append(
                 self.max_deviation.get_new_deadline(exercise_dict[self.max_deviation.exercise_id].closing_time)
             )
+        if len(deadlines) == 0:
```
Mankro, line 172: If the module contains no exercises, then the `deadlines` list is empty. But couldn't you check that there are no exercises already at the start of the method? It would skip most of the function since the code is useless in that case.

Mankro (approved):
Nice! I have two comments below, though they are basically just micro-optimizations.

Mankro (changes requested):
I found a bug. The deadline-triggered reveal rule did not work if the module contained no exercises. It did not take into account that the module has a deadline even if there are no exercises. Basically, you can't return `None` like you had done.

I am fixing this myself now.

Mankro (approved):

PR #1260, opened 2023-09-08
Do not add line numbers to extremely long feedback or code

ihalaij1:
# Description

**What?**

Previously, the inspect submission page froze if line numbers were added to extremely long feedback/code. Line numbers are now added only if the feedback or code is shorter than 5000 lines.

- A page with ~5000 lines of feedback takes about 3 seconds to load with line numbers.
- A page with ~230 000 lines of feedback used to freeze the page. Now it loads fine, but without line numbers.

Fixes #1210

# Testing

**What type of test did you run?**

- [ ] Accessibility test using the [WAVE](https://wave.webaim.org/extension/) extension.
- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Other test. *(Add a description below)*
- [x] Manual testing.

Tested that the inspect submission page no longer freezes.

**Did you test the changes in**

- [x] Chrome
- [ ] Firefox
- [ ] This pull request cannot be tested in the browser.

# Translation

- [ ] Did you modify or add new strings in the user interface? ([Read about how to create translation](https://github.com/apluslms/a-plus/tree/master/doc#running-tests-and-updating-translations))

# Programming style

- [x] Did you follow our [style guides](https://apluslms.github.io/contribute/styleguides/)?
- [ ] Did you use Python type hinting in all functions that you added or edited? ([type hints](https://docs.python.org/3/library/typing.html) for function parameters and return values)

# Have you updated the README or other relevant documentation?

- [ ] documents inside the doc directory.
- [ ] README.md.
- [ ] Aplus Manual.
- [ ] Other documentation (mention below which documentation).

# Is it Done?

- [ ] Reviewer has finished the code review
- [ ] After the review, the developer has made changes accordingly
- [ ] Customer/Teacher has accepted the implementation of the feature

PR #1286, opened 2023-10-12
Add scroll separately button for submitted files on inspect page

### LINKED ISSUE #1111: Separate vertical scroll bars for the submitted file and the grader feedback in the inspect submission view

### LINKED ISSUE #1281: The word wrap button doesn’t work when inspecting student submissions
When inspecting student submissions, these buttons are shown:

image

However, the _Word wrap_ button fails to work, and produces this in the Console:

image

I’ve also had difficulties with the _Copy to clipboard_ button in the same view but that may be related to the specifics of my own setup, so if it seems to work for everyone else, maybe the problem is at my end.

ihalaij1:
# Description

**What?**

* Loading extremely long submission/template/model files and feedback was made faster in #1260 by omitting the line numbers if the data was longer than 5000 lines.

  The performance is now improved even more by utilizing document fragments and chunked loading, where the user can press a button to load the next 5000 lines (with line numbers always included).

  This fixes a potential 'Out of memory' browser error and a bug where the word wrap button was not working.

* The grader feedback already had a `Scroll separately` button, and it is now also added for the submitted files.

* Indentation was fixed in `aplus.js` file.

* Word wrap button now remembers its state and is synchronized across page reloads and with related buttons in other (hidden) tabs.

image

Fixes #1281
Fixes #1111
Closes #1172

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that loading long template files, model files, submitted files, and feedback works correctly.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


Mankro:
@ihalaij1 There is some confusion here.
* The commit that fixes the issue #1281 also contains a lot of code that is not part of the bug issue (the load more button is an additional new improvement). At least, I thought the bug in #1281 is in the definition of the `list` variable since its scope has changed and thus, it is not defined anymore when it is used later in the code. If you define a `const` inside an `if` branch, then the variable is scoped only inside the `if` branch, right? Then, the variable can not be used outside the `if` statement. (Python behaves differently than JS in this topic.)
* The commit that adds a "scroll separately" button may collide a lot with the open pull request #1172 and its issue #1111. What do you think of this? #1172 may also include some additional details that are not part of this PR (#1286); I'm not sure of it yet. There were also problems in #1172 that I don't know if you have considered here. At least, synchronizing the multiple "scroll separately" buttons in the page caused challenges.

ihalaij1:
I separated the word wrap button fix to a separate commit.

The PR #1172 does the exact same thing (adds separate scroll bar for submitted files) as this PR, except it uses ~40 lines more code to do it in `assessment.js`, because it doesn't reuse code as efficiently.

So, I think we should close that PR, and merge this one instead, fixing both #1281 and #1111.

Mankro:
I already manually merged the commit for the "word wrap button" fix (#1281).

Mankro:
> I already manually merged the commit for the "word wrap button" fix (#1281).

I am also going to merge the commit for the "load more lines" button ("Load long feedback and submission/template/model files in chunks"), but I didn't finish my testing yet. The code looks good in that commit. After these manual merges, this pull request could focus on the "scroll separately" button in the "inspect submission" view and you could rename the pull request (title).

Mankro:
Now I have manually added the git commit of the "load more lines" button to the master branch as well as the v1.20_stable branch.

ihalaij1:
I have made more changes, the remainder of the PR is now ready for a review @Mankro.

PR #1299, opened 2023-10-27
Always classify files ending with '.pdf' as binary

### LINKED ISSUE #1298: Binaryornot library sometimes incorrectly classifies PDF files as non-binary
When large PDF files are detected as non-binary data, the inspect submission page tries to render them and causes the page to load slowly.

ihalaij1:
# Description

**What?**

When a student submitted a large PDF file and it was incorrectly detected as non-binary data by the binaryornot library, the inspect submission page tried to render it and caused the page to load slowly.

This is now fixed by always classifying files ending with '.pdf' as non-binary data.

Fixes #1298

### DIFF
--- a/exercise/submission_models.py
+++ b/exercise/submission_models.py
@@ -832,6 +832,9 @@ class SubmittedFile(UrlMixin, models.Model):
         return guess_type(self.file_object.path)[0]
 
     def is_passed(self):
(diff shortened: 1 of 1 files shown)


Mankro (approved):
Thanks! The title is incorrect since it claims that PDF files are non-binary, but they are binary.

PR #1263, opened 2023-09-11
Remove animation from new notifications badge

ihalaij1:
# Description

**What?**

The big red bouncing animation when students have new notifications has been distracting to some students. The animation is not really needed and is now removed.

Fixes #1262

PR #1303, opened 2023-11-07
Reintroduce animation to new notifications badge

### LINKED ISSUE #1288: Reintroduce animation for new notifications but make it only run a set number of cycles
Course notifications are critical for course communication. They notify the student of feedback provided by the instructors of the course.

In commit f6afe4e animation was removed from the new notification badge. The original problem was that a notification could not be cleared, not that the animation needed to be removed. 

A suggested fix is to revert the changes in commit f6afe4e and adding a cycle count to the CSS animation.
The animation below only runs twice.

`animation: bounce 10s 2 ease-in-out;`

A separate issue about clearing notifications on past courses is in #1289

ihalaij1:
# Description

**What?**

PR #1263 removed the animation from the new notifications badge.

The animation is now reintroduced with some tweaks to make it less annoying and in-your-face since notification visibility is really important on some courses.

Fixes #1288


PasiSa (approved):
I assume this is tested to work and look as intended, so I think it is good to be merged. Sorry, I don't have anything smarter to comment about this :-)

Mankro (approved):
Good!

PR #1304, opened 2023-11-07
Optimize inefficient database query in the deviation list view

### LINKED ISSUE #1302: Optimize the very inefficient database queries in the deviation list view
#1122 presented performance issues in the deviations list page. This new issue focuses on specific performance issues.

Using the Django Debug toolbar, we have noticed that the database queries in the deviations list page include some very inefficient parts.

* https://github.com/apluslms/a-plus/blob/8f856141f3af955668df001942a72c696f9a9722/deviations/templates/deviations/list_dl.html#L66
  * rendering the module's name causes new queries for each record because it needs to load the course instance for the module numbering settings (that set Arabic or Roman numerals). Has `select_related()` or `prefetch_related()` been used incorrectly since the code tries to prefetch the course instances?
  * https://github.com/apluslms/a-plus/blob/8f856141f3af955668df001942a72c696f9a9722/deviations/viewbase.py#L254-L265
  * does the variable in the template benefit from the prefetching in the queryset or is the queryset's cache somehow lost when the template is rendered in this case? Does the variable refer to the original queryset anymore at that stage?
* https://github.com/apluslms/a-plus/blob/8f856141f3af955668df001942a72c696f9a9722/deviations/viewbase.py#L27-L29
  * when all deviations in the course are queried, the query includes a lot of unnecessary attributes. Loading them from the database consumes a lot of resources. Especially the large JSON fields in the exercise model consume a lot of bandwidth even though those attributes are not used at all in the deviations list.
  * Unwanted attributes can be avoided with the Django queryset methods `only()` and `defer()`.
  * `exercise.exercise_info` is one of the unnecessary attributes in this case.

ihalaij1:
# Description

**What?**

The course module name is no longer queried separately for each deviation. This reduces the number of performed queries significantly. Additionally, some large unnecessary attributes are no longer loaded.

**Why?**

The deviations list page loaded extremely slowly before. This should help with the problem.

Fixes #1302

### DIFF
--- a/deviations/viewbase.py
+++ b/deviations/viewbase.py
@@ -257,6 +257,12 @@ def get_deviation_groups(
             'submitter', 'submitter__user',
             'granter', 'granter__user',
             'exercise', 'exercise__course_module',
+            'exercise__course_module__course_instance',
+        )
+        .defer(
+            'exercise__exercise_info',
+            'exercise__description',
+            'exercise__course_module__course_instance__description',
         )
         # parent is prefetched because there may be multiple ancestors, and
         # they are needed for building the deviation's URL.


deviations/viewbase.py
```diff
@@ -257,9 +257,15 @@ def get_deviation_groups(
             'submitter', 'submitter__user',
             'granter', 'granter__user',
             'exercise', 'exercise__course_module',
+            'exercise__course_module__course_instance',
         )
         # parent is prefetched because there may be multiple ancestors, and
         # they are needed for building the deviation's URL.
+        .defer(
```
PasiSa, line 264: Minor nit: I think the defer call should be before the comment above, that refers to prefetch_related.

PasiSa (commented):
Looks good to me. Small proposal related to the placement of defer call, though.

Mankro (approved):
Looks good! You can merge this after you have moved the comment as Pasi suggested.

PR #1310, opened 2023-11-30
Users should not be able to edit content hierarchy

### LINKED ISSUE #1306: Users should not be able to edit content hierarchy
Typically the A+ course content structure is configured from a JSON file generated by gitmanager. However, the teachers have possibility to edit the attributes also in the Edit course / Content view, including the parent learning object. By editing the parent selection it is possible to create a circular reference loop between two learning objects pointing to each other as a parent, that leads to excessive number of database operations, and as a result, leading the system to become unusable due to heavy database load.

Easy approach would be to just disable the possibility to modify the parent selection. Later, it might be useful to think more thoroughly how important it is allow editing the content structure in this view, while the main form of course configuration should be through the JSON configuration.

khattam2:
Fixes #1306

# Description

**What?**

Users should not be able to edit content hierarchy

**Why?**

By editing the parent selection it is possible to create a circular reference loop between two learning objects pointing to each other as a parent, that leads to excessive number of database operations, and as a result, leading the system to become unusable due to heavy database load.

**How?**

Now the parent field has been disabled so the case should not occur again.

Fixes #1306

# Testing

image

- [ ] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

The parent field in the hierarchy has been disabled. It was tested and is no longer editable as seen in the picture.

- [ ] Chrome

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/edit_course/exercise_forms.py
+++ b/edit_course/exercise_forms.py
@@ -55,6 +55,8 @@ class LearningObjectMixin:
         self.fields["parent"].queryset = LearningObject.objects\
             .exclude(id=self.lobject.id)\
             .filter(course_module=self.lobject.course_module)
+        self.fields['parent'].widget.attrs.update(
+            {'readonly': True, 'disabled': True})
 
     @property
     def remote_service_head(self):


ihalaij1 (approved):
Excellent!

PR #1329, opened 2024-01-16
Fix issue with querying user submissions that could cause some submissions to be hidden

murhum1:
# Description

Fixed an issue where not all user submissions would be shown for an exercise when the user had made submissions to another exercise in between. This was caused by unsorted data provided for the `groupby` function on the subsequent line - it requires data to be sorted by the key that's used for grouping.

# Testing

Ran manual tests at minus.cs to verify that all user submissions are now shown.


ihalaij1 (approved):
Nice!

I made the commit message slightly shorter.

PR #1341, opened 2024-02-16
Save deadline deviations as seconds instead of minutes

### LINKED ISSUE #1138: Support 23:59:59 deadline deviations better, one way or another
This is not terribly important, but it’s an actual thing.

**Background, Part 1/3:** 

Teachers often set deadlines or (in my case) deadline deviations such that students can submit by end-of-day. Some teachers deliberately avoid marking a deadline as a (student-visible) 00:00, because that gets displayed as something like "00:00 Tuesday" and invariably results in belated students complaints along the lines of "I thought it was until end of Tuesday, please gimme more time!". So these teachers instead prefer 23:59:59, rendered as something like "23:59 Monday". 

**Background, Part 2/3:**

Students often submit their work very close to the deadline. Perhaps even at literally the last minute, it (too) frequently happens.

**Background, Part 3/3:**

However, A+’s support for this is a bit problematic, at least where deadline extensions are considered (and possibly otherwise too, IDK). A teacher can set a deadline as 23:59:59 with the time-selection widget on the deadline deviations page, but that’s misleading: the seconds get discarded. So the student will see "23:59 Monday" but they will not be able to submit later than "23:59:00".

**Consequent Suggestion:**

Support 23:59:59 deadlines somehow. Off the top of my head, this could be done in a whole bunch of different ways, including the following, all of which I’d be OK with.

* In the UI, render all "00:00 Tuesday" deadlines as "23:59 Monday". 
* Always add one minute to all "23:59" deadlines, all "nn:59" deadlines, or even to all deadlines.
* If a deadline (or DL extension) were to fall on midnight, always treat it as falling on 23:59:59 instead. (Everywhere, including the UI.) 
* If #1137 results in changes, make use of that, and actually record the seconds in the system.

khattam2:
# Description

**What?**

Saving deadline deviations as seconds instead of minutes
allows teachers to better set midnight deadlines, such as
23:59:59.

The minutes are no longer discarded, when a deadline deviation is made using a date.

**Testing**

Tested that the deviations work as expected using a docker image [ihalaij1/run-aplus-front:faketime](https://github.com/ihalaij1/run-aplus-front/commit/335f9ad477ee639d71cf67649953a1ba2cdf47f0) that allows changing of the internal time.

docker-compose.yml:
```
plus:
  image: ihalaij1/run-aplus-front:faketime
  environment:
    FAKETIME: "@2024-01-01 23:55:00"
```

Fixes #1138

### DIFF
--- a/deviations/admin.py
+++ b/deviations/admin.py
@@ -20,7 +20,7 @@ class DeadlineRuleDeviationAdmin(admin.ModelAdmin):
     list_display = (
         'submitter',
         'exercise',
-        'extra_minutes',
+        'extra_seconds',
         'granter',
         'grant_time',
     )
--- a/deviations/forms.py
+++ b/deviations/forms.py
@@ -80,11 +80,11 @@ class BaseDeviationForm(forms.Form):
 
 
 class DeadlineRuleDeviationForm(BaseDeviationForm):
-    minutes = DurationField(
+    seconds = DurationField(
         required=False,
         min_value=1,
-        label=_('LABEL_MINUTES'),
-        help_text=_('DEVIATION_EXTRA_MINUTES_HELPTEXT'),
+        label=_('LABEL_SECONDS'),
+        help_text=_('DEVIATION_EXTRA_SECONDS_HELPTEXT'),
     )
     new_date = forms.DateTimeField(
         required=False,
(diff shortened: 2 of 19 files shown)


ihalaij1:
@khattam2 I made changes to this PR to change the whole deadline deviation system so that it uses seconds instead of minutes for saving the deviations.

PR #1345, opened 2024-03-07
Fix course instance editor breaking when using "instances" as URL.

### LINKED ISSUE #782: Course page is rendered with invalid URLs even when the validation rejects the new URL value
Under A+ edit course, course tab, when you try to enter the instance key/url "instances", it does not pass form validation, but the links in the page are anyway rendered with the new value "instances". Those links do not work since there is no course instance with that url.

This issue originates from #772.

Edit:
The form has validation and the url value "instances" does not pass the validation. However, when you see the validation error in the form, you can see that the links to the course already use the instance url "instances" even though that value has not been written to the database. If you try to open the "content" tab under the A+ edit course section, you get HTTP 404 Not found. The URL clearly has the course instance url "instances", which is very wrong.

sayravai:
Fixes #782

# Description

**What?**

When editing a course instance, if one used 'instances' as the URL, it did not get saved into the database due to being a reserved word, but the returned HTML still contained the broken path for all links on the page.

**Why?**

The URL was only checked for disallowed strings on the model level, which is only run after the URL validation process. This caused the links in the returned HTML to contain the broken URL, as no ValidationErrors were raised.

**How?**

A test similar to the one on the model level was added to the URL validation (CourseInstanceForm -> clean_url). Also a static class variable RESERVED_URLS was added to the CourseInstance class, so the forbidden strings list can be referenced from the CourseInstanceForm class without being redeclared in multiple files.

Fixes #782

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

- [x] Chrome
- [ ] Firefox

It's unlikely that this PR would be browser-dependent in any way.

# Translation

I did not add a new translation string, but reused the ERROR_URL_ALREADY_TAKEN.

# Programming style

No new functions were added in this PR.

- [ ] README.md.
- [ ] Aplus Manual.

This PR does not create any new functionality that should be documented.

# Is it Done?

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -592,6 +592,7 @@ class CourseInstance(CourseInstanceProto, models.Model):
         ('ROMAN', 2, _('NUMBERING_ROMAN')),
         ('HIDDEN', 3, _('NUMBERING_HIDDEN_ARABIC')),
     ])
+    RESERVED_URLS = ("instances",) # strings which cannot be used as course instance URL identifiers
 
     course = models.ForeignKey(Course,
         verbose_name=_('LABEL_COURSE'),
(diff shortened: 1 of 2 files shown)


ihalaij1 (approved):
Nice!

I just removed the period from the end of the commit message.

PR #1348, opened 2024-03-13
Use setUpTestData in tests

### LINKED ISSUE #1233: Use setUpTestData in tests
`setUp` is ran before every test method but `setUpTestData` is run only once for the whole test class. Each test method is run in a transaction and rolled back at the end, so there is no need to re-create the database data for each method separately. Most of the time in unit tests seems to be spent on creating data for the tests, so it would be a massive speed up to reuse the data by using `setUpTestData`.

`setUpTestData` takes the class as an argument but it can be used similarly to `setUp`: just assign the properties to the class (e.g. `cls.instance = CourseInstance...`) and the property can be accessed normally through the object (e.g. `self.instance`).

sayravai:
Fixes #1233

Use setUpTestData instead of setUp for creating test data for a-plus unit tests.

**What?**

Unit tests in a-plus have used the setUp method for creating database entries for tests, which causes them to be created and destroyed separately for each test inside a test class. There is an alternative method called setUpTestData, which only creates the entries once and reuses the data for the whole test class, resulting in a considerable speedup when running tests.

**Why?**

Current a-plus unit tests require roughly 1/3 of the time required before this change. On my laptop:
Using setUp: Ran 245 tests in 232.683s
Using setUpTestData: Ran 245 tests in 70.508s

**How?**

Basically the only thing changed is replacing the setUp instance methods with setUpTestData class methods. In a few cases, the api tests used the same dataset as "normal" tests (course, exercise), so here a custom testcase class (with data definitions) was defined and inherited to both classes instead of TestCase. The edit_course > operations was not converted due to issues, but those tests take like a second anyway, so probably not worth a deeper investigation at this point.

Fixes #1233

# Testing

The unit test suite was successfully ran using the updated code, but of course I cannot be sure these changes do not cause any tests to pass when they should not.

- [x] Django unit tests.
- [ ] Selenium tests.
- [ ] Manual testing.

- [ ] Chrome
- [ ] Firefox
- [x] This pull request cannot be tested in the browser.

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?

### DIFF
--- a/apps/tests.py
+++ b/apps/tests.py
@@ -13,25 +13,26 @@ IFRAME_PLUGIN_ADDRESS = "address13123"
 
 class AppsTest(TestCase):
 
-    def setUp(self):
-        self.client = Client()
-        self.course = Course.objects.create(name="Test", code="test", url="test")
-        self.instance = CourseInstance.objects.create(course=self.course,
+    @classmethod
+    def setUpTestData(cls):
+        cls.client = Client()
+        cls.course = Course.objects.create(name="Test", code="test", url="test")
+        cls.instance = CourseInstance.objects.create(course=cls.course,
                                                       instance_name="Ins",
                                                       starting_time="2000-01-01T12:00:00.000Z",
                                                       ending_time="2020-01-01T12:00:00.000Z")
 
-        self.html_plugin = HTMLPlugin.objects.create(container=self.instance,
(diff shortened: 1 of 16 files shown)


ihalaij1 (approved):
Excellent!

PR #1362, opened 2024-05-14
Add module opening and closing times to API

### LINKED ISSUE #1361: API: Add closing_time and opening_time to modules
Right now, there is no way to get the "closing_time" and "opening_time" properties for modules with any api call.

It would be useful to have it at least in the `GET /courses/<course_id>/exercises/` endpoint

ihalaij1:
# Description

**What?**

Add module opening and closing times to API

**Why?**

Requested for the needs of the Aalto Grades project

Fixes #1361

### DIFF
--- a/course/api/views.py
+++ b/course/api/views.py
@@ -245,6 +245,9 @@ class CourseExercisesViewSet(NestedViewSetMixin,
                 module.opening_time,
                 module.closing_time,
             ),
+            'reading_opening_time': module.reading_opening_time,
+            'opening_time': module.opening_time,
+            'closing_time': module.closing_time,
         }
         module_dictionary['exercises'] = self.__recurse_exercises(module, [])
         return module_dictionary


murhum1 (commented):
Is there any reason why `opening_time` and `reading_opening_time` shouldn't both be available in the api? The suggested implementation would be very ambiguous to API consumers.

ihalaij1:
@murhum1 I followed the logic that is used to define the `is_open` field:
https://github.com/apluslms/a-plus/blob/0086e01f23d17d0916f6d1e7500985efa7066db7/course/models.py#L1397-L1402

Do you think we should just provide both opening times? I think if `reading_opening_time` is defined then it has the same value as `opening_time`.

murhum1:
Okay, I see - but yes, I'd still rather provide both to avoid ambiguity.

PR #1368, opened 2024-05-22
Add group_work_allowed boolean to CourseInstance to avoid expensive l…

### LINKED ISSUE #1367: Inefficient way of checking whether the "Form a group" menu item should be shown or not.
The current code goes through all the learning objects to look for the one with the largest allowed group size. This seems very inefficient as it creates a lot of extra queries for every page load. It would probably be a good idea to replace this by simply adding a "Students can form groups" boolean directly to the course instance model. This could be on by default, and if the teacher wants to hide the group forming functionality from students for the course instance, he/she could just uncheck a box in Edit course -> Course.

sayravai:
…ookup on every page load

Fixes #1367

# Description

**What?**

This PR replaces the automated logic for showing/hiding the "Form a group" menu item with a manual boolean setting, which is set at the course instance level. Note, that this PR requires a migration.

**Why?**

The previous implementation created way too many queries for each page load just to decide whether to show a menu item or not. The result should at least be cached if a manual setting is considered too much work for the course staff.

**How?**

A new boolean property group_work_allowed is added into the CourseInstance class. This is referenced in the course/templates/course/_course_menu.html template instead of checking whether the instance_max_group_size method returns a value greater than one.

Fixes #1367

This PR should be completed by adding checks to set the new boolean according to whether there are group-allowing assignments when data is imported via gitmanager. Also disabling the boolean could be prevented when an existing course has group-allowing assignments. User documentation would also need to be updated to include the new setting.

# Testing

- [x] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested changing the "Students can form groups" setting in Edit course -> Course, and that the "Form a group" appears/disappears when clicking Save. Checked cloning the course instance, and that the setting from source instance persists.

- [x] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


ihalaij1 (changes requested):
@sayravai Are the `instance_max_group_size` properties still needed in [course/views.py](https://github.com/sayravai/a-plus/blob/Fix_1367/course/views.py#L173-L175), [course/viewbase.py](https://github.com/sayravai/a-plus/blob/Fix_1367/course/views.py#L173-L175), and [CourseInstanceBaseMixin](https://github.com/sayravai/a-plus/blob/Fix_1367/course/viewbase.py#L44-L46)?

I tried commenting them out and didn't see anything break.

sayravai:
> @sayravai Are the `instance_max_group_size` properties still needed in [course/views.py](https://github.com/sayravai/a-plus/blob/Fix_1367/course/views.py#L173-L175), [course/viewbase.py](https://github.com/sayravai/a-plus/blob/Fix_1367/course/viewbase.py#L69-L71), and [CourseInstanceBaseMixin](https://github.com/sayravai/a-plus/blob/Fix_1367/course/viewbase.py#L44-L46)?
> 
> I tried commenting them out and didn't see anything break.

I don't see any references to those in the repo, but I don't know if they can be accessed from outside. There's also a similar instance_min_group_size which was not even used in any templates.

ihalaij1:
> > @sayravai Are the `instance_max_group_size` properties still needed in [course/views.py](https://github.com/sayravai/a-plus/blob/Fix_1367/course/views.py#L173-L175), [course/viewbase.py](https://github.com/sayravai/a-plus/blob/Fix_1367/course/viewbase.py#L69-L71), and [CourseInstanceBaseMixin](https://github.com/sayravai/a-plus/blob/Fix_1367/course/viewbase.py#L44-L46)?
> > I tried commenting them out and didn't see anything break.
> 
> I don't see any references to those in the repo, but I don't know if they can be accessed from outside. There's also a similar instance_min_group_size which was not even used in any templates.

Well, it looks like these were added in [this commit](https://github.com/apluslms/a-plus/commit/aa49e4aca864a989364f417c325978c4e2703fcd). I also checked and didn't find any references to `instance_max_group_size` or `instance_min_group_size` in mooc-grader, mooc-jutut, or gitmanager repos.

Do you agree that it is probably safe to delete these? @sayravai

ihalaij1 (approved):
Made remaining changes and squashed commits.

PR #1393, opened 2024-08-20
Open menu links to external websites in a new tab

### LINKED ISSUE #1392: Menu links to external websites should open in a new tab
Menu links to external websites, e.g., https://google.com open in the current tab, when they should open in a new tab.
Links to preconfigured external services open correctly in a new tab.

ihalaij1:
# Description

**What?**

Open menu links to external websites, such as https://google.com in a new tab, instead of in the current tab.

**Why?**

This problem was pointed out by a teacher.

Fixes #1392

### DIFF
--- a/course/templates/course/_course_menu.html
+++ b/course/templates/course/_course_menu.html
@@ -110,7 +110,8 @@
 {% for entry in group.items %}
 {% if entry.enabled %}
 <li>
-	<a href="{{ entry.url }}" {% if entry.blank %} target="_blank" {% endif %}>
+	{% is_external_menu_url entry.url as is_external_menu_url_flag %}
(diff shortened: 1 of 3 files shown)


murhum1 (approved):
Seems to work fine - having to assign the separate `is_external_menu_url_flag` variable just for one if statement feels a bit clunky, but at least according to a quick google search it doesn't look like there's a better alternative :/

ihalaij1:
> Seems to work fine - having to assign the separate `is_external_menu_url_flag` variable just for one if statement feels a bit clunky, but at least according to a quick google search it doesn't look like there's a better alternative :/

Yes, I tried to do it on one line first, but it didn't work...

PR #1402, opened 2024-09-12
Add limit for maximum number of retries for automatic regrading, and …

### LINKED ISSUE #1147: Automatic regrade: add maximum limit on retry attempts and rotate the submission to be resent
When the Pending Submissions queue grows longer than GRADER_STABLE_THRESHOLD retries, the system goes into "UNSTABLE" mode where it only retries one submission per time window. The original thinking was to not unnecessarily load the system that may be in unstable situation. Currently the submission that is retried in this situation is always the same submission, causing that particular submission to be retried several times, and others not at all. The retried submission should be rotated so that retries are not accumulated on same submission. Also, there should be an upper limit for submission retries, because the problem may be a faulty grader that is not fixed right away.

sayravai:
…rotate the submission to be sent.

Fixes #1147

# Description

**What?**

Currently the automatic resubmission system always tries to poll the grader using the same submission when it notices grading to be "unstable" (i.e. when the total number of retries over all pending submissions on a course instance grows over the limit set in the GRADER_STABLE_THRESHOLD variable in settings.py). As there is no limit on how many times this submission is retried, the other pending submissions are never even tried.

This PR sets a hard limit on how many times a given submission can be automatically sent to grading before it is considered to be faulty. If the maximum number of automatic resubmission retries is reached without receiving a grade, the pending submission object is removed, and the "Error" state is set for the submission. Also when the grader is considered to be "unstable", the next submission selected for testing the grader is randomly selected from the pool of pending submissions.

When grading has been automatically retried, the user gets a notification of the current (and maximum) amount of retries in the submission status dialog.

**Why?**

Fixing the issues caused by the current regrading system requires manual intervention from the A+ sysadmins, as one faulty submission can prevent others from being retried indefinitely.

**How?**

This PR takes a simple approach to fix the most urgent issues. Ideally, we would measure the average time each autograded assignment spends in grading, and use this value to determine when a given submission should be resubmitted. Here we are just using a fixed value for all assignments, regardless if they take 10 seconds or 10 minutes on average to grade.

Another area of improvement would be in recovery mode: when a submission is successfully graded, and immediate regrading triggered (at set_ready(), see below), we could choose the next submission to be probed from within submissions of the same assignment as the successful one. This would empty the queue faster in cases where a grader for more than one assignment is faulty and one gets fixed before the other. Now the next one selected may be for a faulty assignment, which breaks the probing chain, leaving the rest waiting for new rounds even though they could be regraded immediately.

The regrading process goes as follows:

The PendingSubmissionManager has a method is_grader_stable, which tries to determine if the grader is currently having issues. It does this by counting total retries of all pending submissions of the course instance, and comparing this with the global GRADER_STABLE_THRESHOLD defined in settings.py. If the threshold is not exceeded, we are in "stable" state, and otherwise in a "recovery" state (in which the grader is probed one submission at a time). The Celery process wakes up every SUBMISSION_EXPIRY_TIMEOUT/2 seconds, and executes regrading based on the mode we are in:

1. *Stable state:* we select all pending submission objects, which have a submission time older than SUBMISSION_EXPIRY_TIMEOUT. We go through each and check, that the corresponding exercise can be regraded. Then we check if the pending submission has less than SUBMISSION_RETRY_LIMIT retries, and if not, remove it and set the submission into "Error" state. Finally we check, that the submission is older than (number of retries * SUBMISSION_EXPIRY_TIMEOUT) seconds, which results in those submissions already retried to be resubmitted only after a longer and longer waiting period.
2. *Recovery state:* we randomly select one pending submission and check if it has more than SUBMISSION_RETRY_LIMIT retries. If so, we set the submission into "Error" state, and the pending submission object is deleted. If not, and if the exercise can be regraded, we send the submission for regrading. Note, that if grading is successful while in recovery state, at submission_models.py > set_ready() there is a call to retry_submissions(), causing an immediate reprobe with the next randomly selected submission. If there has been only one broken assignment (grader), this should empty the whole queue quickly. But if there have been several broken ones, and only one gets fixed, it may take a while for the others to get cleared up.

Fixes #1147

# Testing

- [X] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested creating submissions using the generate-submissions.sh script with the grader intentionally broken so that for some submission, it accepts the submission, but never creates a grading pod for it or returns a grade (by modifying mooc-grader/access/types/stdasync.py > _acceptSubmission() ). I tested the following conditions:
* Failed 75% of all submissions (general grader issues)
* Failed all submissions for one assignment (grader broken for one assignment)
* Failed 75% of submissions for one assignment (issues with grader for one assignment)

I also tested submitting an assignment manually. I checked that the status remained "In grading" while the submission was in the retry loop, and that the status changed into "Error" when maximum retries were exceeded. I checked that in the grading status dialog displayed a notification about the submission being retried automatically, and that the message correctly indicated the amount of tries as well as the maximum number of retries.

- [X] Chrome
- [ ] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

Did not find any documentation on automatic regrading.

# Is it Done?

### DIFF
--- a/aplus/celery.py
+++ b/aplus/celery.py
@@ -57,15 +57,15 @@ def retry_submissions():
         random_choice = choice(submission_ids)
         pending = PendingSubmission.objects.get(pk=random_choice)
         if pending.num_retries >= settings.SUBMISSION_RETRY_LIMIT and settings.SUBMISSION_RETRY_LIMIT > 0:
-            # pylint: disable-next=logging-fstring-interpolation
-            logger.info(f"Recovery state: submission retry limit exceeded for submission {pending.submission} - removing from pending")
+            logger.info("Recovery state: submission retry limit exceeded for submission %s - removing from pending",
+                        pending.submission)
             pending.submission.set_error()
             pending.submission.save()
             pending.delete()
         else:
             if pending.submission.exercise.can_regrade:
-                # pylint: disable-next=logging-fstring-interpolation
-                logger.info(f"Recovery state: retrying expired submission {pending.submission} (retries: {pending.num_retries})")
+                logger.info("Recovery state: retrying expired submission %s (retries: %s)",
+                            pending.submission, pending.num_retries)
                 pending.submission.exercise.grade(pending.submission)
         return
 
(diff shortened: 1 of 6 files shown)


murhum1:
Thanks, looks good!

I fixed the lint issues - the currently failing test in the PR is an intermittent issue and tests passed properly in the fork.

PR #1403, opened 2024-09-13
Add courses where the user is an assistant to the API

### LINKED ISSUE #1397: List courses where the user is an assistant in the API
The endpoint https://plus.cs.aalto.fi/api/v2/users/me does not provide information about the courses where the user is an assistant. If the user is not enrolled in a course as a student, the course is not listed in the endpoint at all.

Our plugin checks this endpoint to determine whether the user has access to a course, but due to this it does not work correctly for course assistants.

ihalaij1:
# Description

**What?**

Add courses where the user is an assistant to the API.
Previously, only the courses where the user is a teacher were listed in the API.

**Why?**

This was requested by a course.

Fixes #1397

### DIFF
--- a/userprofile/api/full_serializers.py
+++ b/userprofile/api/full_serializers.py
@@ -22,11 +22,15 @@ class UserSerializer(UserBriefSerializer):
     email = serializers.CharField(source='user.email')
     enrolled_courses = SerializerMethodField()
     staff_courses = SerializerMethodField()
+    teacher_courses = SerializerMethodField()
+    assistant_courses = SerializerMethodField()
 
     class Meta(UserBriefSerializer.Meta):
         fields = (
             'enrolled_courses',
             'staff_courses',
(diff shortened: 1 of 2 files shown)


ihalaij1:
We should agree on if renaming the old `staff_courses` to `teacher_courses` is okay, because it is a breaking change.

`staff_courses` seems confusing to me if assistants aren't included in it, since assistants are also considered staff.
We need a way to get only the teachers via the API for some other purposes, so including both in the same list isn't an option.

See the conversation here: https://github.com/apluslms/a-plus/issues/1397

murhum1:
Could we include staff_courses as well, which would list courses where the user is a teacher or an assistant? That would also fix the issue for the course plugin without them having to make any changes, and would make sense.

ihalaij1:
Made the changes! @murhum1

murhum1 (approved):

PR #1443, opened 2025-02-11
Add cookie notice

### LINKED ISSUE #1400: Add cookie notice
A+ saves cookies to the users' browser for tracking (Matomo) and for functionality. In accordance to GDPR cookie policies we should inform the user about this and allow them to decline tracking cookies if they so wish. We should also update our privacy policy to show more accurate information about our usage of cookies.

murhum1:
Added a cookie notice page and related placeholder texts and translations. Additionally there's CSS for a cookie-notice element, which should be included in the TRACKING_HTML django setting if needed.

**Why?**

There's currently nothing to support notifying users about cookies, and this is supposed to alleviate that.

Fixes #1400 

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [X] Manual testing.

Tested that navigating to the cookie notice page works both in Finnish and English. Also had a temporary cookie banner while testing its CSS, but removed it from the PR since it should come from TRACKING_HTML in production, as the functionality and text will be different per organization / deployment.

- [ ] Chrome
- [X] Firefox

# Translation

- [X] Did you modify or add new strings in the user interface?


ihalaij1 (approved):
Nice!

PR #1445, opened 2025-02-20
Fix 'Continue where you left off' exercise name HTML escape

### LINKED ISSUE #1432: The exercise names within the "Continue where you left off" block render incorrectly
- Problem description: when the exercise name contains quotation marks, they are not correctly escaped within the string within the "Continue where you left off" block.

- Reproduce: Login as a user -> Go to `Def. Course` -> Click into a exercise whose name contains `' `(quotation mark), for example `6.13 Questions about learners' code` -> Return to the main page of the course -> Observe the "Continue where you left off" part

image

ihalaij1:
# Description

**What?**

Fix 'Continue where you left off' exercise name HTML escape.

image

**Why?**

This broke, for example, apostrophes in exercise names.

**How?**

Use django's `|safe` filter instead of a javascript function for escape.

Fixes #1432

### DIFF
--- a/course/templates/course/course.html
+++ b/course/templates/course/course.html
@@ -30,7 +30,7 @@
 				// Add link child
 				const link = $("<a></a>");
 				link.attr("href", escapeHtml(lastVisited.url));
-				link.text(escapeHtml(lastVisited.title));
+				link.text(lastVisited.title);
 				alert.append(link);
 			}
 		}


murhum1 (changes requested):
The link is actually created in JS, so any escaping or whatnot needs to be done in in JS as well. The `|safe` filter has been added to the "last visited" alert's title here instead of the link.

That being said, you could keep in the `escapeHtml` script, and process *only* the `href` attribute through that, and leave the `.text()` call without the `escapeHtml`, as quick googling says it's XSS safe. So essentially the only change made to the original code would be to remove `escapeHtml` from line 33 of the original code.

ihalaij1:
@murhum1 You're right, for some reason I thought that the URL was included as an argument to the translation string.

murhum1 (approved):

Issue #1448, opened 2025-02-25
Improve "All submissions" view


murhum1:
Currently the new "All submissions" view is very slow, on courses with only 500 submissions taking ~10sec to load. Figure out why and improve it.

Related: https://github.com/apluslms/a-plus/issues/1409

ihalaij1:
PR #1449 improves the performance of the page, but pagination should still be added to make the page usability better.

mikaelGusse:
One thing to note is that you cannot use the filters if you are in the default "Show only 50" submissions view. This would be quite crucial to get working. 

Another suggestion from Otto was to make it so that you could use URL get parameter to filter the page so e.g. https://minus.cs.aalto.fi/aplus-manual/master/teachers/all-submissions/?status=Error. Would show all submission with Error status.

PR #1449, opened 2025-02-28
Improve performance of "All submissions" view

ihalaij1:
# Description

**What?**

* Limit the table rows to 50 by default and fetch exercise objects in the same query with the submissions.
* Pagination should be added to the page later.

Adding `select_related('exercise')` improved the performance, when loading the page limited to 50 table rows.
In local testing it loaded ~2x faster with around 100 submissions with `select_related`, vs. without it.

Part of #1448

### DIFF
--- a/course/templates/course/_course_menu.html
+++ b/course/templates/course/_course_menu.html
@@ -239,7 +239,7 @@
 	</a>
 </li>
 <li class="menu-all-submissions">
-	<a href="{{ instance|url:'all-submissions' }}">
+	<a href="{{ instance|url:'all-submissions' }}?limited=true">
 		<span class="glyphicon glyphicon-list" aria-hidden="true"></span>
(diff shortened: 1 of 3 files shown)


murhum1 (approved):
Good start! Approving this, let's see if we can further improve the performance with pagination or such.

murhum1:
I did some more testing on this with a sample of 2000 generated submissions, and to me it looks like the queries aren't an issue, but the large number of url filters on the template side is. Removing the two last <td> entries, with inspect submission link / link to exercise, cuts the page load time by ~80%. I tried to add url fetching to the view instead of the template, but it didn't really seem to matter, and was slow as well. Need to figure out why url generation is slow.

murhum1:
Got rid of lots of extra queries with some added prefetch_related calls, and the page loaded fast with 1000-ish submissions. Merging and closing this for now, let's check out pagination and such in a separate PR.

PR #1453, opened 2025-03-14
Add API endpoints for adding and deleting submission taggings

### LINKED ISSUE #1430: API endpoint to modify submission tags
It would be useful if the submission tags would be configurable through the API for submissions. This would allow for programmatic ways of filtering different kinds of submissions. For example, the grader could tag every submission that has compilation issues or style problems. This could also be used for plagiarism detection so that a submission could for example be tagged as "suspected plagiarism" if some tool recognizes it as such.

ihalaij1:
# Description

**What?**

* Add SubmissionTag and SubmissionTagging to Django admin site
* Add API endpoint for adding and deleting submission taggings

API usage:

`POST /submissions/<submission_id>/tag/`: tags a submission with a submission tag.
- Request body:
    - `tag_slug`: the slug of the tag to be added.

`DELETE /submissions/<submission_id>/tag/`: removes a submission tag from a submission.
- Request body:
    - `tag_slug`: the slug of the tag to be removed.

Fixes #1430

# Testing

- [ ] Django unit tests.
- [ ] Selenium tests.
- [x] Manual testing.

Tested that the API endpoints seem to work by sending POST and DELETE requests with Postman.

### DIFF
--- a/course/admin.py
+++ b/course/admin.py
@@ -11,6 +11,7 @@ from course.models import (
     LearningObjectCategory,
     UserTag,
     UserTagging,
+    SubmissionTag,
 )
 from lib.admin_helpers import RecentCourseInstanceListFilter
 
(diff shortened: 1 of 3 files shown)


murhum1 (commented):
Works nicely!

I was thinking about the combination of URLs, HTTP methods and request parameters. In my opinion it would make sense to only have one url pattern, e.g. the `/submissions/<submission_id>/tag/`, which would accept both POST and DELETE requests, and both of which would get the `tag_slug` parameter from POST data instead of query parameters. That's at least the REST way of handling adding / deleting things, but what do you think?

ihalaij1:
> Works nicely!
> 
> I was thinking about the combination of URLs, HTTP methods and request parameters. In my opinion it would make sense to only have one url pattern, e.g. the `/submissions/<submission_id>/tag/`, which would accept both POST and DELETE requests, and both of which would get the `tag_slug` parameter from POST data instead of query parameters. That's at least the REST way of handling adding / deleting things, but what do you think?

You're right, that does sound cleaner. I will make the changes.

murhum1 (approved):

PR #1507, opened 2026-03-09
Upgrade to Django 5.2 LTS

sayravai:
# Description

Django 4.2 LTS support ends in April 2026, which is next month. We should upgrade all production Django projects to 5.2, so let's kick this off by this PR, which is essentially changes made by the [django-upgrade](https://github.com/adamchainz/django-upgrade) script in the current a-plus repo.

### DIFF
--- a/api/urls_v2.py
+++ b/api/urls_v2.py
@@ -1,7 +1,7 @@
 import copy
 
 from django.conf import settings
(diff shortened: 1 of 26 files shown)


ihalaij1 (approved):
Let's test this more after it is merged to the master branch.

PR #1525, opened 2026-06-04
Don't retain inputs for feedback exercises

murhum1:
mooc-jutut submissions were being autofilled to have the previous submission's input, which was potentially confusing for students. This fixes that by excluding exercises with the "feedback" class.


ihalaij1 (approved):
Good!

PR #1530, opened 2026-06-29
Improve clarity of deadlines extension badges

### LINKED ISSUE #1494: Students have trouble noticing the details of DL extensions
Personal DL extensions are marked on the Your Points page, with the details shown in a tooltip when hovering over the little clock symbols.

It has happened many times that a student is confused about the personalized deadline, because they fail to find the symbols or fail to mouse over them.

It would be better to make the details of the extensions more prominent. For example, some/all of the following might help, as might other tricks:

* Show the details when mousing over the "Tehtäviä henkilökohtaisella määräaikapidennyksillä" button. 
* Show the details when *pressing* that button.
* Show the details where that button is, without pressing/mousing over anything (when the details fit there, which should be often).
* (For all of the above: when the extension applies equally to all the assignments in a module — which is a very common case — summarize the extension accordingly instead of listing all the assignments separately.)
* Visually highlighting the button/symbols better somehow

murhum1:
# Description

**What?**

In both "Your points" page and table of contents:
- Show extension details in badge if all exercises in module have the same updated DL
- Otherwise, show a generic message and hide details in a tooltip

**Why?**

Students seem to have had confusion about finding deadline extension details. These should make those easier to notice.

**How?**

Example image showing both a module with all exercises having the same deadline, and another module that has varying deadline extensions:
image

Fixes #1494

### DIFF
--- a/course/templatetags/course.py
+++ b/course/templatetags/course.py
@@ -87,6 +87,33 @@ def deadline_extended_exercises_open(entry):
     )
 
 
+# List exercises in a module by deadline extension, to display in tooltip details
+@register.filter
+def deadline_extended_exercises_by_deadline(entry):
+    deadlines_map = {}
+    for e in entry.flatted:
+        if isinstance(e, ExercisePoints) and e.personal_deadline is not None:
+            if e.personal_deadline not in deadlines_map:
+                deadlines_map[e.personal_deadline] = []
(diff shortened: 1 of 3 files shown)


ihalaij1 (approved):
Excellent!

PR #1531, opened 2026-06-29
Add tabindex="-1" to base.html modals

### LINKED ISSUE #1013: It's not possible to close pop-up windows using the ESC key
It should be possible to close the pop-up windows using the ESC key, like on most other websites.

murhum1:
This makes the ESC key close `base.html` modals (which is most of the modals created by the site).

Fixes #1013, although the double-modal issue described in a later comment of the issue still remains. It's a very niche situation, though.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -261,7 +261,7 @@
 		</footer>
 		{% endblock %}
 	</div>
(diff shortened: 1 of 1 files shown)


ihalaij1 (approved):
Nice!

PR #1534, opened 2026-07-06
Add submission invalidation

### LINKED ISSUE #1479: Invalidating a graded submission
There are a number of use cases where a graded submission might be invalidated

- Grading result is wrong
- Test submission
- Other reasons for not accepting a submission etc.

It would be useful if the staff could mark a submission as invalidated, leaving the submission and grading information as-is, but marking that it should not be used for point sums, statistics, etc.

This is a small change to the submission model

reppuli92:
# Description

**What?**

Add a new 'invalidated' status for submissions. Teachers can manually set this status, and invalidated submissions do not count towards point totals.

**Why?**

This was requested by teachers, as it is useful for e.g. invalidating AI submissions. 

**How?**

Updated the submission database schema to include the new status. This includes a Django migration. 

Also created two new API endpoints to invalidate the submission from status Ready and re-validate the submission back to status Ready. These endpoints are only accessible by teachers and only work if the submission status is initially Ready or Invalidated (respectively. 

Also created buttons in the Inspect submission view where this endpoint can be accessed. The buttons also only show if the user is a teacher and the submission has correct status.

Fixes #1479 

# Testing

- [x] Django unit tests.
- [x] Playwright tests.
- [x] Manual testing.

Added a couple new unit tests for the invalidations. Also tested in the UI that invalidation and re-validation works and affects the points totals correctly.

- [ ] Chrome
- [x] Firefox

# Translation

# Programming style

- [ ] README.md.
- [ ] Aplus Manual.

# Is it Done?


ihalaij1:
We should figure out if we want to remove the invalidated status from `exclude_errors()` and make a separate `exclude_invalidated()` method. The main concern is if invalidating a submission should refund a submission attempt to the student.

The following explains the ramifications of keeping the new invalidated status in `exclude_errors()` (from Copilot):

`INVALIDATED` should be included only if an invalidated submission is meant to be treated as a non-counting attempt. It is not technically an error, and adding it to `exclude_errors()` has wider effects than just hiding it from results.

In the current code, it affects:

- `get_submissions_for_student(..., exclude_errors=True)`: invalidated submissions disappear.
- Submission limits in `one_has_submissions()` and `no_submissions_left()`: an invalidated attempt does not consume a submission slot, so the student may submit again.
- `ordinal_number()`: invalidated submissions are not counted when generating the attempt number sent to the grader.
- Batch approval: invalidated submissions are excluded from the submissions eligible for approval.
- Any future caller of `exclude_errors()`: it will silently treat invalidation like a grader error or rejected submission.

It does not affect callers that use the default `get_submissions_for_student()` or query `Submission` directly. The CSV/API result queries already explicitly exclude `INVALIDATED` independently.

I would avoid putting it in `exclude_errors()` because the name is now misleading. Prefer separate semantics:

```python
def exclude_errors(self):
    return self.exclude(status__in=(
        Submission.STATUS.ERROR,
        Submission.STATUS.REJECTED,
    ))

def exclude_invalidated(self):
    return self.exclude(status=Submission.STATUS.INVALIDATED)
```

Then use both filters where appropriate. If the intended rule is that invalidation voids the attempt and permits a replacement submission, keep `INVALIDATED` in a separately named “non-counting submissions” filter. If invalidation is disciplinary and must still consume the attempt limit, it should not be part of the filter used by submission-limit calculations. The relevant shared filter is in submission_models.py, and the limit logic is in exercise_models.py.

reppuli92:
I changed the filtering logic as above - now exclude_errors() doesn't include the invalidated status and there's a separate method exclude_invalidated(). Invalidated submissions also now consume an attempt. I also changed the form loading logic so that erroneous or invalidated submissions aren't loaded on questionnaire forms.

There's now a bunch of commits on this PR, maybe it's best for you to review this first @ihalaij1 and then I'll merge the commits?

reppuli92:
Added new commit to also remove explicit final markings upon submission invalidation. This way the invalidated submissions will never count towards point totals. Also fixed the typo suggested by Copilot. 

The question whether to count invalidated statistics towards submissions remains. Currently invalidated submissions, as they consume a submission, count towards submission counts in exercise points data, submission limit calculations, exercise satatistics endpoints and numbers of submitters. In my view these probably should count invalidated submissions as well, since they must consume a submission? What do you think @ihalaij1 or should for example the statistics endpoints exclude invalidated submissions?

Invalidated submissons don't currently count towards points, and csv:s/other aggregate submission exports.

ihalaij1:
> Added new commit to also remove explicit final markings upon submission invalidation. This way the invalidated submissions will never count towards point totals. Also fixed the typo suggested by Copilot.
> 
> The question whether to count invalidated statistics towards submissions remains. Currently invalidated submissions, as they consume a submission, count towards submission counts in exercise points data, submission limit calculations, exercise satatistics endpoints and numbers of submitters. In my view these probably should count invalidated submissions as well, since they must consume a submission? What do you think @ihalaij1 or should for example the statistics endpoints exclude invalidated submissions?
> 
> Invalidated submissons don't currently count towards points, and csv:s/other aggregate submission exports.

That sounds fine. Did you check the `regrade_exercises(..., 'all')` function that Copilot mentions also regrades invalidated submissions and changes the status back to `READY`?

reppuli92:
Whoops, didn't notice that one :) made new commit so that mass re-grading will always exclude invalidated submissions (whether re-grade type is incomplete or not) and also disabled regrading for single invalidated submissions. The button is also disabled in the inspect submission UI when inspecting an invalidated submission. Also added amention of this in the UI tooltip.

reppuli92:
Added one more commit to disable also manual regrading for invalidated submissions. This also prevents assistants from regrading invalidated submissions. Also noticed incorrect tooltip behavior, as previously if the best submission was invalidated, the second-best (which then determined the grade) had a tooltip in the inspect view saying incorrectly that another submission determines the grade. Fixed this and also added a tooltip saying this to invalidated submissions. 

The second issue raised by Copilot doesn't make sense to me. If a submission is marked as best and manually graded, a teacher probably won't invalidate it in the first place. If for some reason the teacher wants to, though, (if there is suspicion of AI use etc.) the best flagging should also be removed, since invalidated submissions can't count towards the grade. If the submission is then revalidated (which is already 3 steps of a quite weird user flow that probably wouldn't happen under normal circumstances), the teacher should in my opinion manually regrade the submission again instead of the system carrying the flags somehow over. What do you think @ihalaij1

